PR #1, opened 2020-10-05
chore: use structopt wrapper instead of building clap by hand

conradludgate:
Type safe wrapper around clap. Supports everything clap does. Has exactly the same command output as it did before


ellie (approved):

ellie:
wow `structopt` is a lot neater :heart_eyes: 

Thank you!

PR #2, opened 2020-10-05
feat: use directories project data dir

conradludgate:
There's a bit to this PR.

First, add a `--db` flag to set the db file, otherwise it defaults to a sensible data dir (on linux, it uses `$HOME/.local/share/shync/history.db`, different to the `$HOME/.history.db` file you were originally using)

Second, I ran `cargo fmt` to make the formatting consistent

Third, I changed some of the function arguments to take more sensible types. `AsRef<Path>` allows for a much more friendlier usage, allowing `&str`, `String`, `Path`, `PathBuf` etc.


ellie (approved):
:rocket:

src/local/database.rs
```diff
@@ -19,35 +18,37 @@ pub struct SqliteDatabase {
...
-
         debug!("opening sqlite database at {:?}", path);
 
-        let create = !Path::new(path).exists();
+        let create = !path.exists();
+        if create {
+            if let Some(dir) = path.parent() {
+                std::fs::create_dir_all(dir)?;
+            }
+        }
```
conradludgate, line 31: Turns out the `setup_db` function below doesn't create the dir, so this is necessary.

ellie: makes sense

PR #8, opened 2021-02-14
fix history importing

conradludgate:
zsh bin is sometimes /usr/bin/zsh or might be elsewhere too
zsh also uses ~/.zsh_history
get better errors for not found history file


src/command/import.rs
```diff
@@ -61,20 +61,33 @@ fn import_zsh(db: &mut Sqlite) -> Result<()> {
...
+            match candidates.next() {
+                Some(candidate) => {
+                    let histpath = home_dir.join(candidate);
+                    if histpath.exists() {
+                        break histpath;
+                    }
+                }
+                None => return Err(eyre!("Could not find history file. try setting $HISTFILE")),
+            }
+        }
```
conradludgate, line 89: Bet you didn't know you could break from a loop with a value ;)

src/command/import.rs
```diff
@@ -39,7 +39,7 @@ impl Cmd {
             Self::Auto => {
                 let shell = env::var("SHELL").unwrap_or_else(|_| String::from("NO_SHELL"));
 
-                if shell.as_str() == "/bin/zsh" {
+                if shell.ends_with("/zsh") {
```
ellie, line 42: :+1:

src/command/import.rs
```diff
@@ -61,20 +61,33 @@ fn import_zsh(db: &mut Sqlite) -> Result<()> {
...
-            "Could not find history file at {}, try setting $HISTFILE",
-            histpath.to_str().unwrap()
-        ));
-    }
+        histpath
+    } else {
+        let user_dirs = UserDirs::new().unwrap();
+        let home_dir = user_dirs.home_dir();
+
+        let mut candidates = [".zhistory", ".zsh_history"].iter();
```
ellie, line 78: :clap:

ellie (approved):

PR #9, opened 2021-02-14
Add stats command

ellie:
For example

atuin stats day yesterday
atuin stats day last friday
atuin stats day 01/01/21

### DIFF
--- a/src/command/stats.rs
+++ b/src/command/stats.rs
@@ -1,48 +1,98 @@
-use chrono::{Duration, Timelike, Utc};
+use std::collections::HashMap;
+
+use chrono::prelude::*;
+use chrono::{Duration, Utc};
 use chrono_english::{parse_date_string, Dialect};
-use eyre::Result;
+
+use cli_table::{format::Justify, print_stdout, Cell, Style, Table};
+use eyre::{eyre, Result};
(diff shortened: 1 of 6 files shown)


src/command/stats.rs
```diff
@@ -0,0 +1,51 @@
...
+
+                debug!("stats for {}", datetime);
+
+                let datetime = datetime
+                    .with_hour(0)
+                    .unwrap()
+                    .with_minute(0)
+                    .unwrap()
+                    .with_second(0)
+                    .unwrap();
```
conradludgate, line 39: Could possibly use `datetime.date()`. Also, how would timezones be handled?

conradludgate (approved):

src/command/stats.rs
ellie: (should have marked this as WIP :see_no_evil:)

Still need to work out handling timezones :/ Currently it's all stored and worked with as UTC, which will hopefully make things simpler

conradludgate: I think you can use the `Local` type to deal with the local tz stuff. Then convert to Utc by using the  `.into::<DateTime<Utc>>()` implementation

Issue #17, opened 2021-03-10
Add searching


ellie:
Allow searching for certain strings in history. Would be neat to include regex too!

Something like 

`atuin search grep`

could list all commands that include the word grep, with

`atuin search ^grep`

including all commands that *begin with* grep

ellie:
`atuin history search grep` works!

No regex though. I'll do that later!

`atuin search` opens the search TUI

Issue #16, opened 2021-03-10
Unique output for listing


ellie:
Don't have dupes! Perhaps add an option to include duplicates

PR #21, opened 2021-03-20
Add TUI, resolve #19, #17, #16

### LINKED ISSUE #19: Do away with FZF usage
We already have everything in a database! It shouldn't be too much work to use a SQL query instead of listing everything and piping to FZF, and we will also gain a lot in performance 

Also! Considering using the SQLite FTS module :)

ellie:
This totally removes the requirement for FZF! Also no longer lists all
the database contents and pipes them to something, and instead uses
SQLite for searching. Again, SQLite still needs optimizing! We're
literally using a LIKE query here, and something like the FTS module
would make more sense I think.

On top of this, at the moment the TUI is very... "initial
implementation". Ie, the code is a mess. It needs restructuring +
rethinking, but it works :) I'd like to include more stats in the TUI at
some point, without slowing it down too much!

Also automatically search for whatever the current command buffer is.

### DIFF
--- a/src/command/search.rs
+++ b/src/command/search.rs
@@ -94,7 +94,7 @@ fn key_handler(input: Key, db: &mut impl Database, app: &mut State) -> Option<St
 // for now, it works. But it'd be great if it were more easily readable, and
 // modular. I'd like to add some more stats and stuff at some point
 #[allow(clippy::clippy::cast_possible_truncation)]
-fn select_history(db: &mut impl Database) -> Result<String> {
+fn select_history(query: Vec<String>, db: &mut impl Database) -> Result<String> {
     let stdout = stdout().into_raw_mode()?;
     let stdout = MouseTerminal::from(stdout);
     let stdout = AlternateScreen::from(stdout);
@@ -105,7 +105,7 @@ fn select_history(db: &mut impl Database) -> Result<String> {
     let events = Events::new();
 
     let mut app = State {
-        input: String::new(),
+        input: query.join(" "),
         results: Vec::new(),
         results_state: ListState::default(),
     };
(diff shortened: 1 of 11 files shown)


src/command/search.rs
```diff
@@ -0,0 +1,220 @@
...
+        Key::Down => {
+            let i = match app.results_state.selected() {
+                Some(i) => {
+                    if i == 0 {
+                        0
+                    } else {
+                        i - 1
+                    }
+                }
+                None => 0,
```
conradludgate, line 70:
```suggestion
                None | Some(0) => 0,
                Some(i) => i - 1,
```

src/command/search.rs
```diff
@@ -0,0 +1,220 @@
...
+                .enumerate()
+                .map(|(i, m)| {
+                    let mut content = Span::raw(m.command.to_string());
+
+                    if let Some(selected) = app.results_state.selected() {
+                        if selected == i {
+                            content.style =
+                                Style::default().fg(Color::Red).add_modifier(Modifier::BOLD);
+                        }
+                    }
```
conradludgate, line 181:
```suggestion
                    if app.results_state.selected() == Some(i) {
                        content.style = Style::default().fg(Color::Red).add_modifier(Modifier::BOLD);
                    }
```

conradludgate (approved):

conradludgate:
> (off topic) @conradludgate how in the world did you approve after the PR was merged...?! 😄 

You can approve any PR post merge :)

bl-ue (commented):
Really! I can't, and have never been able to, even on repos that I have write access to, e.g. https://github.com/tldr-pages/tldr 🤔

PR #23, opened 2021-03-20
Implement server

ellie:


.env
```diff
@@ -0,0 +1 @@
+DATABASE_URL=postgres://atuin:aeBafah6AiJu@localhost/atuin
```
ellie, line 1: not actually a secret password, don't worry. I'm going to write up some "how to dev" docs later and it can be a default for local dev. Nothing secret :)

PR #46, opened 2021-05-07
Fix ubuntu install

ismith:
The filename for the deb is versioned like '0.6.4', not 'v0.6.4'.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -50,7 +50,7 @@ __atuin_install_arch(){
 __atuin_install_ubuntu(){
 	echo "Ubuntu detected"
 	# TODO: select correct AARCH too
-	ARTIFACT_URL="https://github.com/ellie/atuin/releases/download/$LATEST_VERSION/atuin_${LATEST_VERSION}_amd64.deb"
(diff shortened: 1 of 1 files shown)


ismith:
I'm not sure how you are managing versioning/releases, so that's untouched.

ellie (approved):
Thank you for sorting this out! :sparkling_heart: 

Could do with updating the artifact build so that the name matches the other artifacts (and has the v prefix), but I'm happy with this change as it fixes the current release.

ismith:
Yeah, fixing the name so all the versions match would be a fine solution as well, I just couldn't find where cargo (I assume) is configured to build the .deb.
 
 Thanks for making this, it's what I've always wanted!

ellie:
> Yeah, fixing the name so all the versions match would be a fine solution as well, I just couldn't find where cargo (I assume) is configured to build the .deb.

In case anyone else is wondering, currently it's done here: https://github.com/ellie/atuin/blob/main/.github/workflows/release.yaml#L141

`cargo-deb` might have an option for setting the filename, otherwise I guess the pipeline could just rename it or something :woman_shrugging: 

> Thanks for making this, it's what I've always wanted!

Glad you like it! It's something I'd been meaning to make for a while haha

PR #48, opened 2021-05-07
remove dyn Reply

conradludgate:
Creates an enum `ReplyEither` which is essentially as `Result` which implements `Reply`. Provides some helper functions to turn standard reply types into a `Result<ReplyEither<T, E>, Infallible>` aka `ReplyResult<T, E>`.


ellie (approved):
Neat! Thank you :)

LGTM!

PR #49, opened 2021-05-07
cache builds

conradludgate:
Add some cache steps to speed up CI

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -58,14 +58,6 @@ jobs:
         cargo -V
         rustc -V
 
-    - uses: actions/cache@v2
-      with:
-        path: |
-          ~/.cargo/registry
-          ~/.cargo/git
-          target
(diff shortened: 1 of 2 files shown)


ellie (dismissed):
Thank you! :rocket:

I think we could also get away with disabling the mac builds, at least for PRs. They're not really needed + are quite slow

conradludgate:
From 10 minutes to 2 minutes. Not bad

.github/workflows/release.yaml
```diff
@@ -58,6 +58,14 @@ jobs:
         cargo -V
         rustc -V
```
ellie, line 60: I'm happy with the rest, but can we remove this one?

This workflow is only ever triggered when I push a release tag, and for *paranoid reasons* I'd like to ensure releases are built from scratch rather than having dependencies loaded from the cache.

ellie (approved):
Fantastic, thanks! LGTM :rocket:

Issue #56, opened 2021-05-08
possibly a bug: install script broke my .zshrc


avinassh:
Hey there 👋  This is really a nice project and just like what I needed. I did the installation using the install script, at the end my `~/.zshrc` ended up like this:

```shell
alias some_cmd="some args"eval "$(atuin init zsh)"
```

This happened because my earlier version must have not ended with a newline. May be I edited it and forgot to add a newline. I am happy to send a patch which manually adds a new line, but I was not so sure. Because this is not a bug if `~/.zshrc` files are expected to ended with a newline. Also, I am not really sure how do other install scripts behave.

ellie:
> This is really a nice project and just like what I needed.

Thank you!

> Because this is not a bug if ~/.zshrc files are expected to ended with a newline

Nope, this is totally a bug! 

If you want to open a PR to fix this that would be awesome, otherwise I'll do it later today 😊

Thank you for reporting this!

avinassh:
@ellie sent :D

conradludgate:
Solved in #60 :tada: Thanks @avinassh

PR #60, opened 2021-05-08
Add a newline when editing shell rc files

avinassh:
Fixes issue #56 

I have added same change to bashrc file as well, but I don't use bash. If its not correct, let me know I will remove it.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -148,5 +148,5 @@ esac
 printf '\neval "$(atuin init zsh)"' >> ~/.zshrc
 
 curl https://raw.githubusercontent.com/rcaloras/bash-preexec/master/bash-preexec.sh -o ~/.bash-preexec.sh
-echo '\n[[ -f ~/.bash-preexec.sh ]] && source ~/.bash-preexec.sh' >> ~/.bashrc
+printf '\n[[ -f ~/.bash-preexec.sh ]] && source ~/.bash-preexec.sh' >> ~/.bashrc
 echo 'eval "$(atuin init bash)"' >> ~/.bashrc
--- a/install.sh
+++ b/install.sh
@@ -145,8 +145,8 @@ case "$OSTYPE" in
 esac
 
 # TODO: Check which is in use
-echo 'eval "$(atuin init zsh)"' >> ~/.zshrc
+echo '\neval "$(atuin init zsh)"' >> ~/.zshrc
 
 curl https://raw.githubusercontent.com/rcaloras/bash-preexec/master/bash-preexec.sh -o ~/.bash-preexec.sh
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -145,8 +145,8 @@ case "$OSTYPE" in
 esac
 
 # TODO: Check which is in use
-echo 'eval "$(atuin init zsh)"' >> ~/.zshrc
+echo '\neval "$(atuin init zsh)"' >> ~/.zshrc
```
bl-ue, line 148: This line doesn't work for me.  It adds the literal line `\neval "$(atuin init zsh)"` to the end of my `~/.zshrc`, and it doesn't add a newline.

Shellcheck warning [SC2028](https://github.com/koalaman/shellcheck/wiki/SC2028) recommends to use [`printf`](https://manned.org/printf.1) instead of echo, for this very reason.

```suggestion
printf '\neval "$(atuin init zsh)"' >> ~/.zshrc
```
I tested that, and it works! 🎉

FYI—my bash and OS version are the following:
```console
$ bash --version
GNU bash, version 5.1.8(1)-release (x86_64-apple-darwin20.3.0)
Copyright (C) 2020 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
$ sw_vers
ProductName:	macOS
ProductVersion:	11.3.1
BuildVersion:	20E241
```

install.sh
```diff
@@ -145,8 +145,8 @@ case "$OSTYPE" in
...
 
 curl https://raw.githubusercontent.com/rcaloras/bash-preexec/master/bash-preexec.sh -o ~/.bash-preexec.sh
-echo '[[ -f ~/.bash-preexec.sh ]] && source ~/.bash-preexec.sh' >> ~/.bashrc
+echo '\n[[ -f ~/.bash-preexec.sh ]] && source ~/.bash-preexec.sh' >> ~/.bashrc
```
bl-ue, line 151:
```suggestion
printf '\n[[ -f ~/.bash-preexec.sh ]] && source ~/.bash-preexec.sh' >> ~/.bashrc
```

bl-ue (changes requested):

install.sh
conradludgate: I was going to suggest the same thing. Echo with newlines is not always supported

ellie:
Thank you for submitting this! Really appreciate it

If we get the changes suggested by @bl-ue in, I'm happy to approve this and get it merged 😊

bl-ue (approved):

ellie (approved):
Thank you! 😃

PR #62, opened 2021-05-08
Add ATUIN_NOBIND

### LINKED ISSUE #51: Disabling the up-arrow key rebinding?
I like <kdb>Ctrl+R</kbd> to be rebinded to use `atuin`, but I'd rather the up arrow keep its current functionality.  Is there a way to achieve thise?

### LINKED ISSUE #59: Option to disable rebinds
`atuin init <shell>` rebinds up and ctrl+r by default.
There should be an option to disable one of them.

ellie:
This should resolve #59 and resolve #51

I would still like to have more configurable key bindings! Though this could certainly happen within the respective shell config instead.

### DIFF
--- a/docs/key-binding.md
+++ b/docs/key-binding.md
@@ -1,6 +1,6 @@
 # Key binding
 
-By default, Atuin will rebind both Ctrl-r and the up arrow. If you do not want
+By default, Atuin will rebind both <kbd>Ctrl-r</kbd> and the up arrow. If you do not want
 this to happen, set ATUIN_NOBIND before the call to `atuin init`
 
 For example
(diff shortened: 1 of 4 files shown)


docs/key-binding.md
```diff
@@ -0,0 +1,38 @@
+# Key binding
+
+By default, Atuin will rebind both Ctrl-r and the up arrow. If you do not want
```
bl-ue, line 3: Hey, did you know that GitHub allows you to use the HTML tag `kbd` to make text look like a <kbd>keyboard button</kbd>?
```suggestion
By default, Atuin will rebind both <kbd>Ctrl-r</kbd> and the up arrow. If you do not want
```

ellie: I did not! That looks much better, thank you

conradludgate (approved):

PR #69, opened 2021-05-08
Add importer for resh_history file

Sciencentistguy:
A quick parser and importer for the history file generated by resh (https://github.com/curusarn/resh).


atuin-client/src/import/resh.rs
```diff
@@ -0,0 +1,56 @@
...
+    pub os_release_pretty_name: String,
+    pub resh_uuid: String,
+    pub resh_version: String,
+    pub resh_revision: String,
+    pub parts_merged: bool,
+    pub recalled: bool,
+    pub recall_last_cmd_line: String,
+    pub cols: String,
+    pub lines: String,
+}
```
ellie, line 56: I'd love to be able to get some of this stored in Atuin too :thinking: Obviously not required for this PR, but definitely a future thing

src/command/import.rs
```diff
@@ -56,10 +64,75 @@ impl Cmd {
...
+        .split('\n')
+        .map(str::trim)
+        .map(|x| serde_json::from_str::<ReshEntry>(x))
+        .filter_map(Result::ok)
+        .map(|x| {
+            #[allow(clippy::cast_possible_truncation)]
+            #[allow(clippy::cast_sign_loss)]
+            let timestamp = {
+                let secs = x.realtime_before.floor() as i64;
+                let nanosecs = (x.realtime_before.fract() * 1_000_000_000_f64).round() as u32;
```
ellie, line 87: One small thing to confirm here. Is there any chance that Resh will have many history items with the same timestamp?

Due to how sync works at the moment, if too many items have the same timestamp, they will be missed while syncing. 

This is why the ZSH importer applies a small millisecond offset while importing data: https://github.com/ellie/atuin/blob/main/atuin-client/src/import/zsh.rs#L116

In the future I'd like to remove this by either 

1) Move the responsibility for this offset away from the importer
2) Improve sync so that this isn't an issue

ellie (commented):
Thank you so much for this! I actually love the amount of data Resh can track so maybe we could take some inspiration there

Just a couple of comments but otherwise looks great :rocket:

src/command/import.rs
Sciencentistguy: After having a look through my 150MiB resh_history.json, it appears that no two entries share a realtimeBefore, so this is not an issue.

ellie: Fantastic! thanks

ellie (approved):

PR #75, opened 2021-05-08
Support fulltext search of commands

yuvipanda:
Support fulltext search mode.

Fulltext search will report partial matches of commands, not just prefix
matches. Matches ctrl-r default behavior (on zsh at least).

TODO:

- <s>Is the way I'm doing string concat idiomatic rust?</s> It wasn't, but it is now
- <s>Should this be toggleable behind a config options?</s> Yes
- <s>Should this be a param to db.query method, toggling fulltext vs prefix search?</s> Yes

Fixes https://github.com/ellie/atuin/issues/72

### DIFF
--- a/src/command/search.rs
+++ b/src/command/search.rs
@@ -16,6 +16,7 @@ use unicode_width::UnicodeWidthStr;
 
 use atuin_client::database::Database;
 use atuin_client::history::History;
+use atuin_client::settings::{Settings, SearchMode};
 
 use crate::command::event::{Event, Events};
 
@@ -130,10 +131,10 @@ impl State {
     }
 }
 
-async fn query_results(app: &mut State, db: &mut (impl Database + Send + Sync)) -> Result<()> {
+async fn query_results(app: &mut State, search_mode: SearchMode, db: &mut (impl Database + Send + Sync)) -> Result<()> {
     let results = match app.input.as_str() {
         "" => db.list(Some(200), true).await?,
-        i => db.search(Some(200), &format!("*{}", i)).await?,
+        i => db.search(Some(200), search_mode, i).await?,
     };
 
     app.results = results;
@@ -149,6 +150,7 @@ async fn query_results(app: &mut State, db: &mut (impl Database + Send + Sync)) 
 
 async fn key_handler(
     input: Key,
+    search_mode: SearchMode,
     db: &mut (impl Database + Send + Sync),
     app: &mut State,
 ) -> Option<String> {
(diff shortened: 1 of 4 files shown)


src/command/search.rs
```diff
@@ -133,7 +133,7 @@ impl State {
 async fn query_results(app: &mut State, db: &mut (impl Database + Send + Sync)) -> Result<()> {
     let results = match app.input.as_str() {
         "" => db.list(Some(200), true).await?,
-        i => db.search(Some(200), i).await?,
+        i => db.search(Some(200), &("*".to_owned() + &i)).await?,
```
conradludgate, line 136: Personally, I would prefer

```suggestion
        i => db.search(Some(200), &format!("*{}", i)).await?,
```

yuvipanda: Done!

yuvipanda:
Putting this behind a flag makes sense! This is my first rust PR (yay?) so I'll poke around and see how I can do that.

ellie:
Please let us know if you need any help! :) Thanks for opening this

I imagine it would make most sense for this to be a config option, here are a few pointers:

The client settings are defined [here](https://github.com/ellie/atuin/blob/main/atuin-client/src/settings.rs#L14), so we could add a `search_mode`. This could default to "prefix" for the current way of searching, or be "full" for what you propose. You can see the defaults set [here](https://github.com/ellie/atuin/blob/main/atuin-client/src/settings.rs#L96)

Then we could add another parameter to the `search` function defined [here](https://github.com/ellie/atuin/blob/main/atuin-client/src/database.rs#L271). This would be for setting the search mode used, and the query could be constructed appropriately.

Note that this part of the code is taking whatever the query is, and appending a `%` - https://github.com/ellie/atuin/blob/main/atuin-client/src/database.rs#L278

> This is my first rust PR (yay?) 

Yay indeed! hopefully the first of many :D

yuvipanda:
That was a very useful comment, @ellie. I've implemented it as you mentioned and updated the PR!

conradludgate (commented):
Looks good to me! Just a few formatting issues. I'll wait for @ellie to take a look too.

yuvipanda:
Thanks for the review, @conradludgate! I rebased the branch + ran rustfmt.

ellie (approved):
LGTM! Thank you so much for getting this sorted out 😊

yuvipanda:
Yay, thanks for the merge, @ellie :)

stratosgear:
:+1:  Looking forward for a new release, so I can get that feature too.  Just came here to report the same feature, and I find it already PRed!  Thanks!

PR #76, opened 2021-05-08
Retain the query entered into the TUI

### LINKED ISSUE #73: If query is not found, should insert it in shell on enter
Started using atuin on my terminal, it's a very nice tool!

What would be great is if one enters a command that's not found in the history and then presses [ENTER] to go back to the terminal, the entered text in the history prompt should appear again on the command line prompt. I miss this feature because it's the default behavior of CTRL+R reverse search in ZSH (which this tool also overrides).

ellie:
If no results are found, and the user presses enter, keep the query!

Resolves #73


conradludgate (dismissed):

src/command/search.rs
```diff
@@ -160,7 +160,7 @@ async fn key_handler(
             return Some(
                 app.results
                     .get(i)
-                    .map_or("".to_string(), |h| h.command.clone()),
+                    .map_or(app.input.to_string(), |h| h.command.clone()),
```
conradludgate, line 163: Wait - why are you calling `.to_string` on a `String` type? :P

conradludgate: Right, it needs to be cloned. I think `app.input.clone()` would be more appropriate

ellie: Blindly replace the `""` without entirely considering the context? :eyes: never! haha thanks for the catch, sorted it

conradludgate (approved):

PR #79, opened 2021-05-09
Implement 'quick access' via Alt-<n>

yuvipanda:
Puts numbers 0-9 next to commands *above* current selection.
Ctrl-<number> should activate them - but since Ctrl-<num> are
reserved by terminal, this does not currently work. Need to
find different sets of keyboard shortcuts.

Numbers are *above* current selection, since the user must use
the arrow keys to go over the commands below current selection
before reaching selection.

image

### DIFF
--- a/src/command/search.rs
+++ b/src/command/search.rs
@@ -99,18 +99,19 @@ impl State {
 
                 let selected_index = match self.results_state.selected() {
                     None => Span::raw("   "),
-                    Some(selected) => {
-                        match i.checked_sub(selected) {
-                            None => Span::raw("   "),
-                            Some(diff) => {
-                                if 0 < diff && diff < 10 {
-                                    Span::styled(format!(" {} ", diff), Style::default().fg(Color::DarkGray))
-                                } else {
-                                    Span::raw("   ")
-                                }
+                    Some(selected) => match i.checked_sub(selected) {
+                        None => Span::raw("   "),
+                        Some(diff) => {
+                            if 0 < diff && diff < 10 {
+                                Span::styled(
+                                    format!(" {} ", diff),
+                                    Style::default().fg(Color::DarkGray),
+                                )
+                            } else {
+                                Span::raw("   ")
                             }
                         }
-                    }
+                    },
                 };
 
                 let duration = Span::styled(
@@ -131,8 +132,14 @@ impl State {
                     }
                 }
 
-                let spans =
-                    Spans::from(vec![selected_index, duration, Span::raw(" "), ago, Span::raw(" "), command]);
+                let spans = Spans::from(vec![
+                    selected_index,
+                    duration,
+                    Span::raw(" "),
+                    ago,
+                    Span::raw(" "),
+                    command,
+                ]);
 
                 ListItem::new(spans)
             })
@@ -189,9 +196,7 @@ async fn key_handler(
             let c = c.to_digit(10)? as usize;
             let i = match app.results_state.selected() {
                 None => return None,
-                Some(selected) => {
-                    c + selected
-                }
+                Some(selected) => c + selected,
             };
 
             return Some(
--- a/src/command/search.rs
+++ b/src/command/search.rs
@@ -185,10 +185,14 @@ async fn key_handler(
                     .map_or(app.input.clone(), |h| h.command.clone()),
             );
         }
-        Key::Ctrl(c) if ['0', '1', '2', '3', '4', '5', '6', '7', '8', '9'].contains(&c) => {
-            // FIXME: This doesn't work since Ctrl+<num> are reserved and don't pass through to programs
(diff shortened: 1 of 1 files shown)


yuvipanda:
Can't think of a proper keybinding tho :(

yuvipanda:
Hmm, Using `Key::Alt` doesn't seem to work either - and I'm not sure how to debug these either. Any `println` calls are cleared when the tui exits.

ClashTheBunny:
> Hmm, Using `Key::Alt` doesn't seem to work either - and I'm not sure how to debug these either. Any `println` calls are cleared when the tui exits.

can you print to stderr and then do redirection of `2>atuin.log` when running atuin?

yuvipanda:
Yay, that helped @ClashTheBunny. This works now \o/

Is https://github.com/ellie/atuin/blob/e43e5ce74a85d87a625295b9b089a1b5b8e26fab/src/command/search.rs#L354 how the selected command is passed back to the shell? My redirecting stderr seemed to cause weird things to not work, so am trying to understand the mechanism by which this works.

yuvipanda:
`cargo run search -i` doesn't seem to work anymore, but if I do a `cargo build --release` it works fine

ellie:
> how the selected command is passed back to the shell?

In an extremely gross + awful way that I could find no alternative for :/ 

https://github.com/ellie/atuin/blob/de2e34ac500c17e80fe4dcf2ed1c08add8998fa3/src/shell/atuin.zsh#L29

This is needed (afaik) because in order for the TUI to work, it needs to use stdout.

Are you getting any kind of error? `cargo run -- search -i` works ok for me! The arrow keys might not work properly because the terminal could be in the wrong mode, but otherwise it's fine

yuvipanda:
Hmm, `cargo run -- search -i` continues to not work for me - if I press enter on a command, it's not actually showing up. Not sure why. I ran `cargo clean` first as well.

yuvipanda:
I've removed '0' since it's the same as Enter.

src/command/search.rs
```diff
@@ -169,6 +192,19 @@ async fn key_handler(
                     .map_or(app.input.clone(), |h| h.command.clone()),
             );
         }
+        Key::Alt(c) if ['1', '2', '3', '4', '5', '6', '7', '8', '9'].contains(&c) => {
+            let c = c.to_digit(10)? as usize;
+            let i = match app.results_state.selected() {
+                None => return None,
+                Some(selected) => c + selected,
+            };
```
conradludgate, line 200:
```suggestion
            let i = c + app.results_state.selected()?;
```

src/command/search.rs
```diff
@@ -169,6 +192,19 @@ async fn key_handler(
                     .map_or(app.input.clone(), |h| h.command.clone()),
             );
         }
+        Key::Alt(c) if ['1', '2', '3', '4', '5', '6', '7', '8', '9'].contains(&c) => {
```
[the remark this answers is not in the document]
conradludgate: Docs: https://doc.rust-lang.org/std/ops/struct.RangeInclusive.html#method.contains

yuvipanda:
Very helpful @conradludgate! I've updated the PR

conradludgate:
Nice one! Looks good to me, will review properly in the morning

src/command/search.rs
```diff
@@ -169,6 +192,16 @@ async fn key_handler(
                     .map_or(app.input.clone(), |h| h.command.clone()),
             );
         }
+        Key::Alt(c) if ('1'..='9').contains(&c) => {
```
conradludgate, line 195: I've thought of an even better way of doing this :D

```suggestion
        Key::Alt('1'..='9') => {
```

Rust pattern matching is so powerful

conradludgate: If you want to read up more about patterns, see https://doc.rust-lang.org/book/ch18-03-pattern-syntax.html

yuvipanda: @conradludgate if I use the range in the pattern, how do I get access to `c`?

conradludgate: Hmm. Good point. I think maybe

`Key::Alt(c @ '0'..='9')`

ellie: While this is cool, I think for the sakes of clarity the current approach might be best. Contributors new to Rust (and many not new to Rust) might find it confusing or unclear

bl-ue: [(Rust is crazy! 🤣)](https://doc.rust-lang.org/book/ch18-03-pattern-syntax.html#-bindings)

conradludgate: For now, ignore this. I think I'll go in for a round 2 cleaning up some code later. I also want to start working on a minimum supported rust version and getting that enforced

ellie:
Indeed, I can get one of those sorted once the release is out! Unless anyone else wants to ;P

If you'd rather not have this feature then we could add an opt-out? I'd prefer to have it be the default though 😄

PR #80, opened 2021-05-09
Use an enum for dialect

yuvipanda:


atuin-client/src/settings.rs
```diff
@@ -19,9 +19,19 @@ pub enum SearchMode {
...
 
+// FIXME: Can use upstream Dialect enum if https://github.com/stevedonovan/chrono-english/pull/16 is merged
+#[derive(Clone, Debug, Deserialize, Copy)]
+pub enum Dialect{
+    #[serde(rename = "us")]
+    Us,
+
+    #[serde(rename = "uk")]
+    Uk,
+}
```
conradludgate, line 30: I think it would be better to implement the Into<Dialect> for this type https://doc.rust-lang.org/std/convert/trait.Into.html

src/command/stats.rs
```diff
@@ -84,9 +84,9 @@ impl Cmd {
                     words.join(" ")
                 };
 
-                let start = match settings.dialect.to_lowercase().as_str() {
-                    "uk" => parse_date_string(&words, Local::now(), Dialect::Uk)?,
-                    _ => parse_date_string(&words, Local::now(), Dialect::Us)?,
+                let start = match settings.dialect {
+                    atuin_client::settings::Dialect::Uk => parse_date_string(&words, Local::now(), Dialect::Uk)?,
+                    atuin_client::settings::Dialect::Us => parse_date_string(&words, Local::now(), Dialect::Us)?,
```
conradludgate, line 89: This could then become
```suggestion
                let start = parse_date_string(&words, Local::now(), settings.dialect.into())?;
```

yuvipanda:
Wow, this way of converting types seems pretty damn good, @conradludgate! I've updated the PR to do that

yuvipanda:
Clippy tells me to implement `From` rather than `Into`, and this seems to work. I'm going to figure how this works yet, but this seems to be the right thing.

conradludgate (approved):
> Clippy tells me to implement `From` rather than `Into`, and this seems to work. I'm going to figure how this works yet, but this seems to be the right thing.

Yeah, `From` and `Into` are essentially the same, just reversed. But `Into` is generated automatically from a `From`: https://doc.rust-lang.org/std/convert/trait.Into.html#impl-Into%3CU%3E

ellie:
Thanks @yuvipanda! This is much nicer 😌

yuvipanda:
Yay :)

Issue #64, opened 2021-05-08
Use shellcheck on all shell scripts


ellie:
And also add a CI step that uses it!

There's been a couple of install script issues that could have been totally avoided. 

I'd also like to make sure the init scripts are correct

PR #81, opened 2021-05-09
Run shellcheck on bash file

yuvipanda:
Ref #64

### DIFF
--- a/src/shell/atuin.bash
+++ b/src/shell/atuin.bash
@@ -9,14 +9,14 @@ _atuin_precmd() {
 	[[ -z "${ATUIN_HISTORY_ID}" ]] && return
 
 
-	(RUST_LOG=error atuin history end $ATUIN_HISTORY_ID --exit $EXIT &) > /dev/null 2>&1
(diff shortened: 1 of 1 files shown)


yuvipanda:
shellcheck doesn't support zsh unfortunately

ellie:
> shellcheck doesn't support zsh unfortunately

that's a shame! 

Thank you for this one :rocket:

ellie (approved):

ismith:
Shellcheck claims to, or at least https://github.com/ludeeus/action-shellcheck includes it in their list of suffixes - I'll take a pass at it.

ismith:
... I stand corrected :sadpanda:

```
In src/shell/atuin.zsh line 1:
#!/bin/zsh
^-- SC1071: ShellCheck only supports sh/bash/dash/ksh scripts. Sorry!
```

(Shellcheck will run on the file if it lacks a shellbang, but ... I guess it's assuming it's bash? So I'm going to exclude it from files checked by the shellcheck action.)

PR #71, opened 2021-05-08
Generic Importer Trait

conradludgate:
Creates a new trait
```rust
pub trait Importer: IntoIterator<Item = Result<History>> + Sized {
    const NAME: &'static str;
    fn histpath() -> Result<PathBuf>;
    fn parse(path: impl AsRef<Path>) -> Result<Self>;
}
```
and a function `import` which will use the provided `Importer` type to perform a generic import.

It will:
1. Check for a `$HISTFILE` or call `I::histpath()?`
2. Check that the history file is a file (#66)
3. Parse the history file
4. Iterate over the parsed contents, saving them to the DB in batches

TODO: Tests :)

PR #82, opened 2021-05-09
Fix resh importer, and change error handling method

Sciencentistguy:
When I wrote this importer (#69) I failed to notice that one of my last changes broke it. This fixes that.


bl-ue:
@Sciencentistguy you might to need to update #71 somehow 🤔

Sciencentistguy:
In my opinion its worth merging this now, and then I'll update it to use the generic trait when that is merged.

conradludgate:
You can merge this in before, I'll update for my PR later. I still have more testing to do before hand

conradludgate (approved):

PR #91, opened 2021-05-09
Add logout, only login if not already logged in

### LINKED ISSUE #85: Key creation order
I found the key creation order to be a little confusing.  It isn't created when you register, but when you sync the first time.  So I tried to register, then login, and I couldn't find where the key was that I needed to use.

### LINKED ISSUE #86: Have `login` check for current session
It would be nice if login would print that "You're already logged in with a valid session token" if already logged in.  Maybe either have a command or option to log out, or a `[Y/n]` question to log out if logged in.

ellie:
Also:
- Ensures that a key is generated as soon as a user registers!
- Ensures that "atuin key" will generate a key if needed, and doesn't
  double base64 encode data

And a few other little fixes :)

Resolves #85
Resolves #86

### DIFF
--- a/atuin-client/src/encryption.rs
+++ b/atuin-client/src/encryption.rs
@@ -24,23 +24,30 @@ pub struct EncryptedHistory {
     pub nonce: secretbox::Nonce,
 }
 
+pub fn new_key(settings: &Settings) -> Result<secretbox::Key> {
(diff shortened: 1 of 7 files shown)


conradludgate (approved):

ClashTheBunny:
Rockstar!  Thanks!

PR #92, opened 2021-05-09
Fix resh importer crashing on end of file

Sciencentistguy:
This change was made in #82 , but was missed out when it was converted to use the generic trait (in #71).


ellie (approved):

ellie:
Thank you! :blush:

PR #94, opened 2021-05-09
Use format! instead of string concat

yuvipanda:
Cleaner


yuvipanda:
Just minor clean up as I try to learn more about the language itself :)

ellie (approved):
Thanks! Not entirely sure why I wrote that abomination in the first place :sweat_smile:

conradludgate:
Could also use https://doc.rust-lang.org/std/primitive.str.html#method.repeat instead of the while loop

PR #95, opened 2021-05-09
Fix sync

ellie:
The data part of the add history request is actually a string. I don't
want to introduce any structure here, and would rather keep it as "just
a blob". Even if that blob has structure secretly!

My fault for missing this in the last review

### DIFF
--- a/atuin-client/src/sync.rs
+++ b/atuin-client/src/sync.rs
@@ -107,6 +107,8 @@ async fn sync_upload(
 
         for i in last {
             let data = encrypt(&i, &key)?;
+            let data = serde_json::to_string(&data)?;
+
             let add_hist = AddHistoryRequest {
                 id: i.id.into(),
                 timestamp: i.timestamp,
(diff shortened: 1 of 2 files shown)


atuin-client/src/sync.rs
```diff
@@ -107,6 +107,8 @@ async fn sync_upload(
 
         for i in last {
             let data = encrypt(&i, &key)?;
+            let data = serde_json::to_string(&data)?;
```
bl-ue, line 110: `let data` twice??!

ellie: Yup! Imo it's cleaner to read that way, and is something Rust allows you to do :)

ellie:
Thought it worth getting this merged quickly to fix `main`

conradludgate:
Ah oops. Might as well get rid of the type param then too. Thought I was being clever but of course json wouldn't be a valid string :facepalm:

ellie:
I think we could keep it around for now, it might be useful in the future in case what we store changes :thinking:

Issue #98, opened 2021-05-10
zsh multiline import broken


conradludgate:
I entered `echo \<NEWLINE>bar` into my zsh terminal, and the resulting `~/.zsh_history` entry contained the lines

```
: 1620636746:0;echo \\
bar
```

With two `\` tokens. The current import isn't taking this into account. The command is saved with both `\\` characters and is interpreted differently.

This possibly impacts bash import too.

Note, any new entries don't have this issue. This is only for import

dbaynard:
I've been bitten by this, and worked around it with a hacky solution. It replaces every slash followed by a newline with _just_ the newline. It is not idempotent, and will cause damage if run more than once.

```zsh
sqlite3 --box ~/.local/share/atuin/history.db <<- 'SQL'
begin;
update history
   set command = replace(command, '\' || X'0A', X'0A')
 where command like '%\' || X'0A' || '%'
;
select changes();
rollback;
SQL
```

Change the `rollback` to `commit` to make the changes stick. I then had to reload zsh.

PR #100, opened 2021-05-10
fix zsh import

conradludgate:
See #98

TBD:
* Does this bug impact only extended history or all zsh history?
* Does this bug impact bash too?

### DIFF
--- a/atuin-client/src/import/zsh.rs
+++ b/atuin-client/src/import/zsh.rs
@@ -84,7 +84,10 @@ impl<R: Read> Iterator for Zsh<R> {
 
         self.loc -= 1;
 
-        while self.strbuf.ends_with("\\\n") {
+        while self.strbuf.ends_with("\\\\\n") {
+            let range = (self.strbuf.len() - 3)..;
+            self.strbuf.replace_range(range, "\\\n");
+
             if self.file.read_line(&mut self.strbuf).is_err() {
                 // There's a chance that the last line of a command has invalid
                 // characters, the only safe thing to do is break :/
@@ -196,7 +199,7 @@ mod test {
     #[test]
     fn test_parse_file() {
         let input = r": 1613322469:0;cargo install atuin
-: 1613322469:10;cargo install atuin; \
+: 1613322469:10;cargo install atuin; \\
 cargo update
 : 1613322469:10;cargo :b̷i̶t̴r̵o̴t̴ ̵i̷s̴ ̷r̶e̵a̸l̷
 ";


ellie (approved):
Unix is lovely but sometimes I really hate plaintext things 😭

conradludgate:
Oh dear - this goes deeper...

`echo \\` is a valid 1 line command and will output a single `\` character (escaping the `\`). Unfortunately... the entry in my history file looks like so:

```
: 1620639537:0;echo \\
```

Which, by the current logic, makes it think there's another line for this command. The next line of course starts with a `:` to start a new command :sob:

conradludgate:
I'll try figure out exactly how the non-extended history handles newlines. This might be an extended only problem. Best solution would probably be to keep parsing lines until it starts with a `: `?

ellie:
> Best solution would probably be to keep parsing lines until it starts with a : ?

Or maybe just follow the current logic and break if the next line starts with a `:` in addition?

conradludgate:
Ok. This only impacts extended history. Default history just ignores any \ characters.

As for the extended history. I think splitting entries based on the line ending with `\\` and the next one starting with a `:` is the best we can do. It's possible to insert those yourself but I can't see it being likely.

For instance, I put in
```
$ echo "foo" \
> : 1620662700:0;echo "bar"
```
and the history file contains
```
: 1620662639:0;echo "foo" \\
: 1620662700:0;echo "bar"
```
which makes me very sad

ellie:
> which makes me very sad

me too :sob: 

sounds like the best way ahead in any case, I don't think there's much else we can do

watiko:
FYI(I don't know if this is the right place to put this)

Recently, zsh has [modified](https://github.com/zsh-users/zsh/commit/78958c08bfdb37d2eafaf14a33b93229b1fa9e31) the way writes history, so in the future the contents of the file should not be ambiguous by line feeds. Also, zsh's own parsing method has not been changed at that time, so it is backward compatible.

In my opinion, accurately parsing the contents of a history file needs to be done the same way as possible with zsh.

First, there is the metafy problem. If you use emoji or other characters that are not in the ascii range, the contents of the file will not be valid UTF-8.

Secondly, there is the problem of line feeds.  You need to be able to tell whether a line feed is an entry separator or a line continuation.

These are things I read from the zsh source code when I was recently writing code about zsh history files.

- https://github.com/watiko/zsh-history-utils/blob/master/src/zsh/history.rs
- https://github.com/watiko/zsh-history-utils-deno/blob/master/lib/zsh.ts

The deno code is written in a way that is similar to the zsh code, and some of the rust code is not (such as where it uses nom).

PR #101, opened 2021-05-10
Increase number of optimizations on release builds.

thedrow:
This allows us to produce a faster executable (which is useful in shells) at the expense of rarely spending more compilation time upon releasing a new version.
I do this in all of my rust projects, perhaps because they are too small for compilation time to really matter. We should measure next time we release and see how that goes.


ellie (approved):
I totally agree - we should build releases as fully optimised as possible! 

Thanks 🚀

PR #103, opened 2021-05-10
Release v0.7.0

ellie:
- Update all the crate versions
- Update the demo gif
- Write a changelog
- Adjust the title of the search screen (has the old name still)
- Adjust the colours of the quick-jump numbers (sadly invisible on some
  colour schemes as dark grey :/)


bl-ue:
@ellie maybe consider enabling GitHub discussions on this repo?  GitHub lets you create a discussion dedicated to a particular release, and displays the reactions on the release's page ([example](https://github.com/microsoft/winget-cli/releases/tag/v-0.3.11102-preview)).

ellie:
Considered it, but didn't realise it had discussions for releases! That's neat.

I've enabled it :)

bl-ue:
Yep, there's a checkbox on the page where you create the release:

Screen Shot 2021-04-25 at 8 48 08 AM

ellie:
Ok perfect, looks like the plugin we're using to create releases doesn't support also creating discussions :thinking: 

I'll make the release manually and edit the workflow so it's also triggered on new releases. I want to move away from having CI build the binaries in this way anyway - I'd rather have them all pre-built ready to create the release, rather than have them filter in one-by-one as they're built.

ellie:
Haha, I think they're cool even if just for reactions on releases :) Doesn't hurt to have them!

PR #107, opened 2021-05-10
Re-add macro_use to atuin-common

ellie:
When build as a dependency, the macro is available from another crate.
When you try to build common by itself, the macro is not found. Magic,
huh?


conradludgate (dismissed):

ellie:
> `"/home/user/custom_data/atuin"`,

is github doing something weird?

ellie:
:tada:

conradludgate (approved):

PR #134, opened 2021-05-17
Vendor dependencies

ellie:
Just testing how CI works with this. I tend to prefer vendoring, as it
means that if you have a copy of the code *you can always build it*.
Even if you're 20 years in the future

This is the output of

```
cargo vendor --versioned-dirs
```


conradludgate:
+3 million, no biggie

ellie:
Just give it a skim, doesn't need a thorough review ;)

In seriousness I have mixed feelings on this, might leave it open while I have a think 🤔

bl-ue:
"if you have a copy of the code you can always build it" — that _is_ great but @ellie going from 37 MB → 296 MB & 78 files → 14,884 files seems like a bad idea... 😬

ellie:
Yuuuuup I'm less keen on that :( 

It's much less than that to download + obviously it's 300ish once unpacked, but... I guess I'll avoid this for now :/

conradludgate:
Maybe a vendor repo that automatically syncs with latest versions yada yada submodule yada yada

PR #142, opened 2021-05-31
Add fuzzy text search mode

frankh:
Adds a new search mode "fuzzy" which matches as long as the characters in the query exist somewhere in the command in the same order. This is how e.g. SublimeText cmd+P search works

I'm a little worried about performance with large database files. With a ~1000 item database locally it runs no problem

### DIFF
--- a/atuin-client/Cargo.toml
+++ b/atuin-client/Cargo.toml
@@ -17,7 +17,6 @@ log = "0.4"
 fern = {version = "0.6.0", features = ["colored"] }
 chrono = { version = "0.4", features = ["serde"] }
 eyre = "0.6"
-tokio-test = "*"
 directories = "3"
 uuid = { version = "0.8", features = ["v4"] }
 indicatif = "0.16.0"
@@ -41,3 +40,6 @@ humantime = "2.1.0"
 itertools = "0.10.0"
 shellexpand = "2"
 sqlx = { version = "0.5", features = [ "runtime-tokio-rustls", "uuid", "chrono", "sqlite" ] }
+
+[dev-dependencies]
+tokio-test = "*"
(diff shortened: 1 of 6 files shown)


atuin-client/Cargo.toml
```diff
@@ -17,6 +17,7 @@ log = "0.4"
 fern = {version = "0.6.0", features = ["colored"] }
 chrono = { version = "0.4", features = ["serde"] }
 eyre = "0.6"
+tokio-test = "*"
```
conradludgate, line 20: I think this belongs in `[dev-depenencies]`

atuin-client/src/database.rs
```diff
@@ -286,6 +286,7 @@ impl Database for Sqlite {
         let query = match search_mode {
             SearchMode::Prefix => query,
             SearchMode::FullText => format!("%{}", query),
+            SearchMode::Fuzzy => query.split("").collect::<Vec<&str>>().join("%"),
```
conradludgate, line 289: The itertools crate (which is already added to this project) provides a `join` method to iterators, without needing to collect them into a vec first: https://docs.rs/itertools/0.10.0/itertools/trait.Itertools.html#method.join

conradludgate (commented):
Nice one! Some small changes but it looks good

atuin-client/src/database.rs
conradludgate: An alternative to using `.split("")` too, is to just use `.chars()`

frankh: thanks for the tip with itertools, that's much nicer

split("") actually isn't equivalent to chars() as it adds an empty string at the beginning and end, which we need

conradludgate: Very interesting but also quite unintuitive 🤔 oh well

conradludgate (approved):

ellie:
Thanks @frankh! 🎉

PR #171, opened 2021-08-15
Fix typos in the docs.

mundry:
This PR fixes some typos in the docs.


ellie (approved):

ellie:
I think clippy was only failing here because it was a bit broken in the past, it's only a README change so is fine

Thank you for the contribution here :)

PR #181, opened 2021-09-10
use sqlite grouping rather than subquery

mwotton:
you probably don't want to keep the actual test in, but it was quite significantly faster on my machine.


ellie:
Oh that's fantastic! Thank you for your work here :sparkling_heart: 

Would it be possible to get this rebased?

mwotton:
all rebased, hopefully goes through ok.

conradludgate (approved):

conradludgate:
Oh yeah this is a lot faster

PR #182, opened 2021-09-13
feat: nushell import

conradludgate:
- [x] import
- [ ] search
- [ ] history hook

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -144,6 +144,7 @@ dependencies = [
  "reqwest",
  "rmp-serde",
  "rust-crypto",
+ "rustyline",
  "serde 1.0.126",
  "serde_derive",
  "serde_json",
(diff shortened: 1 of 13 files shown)


conradludgate:
Closing this for now since we are missing hook support from nushell to provide the key service that atuin provides. Can re-open in future but it would probably be better to start a fresh at that time

maghsk:
Any progress on this issue?     XD
Nushell team has just released their 0.6.1!

conradludgate:
As far as I can tell, nushell still doesn't support hooks that we need to provide custom history support https://github.com/nushell/nushell/issues/3872

mcheshkov:
I am not an nushell expert, but it seems to me that `pre_execution` hook should be enough for basic integration
https://www.nushell.sh/book/hooks.html

Issue #175, opened 2021-08-26
add a check if sudo is installed


prbond:
Hello,
On debian host, i would like to install atuin but i have an error : 
/dev/fd/63: ligne 59: sudo : commande introuvable

Could you add a check if sudo is installed please ?

I don't use sudo

Pleased to read you

PR #188, opened 2021-09-25
fix: resolve some issues with install.sh

mozzieongit:
Some people don't use sudo or have curl, sed or wget installed by default.
This adds a check, if sudo, curl and sed are installed and changes wget
to curl as this is the mostly used command in the script. If sudo is not
installed it uses su.

The printf was missing a newline at the end, which resulted in the eval
line being appended directly to the end of the previous line. This is 
similar to the problem from  #56.

Closes: #175

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -162,10 +162,10 @@ esac
 # TODO: Check which shell is in use
 # Use of single quotes around $() is intentional here
 # shellcheck disable=SC2016
-printf '\neval "$(atuin init zsh)"' >> ~/.zshrc
+printf '\neval "$(atuin init zsh)"\n' >> ~/.zshrc
 
 curl https://raw.githubusercontent.com/rcaloras/bash-preexec/master/bash-preexec.sh -o ~/.bash-preexec.sh
-printf '\n[[ -f ~/.bash-preexec.sh ]] && source ~/.bash-preexec.sh' >> ~/.bashrc
+printf '\n[[ -f ~/.bash-preexec.sh ]] && source ~/.bash-preexec.sh\n' >> ~/.bashrc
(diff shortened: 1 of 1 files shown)


mozzieongit:
I should probably wait for the dependabot PRs to be merged.

mozzieongit:
Why does `cargo test -p atuin-common --lib` suddenly fail? It failed on my machine, too. 

But now it works again. I just re-ran it. 
Weird.

ellie:
Could you share the logs from the failures please? Maybe some of the env var stuff is being non deterministic

mozzieongit:
```
   Finished test [unoptimized + debuginfo] target(s) in 0.07s
     Running unittests (target/debug/deps/atuin_common-284a7ca8554b08af)

running 4 tests
test utils::tests::test_config_dir ... ok
test utils::tests::test_config_dir_xdg ... ok
test utils::tests::test_data_dir_xdg ... ok
test utils::tests::test_data_dir ... FAILED

failures:

---- utils::tests::test_data_dir stdout ----
thread 'utils::tests::test_data_dir' panicked at '$HOME not found: NotPresent', atuin-common/src/utils.rs:37:38


failures:
    utils::tests::test_data_dir

test result: FAILED. 3 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

error: test failed, to rerun pass '-p atuin-common --lib'
```

mozzieongit:
Yay. The test finished successfully.

ellie (approved):
Fantastic, thank you for your work here! :sparkling_heart:

prbond:
Thanks

PR #202, opened 2021-11-16
Gate the various functionality using features

lespea:
Instead of forcing the user(s) to build a binary that contains both the
client and the server, we can use features to only compile what is
needed.

Furthermore, not all users want the "syncing" functionality in their
client and this allows the user to disable that functionlity from even
being compiled and gives more piece of mind it will never attempt to
make external calls.

The dependency graph shrinks dramatically as well when only building the
client without "sync".

For now, the defaults still include all of the features to retain
existing functionality by default.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /target
 */target
 .env
+/.idea/
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,20 +4,20 @@ version = 3
 
 [[package]]
 name = "ahash"
-version = "0.7.2"
+version = "0.7.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7f200cbb1e856866d9eade941cf3aa0c5d7dd36f74311c4273b494f4ef036957"
+checksum = "fcb51a0695d8f838b1ee009b3fbf66bda078cd64590202a864a8f3e8c4315c47"
 dependencies = [
- "getrandom 0.2.2",
+ "getrandom",
  "once_cell",
  "version_check",
 ]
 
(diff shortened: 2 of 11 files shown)


conradludgate:
The idea is good but it needs some more thought.

First. I think those features need to be enabled by default. It would be confusing otherwise to existing users and documentation if their install suddenly stopped syncing because they didn't have a feature.

Second. I don't like the cfg attrs being sprinkled everywhere. Instead, I think the code needs to be refactored slightly to have a `sync` library. If the cfg is enabled, that library would function as normal, of not, all the functions will be noops. That way the sync stuff is contained (maybe in its own crate) and the core atuin code doesn't need to think about it

We should also consider what other features we could have. For instance, the server code doesn't always need to exist

ellie:
> I don't like the cfg attrs being sprinkled everywhere

I'm not so keen on this one either, and the fact that this increases the different configurations of builds we would need to test and maintain. 

> Instead of forcing the user(s) to build a binary

If it were necessary for everyone to always build Atuin from source then I might agree with you, however we do provide binary builds that remove that concern. No users are being forced to build anything :)

lespea:
> If it were necessary for everyone to always build Atuin from source then I might agree with you, however we do provide binary builds that remove that concern. No users are being forced to build anything :)

I wasn't saying people were forced to build the binary, I was saying right now the binary has the functionality of the server, client, and external syncing always build into the binary, when some users (myself obviously), not only don't need that, but explicitly don't even want it available.

If you don't want the "complexity" (really isn't complex imo) that's fine I'll just maintain my own fork for myself then.

PR #213, opened 2021-11-21
chore: improve build times

conradludgate:
This simple change ended up improving the install time for me from `2:30s` to `20s`. Warp is very type heavy, and super complex types cause the compiler to be bottlenecked. When profiling, I found that rustc spent almost minute alone on monomorphisation.

You'd think that since this code is in the server crate, that you could just build that once and it would improve the build times in the core atuin bin, but that's sadly not the case.

Since this optimisation helps more in multithreaded workflows, github actions does not take full advantage of this since it uses single threaded VMs so the difference isn't noticeable there

Note:
It's worth noting that this adds a little bit of overhead to the server routes due to dynamic dispatch. But it's probably negligible

### DIFF
--- a/atuin-server/src/router.rs
+++ b/atuin-server/src/router.rs
@@ -66,7 +66,8 @@ pub async fn router(
         .and(warp::path::end())
         .and(with_user(postgres.clone()))
         .and(with_db(postgres.clone()))
-        .and_then(handlers::history::count);
+        .and_then(handlers::history::count)
+        .boxed();
 
     let sync = warp::get()
         .and(warp::path("sync"))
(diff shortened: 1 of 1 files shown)


conradludgate:
If you're interested in the number: https://gist.github.com/conradludgate/70f9d909fb05781fc57c4786660a58d2

(ctrl-f for `monomorphization_collector_graph_walk`, all tests were performed by prior running `cargo clean`)

ellie:
Oh sweet! Personally I'd prefer to avoid the heap allocation, buuut if it means we reduce the build time by that much then it's worth it imo. Thanks! :sparkling_heart:

ellie (approved):

PR #229, opened 2021-12-09
Update install.sh to use `pacman` on Arch Linux

orhun:
This is a follow-up PR for #228. It updates `install.sh` script to use `pacman` for installing `atuin` on Arch Linux, since the package is moved to the community repository.


ellie (approved):

ellie:
Thank you for opening both of these! And for adding atuin to community ☺️

PR #234, opened 2021-12-10
Fish importing

### LINKED ISSUE #57: Fish support
Will be great if fish shell can be supported as well

conradludgate:
Closes #57

So, Fish YAML is not real YAML. So there's no point bringing out a YAML parser for this. Instead, we read line by line.

We use `- ` at the beginning to determine whether the line is a new entry. If we find a new entry line and we've found a `cmd: ...` value, we exit the loop and return our history item.

It's pretty fault tolerant, ignoring bad lines and missing values.

### DIFF
--- a/atuin-client/src/import/fish.rs
+++ b/atuin-client/src/import/fish.rs
@@ -1,0 +1,129 @@
+// import old shell history!
+// automatically hoover up all that we can find
+
+use std::{
+    fs::File,
+    io::{self, BufRead, BufReader, Read, Seek},
+    path::{Path, PathBuf},
+};
+
+use chrono::prelude::*;
+use chrono::Utc;
+use directories::UserDirs;
+use eyre::{eyre, Result};
+use itertools::Itertools;
+
+use super::{count_lines, Importer};
+use crate::history::History;
+
+#[derive(Debug)]
+pub struct Fish<R> {
+    file: BufReader<R>,
+    strbuf: String,
+    loc: usize,
+}
+
+impl<R: Read + Seek> Fish<R> {
+    fn new(r: R) -> Result<Self> {
+        let mut buf = BufReader::new(r);
+        let loc = count_lines(&mut buf)?;
+
+        Ok(Self {
+            file: buf,
+            strbuf: String::new(),
+            loc,
+        })
+    }
+
+    fn new_entry(&mut self) -> io::Result<bool> {
+        let inner = self.file.fill_buf()?;
+        Ok(inner.starts_with(b"- "))
+    }
+}
+
+impl Importer for Fish<File> {
+    const NAME: &'static str = "fish";
+
+    fn histpath() -> Result<PathBuf> {
+        let user_dirs = UserDirs::new().unwrap();
+        let home_dir = user_dirs.home_dir();
+
+        let histpath = home_dir.join(".local/share/fish/fish_history");
+        if histpath.exists() {
+            Ok(histpath)
+        } else {
+            Err(eyre!("Could not find history file. Try setting $HISTFILE"))
+        }
+    }
+
+    fn parse(path: impl AsRef<Path>) -> Result<Self> {
+        Self::new(File::open(path)?)
+    }
+}
+
+impl<R: Read> Iterator for Fish<R> {
+    type Item = Result<History>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        let mut time: Option<DateTime<Utc>> = None;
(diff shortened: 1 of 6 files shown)


atuin-client/src/import/fish.rs
```diff
@@ -0,0 +1,209 @@
...
+}
+
+impl Importer for Fish<File> {
+    const NAME: &'static str = "fish";
+
+    fn histpath() -> Result<PathBuf> {
+        let user_dirs = UserDirs::new().unwrap();
+        let home_dir = user_dirs.home_dir();
+
+        let histpath = home_dir.join(".local/share/fish/fish_history");
```
conradludgate, line 52: There did used to be an alternative path https://github.com/fish-shell/fish-shell/issues/862#issuecomment-19138822 but that's very old (at least pre 2013). It might also have some parsing quirks, so I'm going to forget it exists.

It's worth checking to find out if this is the same on macos too.

panekj: https://fishshell.com/docs/current/interactive.html#searchable-command-history

panekj:
IMHO, keybinds should not a default setting

src/shell/atuin.fish
```diff
@@ -0,0 +1,26 @@
+
+export ATUIN_SESSION=(atuin uuid)
+export ATUIN_HISTORY=(atuin history list)
```
panekj, line 3:
```suggestion
set -Gx ATUIN_SESSION (atuin uuid)
set -Gx ATUIN_HISTORY (atuin history list)
```

conradludgate:
> IMHO, keybinds should not a default setting

This is currently the standard behaviour of all of the shell integrations, I don't think we should handle fish any differently

src/shell/atuin.fish
conradludgate: Good spot, thanks!

src/shell/atuin.fish
```diff
@@ -0,0 +1,26 @@
...
+    set h (RUST_LOG=error atuin search -i (commandline -b) 3>&1 1>&2 2>&3)
+    commandline -f repaint
+    if test -n "$h"
+        commandline -r $h
+    end
+end
+
+bind -k up '_atuin_search'
+bind \eOA '_atuin_search'
+bind \e\[A '_atuin_search'
```
panekj, line 26: I looked at `bash` and `zsh` inits and they check for `$ATUIN_NOBIND`
```suggestion
if test -z $ATUIN_NOBIND
	  bind -k up '_atuin_search'
	  bind \eOA '_atuin_search'
	  bind \e\[A '_atuin_search'
end
```

src/shell/atuin.fish
```diff
@@ -0,0 +1,26 @@
+
+set -Gx ATUIN_SESSION (atuin uuid)
+set -Gx ATUIN_HISTORY (atuin history list)
```
panekj, line 3: sorry, should be `-gx`, I corrected casing but you already committed my typo

```suggestion
set -gx ATUIN_SESSION (atuin uuid)
set -gx ATUIN_HISTORY (atuin history list)
```

src/shell/atuin.fish
conradludgate: Ahh right yeah, I missed that (I need more sleep). Thanks for pairing with me :sweat_smile:

src/shell/atuin.fish
```diff
@@ -0,0 +1,28 @@
...
+    commandline -f repaint
+    if test -n "$h"
+        commandline -r $h
+    end
+end
+
+if test -z $ATUIN_NOBIND
+	  bind -k up '_atuin_search'
+	  bind \eOA '_atuin_search'
+	  bind \e\[A '_atuin_search'
```
panekj, line 27: Fix whitespaces to conform to the rest of the file
```suggestion
    bind -k up '_atuin_search'
    bind \eOA '_atuin_search'
    bind \e\[A '_atuin_search'
```

panekj:
Also missing `fish` in here
https://github.com/ellie/atuin/blob/28f78ba4e1874f6ab2eee2b833fb1b12d50e3ebe/README.md#L63-L67

ellie (approved):
Amaaazing, thanks @conradludgate!

And thanks @panekj for your help too 😊 Really happy to get this issue closed off, I'll release it soon

davinkevin:
So happy to see it ❤️, Thank you @conradludgate. 

I can't wait to see the next release !

PR #235, opened 2021-12-10
Support generating shell completions

### LINKED ISSUE #232: Generate shell completions
Hey!

Is there a way to generate shell completions for `atuin`? I saw this line in the release workflow but I'm not sure if it is related:

https://github.com/ellie/atuin/blob/d3059af815130f102dd97cb1d1e5030920754105/.github/workflows/release.yaml#L119

(Also, the contents of this directory seems to be empty.)

If it doesn't exist yet, do you have any plans to work on this? e.g. using [clap::App::gen_completions](https://docs.rs/clap/2.34.0/clap/struct.App.html#method.gen_completions)

I'd be happy to submit a PR about this 🐻

orhun:
closes #232


conradludgate:
Quick question, who's expected to run these shell completions and where should they be saved to? Obviously through pacman will set them up automatically I presume but what about installing directly through cargo?

orhun:
> Quick question, who's expected to run these shell completions and where should they be saved to? Obviously through pacman will set them up automatically I presume but what about installing directly through cargo?

This is mainly for distro packages, for example, you can install shell completions for Atuin on Arch Linux like this:

```sh
install -Dm 644 "completions/atuin.bash" "/usr/share/bash-completion/completions/atuin"
install -Dm 644 "completions/atuin.fish" -t "/usr/share/fish/vendor_completions.d"
install -Dm 644 "completions/_atuin" -t "/usr/share/zsh/site-functions"
```

More info: https://wiki.archlinux.org/title/Shell_package_guidelines#Shell_completions

Since they have to be placed into an appropriate location based on the current platform, I don't think they can be installed directly through `cargo install`.

conradludgate:
I'm happy with this. Considering whether we should move to running in a build script but overall I have no real preference either way.

Thanks for your contributions, I'll wait for @ellie to take a look

ellie (approved):
Fantastic, thank you! I'm happy with this. I agree with @orhun, it's up to the distro/end user to handle sorting the installation of the completion files, I think it's out of scope for something we should be thinking about

SuperSandro2000:
How can I output the completion to stdout?

conradludgate:
> How can I output the completion to stdout?

You could generate them into a local or a temp dir, then `cat` the contents

PR #242, opened 2021-12-16
remove unused environment var loading entire history into an env var

### LINKED ISSUE #241: After importing my `fish_history` I get an error about argument list being too long
After running `atuin import fish`, when I start a new shell, `atuin init fish | source` now gives the following error:

```
Failed to execute process '/opt/homebrew/bin/starship'. Reason:
The total size of the argument and environment lists 1.5MB exceeds the operating system limit of 1MB.
Try running the command again with fewer arguments.
Failed to execute process '/opt/homebrew/bin/starship'. Reason:
The total size of the argument and environment lists 1.5MB exceeds the operating system limit of 1MB.
Try running the command again with fewer arguments.
```

mrjones2014:
It seems that `ATUIN_HISTORY` environment variable isn't used anywhere, and defining it loads your entire history into an environment variable. If you history gets too big, you'll get an error that your argument list is larger than the OS limit. On macOS this is 1MB.

Fixes #241

### DIFF
--- a/src/shell/atuin.fish
+++ b/src/shell/atuin.fish
@@ -1,6 +1,4 @@
-
 set -gx ATUIN_SESSION (atuin uuid)
-set -gx ATUIN_HISTORY (atuin history list)
 
 function _atuin_preexec --on-event fish_preexec
     set -gx ATUIN_HISTORY_ID (atuin history start "$argv[1]")
@@ -9,7 +7,8 @@ end
 function _atuin_postexec --on-event fish_postexec
     set s $status
     if test -n "$ATUIN_HISTORY_ID"
-        RUST_LOG=error atuin history end $ATUIN_HISTORY_ID --exit $s &; disown
+        RUST_LOG=error atuin history end $ATUIN_HISTORY_ID --exit $s &
+        disown
     end
 end
 
@@ -22,7 +21,7 @@ function _atuin_search
 end
 
 if test -z $ATUIN_NOBIND
-    bind -k up '_atuin_search'
-    bind \eOA '_atuin_search'
-    bind \e\[A '_atuin_search'
+    bind -k up _atuin_search
+    bind \eOA _atuin_search
+    bind \e\[A _atuin_search
 end


src/shell/atuin.fish
```diff
@@ -9,7 +7,8 @@ end
 function _atuin_postexec --on-event fish_postexec
     set s $status
     if test -n "$ATUIN_HISTORY_ID"
-        RUST_LOG=error atuin history end $ATUIN_HISTORY_ID --exit $s &; disown
+        RUST_LOG=error atuin history end $ATUIN_HISTORY_ID --exit $s &
+        disown
```
mrjones2014, line 11: these lines and 24-26 just got auto-formatted by `fish_indent` (a formatter for fish scripts). Let me know if you'd like to revert these lines.

conradludgate:
I think this looks ok. I indeed can't find the use for that variable.

I copied and pasted it from the ZSH file and I must have also mis-interpreted it too because its supposed to be the string `"atuin history list"` and not the actual output of the command. Not sure what that would be for though

mrjones2014:
Yeah I used `ripgrep` to search the entire codebase and that was the only instance of the string "ATUIN_HISTORY" being used anywhere. It also seems like loading your entire history into an environment variable would be something you definitely do not want to do 😅

conradludgate:
It's a relic from long ago: https://github.com/ellie/atuin/commit/41f072a8b443b5a404916598f0c9b7a52f5c05fb

Seems you used to be able to configure what the search prompt showed based on that env var. But no longer used anywhere

> Yeah I used ripgrep to search the entire codebase

Our config lib processes env vars for us automatically, so they would show as `config.get_str("history")`, or as a field in our `Settings` struct. Seems it's not their either though so I'd be happy with accepting this. Thanks again for looking into it, we don't use fish so it's nice to have someone test out code 😅

conradludgate (approved):

mrjones2014:
Happy to help!

PR #243, opened 2021-12-17
Add Alt+backspace and Ctrl+u keybinds for deleting by word and by line, respectively

### LINKED ISSUE #227: Add Alt+Backspace to delete word leftwards
This is a standard in shell and would be nice to have it in atuin to delete a whole word quickly and clear the filter.

mrjones2014:
Fixes #227 

These are both standard shell key binds, so we should be able to use them at the Atuin prompt as well.

### DIFF
--- a/src/command/search.rs
+++ b/src/command/search.rs
@@ -207,6 +207,23 @@ async fn key_handler(
             app.input.pop();
             query_results(app, search_mode, db).await.unwrap();
         }
+        // \u{7f} is escape sequence for backspace
+        Key::Alt('\u{7f}') => {
+            let words: Vec<&str> = app.input.split(' ').collect();
+            if words.is_empty() {
+                return None;
+            }
+            if words.len() == 1 {
+                app.input = String::from("");
+            } else {
+                app.input = words[0..(words.len() - 1)].join(" ");
+            }
+            query_results(app, search_mode, db).await.unwrap();
+        }
+        Key::Ctrl('u') => {
+            app.input = String::from("");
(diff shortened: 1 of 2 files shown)


ellie:
Before approving merging this, I'd like to consider using some sort of readline or other line editor. Or at least see if we can find one 

These are not standard shell bindings - they are eMacs bindings. Usually supported by the line editor your shell uses. They support vi input too :)

mrjones2014:
There's this we could look at: https://github.com/kkawakam/rustyline

mrjones2014:
Some alternatives found in the README of that project as well:

- https://github.com/murarth/linefeed
- https://github.com/redox-os/liner

conradludgate:
Seems a bit ironic loading a history-aware line editor into a history search window and then not using it's history feature 😅

ellie:
Yeah lol. I realised that they wouldn't work with our TUI stuff anyway, sooo 🤷‍♀️

ellie (approved):
Thanks for the work!! Would be cool to consider more eMacs and eventually vi bindings too

conradludgate:
I like the idea of having more keybindings, Perhaps it's worth seeing if we can modularize some of these existing line editors to have a shared library for keybindings

mrjones2014:
Yeah I'm not sure how we'd integrate rustyline with termion being an event-driven type thing. it seems like rustline wants to read the whole line at once.

conradludgate:
Yeah all these readline alternatives take control of stdin/stdout so they don't work in our TUI environment. They're more for implementing a `sqlite` style prompt

PR #248, opened 2021-12-19
Replace dpkg with apt

SuperSandro2000:
dpkg is a backend tool not supposed to be used by end users for installing debs.
Also apt installs potentials dependencies.


conradludgate (approved):
Looks good, thanks

PR #260, opened 2022-01-09
fix: get install.sh working on UbuntuWSL

elsbrock:
I was unable to install `atuin` on Ubuntu WSL. Even though detected properly by the script, it looks like the option was forgotten to be included in the distro selection.

In addition, `apt` refused to install the package from the tempfile unless it was carrying the `.deb` file extension. After adding that it worked. I assume it does not hurt to add it by default to the tempfile name.


ellie (approved):
Thank you for your work! 🎉 Really appreciate it, I've never used UbuntuWSL. 

Does Atuin otherwise perform OK?

PR #278, opened 2022-03-10
Fix `history list --cwd` errors

lmburns:
#### Before the fix

```bash
$ atuin history list --cwd

Error: error returned from database: near "/": syntax error

Caused by:
    near "/": syntax error

Location:
    atuin-client/src/database.rs:316:19
```

1. The query into the database was not quoting the path. This fixes that issue.
2. Unused dependencies are removed:

```
atuin
    - fork

atuin-client
    - fern
    - humantime
    - indicatif

atuin-common
    - eyre
    - rmp-serde

atuin-sever
    - chrono-english
    - directories
    - fern
    - fork
    - indicatif
    - parse_duration
    - rmp-serde
    - unicode-width
    - urlencoding
```

3. I fixed the `zsh` plugin by standardizing it according to `zdharma`. The reference can be found in the file. It results in the following:

```zsh
0="${${ZERO:-${0:#$ZSH_ARGZERO}}:-${(%):-%N}}"
0="${${(M)0:#/*}:-$PWD/$0}"
```

The `$commands` array is also loaded with `zmodload` in case the user doesn't have it loaded for whatever reason

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -83,7 +83,6 @@ dependencies = [
  "crossbeam-channel",
  "directories",
  "eyre",
- "fork",
  "humantime 2.1.0",
  "indicatif",
  "itertools",
@@ -112,9 +111,6 @@ dependencies = [
  "config",
  "directories",
  "eyre",
- "fern",
- "humantime 2.1.0",
- "indicatif",
  "itertools",
  "log",
  "minspan",
@@ -140,8 +136,6 @@ name = "atuin-common"
 version = "0.8.0"
 dependencies = [
  "chrono",
- "eyre",
- "rmp-serde",
  "rust-crypto",
  "serde 1.0.132",
  "serde_derive",
@@ -159,17 +153,10 @@ dependencies = [
  "atuin-common",
  "base64",
  "chrono",
- "chrono-english",
  "config",
- "directories",
  "eyre",
- "fern",
- "fork",
- "indicatif",
  "log",
- "parse_duration",
  "rand 0.8.4",
- "rmp-serde",
  "rust-crypto",
  "serde 1.0.132",
  "serde_derive",
@@ -177,8 +164,6 @@ dependencies = [
  "sodiumoxide",
  "sqlx",
  "tokio",
- "unicode-width",
- "urlencoding",
  "uuid",
  "warp",
  "whoami",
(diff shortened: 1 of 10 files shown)


conradludgate:
Do you mind splitting out the zsh plugin change into a separate PR? I'm gonna need some extra time to digest that.

conradludgate:
Hmm. Looking at the history queries, we really should move to using compile-time checked queries (or at least ones with proper parameter support).

I'll tentatively accept this for now though (once clippy is happy and when the zsh changes are moved out)

lmburns:
I'll take them out. Here is some documentation for you to checkout. The file you guy's have is very simple, it's two lines of code.

#### Documentation
[`zmodload`](https://zsh.sourceforge.io/Doc/Release/Shell-Builtin-Commands.html#index-zmodload)
[`zsh/parameter` module](https://zsh.sourceforge.io/Doc/Release/Zsh-Modules.html#The-zsh_002fparameter-Module)
[`zinit` repo](https://github.com/zdharma-continuum/zinit)
[`zsh` plugin standard](https://zdharma-continuum.github.io/zinit/wiki/zsh-plugin-standard/)
[GitHub search showing several other examples using this](https://github.com/search?q=ZERO+ZSH_ARGZERO+extension%3Azsh&type=)

lmburns:
Clippy lints should be fixed now. I don't know why they weren't already done because they were all in files that I did not modify at all. It must've been merged last time before completing the clippy lints.

 This should be able to be merged now.

conradludgate:
Thanks for this, just the formatting to go 😅.

```
cargo fmt --all
```

Should do the trick so you don't need to do it manually

lmburns:
Maybe a github-workflow could be used to prevent this from happening?

Something similar to [this](https://github.com/tversteeg/emplace/blob/master/.github/workflows/rust.yml) or [this](https://github.com/lmburns/wutag/blob/master/.github/workflows/release.yml). Or [`pre-commit`](https://github.com/pre-commit/pre-commit) could be used, but that would require everyone having this program.

 It would prevent a commit from every happening without passing these checks.

conradludgate:
Seems there's still formatting issues, do you have some custom formatting rules set up globally?

lmburns:
I don't have a `rustfmt.toml` in the directory, but I'm guessing that it may have read it from my home directory

conradludgate (approved):
@ellie, thoughts?

ellie:
LGTM! Thanks for the work here 💖

PR #279, opened 2022-03-17
Add SearchMode fzf.

pmarschik:
Add a new search mode "fzf" that tries to mimic the search syntax of
https://github.com/junegunn/fzf#search-syntax
This search mode splits the query into terms where each term is matched
individually. Terms can have operators like prefix, suffix, exact match
only and can be inverted. Additionally, smart-case matching is
performed: if a term contains a non-lowercase letter the match will be
case-sensitive.

I've also refactored the client database tests to be more readable.

Future improvements:
 * add setting that allows switching the case mode between: smart (implemented), ignore (always case-insensitive) and exact.
 * use https://github.com/lotabout/skim for parsing terms and generate query from there, this would allow braces to refine operator precedence.
 * implement re-ordering of fzf searches where each term contributes to a score (could use skim for scoring as well)


panekj:
You can use `atuin history list --cmd-only | fzf`

ellie:
Thanks for this! 🥳 

 I'll give the code a proper look later, but will try running your branch for a bit today.

pmarschik:
> You can use `atuin history list --cmd-only | fzf`

I'm aware that this is possible. But implementing the search mode at query time allows us to use the other filters like `--before`, `--after`, `--exit`, `--cwd` directly. Also it should be faster constraining the matches on database level than querying everything and then `fzf`ing.

panekj:
IMO, it's a bad idea to implement such features instead of improving interop with external software, as it increases maintenance, binary size and there will be implementation differences.

conradludgate:
I do also think that this extra complexity needs to be taken carefully. I have some (long overdue) clean ups of some of the sql code to remove the use of `format!` and this is going to add to that work too, since this makes even heavier use of that.

However, if this is a feature that's very desirable, then it does make sense to take on said complexity

ellie:
I think I'm happy for us to take on the complexity here. I'd suggest that we actually replace the current fuzzy implementation with this though, rather than adding a new setting called `fzf`. In my usage so far, it's nicer than what we have at the moment :)

I appreciate the concern there @panekj, but I'd rather focus on the use case of users using our TUI. Searching and then piping works _fine_, but it's not as good (and probably never can be as good) as searching with the correct term in the first place

atuin-client/src/database.rs
```diff
@@ -326,6 +396,31 @@ mod test {
...
+                results.len(),
+                $expected,
+                "query \"{}\", commands: {:?}",
+                $query,
+                commands
+            );
+            assert_eq!(commands, $commands);
+        };
+    }
+
```
conradludgate, line 423: Does this need to be a macro? or can it be a regular async function?

atuin-client/src/database.rs
```diff
@@ -286,27 +287,96 @@ impl Database for Sqlite {
...
+                                was_or = true;
+                                continue;
+                            } else {
+                                query_params.push(format!("{glob}|{glob}", glob = glob));
+                            }
+                        }
+                        exact_prefix if query_part.starts_with('^') => query_params.push(format!(
+                            "{term}{glob}",
+                            term = exact_prefix.strip_prefix('^').unwrap(),
+                            glob = glob
```
conradludgate, line 325: Minor nit: this `glob = glob` syntax is unnecessary on the latest version ([since 1.58.0](https://blog.rust-lang.org/2022/01/13/Rust-1.58.0.html#captured-identifiers-in-format-strings))

atuin-client/src/database.rs
```diff
@@ -286,27 +287,96 @@ impl Database for Sqlite {
...
+                    let (operator, glob) = if query_part.contains(char::is_uppercase) {
+                        ("glob", '*')
+                    } else {
+                        ("like", '%')
+                    };
+                    let (is_inverse, query_part) = if query_part.starts_with('!') {
+                        (true, query_part.strip_prefix('!').unwrap())
+                    } else {
+                        (false, query_part)
+                    };
```
conradludgate, line 311:
```suggestion
                    let (is_inverse, query_part) = if let Some(prefix) = query_part.strip_prefix('!') {
                        (true, prefix)
                    } else {
                        (false, query_part)
                    };
```

conradludgate (approved):
Overall, code looks good. Just some nitpicks

atuin-client/src/database.rs
conradludgate: Or maybe even a match statement of similar effect:
```rust
                    let (is_inverse, query_part) = match query_part.strip_prefix('!') {
                        Some(prefix) => (true, prefix),
                        None => (false, query_part),
                    };
```

atuin-client/src/database.rs
pmarschik: removed the `glob = glob`

atuin-client/src/database.rs
pmarschik: used `match`

atuin-client/src/database.rs
pmarschik: Sure, I've updated the code to use `async func` but the requirement to `.await.unwrap()` all calls makes the assertions longer again. Maybe a `try_join!( assertions )` could be used instead?

conradludgate: I see, that's fine IMO

conradludgate (approved):
LGTM

ellie:
Thanks @pmarschik! Can't wait to get this released 😁 🚀

PR #287, opened 2022-04-04
Add rust-version to Cargo.toml

schrej:
This adds `rust-version = "1.59"` to `Cargo.toml` to indicate the minimum required version.

I tried compiling with rust 1.56 and got a bunch of compiler errors due to tuple destructuring being used, which apparently [requires rust 1.59](https://github.com/rust-lang/rust/commit/d258e92900739e7798ed4fccfbbb8cb0fe519acd). With this parameter new contributors should get a much more helpful error.


conradludgate (approved):
Thanks!

ellie:
Awesome, thanks for the work here! 🚀

PR #288, opened 2022-04-04
Add compact mode

schrej:
Using this for a while I frequently ran into issues when the terminal was pretty slim, e.g. at the bottom of an IDE. So I've created a compact mode that saves 8 lines of precious terminal space.

I've added a new config option that allows to switch the style between `full`(the 'old' style), `compact` (the new one added in this PR) and `auto`, which toggles the `compact` style at a height of 14 lines (at 14 lines there are still 5 entries visible with `full` style). `auto` is used by default.

`compact` style
image

`full` style with my VS Code terminal for color reference.
image

### DIFF
--- a/atuin-client/src/settings.rs
+++ b/atuin-client/src/settings.rs
@@ -41,9 +41,22 @@ impl From<Dialect> for chrono_english::Dialect {
     }
 }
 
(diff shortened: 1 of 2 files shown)


ellie:
This is great, thanks for the work here! 🚀

ellie (approved):

PR #289, opened 2022-04-05
Update to clap 3.1.x

Sciencentistguy:
Updated to use clap 3.1.x instead of structopt.

### DIFF
--- a/src/command/history.rs
+++ b/src/command/history.rs
@@ -2,8 +2,8 @@ use std::env;
 use std::io::Write;
 use std::time::Duration;
 
+use clap::Subcommand;
 use eyre::Result;
-use structopt::StructOpt;
 use tabwriter::TabWriter;
 
 use atuin_client::database::Database;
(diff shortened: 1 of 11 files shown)


conradludgate (dismissed):
You beat me to it 😅 Looks good, thanks!

src/command/history.rs
```diff
@@ -2,60 +2,51 @@ use std::env;
...
+        #[clap(long, short)]
         cwd: bool,
 
-        #[structopt(long, short)]
+        #[clap(long, short)]
         session: bool,
 
-        #[structopt(long, short)]
+        #[clap(long, short)]
         human: bool,
```
conradludgate, line 36: Seems this is having some troubles. It's inferring the subcommand as `-h` but it's conflicting with the help option. There might be other cases like this too.

I guess an easy solution would be to extract this list variant into a new struct

Sciencentistguy: I think we have to make a decision here; do we want `atuin history last -h` to be `atuin history last --help` or `atuin history last --human`?

ellie: I'd go for -h being help, it's more intuitive and in line with what people might expect

Sciencentistguy: Agreed

conradludgate:
Seems the same issue applies with the `atuin search` subcommand (again a `--human` flag). I also agree that it should be `--long` only

conradludgate (approved):
Thanks! 🥳

PR #298, opened 2022-04-13
Initial implementation of calendar API

ellie:
This can be used in the future for sync so that we can be more
intelligent with what we're doing, and only sync up what's needed

I'd like to eventually replace this with something more like a merkle
tree, hence the hash field I've exposed, but that can come later

Although this does include a much larger number of count queries, it
should also be significantly more cache-able. I'll follow up with that
later, and also follow up with using this for sync :)


atuin-common/src/utils.rs
```diff
@@ -51,6 +52,22 @@ pub fn data_dir() -> PathBuf {
...
+        match month {
+            12 => 1,
+            _ => month + 1,
+        },
+        1,
+    )
+    .signed_duration_since(NaiveDate::from_ymd(year, month, 1))
+    .num_days()
+}
+
```
conradludgate, line 70: Does chrono/time not expose any better solution for this? 😭

ellie: Chrono does not 🙃 and I'd rather avoid adding another dependency for one single function that's pretty easy to implement - though if we end up needing more from that crate it can be added

ellie: haha makes sense! Rust library landscape is constantly changing 😂

Issue #270, opened 2022-01-22
Show proper error message when key is wrong


SuperSandro2000:
Right now if you login with the wrong key you get the following error which is not helpful at all. A proper error should be displayed suggesting the key is wrong.

```
Error: Invalid byte 10, offset 48.

Location:
    /build/rustc-1.57.0-src/library/core/src/result.rs:1914:27
```

ellie:
Resolved in #300

PR #300, opened 2022-04-13
provide better error messages

conradludgate:
[fs-err](https://docs.rs/fs-err/2.7.0/fs_err/) is a drop in replacement for std::fs which provides better error messages.

This also addresses #270


ellie (approved):
neat!

PR #308, opened 2022-04-19
Add count trigger

ellie:
This can be used in a later PR to optimise our database usage
significantly. Avoiding COUNT will be fantastic for our CPU usage, and
therefore sync times.


conradludgate:
Just so I'm clear, after every insert, this runs the function hook to update the total entry? And there's a migration job to back fill the old ones?

ellie:
Correct. If the entry isn't found, it'll run a COUNT to fill it - and from then on, it's just incremented/decremented when needed

PR #312, opened 2022-04-20
Use the count cache

ellie:
By default read from the count cache - if there is no value there, then
do a full COUNT. The cache will be filled when the user posts up some
more history

Eventually I'd like to replace the full COUNT function with reading from
the cache, but that can only be done once all users have a cached value

### DIFF
--- a/atuin-server/src/handlers/history.rs
+++ b/atuin-server/src/handlers/history.rs
@@ -13,10 +13,17 @@ pub async fn count(
     user: User,
     db: Extension<Postgres>,
 ) -> Result<Json<CountResponse>, ErrorResponseStatus<'static>> {
-    match db.count_history(&user).await {
+    match db.count_history_cached(&user).await {
+        // By default read out the cached value
         Ok(count) => Ok(Json(CountResponse { count })),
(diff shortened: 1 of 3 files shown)


atuin-server/src/handlers/history.rs
```diff
@@ -13,10 +13,17 @@ pub async fn count(
...
-    match db.count_history(&user).await {
+    match db.count_history_cached(&user).await {
+        // By default read out the cached value
         Ok(count) => Ok(Json(CountResponse { count })),
-        Err(_) => Err(ErrorResponse::reply("failed to query history count")
-            .with_status(StatusCode::INTERNAL_SERVER_ERROR)),
+
+        // If that fails, fallback on a full COUNT. Cache is built on a POST
+        // only
+        Err(_) => match db.count_history(&user).await {
```
conradludgate, line 22: Can you filter only for row-not-found errors?

conradludgate: I think to do this, change the DB `Result` to be `sqlx::Result` instead of `eyre::Result`, and then you have access to https://docs.rs/sqlx/0.5.13/sqlx/enum.Error.html#variant.RowNotFound

PR #313, opened 2022-04-20
refactor commands for better separation

conradludgate:
Bit of a random one. I want us to start using tracing on server, but keep log for client for now. This also adds a better separation of concerns to me


ellie:
Happy to merge once the conflicts are sorted! I agree it makes more sense this way :)

ellie (approved):

PR #315, opened 2022-04-20
tracing

conradludgate:
Adds tracing support and some simple instrumentations :)

For now, this doesn't log much new. example log line:

Before:

Screenshot 2022-04-20 at 21 22 39

After:

Screenshot 2022-04-20 at 21 21 20

The more info you add to the spans (and the more spans you have), the more you'll see in these events.

### DIFF
--- a/atuin-server/src/database.rs
+++ b/atuin-server/src/database.rs
@@ -1,6 +1,6 @@
 use async_trait::async_trait;
-use tracing::debug;
 use std::collections::HashMap;
+use tracing::debug;
 
 use sqlx::{postgres::PgPoolOptions, Result};
 
--- a/atuin-server/src/database.rs
+++ b/atuin-server/src/database.rs
@@ -1,4 +1,5 @@
 use async_trait::async_trait;
+use tracing::debug;
 use std::collections::HashMap;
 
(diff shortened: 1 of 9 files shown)


conradludgate:
With just a few lines of code

```diff
pub async fn run(self) -> Result<()> {
+        let tracer = opentelemetry_jaeger::new_pipeline().install_simple()?;
+
         tracing_subscriber::registry()
             .with(fmt::layer())
             .with(EnvFilter::from_default_env())
+            .with(tracing_opentelemetry::layer().with_tracer(tracer))
             .init();
```

we can get nice traces like this :)

Screenshot 2022-04-20 at 21 23 47

atuin-server/src/database.rs
```diff
@@ -77,20 +78,23 @@ impl Postgres {
 
 #[async_trait]
 impl Database for Postgres {
+    #[instrument(skip_all)]
```
conradludgate, line 81: The reasoning for `skip_all` is that it's usually better form to opt_in which fields you want to log (for PII as well as just for noisy structs like db handles).

In tracing 0.2, this will be the default: https://github.com/tokio-rs/tracing/issues/651

ellie (approved):

PR #318, opened 2022-04-21
Fix SQL cache query

ellie:
I just deployed the older version and it was falling back on the full
count. Turns out this is because it won't upcast from INT4 to INT8
automatically, and it has to be manual

At some point the underlying total should be changed to int8, but also I
highly doubt anyone will have enough shell history to fill an int4 lol

### DIFF
--- a/atuin-server/src/database.rs
+++ b/atuin-server/src/database.rs
@@ -120,7 +120,7 @@ impl Database for Postgres {
     }
 
     async fn count_history_cached(&self, user: &User) -> Result<i64> {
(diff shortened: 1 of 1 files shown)


conradludgate (approved):
If they manage to get to 2 billion entries, they deserve to pay you to fix this

ellie:
> they deserve to pay you to fix this

yes exactly 😂 

> At some point the underlying total should be changed to int8, but also I
highly doubt anyone will have enough shell history to fill an int4 lol

I hope this isn't quoted in an incident report in like 20 years time 😂

PR #319, opened 2022-04-21
Treat PopOS as Ubuntu for install.sh

bradrf:


bradrf:
Pop!OS is based on Ubuntu: https://en.wikipedia.org/wiki/Pop!_OS#Release_history

PR #329, opened 2022-04-22
History filter

ellie:
Switch between different search modes to narrow down the history you
want - global search for all history, host for all history from your
current machine, session for the current shell session, and directory
for the current directory

The default can be configured via `filter_mode`

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -63,7 +63,7 @@ jobs:
         key: ${{ runner.os }}-cargo-debug-${{ hashFiles('**/Cargo.lock') }}
 
     - name: Run cargo test
-      run: cargo test --workspace
+      run: ATUIN_SESSION=beepboopiamasession cargo test --workspace
 
   clippy:
     runs-on: ubuntu-latest
--- a/atuin-client/src/database.rs
+++ b/atuin-client/src/database.rs
@@ -19,6 +19,28 @@ use super::history::History;
 use super::ordering;
 use super::settings::{FilterMode, SearchMode};
 
+pub struct Context {
+    session: String,
+    cwd: String,
+    hostname: String,
+}
+
+pub fn current_context() -> Context {
+    let session =
+        env::var("ATUIN_SESSION").expect("failed to find ATUIN_SESSION - check your shell setup");
+    let hostname = format!("{}:{}", whoami::hostname(), whoami::username());
+    let cwd = match env::current_dir() {
+        Ok(dir) => dir.display().to_string(),
+        Err(_) => String::from(""),
+    };
+
+    Context {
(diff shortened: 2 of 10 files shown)


.github/workflows/rust.yml
```diff
@@ -63,7 +63,7 @@ jobs:
         key: ${{ runner.os }}-cargo-debug-${{ hashFiles('**/Cargo.lock') }}
 
     - name: Run cargo test
-      run: cargo test --workspace
+      run: ATUIN_SESSION=beepboopiamasession cargo test --workspace
```
conradludgate, line 66: 😂

atuin-client/src/database.rs
```diff
@@ -179,33 +186,58 @@ impl Database for Sqlite {
...
+        unique: bool,
+    ) -> Result<Vec<History>> {
         debug!("listing history");
+        let session = env::var("ATUIN_SESSION")
+            .expect("failed to find ATUIN_SESSION - check your shell setup");
+        let hostname = format!("{}:{}", whoami::hostname(), whoami::username());
+        let cwd = match env::current_dir() {
+            Ok(dir) => dir.display().to_string(),
+            Err(_) => String::from(""),
+        };
```
conradludgate, line 202: Should this really be in the db layer

conradludgate (commented):
Oh jeez that sql

.github/workflows/rust.yml
ellie: well it's not lying is it? 😏

atuin-client/src/database.rs
ellie: Meh probably not, but also I don't want to get to the point where we're passing around 9999999 parameters to every database function 🙃

ellie:
This 100% highlights the fact we need to bin off SQL, or at least stop using it in this way. Hopefully a query builder or something would work, I am also not a fan of ORMs 😬

atuin-client/src/database.rs
conradludgate: I have some thoughts but I can experiment later

conradludgate:
> This 100% highlights the fact we need to bin off SQL, or at least stop using it in this way. Hopefully a query builder or something would work, I am also not a fan of ORMs 😬 

Yeah, I was looking at the sqlx macros for a bit but decided that query builders really is the way to go for us

atuin-client/src/database.rs
ellie: It's chill, I'm wrapping it up into a struct I think. Those three values are rarely used in isolation, and I should stop being lazy

conradludgate (approved):

conradludgate (approved):

PR #332, opened 2022-04-22
Fix empty queries with filters

ellie:
oops 😳

basically, empty queries just had an `and` or a `where` with no actual specifier after it, which sql hated

### DIFF
--- a/atuin-client/src/database.rs
+++ b/atuin-client/src/database.rs
@@ -242,7 +242,11 @@ impl Database for Sqlite {
             FilterMode::Directory => format!("cwd = '{}'", context.cwd).to_string(),
         };
 
-        let filter = format!("{} {}", join, filter_query);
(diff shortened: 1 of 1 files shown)


atuin-client/src/database.rs
```diff
@@ -242,7 +242,11 @@ impl Database for Sqlite {
             FilterMode::Directory => format!("cwd = '{}'", context.cwd).to_string(),
         };
 
-        let filter = format!("{} {}", join, filter_query);
+        let filter = if filter_query.is_empty() {
+            "".to_string()
```
conradludgate, line 246: you are very lucky that rust is smart enough to avoid a new allocation for these excessive `.to_string()`s

ellie: I thought that if the compiler _wasn't_ smart enough to do that, then I'd be deeply disappointed in Rust

PR #328, opened 2022-04-22
feature-flags

conradludgate:
Purely a curiosity for now. Adds some feature flags to disable functionality

PR #336, opened 2022-04-23
Update Cargo.lock

orhun:
Hello!

I saw that manifest files of the crates are updated in [22a7d88](https://github.com/ellie/atuin/commit/22a7d8866bb82cbf68b4acfc61fba464376730f1) but the versions in `Cargo.lock` are not updated.

I spotted this issue while building the [Arch Linux package](https://archlinux.org/packages/community/x86_64/atuin/):

```sh
$ cargo build --locked
error: the lock file /build/atuin/src/atuin-0.9.0/Cargo.lock needs to be updated but --locked was passed to prevent this
```

`--locked` flag is required for the reproducibility of the package. Read more about it [here](https://wiki.archlinux.org/title/Rust_package_guidelines#Prepare).

This PR updates `Cargo.lock` for matching the versions in `Cargo.toml`. (i.e. I simply ran `cargo build` once for updating the lock file.)

P.S. Congrats on the new release! I see that you've added some nice features. I'm curious about the reason behind switching to `axum` :bear:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -70,7 +70,7 @@ dependencies = [
 
 [[package]]
 name = "atuin"
-version = "0.8.1"
+version = "0.9.0"
 dependencies = [
  "async-trait",
  "atuin-client",
@@ -104,7 +104,7 @@ dependencies = [
 
 [[package]]
 name = "atuin-client"
-version = "0.8.1"
+version = "0.9.0"
 dependencies = [
  "async-trait",
  "atuin-common",
@@ -138,7 +138,7 @@ dependencies = [
 
 [[package]]
 name = "atuin-common"
-version = "0.8.1"
+version = "0.9.0"
 dependencies = [
  "chrono",
  "rust-crypto",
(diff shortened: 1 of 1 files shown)


ellie (approved):

ellie:
Ohh yikes, nice catch - thank you! Can't believe I missed that 🤦‍♀️

conradludgate:
Oh hey, while you're here. Could you add `--all-features` to the pkgbuild? We have added them as defaults for now, but are considering disabling some in future. Obviously we don't want to break any existing arch user installs

conradludgate:
As for the reason why we switched to axum, mostly for a more ergonomic and less magic API, while still being as performant and secure as warp (running on hyper). It also reduces compile times due to reduced monomorphisation requirements

orhun:
> Oh hey, while you're here. Could you add `--all-features` to the pkgbuild? We have added them as defaults for now, but are considering disabling some in future. Obviously we don't want to break any existing arch user installs

Sure, I think it is related to #328. Thanks for letting me know! 🐻 

> As for the reason why we switched to axum, mostly for a more ergonomic and less magic API, while still being as performant and secure as warp (running on hyper). It also reduces compile times due to reduced monomorphisation requirements

Interesting, I'm going to give it a try when I have the opportunity for sure.

PR #340, opened 2022-04-24
history list

conradludgate:
## remove rust-crypto

sha2 is much more trusted and up to date than rust-crypto.

## update history printing code

Our use of TabWriter ended up causing some awkward behaviour. TabWriter tries to make each tab stop be at the same position, and the fact we had very varying sizes in our command stage, caused these tabs stops to be sometimes many lines down.

In my case, I had a single command span 5 lines, this caused all other commands to similarly span those same 5 lines without needing to.

The fix here is to swap the order of the command and the duration.

~~For a little flare, I've added some colour to the duration to be more consistent with some of our other setups~~ removed as I think our primary use case for this subcommand was interop with fzf, which does not support ansi

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -90,7 +90,6 @@ dependencies = [
  "indicatif",
  "itertools",
  "log",
- "owo-colors",
  "pretty_env_logger",
  "serde",
  "serde_json",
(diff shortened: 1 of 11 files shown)


conradludgate:
Showcase of the human list (I'm not 100% happy with the middot, It was just a random idea):

Screenshot from 2022-04-24 21-06-44

ellie (approved):

PR #351, opened 2022-04-26
fix db range query

conradludgate:
I think chrono/sqlx was turning the datetime items into second based timestamps, whereas we store them in nano form


ellie (approved):
Yay nice! I wonder when this broke 🤔

PR #353, opened 2022-04-26
SQLx cannot run this migration OK

ellie:
And also correct a typo :)


conradludgate (approved):
Stop being bad at sql

conradludgate:
Would it be worth setting up server sql tests in CI at some point?

ellie:
So the SQL actually runs OK, but sqlx migrate runs everything inside a transaction + posgres doesn't allow concurrent indices within a transaction 🤦‍♀️ 

And yeah probably, but I think I'm done messing with SQL for now 😂

PR #365, opened 2022-04-28
Docs folder is fully translated to russian

mb6ockatf:
Uff... It took me two hours to do it, according to commits history. I really hope that this PR will be approved.

Things like `true` and `false` cannot be translated to russian. I mean, they can, but the meaning is kinda different. Some other words are left with no translation, too.

[OnePieceJoker](https://github.com/OnePieceJoker) was going to make a translation to chinese. Thanks to you, I got inspired by your [issue](https://github.com/ellie/atuin/issues/360)

If translations for another languages will appear, so we can make a folder for each one. But now I think it still looks OK.

I am sorry for the last commit, you can see it. I am just not really good at resolving conflicts and the file I was editing was corrected right before I did it. So, I accidently deleted a comment...


conradludgate:
Thanks! This is very much appreciated. I'll try review through Google translate when I get home but it looks great so far

ellie:
Omg, thank you so so much for the effort here! Really appreciated ❤️ 

I'd be happy to merge this, but I would like if we could put it into a subfolder if possible. Let's get it started off right :)

So all the Russian translation would live under

```
docs/ru/the_file_name.md
```

Thanks again!!!

OnePieceJoker:
> Omg, thank you so so much for the effort here! Really appreciated ❤️ 
> 
> 
> 
> I'd be happy to merge this, but I would like if we could put it into a subfolder if possible. Let's get it started off right :)
> 
> 
> 
> So all the Russian translation would live under
> 
> 
> 
> ```
> 
> docs/ru/the_file_name.md
> 
> ```
> 
> 
> 
> Thanks again!!!

I agree with this file directory format because that's what I'm going to do

docs/stats_ru.md
```diff
@@ -0,0 +1,40 @@
+# `atuin stats`
+
+Atuin также может выводить статистику, основанную на истории. Пока что в очень простом виде,
+но скоро должно появиться больше возможностей.
+
+Статистика выводится пока только на английском
+Statistics in english only 
+# TODO
```
mb6ockatf: Statistics in English only ¯\_(ツ)_/¯

ellie: It'd be awesome to localise everything else at some point too 🤔

ellie (approved):

ellie:
Thank you so much for this! I'm excited to see our documentation is becoming accessible to more people 🎉

Issue #18, opened 2021-03-10
Add stats summary


ellie:
Instead of `atuin stats all`, it should have the same functionality but just for `atuin stats`. A little neater and more intuitive that way.

`atuin stats day 01/01/2021` could then become `atuin stats 01/01/2021`

satyarths:
Hi there 👋🏽 ,

I was playing around with atuin (and rust) and was looking into something that I could pick up. Wanted to check in if thsi issue is still valid.  And if so can you help in fleshing this out a bit more?

From what  I understand;

The `atuin stats`  could be:

```shell
> atuin stats last friday 
# displays  statistics for last friday

> atuin stats 01/01/2022
# displays  statistics for 1st Jan '22

> atuin stats monday
# displays  statistics for last monday

> atuin stats march
# displays stats for last march (2022-03-01 to 2022-03-31) , this currently displays for 1st March '22 only.
```

conradludgate:
Yeah that looks correct. The stats logic is already present in the Day subcommand, it just needs to be reshuffled a bit

satyarths:
In order to consider reducing parsing complexity ,  I would also propose  simplying how it is currently parsed. 
i.e. `atuin stats friday` can be considered equivalent to `atuin stats last friday`. Any day  of the week would correspond to the the nearest past date and if it needs to be more specific , the user could specify an explicity date. 
wdyt about this?

PR #384, opened 2022-05-03
Adds stats summary 

satyarths:
Solves for #18 .

Modifies `stats` to work without sub command 

This currently works as;

```
> atuin stats
# equivalent to the earlier `atuin stats all`


> atuin stats last friday 
# displays  statistics for last friday

> atuin stats 01/01/2022
# displays  statistics for 1st Jan '22

> atuin stats monday
# displays  statistics for last monday

> atuin stats march
# displays stats for last march (2022-03-01 to 2022-03-31) , this currently displays for 1st March '22 only.
```

### DIFF
--- a/src/command/client/stats.rs
+++ b/src/command/client/stats.rs
@@ -73,13 +73,12 @@ impl Cmd {
         } else {
             self.period.join(" ")
         };
-        let history = match words.as_str() {
-            "all" => db.list(FilterMode::Global, &context, None, false).await?,
-            _ => {
-                let start = parse_date_string(&words, Local::now(), settings.dialect.into())?;
-                let end = start + Duration::days(1);
-                db.range(start.into(), end.into()).await?
-            }
+        let history = if words.as_str() == "all" {
+            db.list(FilterMode::Global, &context, None, false).await?
+        } else {
+            let start = parse_date_string(&words, Local::now(), settings.dialect.into())?;
+            let end = start + Duration::days(1);
+            db.range(start.into(), end.into()).await?
         };
         compute_stats(&history)?;
         Ok(())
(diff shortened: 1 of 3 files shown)


src/command/client/stats.rs
```diff
@@ -72,32 +68,23 @@ impl Cmd {
...
+        let words = if self.period.is_empty() {
+            String::from("all")
+        } else {
+            self.period.join(" ")
+        };
+        let history: Vec<History>;
+        match words.as_str() {
+            "all" => {
+                history = db.list(FilterMode::Global, &context, None, false).await?;
+            }
```
conradludgate, line 80: Nit: It's normally more rusty to do
```suggestion
        let history = match words.as_str() {
            "all" => {
                db.list(FilterMode::Global, &context, None, false).await?
            }
```
That is, a match statement is actually an expression and can produce values. Just omit the `;` at the end and that value will propagate out

conradludgate (dismissed):
Overall, looks good! Thanks :)

src/command/client/stats.rs
conradludgate: Ah, clippy is suggesting this too :)

satyarths: Missed this. Updating right now !

src/command/client/stats.rs
```diff
@@ -72,32 +68,23 @@ impl Cmd {
...
-                    words.join(" ")
-                };
-
+        let words = if self.period.is_empty() {
+            String::from("all")
+        } else {
+            self.period.join(" ")
+        };
+        let history: Vec<History>;
+        match words.as_str() {
```
satyarths, line 77: clippy also recommends using an `if` statement instead of `match` here. My intention of using `match` here is to further extend this to include longer durations like a month, year etc.  Let me know if you would want me to change this to an `if` clause instead,

conradludgate: Yeah, I think we can go for an if for now (esp if clippy is going to complain). Converting it back to a match in future is not much work

conradludgate (approved):
Tysm <3

PR #390, opened 2022-05-05
Add local event log storage

ellie:
Start storing an event log locally! 

This will allow us to eventually start syncing events, rather than pure history items.

When merging an event log, we will pair up CREATEs with DELETEs. If there is a pair, we won't create the history. If there is not a pair (and just a CREATE), then the item is created. I've also started on the server end of this, while I had the context in my head. Obviously it's no use just yet

I've been running this branch locally for a bit, and have had no issues.

### DIFF
--- a/atuin-client/src/database.rs
+++ b/atuin-client/src/database.rs
@@ -13,10 +13,10 @@ use sqlx::{
 };
 
 use super::{
+    event::{Event, EventType},
     history::History,
     ordering,
     settings::{FilterMode, SearchMode},
-    event::{Event, EventType},
 };
 
 pub struct Context {
@@ -62,6 +62,8 @@ pub trait Database: Send + Sync {
 
     async fn update(&self, h: &History) -> Result<()>;
     async fn history_count(&self) -> Result<i64>;
+    async fn event_count(&self) -> Result<i64>;
+    async fn merge_events(&self) -> Result<i64>;
 
     async fn first(&self) -> Result<History>;
     async fn last(&self) -> Result<History>;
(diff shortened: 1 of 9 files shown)


atuin-client/src/database.rs
```diff
@@ -115,6 +118,27 @@ impl Sqlite {
         Ok(())
     }
 
+    async fn save_event(tx: &mut sqlx::Transaction<'_, sqlx::Sqlite>, e: &Event) -> Result<()> {
+        let event_type = match e.event_type {
+            EventType::Create => "create",
+            EventType::Delete => "create",
```
conradludgate, line 124: Typo

atuin-client/src/database.rs
```diff
@@ -302,6 +330,49 @@ impl Database for Sqlite {
         Ok(res)
     }
 
+    async fn event_count(&self) -> Result<i64> {
+        let res: (i64,) = sqlx::query_as("select count(1) from events")
```
conradludgate, line 334: I believe this works
```suggestion
        let res: i64 = sqlx::query_scalar("select count(1) from events")
```

atuin-client/src/database.rs
```diff
@@ -302,6 +330,49 @@ impl Database for Sqlite {
...
+            // But also even if you have a LOT of history that should be fine, and we're only going to be doing this once EVER.
+            let all_the_history = self
+                .list(FilterMode::Global, &no_context, None, false)
+                .await?;
+
+            let mut tx = self.pool.begin().await?;
+            for i in all_the_history.iter() {
+                // A CREATE for every single history item is to be expected.
+                let event = Event::new_create(i);
+                Self::save_event(&mut tx, &event).await?;
```
conradludgate, line 368: From what I understand, you're relying on sqlite to say ignore the errors when inserting duplicate events? From what I know on Postgres, transactions are dead once there is an error

conradludgate: Never mind, the save event method ignores the failures explicitly

ellie: Indeed. Way easier to let the unique index and SQLite handle deduping

atuin-client/src/database.rs
ellie: That's super neat! Thanks

conradludgate (approved):

PR #393, opened 2022-05-06
Noyez zsh histdb import

noyez:
This adds import support for zsh-histdb. There are a few caveats.

## Caveat 1, async

Edit: This has been addressed by the recent code change to add async traits to the importer, thanks @conradludgate !

~~Zsh-histdb uses a sqlite file to store history, which can easily be ready by sqlx-rs. However, the `Importer` trait has methods `parse()` and `histfile()`, none are async (i tried to use `#[async-trait]` with no luck). The `parse()` method needs to read the sqlite database using sqlx-rs async runtime, but that cannot be done from a non-async function. As a workaround, i added a method called `populate()`, which gets called before `parse()` when a zsh-histdb installation is detected. The method `populate()` is async, and will populate a global `Vec<History>`. Then `parse()` will simply read the data from the `Vec<History>`. This is kind of an ugly hack to get around the async issues, but it works.
I'm open to other ideas~~

## Caveat 2, sqlite and non UTF-8 codes.

Edit: I believe this has been addressed by using `Vec<u8>` instead of `String` Thanks to @conradludgate for the suggestion!

~~While testing on my installation, i had some entries in my history that would crash sqlx-rs with 'invalid utf-8 squence'. I'm not sure this is fault of sqlx-rs, and i wasn't sure how to isolate those issues. One such entry was created by running the command "cd noyez �s\ iMac.sparsebundle". I tried to reproduce this in a smaller db with no luck, sqlx-rs worked on that smaller DB. I don't know how this entry got into my history, whether it was from zsh-histdb, or an import script that i used when i converted from regular zsh-history to zsh-histdb. Without more test data it is hard to test.~~

~~It might be possible that you'll see bug reports with this error, and if so, perhaps the sqlx-rs project could help.~~

~~If sqlx-rs cannot handle a particular entry the following error is produced:~~

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: error occurred while decoding column "argv": invalid utf-8 sequence of 1 bytes from index 96

Caused by:
    invalid utf-8 sequence of 1 bytes from index 96
```


conradludgate:
Thanks, I appreciate the effort. I'll try and fix it up for you

conradludgate:
Hey, so I've been refactoring over on #395 if you want to take a look. It should be general enough to support sqlx queries. If not, let me know over there 🙏

ellie:
Hey @noyez! Thanks for all the effort here! 🚀 

Just gonna need a rebase on top of @conradludgate's work I just merged. Hopefully things are easier now :)

noyez:
Great! Should be a lot easier, I guess I should have waited a week :).it’ll take me a few days to get the time, but it’ll get done.

noyez:
Code merged, updated from latest main.

ellie (dismissed):
Woo! Thanks for this! 🚀 <3

atuin-client/Cargo.toml
```diff
@@ -62,6 +62,7 @@ hex = { version = "0.4", optional = true }
 sha2 = { version = "0.10", optional = true }
 rmp-serde = { version = "1.0.0", optional = true }
 base64 = { version = "0.13.0", optional = true }
+tokio = { version = "1", features = ["full"] }
```
conradludgate, line 65: I don't think this is needed anymore

atuin-client/src/import/zsh_histdb.rs
```diff
@@ -0,0 +1,165 @@
...
+#[derive(Debug)]
+pub struct ZshHistDb {
+    histdb: Vec<HistDbEntry>,
+}
+
+
+/// Read db at given file, return vector of entries.
+async fn hist_from_db(dbpath: PathBuf) -> Result<Vec<HistDbEntry>> {
+    let pool = SqlitePool::connect(dbpath.to_str().unwrap()).await?;
+    let query = format!("select history.id,history.start_time,history.duration,places.host,places.dir,commands.argv from history left join commands on history.command_id = commands.rowid left join places on history.place_id = places.rowid order by history.start_time");
```
conradludgate, line 87: Does this need to be a format? I believe you can pass a string literal to the query_as function

atuin-client/src/import/zsh_histdb.rs
```diff
@@ -0,0 +1,165 @@
...
+    pub count: usize
+}
+
+#[derive(sqlx::FromRow, Debug)]
+pub struct HistDbEntry {
+    pub id: i64,
+    pub start_time: NaiveDateTime,
+    pub host: String,
+    pub dir: String,
+    pub argv: String,
```
conradludgate, line 60: If you're having issues with utf8, change this to Vec<u8> and manually convert to String and handle the error (see String::from_utf8). Our standard practice is to drop those entries

conradludgate (commented):
Minor comments, if you could delete the test DB files too please 🙏

atuin-client/src/import/zsh_histdb.rs
ellie: clippy has flagged this up too

ellie:
Although the test actually loads that DB. I'd be happy to keep it as a test fixture

atuin-client/src/import/zsh_histdb.rs
noyez: good idea, i'll see if i can dig up a sqlite db that causes this error to test against.

atuin-client/src/import/zsh_histdb.rs
ellie: Generally for testing sqlite, we'd open an in-memory instance (open the path ":memory:"), insert some test values, and then run queries against that. That way there's no dependency on the local filesystem

noyez:
Perfect. I’ll look into making this change to the tests.

atuin-client/src/import/zsh_histdb.rs
noyez: Test re-written to eliminate local filesystem dependency. I ended up having to split a few of the functions into small chunks to try to exercise as much code as possible during the tests.

noyez:
New code merged, 
 - non FS depending tests
 - using Vec<u8> instead of string.

ellie (approved):
Approved! Tysm for the work here, it's great to get another importer in! 🚀

PR #395, opened 2022-05-07
Importer V3

conradludgate:
The old API was clever, but sacrificed a lot of flexibility.

This is much more naive, simple and should be more flexible and probably more performant too

### DIFF
--- a/atuin-client/src/import/mod.rs
+++ b/atuin-client/src/import/mod.rs
@@ -1,27 +1,46 @@
-use std::{
-    io::{BufRead, BufReader, Read, Seek, SeekFrom},
-    path::{Path, PathBuf},
-};
+use std::io;
 
+use async_trait::async_trait;
 use eyre::Result;
 
 use crate::history::History;
 
 pub mod bash;
-pub mod fish;
-pub mod resh;
-pub mod zsh;
+// pub mod fish;
+// pub mod resh;
+// pub mod zsh;
 
-// this could probably be sped up
-fn count_lines(buf: &mut BufReader<impl Read + Seek>) -> Result<usize> {
-    let lines = buf.lines().count();
-    buf.seek(SeekFrom::Start(0))?;
-
-    Ok(lines)
+fn count_lines(input: &[u8]) -> usize {
+    memchr::memchr_iter(b'\n', input).count()
 }
 
-pub trait Importer: IntoIterator<Item = Result<History>> + Sized {
+#[async_trait]
+pub trait Importer: Sized {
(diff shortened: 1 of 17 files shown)


src/command/mod.rs
```diff
@@ -19,11 +19,11 @@ pub enum AtuinCmd {
...
 
 impl AtuinCmd {
-    pub async fn run(self) -> Result<()> {
+    pub fn run(self) -> Result<()> {
         match self {
-            Self::Client(client) => client.run().await,
+            Self::Client(client) => client.run(),
             #[cfg(feature = "server")]
-            Self::Server(server) => server.run().await,
+            Self::Server(server) => server.run(),
```
conradludgate, line 26: This is a side change. The client does no `tokio::spawn` calls, so it doesn't need to be multi-threaded like the server. So this defers the tokio setup to the client/server respectively.

atuin-client/src/import/mod.rs
```diff
@@ -1,27 +1,46 @@
...
-    fn parse(path: impl AsRef<Path>) -> Result<Self>;
+    async fn new() -> io::Result<Self>;
+    async fn entries(&mut self) -> Result<usize>;
+    async fn load(self, loader: &mut impl Loader) -> Result<()>;
+}
+
+#[async_trait]
+pub trait Loader: Sync + Send {
+    async fn push(&mut self, hist: History) -> eyre::Result<()>;
+}
```
conradludgate, line 28: This is the core API change. It's very simple, the importer can

1. Be created
2. Count the number of entries it may contain
3. Load all of it's entries into a `Loader`. (A `TestLoader` is defined below for testing)

All these methods are async and fallible so I'm not sure how much more flexible we can be :sweat_smile:

src/command/client/import.rs
```diff
@@ -45,126 +44,76 @@ impl Cmd {
...
+#[async_trait]
+impl<'db, DB: Database> Loader for HistoryImporter<'db, DB> {
+    async fn push(&mut self, hist: History) -> Result<()> {
+        self.pb.inc(1);
+        self.buf.push(hist);
+        if self.buf.len() == self.buf.capacity() {
+            self.db.save_bulk(&self.buf).await?;
+            self.buf.clear();
+        }
+        Ok(())
```
conradludgate, line 104: Much simpler batching. Push, check if at capacity, drain. No more iterator fuckery

ellie (approved):
🚀 🚀 🚀 

inb4 a year from now "wow this code I wrote a year ago needs fixing" 😂

PR #400, opened 2022-05-09
Build ARM docker image in GitHub Actions using QEMU

### LINKED ISSUE #269: Build ARM docker image
When I try to run the atuin docker image on my MacBook it fails:

```
> docker run -d -v "atuin:/config" ghcr.io/ellie/atuin:latest server start
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
```

This is a MacBook Pro 2021 with the Pro M1 processor running macOS Monterey 12.1

daniel0611:
This PR updates the GH Action docker workflow to cross-build the docker
image for ARM using QEMU.
This theoretically works but takes about 2.5 hours for a clean build,
which may or may not be tolerable for each commit on `main`.
An action run where the dependencies are already cached needs about 15
minutes with this PR.

Closes #269

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -19,6 +19,9 @@ jobs:
       - name: Check Out Repo
         uses: actions/checkout@v2
 
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+
       - name: Docker meta
         id: meta
         uses: docker/metadata-action@v3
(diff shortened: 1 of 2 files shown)


ellie:
2.5hrs for a clean build/15 mins for a cached one seems OK for me, as it's main and won't block any work! 

Thanks for getting this out, nice work! 🚀

ellie (approved):

PR #404, opened 2022-05-13
Allow to build atuin server without client

jirutka:
I’m packaging atuin for Alpine Linux and I’m building the atuin client and server as separate binaries, provided in separate (sub)package.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,13 +33,14 @@ members = ["./atuin-client", "./atuin-server", "./atuin-common"]
 # TODO(conradludgate)
 # Currently, this keeps the same default built behaviour for v0.8
 # We should rethink this by the time we hit a new breaking change
-default = ["sync", "server"]
+default = ["client", "sync", "server"]
+client = ["atuin-client"]
 sync = ["atuin-client/sync"]
 server = ["atuin-server", "tracing-subscriber"]
 
 [dependencies]
 atuin-server = { path = "atuin-server", version = "0.9.1", optional = true }
-atuin-client = { path = "atuin-client", version = "0.9.1", default-features = false }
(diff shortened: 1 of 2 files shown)


conradludgate (approved):
This does seem sensible to me, but I'll also defer a review for @ellie if this is something we want

> I’m packaging atuin for Alpine Linux

This is very much appreciated. Thank you!

ellie:
I'm happy with this! Thanks for working on the package 🚀

ellie:
hey @jirutka just a quick q; how're you getting Atuin to link + build nicely against musl?

if you have the time + have already figured this out, your input on https://github.com/ellie/atuin/issues/144 would be absolutely amazing! tysm <3

panekj:
Standard `cargo build` should work just fine.
For Alpine Linux, you can: `apk add cargo && cargo build`

jirutka:
> how're you getting Atuin to link + build nicely against musl?

That’s very easy when you’re building it on a musl-native distro, such as Alpine, with Rust toolchain provided by the distro. ;) As @panekj said, it’s just `cargo build`. However, it’s important to mention that our `<arch>-alpine-linux-musl` target, unlike the Rust’s `<arch>-unknown-linux-musl`, uses dynamic linking (as does `<arch>-unknown-linux-gnu` and similar).

You can see the aport in https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/34351/diffs. There are also some patches to get rid of bundled TLS and libsqlite3 libraries in favour of system-provided libraries.

> unlike the Rust’s `<arch>-unknown-linux-musl`.

I’ll drop some notes there later. But basically, you can take inspiration from [this workflow](https://github.com/jirutka/connman-resolvconf/blob/master/.github/workflows/ci.yml#L129).

PR #406, opened 2022-05-13
Don't create config dir for server in default location if not needed

jirutka:
...respect ATUIN_CONFIG_DIR.

The current behaviour is problematic when running atuin server as a
system service with config dir in /etc/atuin.


conradludgate (approved):
This seems reasonable, thanks!

PR #414, opened 2022-05-16
Atuin-Server SQLite Support

RickyGrassmuck:
This is the initial stab at implementing SQLite support for the atuin-server.

This is my first time touching Rust at all so things the implementation is very rudimentary and there are some areas that could be done better (specifically around handling determining which migrations to run and the database connection strings).

As-is, this PR does successfully allow the user to use either SQLite or Postgresql as defined in their server.toml.

I started working on streamlining how the user configures the database connection but couldn't decide on how best that would work and figured I'd leave it in the more verbose configuration state that it's in now and just iterate on that later.

One other area of concern I just thought of would be backwards compatibility in existing configuration files. The new option `db_type` is required to be set so preexisting configs would cause future versions with these exact changes to fail to start. This is one of the things that should probably be discussed regarding streamlining the configuration.


atuin-server/migrations/sqlite/20210425153757_create_users.sql
```diff
@@ -0,0 +1,12 @@
+create table users (
+	id INTEGER primary key,               -- also store our own ID
+	username varchar(32) not null unique,   -- being able to contact users is useful
+	email varchar(128) not null unique,     -- being able to contact users is useful
+	password varchar(128) not null unique,
+  created_at timestamp not null default CURRENT_TIMESTAMP
```
ellie, line 6: totally a nit, but could the indentation and spacing on this one be fixed please? just to make it in line with the rest 🙏 

should probably get a formatter for sql.

atuin-server/server.toml
```diff
@@ -7,5 +7,11 @@
 ## whether to allow anyone to register an account
 # open_registration = false
 
-## URI for postgres (using development creds here)
+## Database Type, valid options are "postgres" and "sqlite"
+# db_type = "sqlite"
```
ellie, line 11: we can default this one to postgres :)

atuin-server/server.toml
```diff
@@ -7,5 +7,11 @@
...
+
+## Database Connection string
+# Sqlite Example
+# db_uri="sqlite:///home/rigrassm/Development/atuin/config/data.db"
```
ellie, line 15: could the code infer the db_type from the URI, and save the user some configuration effort?

ellie (commented):
A few comments, mostly nits though! thanks for the work here :D

PR #418, opened 2022-05-19
Add fish shell to key binding docs

UbiquitousPhoton:
Add the missing fish shell part into the documentation for ATUIN_NOBIND.

Binding to up is pretty problematic in fish, as if you want to navigate the pager for auto-suggestions, you tend to get a bit stuck.

### DIFF
--- a/docs/key-binding.md
+++ b/docs/key-binding.md
@@ -36,3 +36,13 @@ eval "$(atuin init bash)"
 # bind to ctrl-r, add any other bindings you want here too
 bind -x '"\C-r": __atuin_history'
 ```
+# fish
+
+```
+set -gx ATUIN_NOBIND "true"
+atuin init fish | source
+
+# bind to ctrl-r in normal and insert mode, add any other bindings you want here too
+bind \cr _atuin_search
+bind -M insert \cr _atuin_search
+```


ellie:
I don't use Fish, but this is the first time I've heard that our default bindings are problematic. I'd rather we stick to the defualts we have, but perhaps document this as an extra case?

UbiquitousPhoton:
I only recently started using fish, so don't count myself as an expert either, but I then came to potentially start an issue and  thankfully found these docs. 

The up key is used for more than just history in fish, this is the problem. I would dearly love some more input from other fish users, though!

I think, so long as as it gets documented this is definitely enough. I had to figure it out, so thought I might as well share the knowledge!

ellie:
Ahhh I see, thank you!

If you could open an issue explaining the problems that binding the up arrow cause with Fish then maybe we can figure out what's going on properly! 

If it's definitely conflicting with some Fish behaviour then I'd be happy to change our default config to no longer bind that key :D

UbiquitousPhoton:
The documentation in this PR is still valid either way, if you don't want to bind all keys, or change the bound keys. The syntax is still different for fish. I will open an issue, but that's kind of a different matter.

ellie:
Oh of course, you're right 🤦‍♀️ 

Sorry, it's been a long day 🙃

ellie (approved):

UbiquitousPhoton:
Not a problem, it has indeed been a long day! - opened #419 for further discussion on the topic.

Issue #419, opened 2022-05-19
With fish/zsh shell, default binding of the up key to trigger Atuin may be problematic.


UbiquitousPhoton:
When using the fish shell, the up key is used for a more than just to quickly access history, it is (for example) used to navigate through the pager of auto-suggestions/auto-completions. I thus found myself often triggering Atuin unintentionally whilst attempting to do so, which lead me to come here.

Whilst I am thinking of this, the same could potentially be said of zsh, with something like oh-my-zsh and completions enabled (just tested this, and this is indeed the case)

I'm not sure whether this issue exists in bash, I haven't used bash interactively in some time.

This is relatively low priority, as for now the workaround of defining ATUIN_NOBIND works, at least, but it might potentially mar the out of box experience. 

Unless the two different uses can be differentiated against (I'm not entirely sure they can be, reading the fish docs), then its possible that binding the up key by default may not be a good thing(tm).

panekj:
I've said that before https://github.com/ellie/atuin/pull/234#issuecomment-991357343

ellie:
@panekj what you've said, is a matter of personal preference.

What I'd like to establish here is if Atuin conflicts with the _default_ behaviour of the shell, as that is the only thing we can plan for.

panekj:
> @panekj what you've said, is a matter of personal preference.

The matter of personal preference is stemmed from default shell preference and vice versa.
As other maintainer said (paraphrased), "It's annoying in the beginning, but you can get used to it"

It's okay to have different behaviour for different shells, I doubt that there is more than 5 people using differnt shells at the same time (not counting bash/sh since those tend to be default)
So as long as it's documented it will be fine.

I've tested it right now and it the binds are intrusive in places where they shouldn't be, so it does conflicts with fish

ellie:
Your paraphrasing has changed the meaning there 😅

What was [actually said](https://github.com/ellie/atuin/issues/358#issuecomment-1129650505):

>  I found it intrusive in the beginning but now I struggle to use a terminal without it. 

Anyway.

----

I'm very hesitant to remove the up-arrow behaviour from defaults. I'm willing to do so if it really interferes with default shell behaviours, but less so if it's just an edge case. Afterall, disabling our default keybindings is not at all difficult. There will _always_ be people that like the defaults, and people that do not.

Because of features like sync, you will have an entirely different set of history while scrolling with the arrows than you will when pressing ctrl-R, which can cause even more confusion. 

> It's okay to have different behaviour for different shells

[...]

PR #420, opened 2022-05-20
Improve default fish keybindings

s0:
Binding on "up" can conflict with the default fish keybindings as when in tab-completion mode, you are supposed to be able to use arrow keys to navigate the grid of suggestions, however pressing "up" will open the tui instead.

This attempts to work around it by tracking when the user is probably in a tab completion mode by setting a variable that we use to determine whether to open atuin or perform the default fish up action

This should partially address the issues raised in #419

### DIFF
--- a/src/shell/atuin.fish
+++ b/src/shell/atuin.fish
@@ -20,16 +20,41 @@ function _atuin_search
     end
 end
 
+function _atuin_suppress_tui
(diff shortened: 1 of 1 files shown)


ellie:
This looks good to me, thank you!!

I'm glad we figured out a way of fixing the feature rather than just disabling it :)

ellie (approved):

Issue #344, opened 2022-04-26
Registration password


conradludgate:
We support two flows when registering:

Upfront (arguments):
```sh
atuin register -u "..." -e "..." -p "..."
```

Partial (prompt):
```sh
atuin register -u "..." -e "..."
Please enter password: ...
```

I think we should probably suggest the latter (it's uncommon to paste passwords into the command itself)

There's also the issue that the password prompt does not enable any proper password modes, meaning that the prompt will display the password as it's typed. Might not be what is expected

conradludgate:
The same is true for login. Relevant code: https://github.com/ellie/atuin/blob/main/src/command/client/sync/login.rs#L45

s0:
This can be closed since #424 is merged now right?

PR #423, opened 2022-05-23
Adds support for password prompt 

satyarths:
Addresses #344. 

I've added a dependency of rpassword since I figured the library would be able to handle the password input better across different platforms compared to something that I would write.  I'm also not sure if the `Cmd` struct for login and register should remain since there are no longer any "options" to use with them.  
Do let me know if these changes make sense. Thanks!

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -19,6 +19,9 @@ jobs:
       - name: Check Out Repo
         uses: actions/checkout@v2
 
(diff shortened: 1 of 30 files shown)


satyarths:
There are also a few commits since the last PR I raised. I'm not sure why they appear but I think since you squash and merge that should appear on the final tree. Do let me know incase I need to do something else to fix this.

PR #424, opened 2022-05-23
Adds password prompt for register and login

satyarths:
Addresses https://github.com/ellie/atuin/issues/344.

I've added a dependency of rpassword since I figured the library would be able to handle the password input better across different platforms compared to something that I would write. I'm also not sure if the Cmd struct for login and register should remain since there are no longer any "options" to use with them.
Do let me know if these changes make sense. Thanks!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -91,6 +91,7 @@ dependencies = [
  "itertools",
  "log",
  "pretty_env_logger",
+ "rpassword",
  "serde",
  "serde_json",
  "termion",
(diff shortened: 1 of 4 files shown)


satyarths:
There are also a few commits since the last PR I raised. I'm not sure why they appear but I think since you squash and merge that should appear on the final tree. Do let me know incase I need to do something else to fix this.

conradludgate:
Thanks! I'll have to test it to see how rpassword works, but it looks good.

Can you keep keep the original argument mode as well however? At least username and email via flags seems reasonable to me

satyarths:
Sure will look into that. I did not do that because I expected it to be confusing to have a partial prompt and options based input. But let me add that back and then decide.

panekj:
Please leave `--password` as well, most sane shells have been capable of secure input via `read -s` for a long time

satyarths:
So if I were to leave `--password` then would it not be the same as the existing functionality? Alternatively can you elaborate on how do you expect this command to accept inputs so that I'm sure we are on the same page here.

conradludgate:
It would not be the same functionality. Our impl before didn't handle passwords in the prompt safely

satyarths:
Okay, so while the password prompt would be handled safely, then option of adding the password with `-p` would still be plain text. Is that expected?

conradludgate:
Yeah I think this is expected

satyarths:
Thanks, I've updated the PR. Although I'm not happy with the code that I've written and feel that this could be refactored better. Do let me know your suggestions.

conradludgate:
I have some ideas for a refactor too. But that can be in a separate PR if you want. I'm testing out the password behaviour locally but the code looks good. Much appreciated

conradludgate (approved):
Works well. Ty

PR #430, opened 2022-05-31
Noyez fix dir hostname utf8

noyez:
Bug fix for zsh_histdb import where if a directory name or hostnames had non-ascii characters that the import would crash.


ellie (approved):

ellie:
Thanks for this! 🚀

PR #435, opened 2022-06-04
[feature] Add scroll wheel support to interactive history search

frankh:
Allows you to scroll quickly through the history search using mouse wheel/trackpad instead of just with keys

### DIFF
--- a/src/command/client/search.rs
+++ b/src/command/client/search.rs
@@ -3,7 +3,7 @@ use std::{env, io::stdout, ops::Sub, time::Duration};
 use chrono::Utc;
 use clap::Parser;
 use eyre::Result;
-use termion::{event::Key, input::MouseTerminal, raw::IntoRawMode, screen::AlternateScreen};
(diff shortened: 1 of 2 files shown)


src/command/client/search.rs
```diff
@@ -305,14 +305,14 @@ fn remove_char_from_input(app: &mut State, i: usize) -> char {
...
-    input: Key,
+    input: TermEvent,
     search_mode: SearchMode,
     db: &mut impl Database,
     app: &mut State,
 ) -> Option<String> {
     match input {
-        Key::Esc | Key::Ctrl('c' | 'd' | 'g') => return Some(String::from("")),
-        Key::Char('\n') => {
+        TermEvent::Key(Key::Esc | Key::Ctrl('c' | 'd' | 'g')) => return Some(String::from("")),
```
frankh, line 314: kind of horrible the wrapping all the keys in TermEvents

Another option would be first split into Key/Mouse events and handle separately, but that would lead to duplication of the up/down handler

ellie (approved):

ellie:
Happy with this! Once you've formatted ;P

ellie (approved):

PR #447, opened 2022-06-10
Add configurable history length

### LINKED ISSUE #416: Make server history item length configurable
1) change https://github.com/ellie/atuin/blob/main/atuin-server/migrations/20210425153745_create_history.sql#L8 to a text, not a varchar. This will have no limit
2) add a history length here: https://github.com/ellie/atuin/blob/main/atuin-server/src/settings.rs#L11
3) check it when adding history here, error if too long: https://github.com/ellie/atuin/blob/main/atuin-server/src/handlers/history.rs#L75

I'll get to this soon, but also happy to help someone work on it :)

Some users have very large history items, and it would make things easier + more flexible if this were configurable

ellie:
This allows servers to decide the max length of each history item they
want to store! Some users might have much larger history lines than
others.

This setting can be set to 0 to allow for unlimited history length. This
is not recommended for a public server install, but for a private one it
can work nicely.

Resolves #416

### DIFF
--- a/atuin-server/migrations/20220610074049_history-length.sql
+++ b/atuin-server/migrations/20220610074049_history-length.sql
@@ -1,0 +1,2 @@
+-- Add migration script here
+alter table history alter column data type text;
--- a/atuin-server/src/lib.rs
+++ b/atuin-server/src/lib.rs
@@ -19,7 +19,7 @@ pub mod settings;
 pub async fn launch(settings: Settings, host: String, port: u16) -> Result<()> {
     let host = host.parse::<IpAddr>()?;
 
-    let postgres = Postgres::new(settings.db_uri.as_str())
+    let postgres = Postgres::new(settings.clone())
         .await
         .wrap_err_with(|| format!("failed to connect to db: {}", settings.db_uri))?;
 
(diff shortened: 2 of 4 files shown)


atuin-server/migrations/20220610074049_history-length.sql
```diff
@@ -0,0 +1,2 @@
+-- Add migration script here
+alter table history alter column data type text;
```
ellie, line 2: Text has no max size, unlike varchar

atuin-server/src/lib.rs
```diff
@@ -19,7 +19,7 @@ pub mod settings;
 pub async fn launch(settings: Settings, host: String, port: u16) -> Result<()> {
     let host = host.parse::<IpAddr>()?;
 
-    let postgres = Postgres::new(settings.db_uri.as_str())
+    let postgres = Postgres::new(settings.clone())
```
ellie, line 22: Settings is also used below, and I'm lazy 🤐

atuin-server/src/settings.rs
```diff
@@ -33,6 +34,7 @@ impl Settings {
             .set_default("host", "127.0.0.1")?
             .set_default("port", 8888)?
             .set_default("open_registration", false)?
+            .set_default("max_history_length", 8192)?
```
ellie, line 37: 8192 is the old default set in the database schema, so behaviour does not change

conradludgate (dismissed):

PR #448, opened 2022-06-11
Batch key handling

frankh:
Batch all input event handling at once, then trigger the search query once if needed afterward

This simplifies the code (removed hacky bounded channel) and also fixes copy/paste which was broken.

### DIFF
--- a/src/command/client/search.rs
+++ b/src/command/client/search.rs
@@ -307,8 +307,8 @@ fn remove_char_from_input(app: &mut State, i: usize) -> char {
 }
 
 #[allow(clippy::too_many_lines)]
-async fn key_handler(
-    input: TermEvent,
+fn key_handler(
+    input: &TermEvent,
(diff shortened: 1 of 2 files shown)


src/command/client/search.rs
```diff
@@ -307,12 +307,7 @@ fn remove_char_from_input(app: &mut State, i: usize) -> char {
 }
 
 #[allow(clippy::too_many_lines)]
-async fn key_handler(
-    input: TermEvent,
-    search_mode: SearchMode,
-    db: &mut impl Database,
-    app: &mut State,
-) -> Option<String> {
+fn key_handler(input: &TermEvent, app: &mut State) -> Option<String> {
```
ellie, line 310: much simpler 🙌  thank you!

ellie (approved):
This is great, thanks for sorting it out! :D

PR #458, opened 2022-06-15
Add binstall configuration

michaelmior:
This adds support for installation via [binstall](https://github.com/ryankurte/cargo-binstall). It's simply a set of configuration settings that allows the binstall tool to locate precompiled binaries in the existing GitHub releases. This allows users of binstall to skip the compilation step if a prebuilt binary is available.


conradludgate (approved):

michaelmior:
Thanks!

PR #459, opened 2022-06-15
Add Fig as an installation method to the README

ibayramli:
The [Fig Plugin Store](https://fig.io/plugins) supports 1-click install for 400+ shell plugins. We have over 100k users, thousands of whom manage their shell configuration with Fig.

`atuin` is already listed in the store so we'd love to have it listed as a download method on your README.

Thanks so much and please let me know if you have any questions!

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,14 +192,6 @@ And then follow [the shell setup](#shell-plugin)
 Once the binary is installed, the shell plugin requires installing. If you use
 the install script, this should all be done for you!
 
-### Fig
-
-[Fig](https://fig.io) adds apps, shortcuts, and autocomplete to your existing terminal.
-
-Install `atuin` shell plugin in zsh, bash, or fish with just one click. 
-
-<a href="https://fig.io/plugins/other/atuin" target="_blank"><img src="https://fig.io/badges/install-with-fig.svg" /></a>
-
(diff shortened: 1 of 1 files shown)


ellie (changes requested):
Hey! Thank you for adding support for Atuin, but I'd appreciate it if you don't try and use our README as an advertisement

ibayramli2001:
Hi @ellie, thanks for reviewing my PR.

> Could you move Fig down below the shell-specific installs please? We've tried to keep things roughly in the order of what people installing Atuin use most, and Fig is still super new

Ordering by tool popularity makes sense - I addressed it in the latest commit. My reasoning for putting Fig at top was that it's a shell-agnostic installation method that works for zsh, bash, or fish - so it made sense to me to list more general install methods above more niche ones (just as you put the script-installation at the top, for example).

> The description here reads like an ad - I appreciate you listing your install, but our README is not the place to sell a product or boost SEO.

Sorry to hear that the description felt like an ad - I removed it in the final commit. We just wanted to give a short description of what Fig is since we realize some users are not familiar with it.

conradludgate (approved):
Looks good to me now. Thanks for the update

ellie:
Yup this is great, thank you!

PR #473, opened 2022-07-03
better cursor search

conradludgate:
I noticed we were using char indices for cursor support, but we can use utf8 byte indices for better performance. Refactored into a `Cursor` type with proper tests so we don't accidentally break it in future


conradludgate:
Sorry, those last 2 commits are pretty chunky. I'd suggest reviewing by commit.

The last 2 commits are
* `move methods to state` - where I changed methods from taking `app: &mut State` and instead made them `&mut self` methods within `impl State`.
* `refactor search modules` - where I moved the interactive search and the `Events`/`Cursor` type as well into the `search` directory.

conradludgate:
I can split the PR if you want :P

ellie (approved):

PR #484, opened 2022-07-26
Add support for prepending a path to all routes for the server

morguldir:
I didn't want to set up a domain for this, so I attempted to add support for putting it in a different path.

### DIFF
--- a/atuin-server/src/router.rs
+++ b/atuin-server/src/router.rs
@@ -57,7 +57,7 @@ async fn teapot() -> impl IntoResponse {
     (http::StatusCode::IM_A_TEAPOT, "☕")
 }
 pub fn router(postgres: Postgres, settings: Settings) -> Router {
-    Router::new()
+    let routes = Router::new()
         .route("/", get(handlers::index))
         .route("/sync/count", get(handlers::history::count))
         .route("/sync/history", get(handlers::history::list))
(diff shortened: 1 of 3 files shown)


atuin-server/src/router.rs
```diff
@@ -57,15 +57,20 @@ async fn teapot() -> impl IntoResponse {
...
         .route("/history", post(handlers::history::add))
         .route("/user/:username", get(handlers::user::get))
         .route("/register", post(handlers::user::register))
-        .route("/login", post(handlers::user::login))
+        .route("/login", post(handlers::user::login));
+
+    let path = settings.path.as_str();
+    let valid_path = if !path.is_empty() {path} else {"/"};
+
+    Router::new().nest(valid_path, routes)
```
conradludgate, line 73: I'd personally only wrap if a path is specified. 
```suggestion
    if path.is_empty() {
        routes
    } else {
        Router::new().nest(path, routes)
    }   
```

docs/server.md
```diff
@@ -67,6 +67,12 @@ Defaults to `false`.
 
 A valid postgres URI, where the user and history data will be saved to.
 
+### path
+
+A path to prepend to all the routes of the server.
+
+Defaults to `"/"`
```
conradludgate, line 74: I wonder if we should default to "" and clarify that means no prefix. And since the prefix doesn't need a trailing `/`, I think this would save some potential confusion

conradludgate (commented):
Idea seems ok. I think this is usually a job for nginx or some other ingress but I don't see why we can't support this too

docs/server.md
morguldir: Sure, the reason I did it was that nest() panics if you use an empty route

conradludgate (dismissed):
Thanks

conradludgate (approved):

Issue #356, opened 2022-04-27
Support for SQLite Server Database Backend


RickyGrassmuck:
Just got started setting up my own personal server instance and while it works well once everything is setup, having to run a full Postgresql database for my single user seems a bit overkill.

I did do some poking around through the codebase and it appears that the sqlx package used for the Postgres code also supports SQLite. If implemented, the server would be deployable as a single binary with no external dependencies which would be amazing for personal usecases!

I would have put forth a PR for this but unfortunately i'm not familiar with Rust at all and that learning curve is **_steep_**! If I find some free time one of these days I may take a stab at it.

conradludgate:
We've designed our database code to be fairly flexible for the use cases their intended for. So it's certainly possible that we can switch out the database impls on the fly. I suppose the challenge then is which databases should we choose to maintain going forward.

I do like the idea of the sqlite on server for quick deployments. This plus https://github.com/ellie/atuin/issues/301 would make a good combo for one off setups

RickyGrassmuck:
So I poked at it a bit earlier after looking at the sqlx crate documentation and it was surprisingly very little code change to switch to the crates `Any` module which basically acts as a wrapper around the other drivers and relies on the db connection string to determine what flavor of database it's using.

The hardest part was actually redoing the DB Migration files to make them compatible with SQLite and this likely would deter the use of the `Any::` module. My guess is the easiest path would be to have a config option to select the database type you want to use and just maintain 2 different sets of migration files.

I haven't pushed any code (I'm positive that the changes I made to the migrations would actually break the Postgres database lol) but I did grab this [short screen recording](https://user-images.githubusercontent.com/17223437/165663007-d8139363-e57e-4641-b6a9-0352c733cab3.mov) showing the custom build running and working with SQLite.

I can fork and push up a branch with the horrendous changes I made if you're interested in seeing the Any driver and changes needed for the db migrations. Just let me know and i'll get that pushed up and link to it here.

ellie:
Oh awesome, that's fantastic! We don't use anything particularly Postgres specific and I doubt we will do - beyond indexes and the like, but that will all be in migrations.

I'd love to review a PR for this if you have the time to push something up! Thank you!

RickyGrassmuck:
@ellie Just opened #414 for you to look at, it definitely needs some changes around the configuration side of things lol. 

[...]

PR #491, opened 2022-08-04
Add sqlite server support

ekroon:
I tried implementing #356 while also taking into consideration the feedback on #414 

I ended up copying the Postgres implementation to the Sqlite implementation, which is not great for maintenance, but I did it for 2 reasons:
- A generic implementation didn't work out because I got stuck in trait boundary problems I couldn't really solve
- If I could solve those problems, the solution would be void when implementation start to drift.

#### How to use this?
`db_uri = sqlite:///my/path/to/db.db`  or other uri string as described [here](https://docs.rs/sqlx/latest/sqlx/sqlite/struct.SqliteConnectOptions.html).

#### Possible improvements:
- Instead of making the trait generic. The functions implementing the trait could be forwarded to reusable functions. I didn't try this, it might have the same compiler issues? It would save some duplicate code
- As an Sqlite setup is probably for a single user, adding configuration options which register a user could save setup time. Or alternatively, we could remove registration all together. Both solutions would make it easier to use Atuin in a Gitpod or Codespace setup by just SSH forwarding the local server port and syncing every command.


atuin-server/src/handlers/history.rs
```diff
@@ -10,16 +10,18 @@ use tracing::{debug, error, instrument};
...
     calendar::{TimePeriod, TimePeriodInfo},
-    database::{Database, Postgres},
-    models::{NewHistory, User},
+    database::{DatabaseExtension, DatabaseWrapped},
+    models::{NewHistory, User as ModelUser},
 };
 
 use atuin_common::api::*;
 
+type User<T> = DatabaseWrapped<ModelUser, T>;
```
conradludgate, line 19: What's the point of this exactly? I see no reason the user type needs to be coupled to the DB

ekroon: 👋 @conradludgate I agree! I got stuck implementing `FromRequest` for `User` for both Postgres and Sqlite:

https://github.com/ellie/atuin/pull/491/files?diff=unified&w=0#diff-8da1109a66fb471557b348bd52fcbb14936eba4d31a11d3b1fa30b4218be1bf0R28-R31

I couldn't find a way to retrieve the correct type via `extensions` on the request. Maybe boxing the value could work?
There is no real data coupling btw, it is implemented as `PhantomData`.

If you have an idea / suggestion how to fix this I gladly take a stab at it.

ekroon:
While implementing this feature I found: https://github.com/tokio-rs/axum/discussions/358 but I missed the link to [a dependency injection example](https://github.com/tokio-rs/axum/blob/main/examples/error-handling-and-dependency-injection/src/main.rs) in the Axum repo. 

I changed the implementation for this feature to follow that example. While this probably causes a bit of overhead (with the usage of `Arc`), the code is a lot easier to read.

conradludgate:
I think it matters very little - async trait is already full of dynamic dispatch

conradludgate:
Hey, sorry for the silence. With the announcement of https://tokio.rs/blog/2022-08-whats-new-in-axum-0-6-0-rc1 I've figured out a way to make it generic and not need `DynDatabase` by utilising the type-safe state.

Because of that, I'm gonna wait until axum 0.6 releases then this can be the PR to kick that off :)

conradludgate:
On top of that, I've tested both the postgres and the sqlite DB paths. I'm very happy with this, thanks for putting in the effort :)

evanpurkhiser:
Did anything end up blocking this?

I was interested in running my own personal atuin server in docker, but needing a postgres instance was a bit of a no start for me, so this looks awesome!

conradludgate:
Thanks for the nudge! 

> Did anything end up blocking this?

axum 0.6, but we should get that merged soon #674

As for

> I've figured out a way to make it generic and not need DynDatabase by utilising the type-safe state

If you check out 674, you'll see that handlers take `State<AppState>`. This can actually be generic, so we can offer `Postgres` or `Sqlite` handlers each.

I'll try resume this soon when I have some spare time again, unless @ekroon is up to rebase this onto 674 and understands my idea here.

ekroon:
@conradludgate  I think I understand and will give it a shot this weekend.

cocoonkid:
Thank you all for this initiative. sqlite support is so much better for atuin :)

evanpurkhiser:
Hate to nudge again but would still love to see this!

ekroon:
I started some time ago on some integration tests as mentioned here: https://github.com/ellie/atuin/pull/711 

Will try to pick that up to get this moving again.

Issue #475, opened 2022-07-08
performance issues on initial command listing in interactive mode


Byron:
Atuin is a fantastic tool and it changed my workflow to the better! Thanks for that :)!
Now that I use it all the time it's easy to notice even relatively minor issues, so here I go. Note that further down there is also a workaround for the issue mentioned here, even though I am unable to provide a proper solution at this time.

### How to reproduce

- tested on f946644cebb94bb62646cce6fcf3900fe71e2bea, on MacOS with history of 24k items
- the initial load of `atuin` takes about 100ms
- when typing, then erasing to clear the search, the listing is slow with about 100ms duration, but in bad cases it can take up to 800ms, which makes the experience sluggish and frustrating given how often I use it

#### Profile 

Profiling it showed the database is slow, here is a particularly slow case with ~830ms listing time, with drawing taking no considerable time: 

<img width="1025" alt="Screen Shot 2022-07-08 at 09 38 55" src="https://user-images.githubusercontent.com/63622/177900425-32c8c73e-803b-4693-b55a-e2874cf4dfc5.png">

<img width="584" alt="Screen Shot 2022-07-08 at 09 39 05" src="https://user-images.githubusercontent.com/63622/177900455-7c4797c4-f72a-4dba-b511-cbadb6e25b3f.png">

#### Workaround

Out of a whim I applied this patch as `unique` constraints can be slow due to their potential for exponential runoffs. With it `atuin` became as snappy as never before for the workflow described in _How to reproduce_.

```patch
diff --git a/src/command/client/search.rs b/src/command/client/search.rs
index c50c492c..8ad1ea93 100644
--- a/src/command/client/search.rs
+++ b/src/command/client/search.rs
@@ -262,7 +262,7 @@ async fn query_results(
 ) -> Result<()> {
     let results = match app.input.as_str() {
         "" => {
-            db.list(app.filter_mode, &app.context, Some(200), true)
+            db.list(app.filter_mode, &app.context, Some(200), false)
                 .await?
         }
         i => {
```

#### Possible next steps

Unfortunately I don't have the experience to find a way to make `unique` fast even on relatively large databases, but hoped a more suitable index could do the trick, or a different way of writing the SQL statement.

[...]

PR #493, opened 2022-08-06
Add index for interactive search

LovecraftianHorror:
Related #475 (cc @Byron)

This PR adds an index intended to improve the query for interactive search

I was able to artificially create bad startup times for `atuin search --interactive` by adding several thousand entries to my history with the same command. This stresses the `max(timestamp)` for a given command because there is no index to relate timestamps to their respective commands. With this I was able to reproduce an over 2 second startup time which became well under 1 second after this change

Opening as a draft because I want to see if this impact is noticeable for more organic histories. You can test by manually creating this index on your `history.db`

NOTE: for local testing your system installed `atuin` will get angry if extra migrations are applied that it doesn't recognize. You can still manually add the index using `sqlite3` and it won't notice, but running `atuin` built from this PR while still having an older version installed still can give

```text
Error: migration 20220806155627 was previously applied but is missing in the resolved migrations

Caused by:
    migration 20220806155627 was previously applied but is missing in the resolved migrations

Location:
    src/command/client.rs:68:22
```

### DIFF
--- a/atuin-client/migrations/20220806155627_interactive_search_index.sql
+++ b/atuin-client/migrations/20220806155627_interactive_search_index.sql
@@ -1,0 +1,6 @@
+-- Interactive search filters by command then by the max(timestamp) for that
+-- command. Create an index that covers those
+create index if not exists idx_history_command_timestamp on history(
+	command,
+	timestamp
+);


ellie:
I'd like to properly run the numbers for this, but anecdotally it seems faster :)

Is Atuin any faster with this index for you @chrysn?

conradludgate:
Adding this index increased my database from 10M to 12M. Not hugely worrying.

If we have this, do we need our existing indexes? I deleted them locally to no observable downside but maybe my number of entries is too low to see that impact

conradludgate:
However, I do definitely notice a difference in search times with the new index. LGTM

ellie:
I think let's keep the existing for now and review all indices later

PR #506, opened 2022-08-21
Securize Dockerfile

DucretJe:
* Bump base images to fix several critical vuln
* Run the container as non root

I would also suggest to use a tool like [renovate](https://github.com/renovatebot/renovate) to auto bump the base images.


conradludgate (approved):
Thanks!

ellie (approved):

ellie:
Thank you! We already use dependabot, would be nice to have it update the dockerfile too

PR #507, opened 2022-08-22
Add dockerfile in dependabot scope

DucretJe:
Regarding the discussions in https://github.com/ellie/atuin/pull/506, I suggest to add the dockerfile in dependabot's scope.

It will create more PRs but will help keeping everything up to date :)


ellie (approved):
Awesome! Thanks for this 🙌

PR #509, opened 2022-08-23
Fix atuin crashing on commands that start with --

Sciencentistguy:
The use of `--` in the shell scripts prevents clap attempting to parse
the command name as a flag, in the case that it starts with `--`.

Previously, atuin would crash with the following error, if a command starting with `--` is used:

```shell
jamie@chronos (zsh) in atuin on  fix-double-dash via 🦀 v1.63.0 within  rust-nix-shell
<I> ➜ --example
error: Found argument '--example' which wasn't expected, or isn't valid in this context

	If you tried to supply `--example` as a value rather than a flag, use `-- --example`

USAGE:
    atuin history start [COMMAND]...

For more information try --help

```

### DIFF
--- a/src/shell/atuin.bash
+++ b/src/shell/atuin.bash
@@ -2,7 +2,7 @@ ATUIN_SESSION=$(atuin uuid)
 export ATUIN_SESSION
 
 _atuin_preexec() {
-    local id; id=$(atuin history start "$1")
+    local id; id=$(atuin history start -- "$1")
     export ATUIN_HISTORY_ID="${id}"
 }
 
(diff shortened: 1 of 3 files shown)


conradludgate:
Good find - I think if we do as the suggestion says and prefix the command start prehoox with `--` should fix it

conradludgate:
Oh - I thought this was an issue and not a PR 😂

conradludgate (approved):

ellie:
thanks for catching this! appreciate the fix 🚀

PR #519, opened 2022-09-01
tab for filter-mode panes switching

conradludgate:
Tab to switch filter modes (like session/directory/host etc)


ellie:
99% sure I also DMd you about this, but for clarity;

I'd rather reserve the tab key than bind it to this, I have some other things in mind for it. EG, would be nice to use it to open a "menu" on the selected history item (maybe view context on it, have the option to delete it, etc)

Issue #244, opened 2021-12-17
Scrolling should have an offset that keeps the 1-9 labels in view


mrjones2014:
Gif to demonstrate what I mean:

![CleanShot-2021-12-17-at-08 30 44](https://user-images.githubusercontent.com/8648891/146552086-4e07c63a-b05b-4314-b236-d3c276df680f.gif)

It should start scrolling when the `9` in the leftmost column reaches the top line instead of the numbers scrolling off screen.

mrjones2014:
Looks like its not currently possible to fix this but will be at some point in the future.

https://github.com/fdehau/tui-rs/issues/328

conradludgate:
Should probably be fixed but it's definitely a niche issue, so I wouldn't put priority on this

ellie:
Resolved by #524!

PR #524, opened 2022-09-12
custom history list

conradludgate:
This is a pretty fun one. Uses a custom `HistoryList` impl rather than the default `List` that `tui` provides. This means we can output the history rendering lazy rather than eagerly allocate on every render.

Average CPU usage of rendering seems to have halved, which is nice for battery and latency (although I haven't really observed much by eye).

This also has a fix for #244 by using a custom list state that is aware of the 10 row offset.

https://user-images.githubusercontent.com/6625462/189717736-bc1f7b7c-5c70-41b5-9639-f3f23f1055f3.mov

### DIFF
--- a/src/command/client/search/history_list.rs
+++ b/src/command/client/search/history_list.rs
@@ -69,6 +69,7 @@ impl<'a> StatefulWidget for HistoryList<'a> {
             s.time(item);
             s.command(item);
 
+            // reset line
             s.y += 1;
             s.x = 0;
         }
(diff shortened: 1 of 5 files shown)


conradludgate:
> Average CPU usage of rendering seems to have halved

Benchmark: open the interactive search and spam scroll. record average CPU usage.

Old: I averaged around 12% CPU 😬 
New: I averaged around 4% CPU 🚀

ellie:
Oooh 👀 I'd be happy to merge once conflicts are resolved

Looks like the next release is gonna be much faster :)

src/command/client/search/history_list.rs
```diff
@@ -0,0 +1,175 @@
...
+
+struct DrawState<'a> {
+    buf: &'a mut Buffer,
+    list_area: Rect,
+    x: u16,
+    y: u16,
+    state: &'a ListState,
+}
+
+// longest line prefix I could come up with
```
ellie, line 112: long boi 🙇‍♀️

ellie (approved):
LGTM 🔥🚀

PR #525, opened 2022-09-13
[WIP] no client async

conradludgate:
An experiment for now.

our TUI doesn't really need async, but sqlx and reqwest forced our hands. There's alternatives like rusqlite and ureq though that are sync.

Performance seems the same, but builds seem better when used with client+sync features

Old:
Build time: 1m40s
Dependencies: 386
Binary size: 11M

New:
Build time: 1m05s
Dependencies: 205
Binary size: 6M


conradludgate:
Not sure why tests are failing. They work locally

walkert:
For reqwest you could use the `blocking` feature then just swap out `reqwest::Client` for `reqwest::blocking::Client`.

conradludgate:
You could do that, but it still includes tokio and uses the async API underneath, which I dislike heavily

PR #541, opened 2022-09-24
Fix compatability with fish vi key bindings

shreve:
PR #420 addressed the existence of pagination in a fish prompt by trying to track when pagination was enabled and disabled. This introduced atuin-specific bindings for common keys: `\t`, `\e`, `\r`, and `\n` which export a variable and informs the `_autiun_search` function to not show the TUI.

Fish has a commandline function that will instead tell you whether pagination is enabled so the user doesn't need to keep track of that state. This PR uses this function, `commandline -P` to replace the prior TUI supression scheme. Removing these extra function calls allows us to remove the additional bindings which were breaking fish vi key bindings.

Replacing the value for `bind -M insert \e` completely breaks vi mode by blocking the user from exiting insert mode. By removing these extra binds, this PR restores fish vi mode compatability.

### DIFF
--- a/src/shell/atuin.fish
+++ b/src/shell/atuin.fish
@@ -20,19 +20,11 @@ function _atuin_search
     end
 end
 
-function _atuin_suppress_tui
-    set -gx ATUIN_SUPPRESS_TUI "true"
-end
-
-function _atuin_unsuppress_tui
-    set -ge ATUIN_SUPPRESS_TUI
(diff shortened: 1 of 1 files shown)


ellie:
Oh nice! Thanks for cleaning this up, this is a nicer solution 🙌

ellie (approved):

PR #544, opened 2022-09-24
Allow stateless commands to be run without config/database

c-14:
Fixes an issue where gen-completions fails in restricted build environments while trying to create a config directory.

Settings::new() and Sqlite::new() are currently run unconditionally for every client command, even if those commands don't need access to the settings or database. This would cause running ```atuin gen-completions``` in a restricted build environment to fail while trying to create a path to the config directory.

This PR splits the client commands into two subsets, one subset that requires access to the settings/database and one that does not. Settings::new and Sqlite::new are then only called for the subset of commands that need them.

I think this is the most elegant approach to solving this issue, the other alternative would have been to move the calls to Settings::new and Sqlite::new into the commands' individual run methods. The current approach avoids this code duplication and should be easily extensible in the future if necessary.

The use of clap::flatten on the stateful commands means the commandline interface doesn't change in a user-facing way.

### DIFF
--- a/src/command/client.rs
+++ b/src/command/client.rs
@@ -1,24 +1,21 @@
 use std::path::PathBuf;
 
-use clap::{CommandFactory, Subcommand};
-use clap_complete::{generate, generate_to, Shell};
+use clap::Subcommand;
 use eyre::{Result, WrapErr};
 
 use atuin_client::{database::Sqlite, settings::Settings};
-use atuin_common::utils::uuid_v4;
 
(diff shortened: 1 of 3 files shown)


conradludgate:
Thanks! This was a great idea and something I had in the back of my head for a while. I've moved the stateless commands up a level to the AtuinCmd enum, that seems to fit better in my mental model

conradludgate (dismissed):

conradludgate (approved):

PR #545, opened 2022-09-25
Clear RBUFFER when accepting output from atuin

c-14:
Since we pass $BUFFER to atuin search, retaining RBUFFER (the part of the buffer to the right of the cursor) on accepting a history replacement probably doesn't make sense. The advantage of setting RBUFFER and LBUFFER separately instead of setting BUFFER is that the cursor is positioned after the end of LBUFFER instead of remaining where it was before atuin was called.

Alternatively, _atuin_search() could pass $LBUFFER to atuin instead of $BUFFER which would allow the widget to replace parts of command lines (such as the initial parts of a shell pipeline), but I would prefer the default to replace the whole command line from the history.

A widget that only replaces parts of command lines could be added separately if there's interest in it, though users could always just add it themselves (I'm not sure how many users would want such a feature).

### DIFF
--- a/src/shell/atuin.zsh
+++ b/src/shell/atuin.zsh
@@ -38,6 +38,7 @@ _atuin_search(){
 	echoti smkx
 
 	if [[ -n $output ]] ; then
+		RBUFFER=""
 		LBUFFER=$output
 	fi
 


ellie (approved):
Thanks for this! 🚀 

> A widget that only replaces parts of command lines could be added separately if there's interest in it, though users could always just add it themselves (I'm not sure how many users would want such a feature).

Would you be able to explain a workflow or use case where this might be useful?

c-14:
I was thinking something like a ```curl <endpoint> | jq <query>``` pipeline in which one might want to replace e.q. the call to curl with a call to a different endpoint (potentially with other curl options) that you've used previously and would be in atuin's history, but would then want to keep the same jq query at the end.

I don't think it would be useful to the point that very many people would make use of it though. Replacing the entire command line with the match from atuin seems a lot more generally useful.

PR #552, opened 2022-10-05
Add `style` to config.toml and config.md

xfzv:
I came across #288 by accident, wanted to try compact mode but couldn't figure how to. Turns out the option wasn't added to `config.toml`. Hope that makes sense to add it.

Also added to `config.md`.

### DIFF
--- a/atuin-client/config.toml
+++ b/atuin-client/config.toml
@@ -29,4 +29,4 @@
 
 ## which style to use
 ## possible values: auto, full, compact
-style = "auto"
+#style = "auto"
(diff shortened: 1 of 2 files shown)


atuin-client/config.toml
```diff
@@ -26,3 +26,7 @@
 ## which search mode to use
 ## possible values: prefix, fulltext, fuzzy
 # search_mode = "prefix"
+
+## which style to use
+## possible values: auto, full, compact
+style = "auto"
```
conradludgate, line 32: By convention, we leave default configs commented

```suggestion
#style = "auto"
```

conradludgate (commented):
Looks good, just 1 suggestion. Thanks

conradludgate (approved):

PR #555, opened 2022-10-12
Add automatic update checking

ellie:
Check for updates, and show a message if there is one available!

I'm going to be rolling out deletion support over several versions, as it's _complicated_ and changes how sync and history is transferred. This way, we make it more likely that people will follow the updates on time.

My Rust is a teeny bit rustier than I'd like, it's been a little while 😞

### DIFF
--- a/atuin-client/src/settings.rs
+++ b/atuin-client/src/settings.rs
@@ -129,10 +129,7 @@ impl Settings {
 
         let value = fs_err::read_to_string(path);
 
-        match value {
-            Ok(v) => Some(v),
-            Err(_) => None,
-        }
+        value.ok()
     }
 
     fn save_current_time(filename: &str) -> Result<()> {
--- a/atuin-client/src/settings.rs
+++ b/atuin-client/src/settings.rs
@@ -92,6 +92,7 @@ pub struct Settings {
     pub dialect: Dialect,
     pub style: Style,
     pub auto_sync: bool,
+    pub update_check: bool,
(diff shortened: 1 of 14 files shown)


atuin-client/src/settings.rs
```diff
@@ -99,31 +106,68 @@ pub struct Settings {
...
+        }
+
+        let value = fs_err::read_to_string(path);
 
-        if !sync_time_path.exists() {
-            return Ok(Utc.ymd(1970, 1, 1).and_hms(0, 0, 0));
+        match value {
+            Ok(v) => Some(v),
+            Err(_) => None,
         }
```
conradludgate, line 135:
```suggestion
        value.ok()
```

conradludgate (approved):
Looking good 🚀 🦄

ellie:
Any idea why the tests are failing @conradludgate?

It's not liking the api client import but I don't get why 🤔

atuin-client/src/settings.rs
ellie: Wow this is much neater 🤣

conradludgate:
> It's not liking the api client import but I don't get why 🤔

Ahh yeah. Since that's a sync related feature, it doesn't work in the tests which specify `--no-default-features`.

I split those features out a while back so we could have minimal client-only installs. It shouldn't be too hard to get it to work here either (`needs_update` will just always return None if the feature is disabled)

Issue #517, opened 2022-09-01
Incognito mode


conradludgate:
I think it would be cool if we could do
```sh
atuin incognito
```

And spawn a subshell. In this subshell we will have syncing disabled, and we will be working with a local copy of our normal history database.

When you exit the subshell, the database is 'rolled back'. As if the commands were never run.

There could be a `atuin incognito commit` if you decide you want to keep the commands.

Scenario: you are playing around with a REST API in curl and need to use an auth token, but don't want to set up environment variables and don't want the token in your shell history

panekj:
`fish` has private mode

```
-P or --private
       Enables private mode: fish will not access old or store new history.
```

and sets `fish_private_mode` variable to `1`

conradludgate:
I think I would like to see existing history, but that could be a further flag I guess

andrewcrook:
There needs to be something to add the ability to either pause logging or stop displaying history for things like sudo. May be an option to just filter out items containing configurable keywords such as ‘sudo’ or ‘supercooladmintool'

mozzieongit:
For ZSH I use the following function to toggle atuin history saving. Then I can use my terminal without the commands getting persisted in the database. Searching will still work in incognito mode.

Usage:  
`incognito` to enable incognito and disable saving commands to atuin database. `incognito d` to disable incognito and continue saving commands to the database.

ZSH:
```zsh
incognito () {
    if [[ $1 = disable ]] || [[ $1 == d ]]
    then
        add-zsh-hook precmd _atuin_precmd
        add-zsh-hook preexec _atuin_preexec
    else
        add-zsh-hook -d precmd _atuin_precmd
        add-zsh-hook -d preexec _atuin_preexec
    fi
}
```
If I want those commands to also not be saved in the shell history file, I also enter `export HISTFILE=` (to empty the HISTFILE variable, which stores the file to which zsh will write history)

Bash:
```bash
incognito () {
  if [[ $1 = disable ]] || [[ $1 == d ]]
  then
    # disable incognito
    if [[ -n "${BLE_VERSION-}" ]]; then
      echo "no idea what ble is (no incognito mode available here)"
    else
      precmd_functions+=(_atuin_precmd)
      preexec_functions+=(_atuin_preexec)
    fi
  else
    # enable incognito
    if [[ -n "${BLE_VERSION-}" ]]; then
      echo "no idea what ble is (no incognito mode available here)"
    else
[...]

PR #577, opened 2022-10-21
Respect fish shell's builtin private mode

patricksjackson:
Fish shell has a private mode builtin, where commands are not written to the history file. This change only allows `atuin history start` when the fish shell is not in private mode.

See here for details on Private mode in fish: https://fishshell.com/docs/current/interactive.html?highlight=private#private-mode

Related #517


conradludgate (approved):
Seems acceptable

PR #586, opened 2022-10-29
Allow changing search_mode during interactive search

### LINKED ISSUE #494: feat: add option (cli or interactive input) to switch search mode at runtime
as title suggest

iamkroot:
Press `Ctrl+s` to change the search mode.

Closes #494


patricksjackson:
I tried this, and while I don't know if I'm sold on the UI here I do like the ability to switch the search functions. I would consider doing something like _replacing_ the filter mode name with the new search mode name whenever the search mode is switched, and then reverting back on the next user input. The user is rarely going to switch search modes, and the filter mode is more important to provide context.

Alternately, consider switching to this new display for the rest of the search after the user tries to switch search modes for the first time.

Please don't go changing things on my account though. I'd suggest waiting for Ellie or Conrad to give their opinion.

Unrelated, but I like how you implemented the switching for `search_mode`, we should do that for `filter_mode` as well though probably not in this PR.

atuin-client/src/settings.rs
```diff
@@ -43,13 +43,32 @@ pub enum FilterMode {
...
     pub fn as_str(&self) -> &'static str {
         match self {
-            FilterMode::Global => "GLOBAL",
+            FilterMode::Global => "GLOB",
             FilterMode::Host => "HOST",
-            FilterMode::Session => "SESSION",
-            FilterMode::Directory => "DIRECTORY",
+            FilterMode::Session => "SESS",
+            FilterMode::Directory => "DIR",
         }
```
panekj, line 72: Why change this?

iamkroot:
> I don't know if I'm sold on the UI here

Agreed, even I am a bit hesitant to truncate the filter modes like this. Maybe show the search mode on the right side? Open to more suggestions on the display.

> _replacing_ the filter mode name with the new search mode name whenever the search mode is switched, and then reverting back on the next user input

That does sound useful, but I'm not sure how easy it'll be to implement. It might complicate the state logic a lot.

> I like how you implemented the switching for `search_mode`

Thanks!

atuin-client/src/settings.rs
iamkroot: Needed to fit `filtermode | searchmode` in the 10-ish chars width available in the left panel. Happy to revert once we design a better UI.

ellie:
Hey! Sorry for being slow here 🙃 

I like the idea for sure - but like @patricksjackson, the UX I am less of a fan of. The user will be switching search mode far less often (I suspect, anyway), so making it such a prominent part of the UI does not feel right to me. If we could put it elsewhere then it would work?

Or even having a character to indicate the search mode 🤔 `^` for prefix, `*` or `%` or similar for fuzzy, etc

iamkroot:
> replacing the filter mode name with the new search mode name whenever the search mode is switched, and then reverting back on the next user input

this sounded like the best approach to me. I have updated the code to do exactly this. Here's an asciicast for the new UI-
[asciicast](https://asciinema.org/a/MGJ3JY3ooi80JYbkQqe8HZJFE)

(I'm slightly irked by the lack of padding to the left of `SRCH:` when it is shown. Happy to discuss alternative words for it.

atuin-client/src/settings.rs
iamkroot: Reverted.

src/command/client/search/interactive.rs
```diff
@@ -395,9 +402,15 @@ impl State {
     }
 
     fn build_input(&mut self, compact: bool, chunk_width: usize) -> Paragraph {
+        let (pref, pref_width, mode, mode_width) = if self.search.switched_search_mode {
+            ("SRCH:", 6, self.search.search_mode.as_str(), 8)
+        } else {
+            ("", 0, self.search.filter_mode.as_str(), 14)
+        };
+        // just a sanity check to ensure we don't exceed the layout limits
+        debug_assert_eq!(pref_width + mode_width, 14);
```
conradludgate, line 411:
```suggestion
        let (pref, mode) = if self.search.switched_search_mode {
            (" SRCH:", self.search.search_mode.as_str())
        } else {
            ("", self.search.filter_mode.as_str())
        };
        let mode_width = 14 - pref.len();
```

conradludgate (approved):
Tysm!!

patricksjackson:
That looks great! Thanks for finishing this, I'm looking forward to using it :)

We should probably add some documentation for this. I'd do it myself but I'm headed out on vacation soon.

PR #593, opened 2022-11-03
Build ARM64 releases & update runners

Absolucy:
Not entirely sure how to test this workflow.

I updated the runners as the current ones being used are deprecated, and GitHub really seems to want people to move to newer ones.

I'm confused: why weren't ARM64 builds being made? The workflow literally has cases to handle it...

Fixes https://github.com/ellie/atuin/issues/369

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -21,8 +21,10 @@ jobs:
       fail-fast: false
       matrix:
         job:
-          - { os: ubuntu-18.04 , target: x86_64-unknown-linux-gnu    }
-          - { os: macos-10.15  , target: x86_64-apple-darwin         }
+          - { os: ubuntu-20.04 , target: x86_64-unknown-linux-gnu    }
+          - { os: ubuntu-20.04 , target: aarch64-unknown-linux-gnu    }
+          - { os: macos-12     , target: x86_64-apple-darwin         }
+          - { os: macos-12     , target: aarch64-apple-darwin         }
     steps:
     - name: Checkout source code
       uses: actions/checkout@v2


.github/workflows/release.yaml
```diff
@@ -21,8 +21,10 @@ jobs:
       fail-fast: false
       matrix:
         job:
-          - { os: ubuntu-18.04 , target: x86_64-unknown-linux-gnu    }
-          - { os: macos-10.15  , target: x86_64-apple-darwin         }
+          - { os: ubuntu-20.04 , target: x86_64-unknown-linux-gnu    }
+          - { os: ubuntu-20.04 , target: aarch64-unknown-linux-gnu    }
+          - { os: macos-12     , target: x86_64-apple-darwin         }
+          - { os: macos-12     , target: aarch64-apple-darwin         }
```
Absolucy, line 27: if I weren't lazy and sleep-deprived right now, I could probably make a fancy workflow that makes a fat binary containing both architecture... but I'm lazy and sleep-deprived right now.

ellie:
> I'm confused: why weren't ARM64 builds being made? The workflow literally has cases to handle it...

I very vaguely remember there being a case where a dependency wouldn't built for ARM, while on CI. 

ARM for mac was much newer then though

.github/workflows/release.yaml
ellie: Hahaha, either way it's much better off now so thank you!

ellie (approved):
Yay thanks for this! 🚀

I'm happy to merge as is, we can see how it works on the next release. It won't affect any existing binary builds so worst case nothing changes

Absolucy:
I think I figured out why it failed: https://github.com/rust-lang/rust/issues/28924#issuecomment-442260036

ellie:
> I think I figured out why it failed: [rust-lang/rust#28924 (comment)](https://github.com/rust-lang/rust/issues/28924#issuecomment-442260036)

ohh interesting! we can try that for the next release then, thank you!

PR #597, opened 2022-11-05
fix #596: Avoid drawing a command section beyond the width  of the list

### LINKED ISSUE #596: Error thrown in main when printing long history items
It seems there's currently a bug in `main` where attempting to print long history items can cause the following error to be thrown:

```
thread 'main' panicked at 'attempt to subtract with overflow', src/command/client/search/history_list.rs:169:17
```

s0:
For long commands that had sections (i.e. whitespace) that started beyond the width of the list, an underflow error would occur. This fixes this error by avoiding trying to print a command section beyond this boundary.

Fixes #596

### DIFF
--- a/src/command/client/search/history_list.rs
+++ b/src/command/client/search/history_list.rs
@@ -159,7 +159,7 @@ impl DrawState<'_> {
 
         for section in h.command.split_ascii_whitespace() {
             self.x += 1;
-            if (self.x > self.list_area.width) {
+            if self.x > self.list_area.width {
                 // Avoid attempting to draw a command section beyond the width
                 // of the list
                 return;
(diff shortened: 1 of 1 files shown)


src/command/client/search/history_list.rs
```diff
@@ -159,6 +159,11 @@ impl DrawState<'_> {
 
         for section in h.command.split_ascii_whitespace() {
             self.x += 1;
+            if (self.x > self.list_area.width) {
```
conradludgate, line 162:
```suggestion
            if self.x > self.list_area.width {
```
To make clippy happy :D

conradludgate (approved):

conradludgate (approved):

PR #612, opened 2022-11-14
Fix logging in _atuin_preexec

tomasol:
Override `RUST_LOG` in `_atuin_preexec` in order to avoid unexpected logs when `RUST_LOG` is set in bash.

### DIFF
--- a/src/shell/atuin.zsh
+++ b/src/shell/atuin.zsh
@@ -9,11 +9,11 @@
 # Source this in your ~/.zshrc
 autoload -U add-zsh-hook
 
-export ATUIN_SESSION=$(atuin uuid)
+export ATUIN_SESSION=$(RUST_LOG=error atuin uuid)
 export ATUIN_HISTORY="atuin history list"
 
 _atuin_preexec(){
-	local id; id=$(atuin history start -- "$1")
+	local id; id=$(RUST_LOG=error atuin history start -- "$1")
 	export ATUIN_HISTORY_ID="$id"
 }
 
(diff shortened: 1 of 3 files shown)


patricksjackson:
I would suggest making this change for all the `atuin` invocations, and for the other shells as well.

src/shell/atuin.zsh
```diff
@@ -9,11 +9,11 @@
 # Source this in your ~/.zshrc
 autoload -U add-zsh-hook
 
-export ATUIN_SESSION=$(atuin uuid)
+export ATUIN_SESSION=$(RUST_LOG=error atuin uuid)
 export ATUIN_HISTORY="atuin history list"
```
tomasol, line 13: not sure what this is for

tomasol:
> I would suggest making this change for all the `atuin` invocations, and for the other shells as well.

I've made the change to all 3 shells, however it needs to be tested because I do not have fish or zsh installed.

patricksjackson (approved):
I tested that this works without error in fish, zsh, and bash.

src/shell/atuin.zsh
patricksjackson: It's unused. Looks like it's a leftover from an older version of the script.

conradludgate:
Instead of this, I think we said a while back we'd configure the env var to be `ATUIN_LOG`

patricksjackson:
> Instead of this, I think we said a while back we'd configure the env var to be ATUIN_LOG

Oh, I didn't even know you could do that. That would probably be nicer, and looks fairly simple - https://rust-lang-nursery.github.io/rust-cookbook/development_tools/debugging/config_log.html#use-a-custom-environment-variable-to-set-up-logging

ellie:
Closed in favour of: https://github.com/ellie/atuin/pull/642

PR #623, opened 2022-11-19
Handle multiline commands in fish shell

### LINKED ISSUE #622: Can't press up arrow while running multiline command in Fish without opening atuin
Pressing <kbd>^</kbd> in fish while writing a multi line command (to go and edit the previous line) causes atuin to open.
Ideally, atuin should only open when the cursor is on the first line and <kbd>^</kbd> is pressed

patricksjackson:
Clean up the `_atuin_bind_up` function for the fish shell.

* Gives control back to fish in search- or paging-mode
* Correctly handles multiline commands

A user can still enter fish's search-mode by invoking `up-or-search` directly, so we probably want to let them do that.

Tested manually

Closes #622

### DIFF
--- a/src/shell/atuin.fish
+++ b/src/shell/atuin.fish
@@ -23,10 +23,19 @@ function _atuin_search
 end
 
 function _atuin_bind_up
-    if commandline -P
+    # Fallback to fish's builtin up-or-search if we're in search or paging mode
+    if commandline --search-mode; or commandline --paging-mode
         up-or-search
-    else
-        _atuin_search
+        return
+    end
+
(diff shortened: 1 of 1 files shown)


conradludgate (approved):
Makes sense! Thanks!

krobelus:
What's the reason for delegating to fish's native behavior if fish has a search active?
This causes a minor conflict with native search, see https://github.com/fish-shell/fish-shell/issues/10584#issuecomment-2423657650
Maybe we can revert that part of this PR (the other two changes seem fine).
I haven't used atuin myself so I probably don't understand how it's used

PR #629, opened 2022-11-25
client filtering done in query

jean-santos:
As title says, instead of filtering a vec, now it's in the query

### DIFF
--- a/atuin-client/src/database.rs
+++ b/atuin-client/src/database.rs
@@ -25,6 +25,17 @@ pub struct Context {
     pub hostname: String,
 }
 
+#[derive(Default, Clone)]
+pub struct OptFilters {
+    pub exit: Option<i64>,
+    pub exclude_exit: Option<i64>,
+    pub cwd: Option<String>,
+    pub exclude_cwd: Option<String>,
+    pub before: Option<String>,
+    pub after: Option<String>,
+    pub limit: Option<i64>,
+}
+
 pub fn current_context() -> Context {
     let Ok(session) = env::var("ATUIN_SESSION") else {
         eprintln!("ERROR: Failed to find $ATUIN_SESSION in the environment. Check that you have correctly set up your shell.");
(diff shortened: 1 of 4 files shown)


atuin-client/src/database.rs
```diff
@@ -458,6 +469,36 @@ impl Database for Sqlite {
...
+            Some(opts) => (
+                opts.exit,
+                opts.exclude_exit,
+                opts.cwd,
+                opts.exclude_cwd,
+                opts.before,
+                opts.after,
+            ),
+            None => (None, None, None, None, None, None),
+        };
```
conradludgate, line 482: Haven't quite looked into what these queries are doing, but for this I'd suggest making the opt_filter param not optional and instead derive(Default) on the struct. It will default to all None internally which matches the logic you have here

conradludgate (commented):
I like the principal, thanks. I think we were meaning to do something like this for a while

conradludgate:
Are there any other fields we can add to the opt filters? Like limit?

atuin-client/src/database.rs
jean-santos: Something like this? 9597ad5faad8804720a0e23c618b537cd9942b3c

jean-santos:
> to the opt filters? Like limit?

Added limit too on the last commit!

atuin-client/src/database.rs
```diff
@@ -458,6 +469,32 @@ impl Database for Sqlite {
             }
         };
 
+        filter_options
+            .exit
+            .map(|exit| Some(sql.and_where_eq("exit", exit)));
```
conradludgate, line 474: I think you don't need the `Some` in this and the following lines

conradludgate (commented):
Thanks, this is looking a lot better. I'll give it a test later

atuin-client/src/database.rs
jean-santos: Indeed, my mistake! It was some leftovers while trying to use and_then.
Removed on the last commit!
By the way, did you see https://github.com/ellie/atuin/commit/9597ad5faad8804720a0e23c618b537cd9942b3c#r91116514 about `interim::Dialect::Uk` ?

ellie:
This is much nicer! Thanks for the work, and sorry for being slow 😞 

If we get the conflicts resolved I'm happy to get this merged

ellie (approved):

ellie:
Thank you!! 🙏

PR #630, opened 2022-11-25
wip: Highlights results

jean-santos:
Hello!
I'm opening this PR to introduce the idea and get your opinion on this!
 
~I started with the idea of highlighting the output for fzf (which I currently use), while at that, I saw that is possible to use the same strategy to highlight TUI output too. Though I'm not sure if I should keep both on the same PR.~

~What it is somewhat working right now:
`atuin search --fzf-mode QUERY | fzf --ansi`
--fzf-mode invert the order of the output (fzf style : the first line is the most relevant) and highlight the result's~
image

~The current tui `atuin search -i`~
image

Update : I'm working on this again with the goal to only highlight on both tui and --cmd-only

PR #631, opened 2022-11-27
Windows version

jean-santos:
Experiment disabling tui+termion and using fzf as atuin's frontend.

It seems that the basic functionalities are ok!
I've added a powershell script to remap Ctrl+R and Enter so we can log the commands

The requirements are:
- Having fzf installed
- Specify the path inside the powershell script provided
- Put the powershell script on $profile path (On my PC was ~\Documents\WindowsPowerShell)

Todo:
- Save env variables like RUST_LOG and FZF_DEFAULT_COMMAND on the powershell script to restore then afterwards
- Generate the powershell script via atuin cli
- Write the proper instructions on README.md

This is all I can think of right now

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,8 +56,6 @@ directories = "4"
 indicatif = "0.17.1"
 serde = { version = "1.0.145", features = ["derive"] }
 serde_json = "1.0.86"
-tui = { version = "0.19", default-features = false, features = ["termion"] }
-termion = "1.5"
 unicode-width = "0.1"
 itertools = "0.10.5"
 tokio = { version = "1", features = ["full"] }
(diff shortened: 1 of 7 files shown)


jean-santos:
By the way, the most relevant match for fzf is different from what (as you can see on the image below, where the query is ls)
`atuin search --cmd-only QUERY` 
outputs now and that will be fixed when #630 is done!
image

patricksjackson:
Just to add a link here because this mentions Windows, switching the backend from termion to crossterm, #331, is how atuin would fully support Windows.

ellie:
Hey! Thanks for doing the work here, but I don't think we can support windows in any official capacity.

Atm neither @conradludgate nor myself have regular access to a windows box, so releases would not be tested for windows

Given how we can only just support bash (and omg, it has a lot of problems), I don't want to add windows to that list too

I'd be happy to have a build that "might work" for windows, but I wouldn't want to claim that we support it in any way

PR #639, opened 2022-12-07
install.sh: Fallback to using cargo

### LINKED ISSUE #590: Fedora Linux installation fails when `redhat-lsb-core` package is not installed
Installation on Fedora 36 fails unless is the `redhat-lsb-core` package installed.

Without `redhat-lsb-core`:
```
Detected Linux!
Checking distro...
lsb_release could not be found, unable to determine your distribution
If you are using Arch Linux, please get community/lsb-release
```

The installation succeeds with `redhat-lsb-core` and `crate` packages installed.

patricksjackson:
Offers to fallback to installing from cargo when we can't determine the distro.

Closes #590

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -87,8 +87,9 @@ __atuin_install_linux(){
     else
         if ! command -v lsb_release &> /dev/null; then
             echo "lsb_release could not be found, unable to determine your distribution"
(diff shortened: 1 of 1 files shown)


ellie (approved):
Thank you!

PR #642, opened 2022-12-16
Switch to using ATUIN_LOG instead of RUST_LOG

ellie:
Instead of https://github.com/ellie/atuin/pull/612


ellie:
Oh yeah I also ditched pretty_env_logger, I wouldn't say that it's _unmaintained_, just very sleepy

conradludgate (approved):

PR #645, opened 2022-12-18
Updates

ellie:
Just combine a bunch of dependabot updates


conradludgate:
Clippy says your code is bad

ellie:
Well I say that clippy is bad

PR #649, opened 2022-12-20
Add Uninstall section

damascene:


README.md
```diff
@@ -253,6 +253,21 @@ Install `atuin` shell plugin in zsh, bash, or fish with [Fig](https://fig.io) in
...
+
+```
+sudo apt remove atuin
+```
+Then you have to comment out these two lines from your `~/.bashrc` file:
+  
+```
+[[ -f ~/.bash-preexec.sh ]] && source ~/.bash-preexec.sh
+eval "$(atuin init bash)"
+```
```
conradludgate, line 268: This assumes bash. I think we can have the user refer above for their shell to see which lines to remove.

We should also not really recommend users remove the dependencies like preexec incase they need it for something else that might break if removed

damascene: I've did some few changes to address it, please check.

ellie:
A couple comments from me too

1. I'd like to avoid providing instructions specifically targetting one OS, or one shell. Even as an exampke
2. I'd also like this to be a separate section, and not a sub-part of the install instructions

damascene:
Hey @ellie  could you please elaborate on 

> A couple comments from me too
> 
>     1. I'd like to avoid providing instructions specifically targetting one OS, or one shell. Even as an exampke
 
Hey @ellie , do suggest putting instruction for each os and shell or just not mention any specific one?

>     2. I'd also like this to be a separate section, and not a sub-part of the install instructions

Do you mean I shouldn't have the Uninstall section next to Install in the table of content, or that I should have Uninstall in an another page? currently Uninstall section is on a level of its own just before the "...what's with the name?" section in the file

PR #652, opened 2022-12-21
Rework `atuin init`

Sciencentistguy:
This allows users to disable the CTRL-R and Up Arrow bindings, independently from one another.

The default is to bind both, but by adding `--disable-ctrl-r` or `--disable-up-arrow` (or both!) to the invocation of `atuin init <shell>` in the shell rc file, each can be disabled.

### DIFF
--- a/docs/key-binding.md
+++ b/docs/key-binding.md
@@ -1,11 +1,27 @@
 # Key binding
 
-By default, Atuin will rebind both <kbd>Ctrl-r</kbd> and the up arrow. If you do not want
-this to happen, set ATUIN_NOBIND before the call to `atuin init`
+By default, Atuin will rebind both <kbd>Ctrl-r</kbd> and the up arrow.
 
-For example
+You can also disable either the up-arrow or <kbd>Ctrl-r</kbd> bindings indivdually, by passing
+`--disable-up-arrow` or `-disable-ctrl-r` to the call to `atui init`:
 
 ```
+# Bind ctrl-r but not up arrow
+eval "$(atuin init zsh --disable-up-arrow)"
+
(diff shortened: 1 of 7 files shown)


conradludgate (commented):
Interesting idea. I think it makes sense and I don't see a reason this would be breaking

ellie:
I like this, thank you! 

Totally happy to merge this as-is, although if you could also update our documentation that would be great

Perhaps highlighting the disabling of `up` in the main README file too would be good, lots of people ask about this

docs/key-binding.md
```diff
@@ -1,11 +1,27 @@
 # Key binding
 
-By default, Atuin will rebind both <kbd>Ctrl-r</kbd> and the up arrow. If you do not want
-this to happen, set ATUIN_NOBIND before the call to `atuin init`
+By default, Atuin will rebind both <kbd>Ctrl-r</kbd> and the up arrow.
 
-For example
+You can also disable either the up-arrow or <kbd>Ctrl-r</kbd> bindings indivdually, by passing
```
ellie, line 5:
```suggestion
You can also disable either the up-arrow or <kbd>Ctrl-r</kbd> bindings individually, by passing
```

docs/key-binding.md
```diff
@@ -1,11 +1,27 @@
...
+`--disable-up-arrow` or `-disable-ctrl-r` to the call to `atui init`:
```
ellie, line 6:
```suggestion
`--disable-up-arrow` or `-disable-ctrl-r` to the call to `atuin init`:
```

ellie (approved):

ellie:
Thank you for this! 🚀

PR #659, opened 2023-01-10
chore(#621): Exit atuin when pressing down arrow on most recent entry

### LINKED ISSUE #621: Exit atuin when pressing down arrow on most recent entry
It'd be great if pressing <kbd>▼</kbd> when the most recent entry to the history is highlighted would exit atuin and return to the shell prompt, similar to how history works in most shells (press up to check last command, then down to return to prompt)

BapRx:
Fixes #621

Signed-off-by: Baptiste Roux <arte.but.posix@gmail.com>


ellie:
Yeah this totally makes sense to me! Thanks for the work 🚀

ellie (approved):

PR #662, opened 2023-01-11
feat(history): Add new flag to allow custom output format

### LINKED ISSUE #619: Show directory in `atuin history list`
I'd like to see the directory of each command in the history display.

BapRx:
Fixes #619 

Hey, that's an easy one to implement but I'm not sure what's the best way to display it, here's a few examples:
```
❯ cargo run history list 2>/dev/null | tail -n1
2023-01-11 20:35:26     < /home/baptiste/github.com/BapRx/atuin >       cargo run history list 2>/dev/null | tail -n1   0s
❯ cargo run history list 2>/dev/null | tail -n1
2023-01-11 20:35:44     === /home/baptiste/github.com/BapRx/atuin ===   cargo run history list 2>/dev/null | tail -n1   0s
❯ cargo run history list 2>/dev/null | tail -n1
2023-01-11 20:36:00     (/home/baptiste/github.com/BapRx/atuin) cargo run history list 2>/dev/null | tail -n1   0s
❯ cargo run history list 2>/dev/null | tail -n1
2023-01-11 20:36:17     [/home/baptiste/github.com/BapRx/atuin] cargo run history list 2>/dev/null | tail -n1   0s
❯ cargo run history list 2>/dev/null | tail -n1
2023-01-11 20:37:10     /home/baptiste/github.com/BapRx/atuin$  cargo run history list 2>/dev/null | tail -n1   0s
```

If we want this to be optional I can add a flag like `--show-cwd`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -91,10 +91,10 @@ dependencies = [
  "itertools",
  "log",
  "rpassword",
+ "runtime-format",
  "semver",
  "serde",
  "serde_json",
- "strfmt",
  "termion",
  "tokio",
  "tracing-subscriber",
@@ -1649,6 +1649,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "runtime-format"
+version = "0.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b035308411b1af4683acc4fc928366443f08b893bb73e235c85de4c2be572495"
+dependencies = [
+ "tinyvec",
+]
+
+[[package]]
 name = "rustix"
 version = "0.36.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2032,12 +2041,6 @@ dependencies = [
  "tokio",
  "tokio-rustls",
 ]
-
-[[package]]
-name = "strfmt"
-version = "0.2.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "26cdabcdab6da7e8c2ac1160e917ec83e78bbe3e10325e17d532718c67a4828f"
 
 [[package]]
 name = "stringprep"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -94,6 +94,7 @@ dependencies = [
  "semver",
  "serde",
  "serde_json",
+ "strfmt",
  "termion",
  "tokio",
  "tracing-subscriber",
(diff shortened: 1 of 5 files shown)


conradludgate:
I wonder at what point we should provide a formatting flag.

I think at this point I'd be hesitant to change the current format since some users might be piping it into other scripts and that might break (eg cut + fzf)

BapRx:
Yes that's what I thought, a flag is definitely required to avoid breaking anything.

The format flag is a good idea! I have no idea how to code this but I'll give it a try. Thanks for your input 👍

BapRx:
Hey @conradludgate, I implemented a new `--format` flag to allow custom output format. Is that what you had in mind? The default behavior is to use the current format ([human](https://github.com/ellie/atuin/blob/62f001898add1edda5db377ad5e96867b72b9b2b/src/command/client/history.rs#L103) or [regular](https://github.com/ellie/atuin/blob/62f001898add1edda5db377ad5e96867b72b9b2b/src/command/client/history.rs#L115)).

Here's a few examples:
```bash
❯ atuin history last -h
Get the last command ran

Usage: atuin history last [OPTIONS]

Options:
      --human
      --cmd-only         Show only the text of the command
  -f, --format <FORMAT>  Output format, you can use the special variables {command}, {directory}, {duration} and {time}. Example: --format "{time} - [{duration}] - {directory}$\t{command}"
  -h, --help             Print help information
```

```bash
❯ # Use default formatting
❯ atuin history last
2023-01-12 16:31:26     atuin history last -h   250ms
```

```bash
❯ # Use custom format
❯ atuin history last -f "{time} - [{duration}] - {directory}$\t{command}"
2023-01-12 16:31:52 - [265ms] - /home/baptiste/github.com/BapRx/atuin$  atuin history last
```

```bash
❯ # Exit on format error
❯ atuin history last -f "{time} - [{duration}] - {this_variable_doesnt_exist} - {directory}$\t{command}"
ERROR: History formatting failed with the following error: KeyError(Invalid key: this_variable_doesnt_exist)
If your formatting string contains curly braces (eg: {var}) you need to escape them this way: {{var}}.

❯ atuin history last -f "{time} - [{duration}] - {{this_variable_doesnt_exist}} - {directory}$\t{command}"
2023-01-12 16:33:27 - [278ms] - {this_variable_doesnt_exist} - /home/baptiste/github.com/BapRx/atuin$   atuin history last -f "{time} - [{duration}] - {this_variable_doesnt_exist} - {directory}$\t{command}"
```

conradludgate:
Wow that was quick! It looks exactly how I had in mind, thanks!

I'll review the code when I have some spare time. @ellie what do you think of this idea before we continue any further?

ellie:
I really like this! Thank you for the effort 🚀 

I'll leave it to @conradludgate to handle the review heh

conradludgate:
Thanks for this, I made a few changes for performance, and I've added the user/host field as @tuxruffian pointed out. I've tested this quite a bit and it works well. Thanks for taking the time to design this :D

conradludgate (approved):

PR #680, opened 2023-01-30
docs(README): add static activity graph example

eripa:
This PR adds a snapshot of the activity graph instead of relying on live loading it. Not 100% sure, but I think some browsers, or GitHub, isn't allowing loading resources from other domains for security purposes (at least not when it's not an obvious static image link).

resolves ellie/atuin#678


ellie:
Awh that's super annoying, it used to work :(

Anyways, thanks for the fix!

ellie (approved):

PR #683, opened 2023-02-05
Fix CI build badge

s1ck:
see https://github.com/badges/shields/issues/8671

also, nice talk at FOSDEM :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@
 <hr/>
 
 <p align="center">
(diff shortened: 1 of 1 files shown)


conradludgate (approved):

ellie:
Thank you! For both the fix and the kind words ☺️

ellie:
Hmmm I'll check what clippy is on about separately

PR #688, opened 2023-02-05
Allow using existing key file on login

trygveaa:
If the given key on login in empty, keep the existing key file rather
than overwriting it with an empty file. This is useful if you log out
and want to log in again and still use the same key, or if you have
copied over the key file rather than providing it as input.

Before the first commit, an empty key file would be created if key
wasn't specified, and after that commit, the key file would not be
created if the key wasn't specified and stay empty if it was empty.

With the second commit the log command checks the key file if a key is
not specified and exits with an error message if either the key file
couldn't be opened or is empty. If a key is specified, the key file is
just created with it as before.

(Btw, great talk at fosdem!)

### DIFF
--- a/src/command/client/sync/login.rs
+++ b/src/command/client/sync/login.rs
@@ -1,7 +1,7 @@
 use std::io;
 
 use clap::Parser;
-use eyre::{bail, ContextCompat, Result};
+use eyre::{bail, Context, ContextCompat, Result};
 use tokio::{fs::File, io::AsyncWriteExt};
 
 use atuin_client::{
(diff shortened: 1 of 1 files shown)


src/command/client/sync/login.rs
```diff
@@ -40,8 +43,25 @@ impl Cmd {
...
+        if key.is_empty() {
+            let mut file = match File::open(key_path).await {
+                Err(_e) => bail!("key file couldn't be opened, key must be specified"),
+                Ok(f) => f,
+            };
+            let mut bytes = Vec::new();
+            file.read_to_end(&mut bytes).await?;
+            if bytes.is_empty() {
+                bail!("key file empty, key must be specified");
+            }
```
conradludgate, line 59:
```suggestion
            let bytes = tokio::fs::read(key_path)
                .await
                .context("key file couldn't be opened, key must be specified")?;

            if bytes.is_empty() {
                bail!("key file empty, key must be specified");
            }
```
(the `context` here needs `eyre:: WrapErr` to be imported)

conradludgate:
Seems ok, I'm a little skeptical on it's value, but I'm overall not against this

conradludgate:
(you can ignore the clippy failure, being fixed in #686)

src/command/client/sync/login.rs
trygveaa: Thanks! I have written very little rust, so not surprised it could be nicer. I applied the patch and included the import.

trygveaa:
> Seems ok, I'm a little skeptical on it's value, but I'm overall not against this

I'd argue that the first commit is useful at least. After registering, I logged out and in again because I like to confirm that the login works and got the credentials I expect. On the key prompt, I assumed it would just use the existing key if I didn't specify anything, so I ended up with an empty key file.

The second commit could be better. I think it makes sense to verify that the key is valid on login, but it would be better to actually verify that it's a valid key. Just checking that it's not empty isn't that useful, especially since you shouldn't end up with an empty key after the first commit. (And I see now that the sync panics if the key isn't valid, so there shouldn't be any security concern with an empty/weak key here).

If we were to add a check that the key is valid, I think it would make sense to also do that check for key values specified at the prompt.

conradludgate (approved):
I'll also ask for @ellie's judgement, but I am happy to accept this

ellie (approved):
This looks good to me! Definitely an edge case, but probably a very frustrating one if you happen to be caught by it! Thanks for the contribution 😊 

> (Btw, great talk at fosdem!)

Thank you! 🙏

PR #692, opened 2023-02-06
Make the install script more robust

frukto:
Dont't fail is a Linux system does not provide lsb_release, try to read the distro from /etc/os-release in this case.


ellie (approved):

ellie:
Thank you! 🙏

Issue #386, opened 2022-05-03
Improvements to key sharing


conradludgate:
Logging in on a new device is currently a bit awkward. You have to somehow get your encryption key from one device to the other.

You could store the key in a password manager (which does seem like the best option to promote)

<details>
  <summary>Tangent...</summary>
Putting the key next to your password can be a security risk, although usually if they have your password manager, you have bigger worries.

But anyway, maybe we could switch to a PBKDF? - this does reduce security considerably though if you have crap passwords, or someone cracks your password from the DB hash.
</details>

Another alternative for users without PMs could be some kind of LAN HTTP server that can acts as a share mechanism.

Scenario: Device 1 is logged in to the sync server, Device 2 is not.

Device 2:
```sh
$ atuin login --remote # feel free to bikeshed

Setting up remote login
From the device, run
atuin login --remote 192.168.1.123:8723 123456
```

This sets up a HTTP server listening on a random port, and shares a OTP.

Device 1 then the displayed command
```sh
$ atuin login --remote 192.168.1.123:8723 123456
```

This will make a HTTP request to that host
```http
GET /

HTTP/1.0 200 OK

123456
```

Device 1 will confirm that the responded code matches the one provided on the CLI. This confirms that the connection is most likely correct.

<details>
  <summary>Security...</summary>
We probably want some encryption for this communications, although it should be secure over LAN...

I don't think TLS would work to solve MITM attacks since there is no CA to confirm the certs.

Maybe the OTP could act as part of a DH exchange. This is probably over-engineering this though
</details>

Device 1 will then follow up by sending the username and key through that API.

Now that Device 2 has this data, it will prompt the user for their password to complete the login.

ellie:
Yeah this definitely needs improving! I chose not to solve this problem earlier on 😂 

I think we can get at least some of the way by making the key type-able and more memorable. Right now if you try and type the output of `atuin key`, you're not going to get anywhere. I think if we do something similar to [bip39](https://docs.rs/tiny-bip39/latest/bip39/), we'd at least reduce the difficulty here.

I'm also wondering how many of our users are on a lan often, vs an install on their laptop, a VPS, work computer, etc etc.

conradludgate:
Transfer to VPS seems easy enough through SSH. My flow was thinking of a work laptop brought home, next to their personal workstation.

Textual keys would definitely help though, if they need to be written down on paper and taken elsewhere

ellie:
> Transfer to VPS seems easy enough through SSH

[...]

PR #694, opened 2023-02-06
display mnemonic key

conradludgate:
Demo of #386


ellie (approved):
Very nice! Using the bip makes sense

PR #695, opened 2023-02-06
skim-demo

conradludgate:
If you are not satisfied with the current fuzzy search, please try this out and let us know if you find the results any better. This is ultimately a very subjective issue, but it's worth considering.

Install using
```sh
cargo install --locked --git https://github.com/ellie/atuin --branch skim-demo
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -38,6 +38,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b9a8f622bcf6ff3df478e9deba3e03e4e04b300f8e6a139e192c05fa3490afc7"
 
 [[package]]
+name = "arrayvec"
+version = "0.7.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


ellie:
I think it would make sense to get this merged as a new option in the next release. We can then get more people trying it out, and potentially replace the current fuzzy. Or at least make it the default!

ellie:
I think it would be good to get this released asap under search_mode "skim"

We can then try to get a bunch of people trying it - if they like it, we can consider replacing the other fuzzy mode, or changing the default mode

conradludgate:
Over-engineering, I can imagine restructuring the commands (**just for indexing, not as a source of truth**) to be

```
command_text{ id, text, count }
directories{ id, path }
host{ id, name }
executed { command_id, directory_id, session_id, host_id, timestamp }
```

To play with at some other time though. I cba now

ellie (approved):

conradludgate:
Was asked to add cwd weighting but that can be added later, gonna merge as is for now so it's easier to play with

PR #699, opened 2023-02-07
fish: Fix `atuin init` for the fish shell

hunger:
When I try to do `atuin init` I get a error from fish, suggesting to replace `$*` with `$argv`.

This patch does that and makes fish accept with `atuin init fish --disable-up-arrow`. I did not do much testing after that though :-)


ellie (approved):
Thank you for the catch! 🙏

adamdicarlo0:
This works for me 👍🏻

conradludgate (approved):

PR #706, opened 2023-02-10
[rebase] make update_needed check lazy

ellie:
@conradludgate Pretty sure that's all we need 🤷‍♀️ 

didn't want to clobber your branch though


ellie:
wtf how did this end up just being me on the commit when merged 😭 sorry for stealing it :(

PR #711, opened 2023-02-11
Add support for generic database in AppState

ekroon:
This is a first step to use https://github.com/ellie/atuin/pull/491 to implement https://github.com/ellie/atuin/issues/356 on top of the Axum 0.6. I thought it might be easier to review with smaller PRs. 

I think that adding integration tests might be a good next step, to have less chance of defects?


conradludgate (approved):

ellie:
> I think that adding integration tests might be a good next step, to have less chance of defects?

Sounds good to me! Thank you for this!

PR #712, opened 2023-02-12
Weight first word match higher

frankh:


conradludgate:
I wonder why skim thinks ssh-key matches that cargo install.

I did apply an artificial weighting to newer entries, but I could be smarter about the boundaries. This is probably what's causing the awkward weights currently

frankh:
> I wonder why skim thinks ssh-key matches that cargo install

It's just "ssh-ke" so each letter does have a match in that command it's just very spread out

conradludgate:
@ellie why ya merging stuff into my branch >:(

ellie:
Oh oops, sorry 😭 I thought `main` was the target 😬

PR #716, opened 2023-02-13
Add `history_filter` setting to exclude commands from history

### LINKED ISSUE #515: Excluding commands from being added to history
What is the best way to exclude commands to be added to history?

Using fish I modified `_atuin_preexec` like follows

```
function _atuin_preexec --on-event fish_preexec
    if string match -qr '^(cd|mpv|mplayer)' -- $argv
        return
    end
    set -gx ATUIN_HISTORY_ID (atuin history start "$argv[1]")
end
```
This seems to work. 

Question: Is this a good way to do this, or could it be done in a better way?

jbaiter:
Adds a new `history_filter` setting through which users can specify a list of regular expressions that match commands that should not be recorded in the history.

Example configuration:
```toml
## prevent commands matching any of these regexes from being written to history.
## Note that these regular expressions are unanchored, i.e. if they don't start
## with ^ or end with $, they'll match anywhere in the command.
## For details on the supported regular expression syntax, see
## https://docs.rs/regex/latest/regex/#syntax
history_filter = [
   "^secret-cmd",
   "^innocuous-cmd .*--secret=.+"
]
```

Fixes #515

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -133,6 +133,7 @@ dependencies = [
  "semver",
  "serde",
  "serde_json",
+ "serde_regex",
  "sha2",
  "shellexpand",
  "sodiumoxide",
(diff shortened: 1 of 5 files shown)


conradludgate:
Thanks!

ellie (approved):
Very nice - thank you!

Issue #610, opened 2022-11-14
[Feature Request] Custom keyboard shortcuts


stevenxxiu:
Thanks for the great project!

It would be great for there to be custom keyboard shortcuts for each action. e.g. I often use `Ctrl + LeftArrow`, `Ctrl + RightArrow`, `Home`, `End`, `Delete`, which don't work in the editor. I'm sure other people might also have their own preferences.

This can be done by allowing configurable escape sequences, or one of the libraries in [Comprehensive keyboard handling in terminals - kitty](https://sw.kovidgoyal.net/kitty/keyboard-protocol/). I know that [Broot](https://github.com/Canop/broot)  uses [Crossterm](https://github.com/crossterm-rs/crossterm/).

patricksjackson:
To clarify, I'm assuming you mean custom keyboard shortcuts inside the interactive search ui, `atuin search --interactive`?

If there are shortcuts you think make sense to add universally, it should be fairly simple to add them. The inputs are handled [with this match statement](https://github.com/ellie/atuin/blob/2b7ecabec22e2c127a09498efb894a1ae2b70923/src/command/client/search/interactive.rs#L71).

However adding the ability to have _user-defined_ keyboard shortcuts would be a decent chunk of work, since you'd need to define some sort of API for the UI and commit to supporting it. Personally I don't think it's worth the effort.

stevenxxiu:
> To clarify, I'm assuming you mean custom keyboard shortcuts inside the interactive search ui, `atuin search --interactive`?

Yup that's what I mean

> However adding the ability to have _user-defined_ keyboard shortcuts would be a decent chunk of work, since you'd need to define some sort of API for the UI and commit to supporting it. Personally I don't think it's worth the effort.

I'm not too sure on what you mean here on API for the UI, but fair enough. The keyboard shortcuts I use are the pretty usual ones, so I think adding them universally makes sense.

The ones I want are pretty self-explanatory:

- `Ctrl + LeftArrow`: Previous word.
- `Ctrl + RightArrow`: Next word.
- `Home`: Go to start of line.
- `End`: Go to end of line.
- `Delete`: Delete next character.
- `Ctrl + Backspace`: Delete previous word.
- `Ctrl + Delete`: Delete next word.

[...]

PR #719, opened 2023-02-17
feat: add common default keybindings

stevenxxiu:
For [[Feature Request] Custom keyboard shortcuts · Issue #610 · ellie/atuin](https://github.com/ellie/atuin/issues/610).

### DIFF
--- a/atuin-client/config.toml
+++ b/atuin-client/config.toml
@@ -44,6 +44,9 @@
 ## characters that count as a part of a word
 # word_chars = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
 
+## number of context lines to show when scrolling by pages
+# scroll_context_lines = 1
+
 ## prevent commands matching any of these regexes from being written to history.
 ## Note that these regular expressions are unanchored, i.e. if they don't start
 ## with ^ or end with $, they'll match anywhere in the command.
--- a/atuin-client/config.toml
+++ b/atuin-client/config.toml
@@ -38,6 +38,9 @@
 ## possible values: return-original, return-query
 # exit_mode = "return-original"
 
+## characters that count as a part of a word
+# word_chars = "./\\()\"'-:,.;<>~!@#$%^&*|+=[]{}`~?"
+
 ## prevent commands matching any of these regexes from being written to history.
 ## Note that these regular expressions are unanchored, i.e. if they don't start
 ## with ^ or end with $, they'll match anywhere in the command.
--- a/atuin-client/config.toml
+++ b/atuin-client/config.toml
@@ -38,8 +38,11 @@
 ## possible values: return-original, return-query
 # exit_mode = "return-original"
 
+## possible values: emacs, subl
+# word_jump_mode = "emacs"
+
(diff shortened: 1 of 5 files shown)


trygveaa:
I think it would be beneficial if this behaved in the same way as bash and zsh. The differences I see is that they always move to the beginning/end of the next word with ctrl-left/right. So with e.g. `test()-test` and the cursor at the end, ctrl-left first moves after `-` and then to the beginning. And then ctrl-right moves before `(` and then to the end. ctrl-delete and alt-backspace does the same movements, but deletes the words instead (ctrl-backspace doesn't do anything in me for bash or zsh, and just moves the cursor in this branch for me).

Another difference is that `_` is counted as a word separator in bash and zsh (I haven't checked all other symbols).

(I'm not affiliated with atuin btw, so these are just my personal opinions).

I also notice that if I press ctrl-right or ctrl-delete on an empty prompt, atuin panics with this branch.

stevenxxiu:
> I also notice that if I press ctrl-right or ctrl-delete on an empty prompt, atuin panics with this branch.

Thanks for testing. Should be fixed now. I added a test case for that.

> I think it would be beneficial if this behaved in the same way as bash and zsh.

I made the word skipping action identical to *Sublime Text*, which is what I'm used to. In fact, I configured my *Zsh* line editor to behave the same way.

I tried your example of `test()-test`. The current behavior also matches my browser (*Vivaldi*).

I think word jumping behavior is up to personal preference. It makes sense to add some configuration option for this, but I'm not clear on what options would make most people happy. The only thing I could think of is to configure the value of `WORD_SEPARATORS`.

trygveaa:
Ah, right, emacs-forward/backward-word and vi-forward/backward-word behave differently in zsh and none of them are the default. I think bash behaves the way I said by default though. I think it would make sense to stick close to the default behavior of common shells, since this is a tool for shell history. I'll defer to the maintainers here to decide this and to decide whether a configuration option should be added.

conradludgate:
We already have a mechanism for deleting words: <kbd>Ctrl</kbd>-<kbd>W</kbd>. I've just tested it and it treats `test()-test` as a single word.

As I don't use this feature in any shell and I'm not an emacs/vim keybinding user I can't really comment on which behaviour works best here

ellie:
> Ah, right, emacs-forward/backward-word and vi-forward/backward-word behave differently in zsh and none of them are the default. I think bash behaves the way I said by default though. I think it would make sense to stick close to the default behavior of common shells, since this is a tool for shell history. I'll defer to the maintainers here to decide this and to decide whether a configuration option should be added.

There doesn't really seem to be an obvious answer, so I think going for the default bash behaviour is the best option - I imagine most people will be used to this!

Otherwise, making things configurable would be 💯

stevenxxiu:
I added `word_jump_mode` (`emacs`/`vi`) and `word_chars` options now.

Hope this matches your expectations of *Emacs* behavior. Please let me know if it's still incorrect in some way.

stevenxxiu:
@ellie Sorry about the nag but have time to look at this?

ellie (approved):
I'm happy for this to go in, thank you for the work! 🚀 And apologies for the slower review :/

PR #733, opened 2023-02-27
[docs] fix some links between config and commands

Carthaca:
Add anchors while at it.

I verified by running the docs with `yarn start` locally


ellie (approved):
Thank you!

PR #736, opened 2023-02-27
Disable 2 tests that shouldn't run in release mode

cyqsimon:
These tests expect `debug_assert!` to panic, so should only run in debug mode:
- `pos_of_panics_on_out_of_bounds`
- `index_of_panics_on_out_of_bounds`

I came across this problem when [my COPR build](https://copr.fedorainfracloud.org/coprs/cyqsimon/el-rust-pkgs/builds/) failed, because I ran `cargo test --release` in the build script.

### DIFF
--- a/src/tui/buffer.rs
+++ b/src/tui/buffer.rs
@@ -480,6 +480,7 @@ mod tests {
     }
 
     #[test]
+    #[cfg(debug_assertions)]
     #[should_panic(expected = "outside the buffer")]
     fn pos_of_panics_on_out_of_bounds() {
         let rect = Rect::new(0, 0, 10, 10);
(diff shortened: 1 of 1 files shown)


conradludgate:
Curious. Why do you use test --release?

cyqsimon:
> Curious. Why do you use test --release?

Honestly, no particular reason specific to atuin 😅

However I've had situations with other packages where some weird synchronisation issue was only causing crashes in release builds, so I've made a habit to always run my tests in release mode.

conradludgate (approved):

conradludgate:
Fair enough. These aren't really our tests (we're vendoring tui-rs because of lack of maintainance) so I don't have any opinions either way.

PR #477, opened 2022-07-11
Add nix files and install instructions

Sciencentistguy:
Add a flake.nix to facilitate easier development on NixOS, and to allow atuin to be installed using `nix profile`. Also add a nix section to the install instructions in the readme

PR #743, opened 2023-02-28
nix: add flake-compat

0x4A6F:
Support nix (non-flakes), continuation of #477.
- tested changes on top of v12.0.0 (cargo test fails for nix on v13.0.0)

# using nix-shell (legacy)
```
[shell:~/atuin]$ nix-shell
[nix-shell:~/atuin]$ cargo build
[...]
   Compiling atuin-client v12.0.0 (/root/atuin/atuin-client)
   Compiling atuin-server v12.0.0 (/root/atuin/atuin-server)
   Compiling atuin v12.0.0 (/root/atuin)
    Finished dev [unoptimized + debuginfo] target(s) in 1m 34s
[nix-shell:~/atuin]$
```
# using nix-build (legacy)
```
[shell:~/atuin]$ nix-build -A packages.aarch64-linux.atuin
[...]
/nix/store/ri763ifjs3vrba8s3w64lan24j34mm54-atuin
```

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,0 +1,10 @@
+(import
+  (
+    let lock = builtins.fromJSON (builtins.readFile ./flake.lock); in
+    fetchTarball {
+      url = "https://github.com/edolstra/flake-compat/archive/${lock.nodes.flake-compat.locked.rev}.tar.gz";
+      sha256 = lock.nodes.flake-compat.locked.narHash;
+    }
+  )
+  { src = ./.; }
+).defaultNix
--- a/flake.lock
+++ b/flake.lock
@@ -1,5 +1,21 @@
 {
   "nodes": {
+    "flake-compat": {
+      "flake": false,
+      "locked": {
+        "lastModified": 1673956053,
+        "narHash": "sha256-4gtG9iQuiKITOjNQQeQIpoIB6b16fm+504Ch3sNKLd8=",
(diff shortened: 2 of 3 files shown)


ellie:
Thank you for this!

>  (cargo test fails for nix on v13.0.0)

Would you be able to elaborate there?

default.nix
```diff
@@ -0,0 +1,10 @@
+(import
+  (
+    let lock = builtins.fromJSON (builtins.readFile ./flake.lock); in
+    fetchTarball {
+      url = "https://github.com/edolstra/flake-compat/archive/${lock.nodes.flake-compat.locked.rev}.tar.gz";
+      sha256 = lock.nodes.flake-compat.locked.narHash;
+    }
+  )
+  { src = ./.; }
+).defaultNix
```
patricksjackson, line 10: It's probably best to just have either `default.nix` or `shell.nix`, and not both? `nix-shell` supports both, but prefers `shell.nix` - https://nixos.org/manual/nix/stable/command-ref/nix-shell.html

0x4A6F: Yes, we can live without a `shell.nix`. Without `default.nix` packages attributes can't be built.

patricksjackson (approved):

ellie (approved):

PR #748, opened 2023-03-01
Fix race condition in directory tests

cyqsimon:
These 4 tests cannot be run in parallel, because one of them could be unsetting an environment variable when another is expecting it to be set.

For an example, see [here](https://github.com/ellie/atuin/actions/runs/4306003050/jobs/7509233843).


ellie (approved):
Thank you!

PR #753, opened 2023-03-02
docs(README): fix activity graph link

notjedi:


ellie (approved):
totally missed this while moving docs, thank you for fixing it!

PR #754, opened 2023-03-02
Windows support

YummyOreo:
Adds support for windows in atuin.

Things that windows can't support:
- `atuin import auto` This is because windows does not have anything equivalent to `$SHELL`

Ps. I totally get if you guys don't want to support this.


ellie:
> Ps. I totally get if you guys don't want to support this.

😆 

I've been thinking about this a bunch lately. My only issue with us supporting Windows is that neither me nor @conradludgate regularly use a Windows machine, so Windows releases would not get properly tested at all.

However... it has been mentioned, a lot. Here and in a several other PRs and issues. If we're able to find a few users happy to run `main` and keep reasonably up to date, then this is something I'd consider, probably with some caveat in the README that Windows is not as thoroughly tested. Plus with Crossterm in place now, there's probably far fewer issues.

@conradludgate thoughts?

ellie:
Otherwise - regardless of whether we end up supporting Windows or not, thank you so much for the work! <3

ellie:
For continuity - last mentioned in #631

YummyOreo:
> However... it has been mentioned, a lot. Here and in a several other PRs and issues. If we're able to find a few users happy to run `main` and keep reasonably up to date

Yeah, I am happy to run `main` but I know that probably not enough.

src/command/client/search/interactive.rs
```diff
@@ -429,6 +429,11 @@ pub async fn history(
         let initial_input = app.input.as_str().to_owned();
         let initial_filter_mode = app.filter_mode;
 
+        {
+            #[cfg(target_os = "windows")]
+            let _ = event::read();
+        };
+
```
conradludgate, line 436: What's this for?

src/command/client/import.rs
```diff
@@ -45,6 +45,11 @@ impl Cmd {
 
         match self {
             Self::Auto => {
+                if cfg!(windows) {
+                    println!("This feature does not work on windows.");
+                    return Ok(());
+                }
+
```
conradludgate, line 52: In theory, we can compile away the auto arg completely on Windows. But tbh this is more self documenting so I like this better

conradludgate (commented):
I'm surprised so little had to change

src/command/client/search/interactive.rs
YummyOreo: It was reading double input and was reading the Enter input when you enter the command

YummyOreo:
> 
> I'm surprised so little had to change

Same. I'm going to pull main and test the new bash import later today

src/command/client/import.rs
YummyOreo: Another thing that we could be (probably not a good idea though) is to disable importing from shells that don't support windows like zsh. But it would just increase code complexity

ellie: I'm happy with it being self-documenting, though could the message be a little clearer? Guide the user on what they _should_ do, not just tell them what they cannot

Ie "Auto import does not work on Windows, please specify your shell" or similar

src/command/client/search/interactive.rs
```diff
@@ -429,6 +429,13 @@ pub async fn history(
         let initial_input = app.input.as_str().to_owned();
         let initial_filter_mode = app.filter_mode;
 
+        {
+            // We do this because windows does double inputs and catpues the `Enter` when running a
```
ellie, line 433:
```suggestion
            // We do this because windows does double inputs and captures the `Enter` when running a
```

src/command/client/import.rs
YummyOreo: ok yea that would be better

YummyOreo:
can confirm that bash import works and command prev works

ellie:
I have absolutely no way of testing this just yet so will need to get access to a Windows machine before releasing, but am comfortable to say that support is "experimental" for now 🚀 

Thank you so much for this work @YummyOreo, really appreciate it!

ellie (approved):

ellie:
I'll follow up with docs and stuff later, need to figure out a nice way of wording it so people don't expect us to 100% support Windows

YummyOreo:
> I'll follow up with docs and stuff later, need to figure out a nice way of wording it so people don't expect us to 100% support Windows

I can help with the docs. Also it could be worded as partial support or not tested.

PR #757, opened 2023-03-05
fix(client): always read session_path from settings

lukekarrys:
When setting the `session_path` config to a non-default value, `auto_sync` and `sync_frequency` were having no effect for me.

After digging into the code, I noticed that `settings.should_sync()` and the `login` and `lougout` commands were reading `session_path` as `atuin_common::utils::data_dir().join("session")` which was not the same as the value I set in my `config.toml`.

This PR should standardize all the places that read `session_path` to `settings.session_path`, which was already how it was done in a few places including the `register` command.

**Note**: I've basically never written Rust before and while making these changes I'm attempting to get my local env setup for Rust dev, but things are still installing as I type this. So there's a good chance this doesn't even compile but I'll come to fix it up or ask questions once I get it running locally.

### DIFF
--- a/atuin-client/src/settings.rs
+++ b/atuin-client/src/settings.rs
@@ -184,9 +184,7 @@ impl Settings {
     }
 
     pub fn should_sync(&self) -> Result<bool> {
-        let session_path = atuin_common::utils::data_dir().join("session");
-
-        if !self.auto_sync || !session_path.exists() {
+        if !self.auto_sync || !PathBuf::from(self.session_path.as_str()).exists() {
             return Ok(false);
         }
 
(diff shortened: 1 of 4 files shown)


lukekarrys:
I pushed a `fixup` commit after I got the project compiling. I made sure `cargo test` still passed as well.

ellie:
Nice catch, thank you! Would you be able to run the formatter please?

Just "cargo fmt" 😊

lukekarrys:
Thanks for the review! Just pushed another `fixup` for formatting.

As a newbie, `cargo fmt` was giving me the error `error: no such command: fmt` until I realized I had installed `rust` with `homebrew` and not `rustup`. Then it worked as expected.

ellie (approved):
Thank you for the fix! 🚀

PR #760, opened 2023-03-06
Improve import, again

cyqsimon:
## Motive

I am recently distro-hopping and wanted to import my `.bash_history` from another file root that's mounted on `/run/media/OldRoot`. It wasn't initially obvious whether this is supported, so I read the code and found the `$HISTFILE` environment variable. It worked well enough for Bash, but it seems like the implementation is inconsistent across different kind of shells.
- Fish's import code seems to ignore `$HISTFILE` entirely
- Zsh's import code uses `$HISTFILE` for importing plain text history, but uses the largely-undocumented `$HISTDB_FILE` for importing history in SQL db file
- The way dependency injection is done is not very "Rusty", and can be easily overlooked by later implementers

So I thought I'd improve this situation.

## Change summary
1. Improve the `Importer` trait. Now structs just need to implement `Importer::default_source_path`, and `Importer::final_source_path` is automatically provided.
2. `$HISTFILE` is now universally respected for all shell types.
3. `$HISTDB_FILE` is no longer used. Instead the import code will try to infer whether `$HISTFILE` points to a SQL db or a plain text file.
4. Added a `--from-file` flag to the `import` subcommand, which does the same thing as `$HISTFILE` but is more discoverable.
5. Other minor code style changes that improves cleaniness and readability IMO.

## Todo
- [x] implementation
- [ ] documentation

### DIFF
--- a/atuin-client/src/import/bash.rs
+++ b/atuin-client/src/import/bash.rs
@@ -1,12 +1,16 @@
-use std::{fs::File, io::Read, path::PathBuf, str};
+use std::{
+    fs::{self},
+    path::{Path, PathBuf},
+    str,
+};
 
 use async_trait::async_trait;
 use chrono::{DateTime, Duration, NaiveDateTime, Utc};
 use directories::UserDirs;
-use eyre::{eyre, Result};
+use eyre::{bail, Result};
 use itertools::Itertools;
 
-use super::{get_histpath, unix_byte_lines, Importer, Loader};
+use super::{unix_byte_lines, Importer, Loader};
 use crate::history::History;
 
 #[derive(Debug)]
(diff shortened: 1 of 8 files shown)


cyqsimon:
On second thought I may have overengineered part of this...

I think it's a better idea to flat out decline the import if the user sets `$HISTFILE` but didn't explicitly specify the shell type. Makes the whole experience more consistent and reduces the likelihood of misuse (i.e. importing the history file of shell A using the importer for shell B).

cyqsimon:
Sorry, I realised I've made a grave misunderstanding of the `$HISTFILE` environment variable. It seems like it's in fact a variable used by various shells to determine where to place their history files, not just any environment variable used by atuin like I assumed.

I will make corresponding changes and create another PR for this.

PR #767, opened 2023-03-08
Minor documentation updates

onkelT2:
Updated the default search_mode and documented the default filter_mode_shell_up_key_binding to reflect their current default values in the code.


ellie (approved):
Thank you! 🚀

PR #774, opened 2023-03-09
make atuin-server a standalone binary

conradludgate:
I'm not a huge fan of `atuin server`, I think 99% of devices won't need both atuin and atuin-server installed, so I'm starting to split it out even more


ellie:
I've actually had a lot of feedback to the contrary - people temporarily spin up servers for sync, etc 

I'd rather keep it all in one unless we have a really good reason to not

PR #776, opened 2023-03-11
fix(installer): use case-insensitive comparisons

aschey:
On my arch system, the value of `/etc/os-release` is `arch` rather than `Arch`, which caused the OS check to fail. This change uses `tr '[:upper:]' '[:lower:]'` to switch everything to lowercase so we don't need to worry about case-sensitive comparisons.


ellie (approved):
Nice catch! Thank you 🙏

PR #779, opened 2023-03-13
Switch to using uuidv7

ellie:
Why?

uuidv4 is totally random. uuidv7 has a time-based prefix, ie, they
_sort_.

This is important when we start to think about database indices. btree
indices have shit performance when the elements are totally random, but
good insertion + perf when they are sorted. Plus it's just kinda useful
to be able to sort them

I don't think it's worth re-ID-ing past data, but at least from hereon
out future data will be nice

TODO: Update the database schema to use the UUID type, and not just a
string.

Also we need the unstable flags here, as uuidv7 is technically not a finished spec.

### DIFF
--- a/atuin-common/Cargo.toml
+++ b/atuin-common/Cargo.toml
@@ -14,5 +14,3 @@ repository = "https://github.com/ellie/atuin"
 chrono = { version = "0.4", features = ["serde"] }
 serde = { version = "1.0.145", features = ["derive"] }
 uuid = { version = "1.3", features = ["v7"] }
-sha2 = { version = "0.10" }
-hex = { version = "0.4" }
--- a/atuin-common/Cargo.toml
+++ b/atuin-common/Cargo.toml
@@ -13,4 +13,6 @@ repository = "https://github.com/ellie/atuin"
 [dependencies]
 chrono = { version = "0.4", features = ["serde"] }
 serde = { version = "1.0.145", features = ["derive"] }
-uuid = { version = "1.2", features = ["v4"] }
+uuid = { version = "1.3", features = ["v7"] }
+sha2 = { version = "0.10" }
+hex = { version = "0.4" }
--- a/atuin-common/src/utils.rs
+++ b/atuin-common/src/utils.rs
@@ -3,8 +3,19 @@ use std::path::PathBuf;
 use chrono::NaiveDate;
 use uuid::Uuid;
 
-pub fn uuid_v4() -> String {
-    Uuid::new_v4().as_simple().to_string()
+pub fn uuid_v7() -> String {
+    Uuid::now_v7().as_simple().to_string()
+}
+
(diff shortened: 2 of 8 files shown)


atuin-common/src/utils.rs
```diff
@@ -3,8 +3,19 @@ use std::path::PathBuf;
 use chrono::NaiveDate;
 use uuid::Uuid;
 
-pub fn uuid_v4() -> String {
-    Uuid::new_v4().as_simple().to_string()
+pub fn uuid_v7() -> String {
+    Uuid::now_v7().as_simple().to_string()
+}
+
+pub fn hash_str(string: &str) -> String {
```
ellie, line 10: I rebased this from a big WIP branch, sneak peak

atuin-common/src/utils.rs
```diff
@@ -3,8 +3,19 @@ use std::path::PathBuf;
...
+    hash_bytes(string.as_bytes())
+}
+
+pub fn hash_bytes(buf: &[u8]) -> String {
+    use sha2::{Digest, Sha256};
+    let mut hasher = Sha256::new();
+    hasher.update(buf);
+    hex::encode(hasher.finalize())
```
conradludgate, line 18: Snuck in with another change?

atuin-common/Cargo.toml
```diff
@@ -13,4 +13,6 @@ repository = "https://github.com/ellie/atuin"
 [dependencies]
 chrono = { version = "0.4", features = ["serde"] }
 serde = { version = "1.0.145", features = ["derive"] }
-uuid = { version = "1.2", features = ["v4"] }
+uuid = { version = "1.3", features = ["v7"] }
+sha2 = { version = "0.10" }
+hex = { version = "0.4" }
```
conradludgate, line 18: Same here?

conradludgate (commented):
I love it. I need to see if that unstable flag has any consequences. Eg I don't know if it works with `cargo install` from crates.io off the top of my head

atuin-common/Cargo.toml
ellie: Ya, rebased from a WIP breanch

ellie:
> I love it. I need to see if that unstable flag has any consequences. Eg I don't know if it works with `cargo install` from crates.io off the top of my head

Yup for sure, I wasn't certain what implications it might have

conradludgate:
Following from our chat on Tuesday, the unstable config is a problem. However, since we don't rely on stable UUIDs and since the format is so simple, we can utilise the genericness of the uuid crate using https://docs.rs/uuid/1.3.0/uuid/struct.Uuid.html#method.from_fields.

The core implementation is https://docs.rs/uuid/1.3.0/src/uuid/timestamp.rs.html#236-255. 48 bit milliseconds + some random bytes with version `7` and variant `0b10`.

PR #780, opened 2023-03-14
Fixes a bug on windows

YummyOreo:
Fixes the bug that I tried to fix in https://github.com/ellie/atuin/pull/754.

The bug was that it was not distinguishing between `Release` and `Press`. The "fix" *was* to throw away one  input. But this is pragmatic because:
- Git bash does not send both `Release` and `Press`.
- Cgywin compiles it for linux, runs it as linux, thus the fix wont work.

This fixes it by throwing away any `Release` ones.

### DIFF
--- a/src/command/client/search/interactive.rs
+++ b/src/command/client/search/interactive.rs
@@ -94,6 +94,10 @@ impl State {
         input: &KeyEvent,
         len: usize,
     ) -> Option<usize> {
+        if input.kind == event::KeyEventKind::Release {
+            return None;
+        }
+
         let ctrl = input.modifiers.contains(KeyModifiers::CONTROL);
         let alt = input.modifiers.contains(KeyModifiers::ALT);
         match input.code {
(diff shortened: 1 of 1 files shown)


YummyOreo:
For the clippy  warning. Idk if I should split the function up into multiple different ones (I feel like that is outside the scope of this), handle this in the `handle_input` function, or ignore it.

conradludgate (approved):
This makes a lot more sense. Thanks

conradludgate:
> For the clippy warning. Idk if I should split the function up into multiple different ones (I feel like that is outside the scope of this), handle this in the `handle_input` function, or ignore it.

You can add the `allow(clippy::too_many_lines)` for now. I have a refactor in progress anyway

conradludgate (approved):

src/command/client/search/interactive.rs
```diff
@@ -88,12 +88,17 @@ impl State {
         None
     }
 
+    #[allow(clippy::to_many_lines)]
```
conradludgate, line 91:
```suggestion
    #[allow(clippy::too_many_lines)]
```

conradludgate (approved):

PR #789, opened 2023-03-19
fix: many links were broken :memo:

nicoandmee:
This fixes a number of broken links in the documentation.
P.S. Thanks for this project! It's hard to overstate the utility for someone like me who reguarly hops between (many) machines. It's saved me many times from disaster.


conradludgate:
Ah. I think the links in docs/docs do work on the website, but not in github. Hmmm

ellie (approved):
Thank you! Totally working on the website, but not in GitHub - paths are all different I suppose

I just checked and the website still works with this 😁

PR #790, opened 2023-03-19
tantivy demo

conradludgate:
Tantivy 🥳 

It's currently not great. still working out how to make the queries work properly. But it does work, and that's what matters really

To use, run `atuin tantivy refresh-index` to populate the index initially, then it will work from there. Enable the tantivy search-mode and it will be used

### DIFF
--- a/atuin-client/src/tantivy.rs
+++ b/atuin-client/src/tantivy.rs
@@ -1,0 +1,103 @@
+use crate::{database::Database, history::History};
+use eyre::Result;
+use tantivy::{
+    directory::MmapDirectory,
+    doc,
+    schema::{Field, Schema, FAST, STORED, STRING, TEXT},
+    DateTime, Index, IndexWriter,
+};
+
+pub fn schema() -> (HistorySchema, Schema) {
+    let mut schema_builder = Schema::builder();
+
+    (
+        HistorySchema {
+            id: schema_builder.add_text_field("id", STRING),
+            command: schema_builder.add_text_field("command", TEXT | STORED),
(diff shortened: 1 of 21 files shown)


atuin-client/src/tantivy.rs
```diff
@@ -0,0 +1,103 @@
...
+pub fn schema() -> (HistorySchema, Schema) {
+    let mut schema_builder = Schema::builder();
+
+    (
+        HistorySchema {
+            id: schema_builder.add_text_field("id", STRING),
+            command: schema_builder.add_text_field("command", TEXT | STORED),
+            cwd: schema_builder.add_text_field("cwd", STRING | FAST),
+            session: schema_builder.add_text_field("session", STRING | FAST),
+            hostname: schema_builder.add_text_field("hostname", STRING | FAST),
```
ellie, line 19: we should do `| SLOW` so that in a future release we can swap it for `FAST` and claim 1000x performance boost 🚀

atuin-client/src/tantivy.rs
```diff
@@ -0,0 +1,103 @@
...
+
+pub fn bulk_write_history(
+    writer: &mut IndexWriter,
+    schema: &HistorySchema,
+    h: impl IntoIterator<Item = History>,
+) -> Result<()> {
+    for h in h {
+        write_single_history(writer, schema, h)?;
+    }
+    writer.commit()?;
```
ChillFish8, line 67: You may find it useful to also call `writer.wait_merging_threads()` to wait for the background mergers to finish, normally this is very quick but just prevents you from creating a large number of background threads each time bulk_write_history is called.

PR #791, opened 2023-03-19
Add history deletion

### LINKED ISSUE #592: Deleting history
So this has been proposed for a [super long time](https://github.com/ellie/atuin/issues/70).

It's taken a while, partly because deletion with sync is difficult. Deleting on one machine would result in it coming back the next time a sync happens, so we need to ensure deletion across all machines

I'm starting a new issue to track this, as it's actively being worked on

I wrote up my plan in a comment on #70, but copying it here for clarity

------------

tl;dr for the plan, which I'd like to break down a bit more and HOPEFULLY get in for a v12 release. Or at least partly!

## Deleting things, finally, as everyone has been asking me for it

(cue brain dump)

### 1. Switch to event log sync

I started on this here: https://github.com/ellie/atuin/pull/390

The difficulty here is that _actually_ deleting the entire row will change the count of rows. Syncing two append-only logs is much much easier and simpler than syncing two arbitrary length sets of data, especially when each installation of atuin is both read and write (so technically multi-master). 

The premise with this solution is that instead of syncing up history items, as we do at the moment, we sync up events. I'm proposing two kinds of event. A CREATE event, and a DELETE event. For backwards compatibility, we'd want to add a CREATE event for every currently-existing item of history.

To delete an item of history, we push up a DELETE event to the log. This is still only an append, so we can retain our simple sync logic + keep things nice and easy.

If we replay the log on a fresh client, the history item will end up being created + then deleted. Mostly there, and mostly achieves the use case of "I don't want this thing showing up in my history".

Once that works, we could expand it to also blank out the data section of a CREATE event when a DELETE event has been created, though that would cause limitations in future sync methods (I was conisdering something based on merkle trees).

### 2. Implement UI for deletion

I'd like this to be smooth, obvious, and not some arbitrary key combination. But also not easy to do by accident! We may also want the option to delete a single occurence of a command, but not delete all historical occurances of it (and vice-versa).

To allow for this, it would be good to be able to open a "context menu" on a single item of history. I'd suggest using the tab character for this. 

Scroll back to the item you wish to delete. Hit tab, and we switch to the menu item for that history. As well as showing extra information about it (directory executed in, some other context, etc), we can have an option for deleting it!

Might want some kind of UI hint to make it obvious that tab does stuff

### 3. done...?

Probably not, I imagine some things will break. I'd like to roll this out gradually, as we need to maintain backwards compatibility. Will probably save old style history rows and event rows side-by-side, and start by just syncing CREATE events up and not even implementing DELETE until the next version.

ellie:
Allow for deleting history!

Resolves #592

I have ditched the idea of the event sync. I do intend on coming back to it, but I have different thoughts on how I'd like it to work, and I don't want to shove it in + design it around deletions

Firstly, drop the events table. Not needed, and will improve insertion speed + disk usage to just not have it

Add an API on the server to delete history by ID, and a new status endpoint to replace coutn. The latter will also list the currently deleted history

Then make sync use this new api + reconcile


atuin-server/src/database.rs
```diff
@@ -141,6 +149,46 @@ impl Database for Postgres {
         Ok(res.0 as i64)
     }
 
+    async fn delete_history(&self, user: &User, id: String) -> Result<()> {
+        sqlx::query(
+            "update history
+            set deleted_at = $3
+            where user_id = $1
+            and client_id = $2
+            and deleted_at is null", // don't just keep setting it
```
conradludgate, line 158: Interesting. I assume this can avoid a write flush

atuin-server/src/database.rs
```diff
@@ -141,6 +149,46 @@ impl Database for Postgres {
...
+        .fetch_all(&self.pool)
+        .await?;
+
+        let res = res
+            .iter()
+            .map(|row| row.get::<String, _>("client_id"))
+            .collect();
+
+        Ok(res)
+    }
```
conradludgate, line 190: no pagination/cursors?

ellie: We can list thousands+ of IDs before the status page becomes large, and idk if people are even going to delete that much. I can add it if you're concerned, but tbh I'd like to see how it's used first

atuin-server/src/database.rs
ellie: Probably, but also means deleted_at won't be set if it has already been set

conradludgate (approved):

PR #793, opened 2023-03-20
fix: paste into terminal after switching modes

hhamud:
Fixes this [issue](https://github.com/ellie/atuin/issues/786)


src/command/client/search/interactive.rs
```diff
@@ -133,6 +134,14 @@ impl State {
         None
     }
 
+    fn handle_paste_input(&mut self, input: &String) -> Option<usize> {
+        let n: Vec<char> = input.chars().collect();
+        for i in n {
```
conradludgate, line 139: No need to collect here, we can use the iterator directly
```suggestion
        for i in input.chars() {
```

PR #794, opened 2023-03-20
Support old msgpack

ellie:
I forgot it isn't backwards compatible... This should fix any sync issues resulting from the deletion PR

Will switch to a new encoding in the future


conradludgate (dismissed):
Smh ellie

PR #805, opened 2023-03-23
chore: uuhhhhhh crypto lol

conradludgate:
I have tested to make sure that old hashed/encrypted blobs still work under the new libraries. All the tests pass but I want to be extra thorough so let's hold for now...

We originally intended to use https://github.com/RustCrypto/nacl-compat/tree/master/crypto_box for the secretbox replacement, but the have their own keygen on top (they use a public+private key pair to generate nonce and to generate a symmetric key. We already have the symmetric key so that abstraction on top does not work). The underlying impl is the one we are using here. nacl-compat has been audited, which would imply that `xsalsa20poly1305` also has been audited successfully.

`argon2` crate is maintained by the same `RustCrypto` group. This has no claim to be audited. I will review the code myself, but it has quite a lot of downloads on crates.io and no reported vulnerabilities in it's lifetime.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,16 @@
 version = 3
 
 [[package]]
+name = "aead"
+version = "0.5.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d122413f284cf2d62fb1b7db97e02edb8cda96d769b16e443a4f6195e35662b0"
+dependencies = [
+ "crypto-common",
+ "generic-array",
+]
+
+[[package]]
 name = "ahash"
 version = "0.7.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


ellie:
Is this compatible with secretbox?

conradludgate:
> Is this compatible with secretbox?

I haven't found any incompatibilities yet. `argon2id13` is the same algorithm, and according to [the libsodium docs](https://doc.libsodium.org/secret-key_cryptography/secretbox#algorithm-details)

> Algorithm details:
Encryption: XSalsa20 stream cipher
Authentication: Poly1305 MAC

Which `XSalsa20Poly1305` is an implementation of

conradludgate:
Nice benefit: since macos and libsodium were not good friends, full builds are down from 3m 18s to 1m 47s :D

ellie:
Ohhhh wow very nice 👏 

I'll give this a test tomorrow. Maybe see if we can get some of the community to try it as well? I imagine if it decrypts our history fine then it'll work for everyone, but...

conradludgate:
Did a full re-sync, it decrypts everything fine 😌

ellie:
Just deleted my history and re-synced on this branch, it all worked ✨

ellie (approved):
LGTM! what's the worst that could happen?

PR #811, opened 2023-03-26
Bind keys in vi mode too

ellie:


ellie:
thanks for the catch @YummyOreo!

PR #813, opened 2023-03-26
Fixes nushell recording empty commands

YummyOreo:
If you would press `Enter` it would record it. I don't think that this is intended.


ellie (approved):
Awesome! Thanks!

PR #821, opened 2023-03-27
fix: record negative exit codes

stevenxxiu:
Without the `=` I get the error on pressing `Ctrl + C`:

```
error: Found argument '-1' which wasn't expected, or isn't valid in this context

  If you tried to supply '-1' as a value rather than a flag, use '-- -1'

Usage: atuin history end --exit <EXIT> <ID>

For more information try '--help'
```


ellie (approved):
Nice catch! Thanks

PR #823, opened 2023-03-28
Refactor/nu remove dep on sh

stevenxxiu:
Remove dependency on `sh` per [The `complete` command doesn't work with interactive commands · Issue #8539 · nushell/nushell](https://github.com/nushell/nushell/issues/8539).


ellie (approved):
Woohoo! Thank you! 🚀

PR #825, opened 2023-03-28
Add `--offset` flag to `atuin search`

takac:
This flag allows the user to continue searching at an offset. This is useful for building tools that use atuin to search for previous commands and return only one result.

```
# assuming atuin isn't recording the next few commands..

# return first result for "ssh"
atuin search --limit 1
# return second result
atuin search --limit 1 --offset 1
# so on..
atuin search --limit 1 --offset 2
```

### DIFF
--- a/atuin-client/src/database.rs
+++ b/atuin-client/src/database.rs
@@ -34,6 +34,7 @@ pub struct OptFilters {
     pub before: Option<String>,
     pub after: Option<String>,
     pub limit: Option<i64>,
+    pub offset: Option<i64>,
 }
 
 pub fn current_context() -> Context {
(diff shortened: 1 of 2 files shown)


ellie (approved):
Great work, thank you!

ellie:
Oh hang on - I always forget this 🤦‍♀️ 

would you be able to update the docs with this as well please? 🙏 

https://github.com/ellie/atuin/blob/main/docs/docs/commands/search.md

takac:
Sure, will follow up on that

takac:
https://github.com/ellie/atuin/pull/828

PR #826, opened 2023-03-28
Add zsh incremental search widget for atuin

takac:
Add a bash/zsh like reverse history lookup widget that uses `atuin` as
the backend. This widget tries to stay true to default behaviour of
`ctrl-r` found in bash and zsh.

To activate the plugin in zsh, source the file, `source
atuin.isearch.zsh`. Then use `ctrl-r` to activate reverse search. Use
`ctrl-r` again while in search mode to cycle to previous hits, use
`ctrl-f` to cycle filter mode (prefix, fuzzy, full-text, skim). Press
`enter` to select the search hit or `ctrl-c` or `esc` to close.

Screenshot 2023-03-28 at 15 23 46

### DIFF
--- a/.github/workflows/shellcheck.yml
+++ b/.github/workflows/shellcheck.yml
@@ -16,3 +16,6 @@ jobs:
       uses: ludeeus/action-shellcheck@master
       env:
         SHELLCHECK_OPTS: "-e SC2148"
+      ignore:
+        # can't check zsh with shellcheck
+        - atuin.plugin.zsh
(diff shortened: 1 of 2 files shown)


ellie:
Nice! A whole bunch of people have been asking for something like this

I'm not sure it would work well in the root of the repo though. We've actually had a few neat shell integrations shared lately, so maybe it's best to centralize them somewhere?

It might be good to have a page in the [docs](https://github.com/ellie/atuin/tree/main/docs/docs) for "integrations" or similar, and we can detail that there! It'll be published to https://atuin.sh once merged, so discoverability will be much better than a file in a git repo.

ellie:
Hey! I was just wondering if you planned to work on this any further? 🙏

takac:
I use this plugin daily, it works well enough for my own use. Did you want me to put in somewhere specific in the repo, happy to tidy up what I've done.

PR #833, opened 2023-03-30
feat: add github action to test the nix builds

patricksjackson:
This adds an action that will run a `nix flake check`, which checks the nix configuration, and `nix build`, which builds atuin using the nix config. Failures in this workflow will indicate that flake.lock needs to be updated with `nix flake update` or a change is needed in the configuration files.

This requires #832 to pass. I verified locally using [act](https://github.com/nektos/act) that it succeeds with that change.


ellie (approved):
Thank you! Appreciate you keeping our Nix stuff going 🙏

PR #839, opened 2023-04-03
Updated client config docs

cyqsimon:
- Example `config.toml` now includes all the newest options
- Example `config.toml` now have default values consistent with the actual default (except for the paths)
- `settings.rs`, `config.toml`, and `config.md` now have uniform option order

## Todo
- [ ] 3 options are still missing from `config.md`: `word_jump_mode`, `word_chars`, and `scroll_context_lines`. I don't know exactly what these do. Help wanted here.

### DIFF
--- a/atuin-client/config.toml
+++ b/atuin-client/config.toml
@@ -10,7 +10,7 @@
 # session_path = "~/.key"
 
 ## date format used, either "us" or "uk"
-# dialect = "uk"
+# dialect = "us"
 
 ## enable or disable automatic sync
 # auto_sync = true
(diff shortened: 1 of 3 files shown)


atuin-client/config.toml
```diff
@@ -10,29 +10,46 @@
 # session_path = "~/.key"
 
 ## date format used, either "us" or "uk"
-# dialect = "uk"
+# dialect = "us"
```
ellie, line 13: I believe the US is the only country that uses the US format of mm/dd/yy, vs most of the rest of the world in using dd/mm/yy

```suggestion
# dialect = "uk"
```

ellie (approved):
Thank you! It's needed updating for a while, appreciate you sorting it out 🙏

PR #840, opened 2023-04-03
update `str substring` usage to use range parameter

WindSoilder:
Hi, nushell's `str substring` doesn't accept string argument, I'd like to propose a update to use nushell range instead


zephraph:
I reported a related failure in #848 but close that in favor of this PR.

ellie (approved):

ellie:
Thank you for fixing this! Appreciate it 🙏 

Also, as you're a first-time contributor, if you'd like a holographic contributor-only sticker then please fill out this form: https://notionforms.io/forms/contributors-stickers

happysalada:
if possible, could we get a point release including this ?

ellie:
> if possible, could we get a point release including this ?

yup - trying to get one out this weekend

PR #844, opened 2023-04-03
Fix deleting history that doesn't exist yet

ellie:
This can occur if history has been added + then deleted on a machine before it has a chance to be synced to a new one.

### DIFF
--- a/atuin-client/src/sync.rs
+++ b/atuin-client/src/sync.rs
@@ -95,8 +95,14 @@ async fn sync_download(
     for i in remote_status.deleted {
         // we will update the stored history to have this data
         // pretty much everything can be nullified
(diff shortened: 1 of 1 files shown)


atuin-client/src/sync.rs
```diff
@@ -95,8 +95,14 @@ async fn sync_download(
     for i in remote_status.deleted {
         // we will update the stored history to have this data
         // pretty much everything can be nullified
-        let h = db.load(i.as_str()).await?;
-        db.delete(h).await?;
+        if let Ok(h) = db.load(i.as_str()).await {
+            db.delete(h).await?;
+        } else {
+            info!(
```
ellie, line 101: it's not technically a problem, hence info

conradludgate (approved):

PR #845, opened 2023-04-04
Handle empty lines when importing from Bash

### LINKED ISSUE #167: `atuin import` should support bash `HISTTIMEFORMAT="%F %T "`
# Setup

This is a common bash config to enable storing timestamps, and is recommended by tools like[ hstr](https://github.com/dvorka/hstr). It result in a history file looking like
```
#1627034544
ls
#1627034556
echo <<EOF
multi
line
EOF
```

# Observed
When running `atuin import bash`, all the timestamps end up imported as actual commands.

# Expected
`atuin import` should at least not confuse timestamps as commands, and ideally parse those timestamps as useful metadata.

cyqsimon:
Amends #747. Fixes #167.

### DIFF
--- a/atuin-client/src/import/bash.rs
+++ b/atuin-client/src/import/bash.rs
@@ -166,9 +166,7 @@ cargo :b̷i̶t̴r̵o̴t̴ ̵i̷s̴ ̷r̶e̵a̸l̷
                 "cargo :b̷i̶t̴r̵o̴t̴ ̵i̷s̴ ̷r̶e̵a̸l̷",
             ],
         );
-        assert!(is_strictly_sorted(
-            loader.buf.iter().map(|h| h.timestamp.timestamp())
-        ))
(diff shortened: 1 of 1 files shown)


atuin-client/src/import/bash.rs
```diff
@@ -59,7 +59,11 @@ impl Importer for Bash {
 
         // if no timestamp is recorded, then use this increment to set an arbitrary timestamp
         // to preserve ordering
-        let timestamp_increment = Duration::seconds(1);
+        // this increment is deliberately very small to prevent particularly fast fingers
+        // causing ordering issues; it also helps in handling the "here document" syntax,
+        // where several lines are recorded in succession without individual timestamps
+        let timestamp_increment = Duration::milliseconds(1);
```
conradludgate, line 65: I think our other importers use nanoseconds

conradludgate (approved):

PR #849, opened 2023-04-05
fix(nushell): use ranges for `str substring`

stevenxxiu:
Fix for *Nushell* v0.78.0.


ellie:
Closing in favour of #840, as it was there first 🙈 

But thank you for the fix, either way!

stevenxxiu:
The project's getting so popular!

PR #857, opened 2023-04-09
Add some emacs movement keys

majutsushi:
This adds some common Emacs movement keys. Since Emacs mode is also the default for most shells the keys should be familiar to a lot of people. I kept accidentally adding `f`s and `b`s to the search buffer when trying to jump around ...

I also added an exception for a clippy lint caused by the additional match branches. It's not ideal but I'm not sure breaking up the function would be better.

### DIFF
--- a/src/command/client/search/interactive.rs
+++ b/src/command/client/search/interactive.rs
@@ -86,6 +86,7 @@ impl State {
     }
 
     #[allow(clippy::too_many_lines)]
+    #[allow(clippy::cognitive_complexity)]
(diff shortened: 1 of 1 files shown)


ellie (approved):
Thank you!

Seeing as this is your first time contributing, if you would like a holographic contributors-only Atuin sticker, then please fill out this form! https://notionforms.io/forms/contributors-stickers

We have a [Discord](https://discord.gg/Fq8bJSKPHh), if you're like to ask any questions! Or just hang out

PR #861, opened 2023-04-10
Add status bar

ellie:
Add a status bar to display extra info about a command - currently the directory + times ran

Also introduces the "HistoryResult" to the code. It replaces returning history in a number of places, to also allow us to return the number of times that command has been seen. It could also be used for additional context in the future.

I'd like to follow up with a theming PR. Originally it was done here too, but was getting pretty fat.

Screenshot 2023-04-10 at 19 11 39

I've messed around with some of our queries a bit too. It seems OK for me locally, but would appreciate extra close 👀 there!

### DIFF
--- a/atuin-client/src/result.rs
+++ b/atuin-client/src/result.rs
@@ -1,0 +1,11 @@
+use crate::history::History;
+
+// Return a single instance of history, but include some context.
+// EG how many other history items we have that are the same
+// Can be extended to include match context, when we use actual
+// seach engines
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct HistoryResult {
+    pub history: History,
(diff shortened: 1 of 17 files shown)


atuin-client/src/result.rs
```diff
@@ -0,0 +1,11 @@
+use crate::history::History;
+
+// Return a single instance of history, but include some context.
+// EG how many other history items we have that are the same
+// Can be extended to include match context, when we use actual
+// seach engines
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct HistoryResult {
```
ellie, line 8: I'm not super sure on the name, `Result<HistoryResult>` is... odd. But also can't think of much better, because it's like a search result but history

PR #863, opened 2023-04-10
syntax highlighting

conradludgate:
- works
- modify
- stash
- change back to no extra crate
- speed up latency


ellie:
This would also be super cool to have, but will need a (probably annoying) rebase. Closing for now, I'll get to it later

PR #806, opened 2023-03-24
replace chrono with time

conradludgate:
Kinda useless, but I've wanted to do this one for a while nice. In my opinion, `time` has a much cleaner (and safer by design) API compared to `chrono`.

Still need to test and review everything

PR #868, opened 2023-04-12
Workspace reorder

utter-step:
* moved main crate (`atuin`) to be on the same level with other crates
* extracted common deps to workspace definition, so it would be easier to avoid having same multiple dependencies with slightly different versions (base64 was an example)
* slight fix to base64 usage, to comply with deprecation warning from v0.21

As a nice side-effect, `cargo test`/`cargo nextest run` now runs tests from all the crates in a workspace, compared to the main crate only, as before :)

What do you think?


utter-step:
https://github.com/ellie/atuin/pull/868/commits/993e60f8dea81a1625a04285a617959ad09a0866 this commit updates deps and fixes chrono deprecations. It added a bunch of unwraps, but in each case – previously the code would also panic in the same conditions, it's just more obvious now :)

Aside from that – zero meaningful changes

conradludgate:
You can revert the last commit as we'll be going with #806 soon

utter-step:
> You can revert the last commit as we'll be going with https://github.com/ellie/atuin/pull/806 soon

Great! Reverted :)

ellie (approved):
This is great - thank you!

PR #872, opened 2023-04-13
Bug/fix gh action aarch64 tarball

m2nx:
this PR is going to fix a bug that github action will not create aarch arm64 tar file correctly.

you can see the error here https://github.com/ellie/atuin/actions/runs/4584398463/jobs/8095857446
image

#### how to fix
 the first commit will fix this problem.  just add a RUSTFLAGS to set linker aarch64-linux-gnu-gcc
`RUSTFLAGS=-C linker=aarch64-linux-gnu-gcc`

the second commit is to bypass generate completion job because you can't run arm format on a Github hosted runner which is x86-64. we need to discuss how to fix the completion file. 

here's result:

image

FYI:

https://crates.io/crates/grb/1.3.0
https://users.rust-lang.org/t/unable-to-compile-to-aarch64-unknown-linux-gnu/53517

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -34,7 +34,7 @@ jobs:
       run: |
         case ${{ matrix.job.target }} in
           arm-unknown-linux-gnueabihf) sudo apt-get -y update ; sudo apt-get -y install gcc-arm-linux-gnueabihf ;;
-          aarch64-unknown-linux-gnu) sudo apt-get -y update ; sudo apt-get -y install gcc-aarch64-linux-gnu ;;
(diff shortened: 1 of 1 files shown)


ellie:
Hey! Thank you for working on this!

> the second commit is to bypass generate completion job because you can't run arm format on a Github hosted runner which is x86-64. we need to discuss how to fix the completion file.

It would be nice to solve this 🤔 

Maybe we should virtualize ARM for generating completions?

m2nx:
got it. I will give it a try to fix this.

m2nx:
image
image

 plz let me know if there is any issue.

m2nx:
@ellie

m2nx:
Hi @ellie is there anything else I can do about this PR?

ellie (approved):
Thank you for this! And sorry for being so slow to review it 😓 

Issue #850, opened 2023-04-05
Could not find ZSH history file


ivovegter:
I just installed atuin, and when trying to import my history from ZSH, it breaks. The history file exists, $HISTFILE is set correctly, but atuin says it cannot find it. See:

`$ la .histfile
Permissions Size User    Group   Date Modified Name
.rw-------  244k captain captain  5 Apr 13:03  .histfile

$ echo $HISTFILE
/home/captain/.zsh/.histfile

$ atuin import auto
        Atuin         
======================
          🌍          
       🐘🐘🐘🐘       
          🐢          
======================
Importing history...
Detected ZSH
Importing history from zsh
Error: Could not find history file. Try setting $HISTFILE

Location:
    atuin-client/src/import/zsh.rs:35:31
`

conradludgate:
Just taken a look through our ZSH import code. You should only get that error if the `HISTFILE` variable could not be found - or if it contains invalid utf-8.

ivovegter:
14.0.0-1 installed from Arch community repo

majutsushi:
Maybe the HISTFILE variable isn't exported? In that case zsh can see it, but atuin can't. Try running `export HISTFILE` to see if that fixes it.

Kadrian:
Thank you! That fixed it for me as well:

```
export HISTFILE && atuin import auto
```

 I'm on zprezto, so I wonder whether that was the reason HISTFILE wasn't exported?!

simon-b:
@Kadrian standard `zsh-newuser-install` adds:
```bash
HISTFILE=~/.histfile
```
to a new `~/.zshrc` file, with no `export`. So this will probably happen to many folks (myself included).

robertpeteuil:
Just happened to me as well, and an `export` fixed it.  

Perhaps you could add an info box to the import doc page (https://docs.atuin.sh/guide/import/) and suggest running the command as: `export HISTFILE && atuin import auto`

PR #876, opened 2023-04-14
Improve error message for issue #850.

postmath:
I just had the same thing as the person in issue #850: I hadn't exported my $HISTFILE variable (which I think is probably not an uncommon thing to do; for users who *don't* use atuin it's usually only the shell, not other programs, that interact with the history file). I was able to find the github issue that described it, but figured it would be helpful to include this suggestion in the error message.


ellie:
Hey! Thanks for clearing this up - if you can give it a `cargo fmt` then I'm happy to get it merged!

ellie (approved):
Awesome! Thank you 🙏 

PR #878, opened 2023-04-14
Avoid accidentally deleting all history, but allow it if intended

ellie:
It would make more sense to have an "atuin delete" or something for blowing away the whole lot. But, users will logically conclude that passing --delete with no query should delete the lot. So this puts a much more explicit switch to do what they want in the same place.


mentalisttraceur:
Strongly endorsed, per my reasoning in https://github.com/ellie/atuin/issues/877#issuecomment-1509498047 more or less.

conradludgate (approved):

PR #882, opened 2023-04-16
[feature] store env variables in History records

### LINKED ISSUE #608: Record the shell used for each command
I would find it useful if atuin recorded the shell used for each command.

For various reasons, I occasionally switch between shells for different purposes, and sometimes a command that works in one shell (bash/zsh) doesn't work in another (fish). This is even more of an issue if atuin grows to support more of the exotic shells (nushell, elvish, etc).

Having the ability to know which shell was used for a command would let me filter by commands for the current shell environment.

utterstep:
_Whew! That's something..._

Store env-variables (more precisely, configurable subset) which were set at the start of a command.
Also closes #608, storing interpreter and allowing to filter by it.

Here is summary, broken down by commits. Also, I've left some comments here in this PR, hope it'll help to review.

* 8e99b9b8f8ca179427fe344cef0ab111f8c9790e: remove `HistoryWithoutDelete`, add some docstrings, tests (now in #933)
* 772429ef284d327a32a4bd1b9ef68b9c3df5c24a: (minor) clear apt cache in Dockerfile, to reduce resulting image size a bit (now in #931)
* 0490de0da657a3f15e4914ea7406bb4c564605c4: create set of builders for `History` struct to simplify creation, while still having compile-time required fields, using [`typed-builder`](https://github.com/idanarye/rust-typed-builder) (now in #933)
* 81ae5a2189946714086fa61c04be20c60ae38bb2: add `interpreter` field to `History` and allow to filter by it
* a5f23aa5415441a58156676d7adccd663b3918ea: add `context` field to `History`, currently storing only environment variables, and add configuration to specify, which env vars to filter out
* 44d897e8d64cd01c96e87a47237188bfffee8564: (minor) move contributors to separate text file in root (now in #932)
* 96680fb73405422abe93191bb88edbe6f5cddaff: (minor) update History docs, listing available builders (now in #933)

PS Hope that large +diff would not scare you – I've also added bunch of docs and extracted CONTRIBUTORS in separate file, so not all of ~+900 is a pure code :)


CONTRIBUTORS
utterstep, line 1: Extracted content from `atuin/src/command/contributors.rs` to separate text file in root, so it would be more accessible to anyone reading/editing source code

Dockerfile
```diff
@@ -19,8 +19,9 @@ RUN cargo build --release --bin atuin
 FROM debian:bullseye-20230320-slim AS runtime
 
 RUN useradd -c 'atuin user' atuin && mkdir /config && chown atuin:atuin /config
-RUN apt update && apt install ca-certificates -y # so that webhooks work
-WORKDIR app
+# Install ca-certificates for webhooks to work
+RUN apt update && apt install ca-certificates -y && rm -rf /var/lib/apt/lists/*
```
utterstep, line 23: small enhancement – reduces size of a resulting image a bit

atuin-client/config.toml
```diff
@@ -31,7 +31,7 @@
 # search_mode = "fuzzy"
 
 ## which filter mode to use
-## possible values: global, host, session, directory
+## possible values: global, host, session, directory, interpreter
```
utterstep, line 34: new filtering mode, to show only entries from the same interpreter

atuin-client/config.toml
```diff
@@ -73,3 +73,13 @@
...
+
+## should env-vars be captured in the history
+# capture_env = false
+
+## prevent env-vars matching any of these regexes from being written to history.
+## As with history_filter, these regexes are unanchored.
+# env_filter = [
+#   "^SECRET_.*",
+#   "^AWS_SECRET_ACCESS_KEY$"
+# ]
```
utterstep, line 85: Allow capturing env-variables.
`false` by default, as there are frequently some secret-like values stored in env, so not everyone might be comfortable discovering that they are recorded by default.

Also allow to configure black list of vars.

utterstep: Maybe it should be white list?

atuin-client/src/database.rs
```diff
@@ -38,18 +42,32 @@ pub struct OptFilters {
...
 
-pub fn current_context() -> Context {
+pub fn current_context(settings: &Settings) -> Context {
     let Ok(session) = env::var("ATUIN_SESSION") else {
         eprintln!("ERROR: Failed to find $ATUIN_SESSION in the environment. Check that you have correctly set up your shell.");
         std::process::exit(1);
     };
     let hostname = format!("{}:{}", whoami::hostname(), whoami::username());
     let cwd = utils::get_current_dir();
+    let interpreter = env::var("ATUIN_INTERPRETER").ok();
```
utterstep, line 52: I've decided to pass interpreter through environment variable, not as the parameter. This felt more natural, as interpreter is used in `atuin history start` and `atuin history search` now, and may be will be in other commmands

atuin-client/src/database.rs
```diff
@@ -138,10 +156,35 @@ impl Sqlite {
     }
 
     async fn save_raw(tx: &mut sqlx::Transaction<'_, sqlx::Sqlite>, h: &History) -> Result<()> {
-        sqlx::query(
-            "insert or ignore into history(id, timestamp, duration, exit, command, cwd, session, hostname, deleted_at)
-                values(?1, ?2, ?3, ?4, ?5, ?6, ?7, ?8, ?9)",
-        )
+        sqlx::query(indoc::indoc! {
```
utterstep, line 159: added [`indoc`](https://crates.io/crates/indoc) to format query to more readable way, while removing unnecessary spaces from it.

atuin-client/src/database.rs
```diff
@@ -150,27 +193,35 @@ impl Sqlite {
...
-            exit: row.get("exit"),
-            command: row.get("command"),
-            cwd: row.get("cwd"),
-            session: row.get("session"),
-            hostname: row.get("hostname"),
-            deleted_at: deleted_at.map(|t| Utc.timestamp_nanos(t)),
-        }
+        let context: Option<String> = row.get("context");
+        // this currently ignores any invalid json quietly, maybe we should log it?
+        let context: Option<HistoryContext> = context.and_then(|c| serde_json::from_str(&c).ok());
```
utterstep, line 209: I've chosen to skip errors here

atuin-client/src/database.rs
```diff
@@ -376,6 +442,10 @@ impl Database for Sqlite {
             FilterMode::Host => sql.and_where_eq("hostname", quote(&context.hostname)),
             FilterMode::Session => sql.and_where_eq("session", quote(&context.session)),
             FilterMode::Directory => sql.and_where_eq("cwd", quote(&context.cwd)),
+            FilterMode::Interpreter => match &context.interpreter {
+                Some(i) => sql.and_where_eq("interpreter", quote(i)),
+                None => sql.and_where_is_null("interpreter"),
+            },
```
utterstep, line 448: Also filtering

atuin-client/src/history.rs
```diff
@@ -5,38 +5,67 @@ use serde::{Deserialize, Serialize};
...
-// Any new fields MUST be Optional<>!
-#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, sqlx::FromRow)]
+mod builder;
+mod context;
+
+pub use context::Context;
+
+/// A marker type used to seal the `History` struct, preventing it from being constructed directly.
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq)]
+pub(self) struct HistorySeal;
```
utterstep, line 15: Created `HistorySeal` ZST, which can be constructed from this module, but is unavailable for any external ones. This assures that now only builders can be used to created `History` struct, and these are type-safe, so incorrect `History` object cannot be produced.

atuin-client/src/history.rs
```diff
@@ -5,38 +5,72 @@ use serde::{Deserialize, Serialize};
...
+
+/// Client-side history entry.
+///
+/// Client stores data unencrypted, and only encrypts it before sending to the server.
+///
+/// To create a new history entry, use one of the builders:
+/// - [`History::import()`] to import an entry from the shell history file
+/// - [`History::capture()`] to capture an entry via hook
+/// - [`History::from_db()`] to create an instance from the database entry
```
utterstep, line 24: This is new ways to construct `History` objects, suitable for different scenarios. As such, there are a less required (or possible) fields while capturing new entry, then there are fields for an imported one. And entry, queried from DB, should have all the fields.

atuin-client/src/history.rs
```diff
@@ -5,38 +5,72 @@ use serde::{Deserialize, Serialize};
...
-    pub timestamp: chrono::DateTime<Utc>,
-    pub duration: i64,
-    pub exit: i64,
-    pub command: String,
-    pub cwd: String,
-    pub session: String,
-    pub hostname: String,
+    /// Interpreter (usually some kind of shell) that was used to run the command.
+    #[serde(default)]
+    pub interpreter: Option<String>,
```
utterstep, line 56: I've decided to make it a distinct parameter in the end, but that's my take, I'm totally fine with reconsidering it, given the arguments :)

For now it just felt more natural – for example, it makes sense to enable interpreter capture by default, but disable env vars capture.

atuin-client/src/history.rs
```diff
@@ -5,38 +5,72 @@ use serde::{Deserialize, Serialize};
...
+    /// [`History::import()`], [`History::from_db()`] or [`History::capture()`].
+    #[doc(hidden)]
+    #[serde(skip)]
+    _seal: HistorySeal,
 }
 
 impl History {
     #[allow(clippy::too_many_arguments)]
-    pub fn new(
+    fn new(
```
utterstep, line 73: `History::new` is private now, used only by builders, so it hurts less to add more parameters to it. :)

atuin-client/src/history.rs
```diff
@@ -62,10 +98,241 @@ impl History {
             session,
             hostname,
             deleted_at,
+            interpreter,
+            context,
+            _seal: HistorySeal,
         }
     }
 
+    /// Builder for a history entry that is imported from shell history.
```
utterstep, line 107: here and below are new builders, with some documentation considering their scenarios and usage

atuin-client/src/history.rs
```diff
@@ -62,10 +98,241 @@ impl History {
...
+        pub duration: i64,
+        pub exit: i64,
+        pub command: String,
+        pub cwd: String,
+        pub session: String,
+        pub hostname: String,
+    }
+
+    #[test]
+    fn test_backwards_compatibility() {
```
utterstep, line 263: Here is the test showing that we don't need `HistoryWithoutDelete`

atuin-client/src/history.rs
```diff
@@ -62,10 +98,241 @@ impl History {
...
+        let serialized = rmp_serde::to_vec(&history_without_delete).expect("Failed to serialize");
+        let deserialized: History =
+            rmp_serde::from_slice(&serialized).expect("Failed to deserialize");
+
+        assert!(deserialized.deleted_at.is_none());
+    }
+
+    #[test]
+    #[should_panic = "Failed to deserialize: LengthMismatch(8)"]
+    fn test_forwards_compatibility_with_rmp() {
```
utterstep, line 300: But, forwards compatibility is broken with RMP – we can deserialize old object with a new client, but we cannot do same with new object and old client. The same problem was prevalent with old `History`+`HistoryWithoutDelete` approach, this test just highlights it.

atuin-client/src/history/builder.rs
utterstep, line 1: Here all the builders and their required fields are specified.
The [`typed-builder`](https://github.com/idanarye/rust-typed-builder) crate is used, so it would be impossible to forget non-optional field in any scenario.

atuin-client/src/history/context.rs
utterstep, line 1: `Context` struct, currently containing env vars, but it could be extended if needed.

atuin-client/src/history/context.rs
```diff
@@ -0,0 +1,26 @@
...
+    pub env_vars: Option<HashMap<String, String>>,
+}
+
+impl ToString for Context {
+    fn to_string(&self) -> String {
+        serde_json::to_string(self).expect(
+            "JSON serialization failed, but it shouldn't, as all keys are guaranteed to be strings",
+        )
+    }
+}
```
utterstep, line 26: It's `ToString` uses JSON serialization forcefully, so that [JSON functionality](https://www.sqlite.org/json1.html), shipped with SQLite, could be used if needed.

atuin-client/src/import/resh.rs
```diff
@@ -122,18 +122,18 @@ impl Importer for Resh {
...
-            .await?;
+            let imported = History::import()
+                .interpreter(Self::NAME)
+                .command(entry.cmd_line)
+                .timestamp(timestamp)
+                .duration(duration)
+                .exit(entry.exit_code)
+                .cwd(entry.pwd)
+                .hostname(entry.host)
+                // CHECK: should we add uuid here? It's not set in the other importers
```
utterstep, line 133: I left it as it was, but added the comment to highlight, that we don't set `session_id` in other importers.
So in this case all entries, imported from `resh` will have different `session_id`.

atuin/src/command/client/history.rs
```diff
@@ -176,63 +176,127 @@ pub fn print_cmd_only(w: &mut StdoutLock, h: &[History]) {
 }
 
 impl Cmd {
-    pub async fn run(&self, settings: &Settings, db: &mut impl Database) -> Result<()> {
-        let context = current_context();
+    async fn handle_start(
```
utterstep, line 179: I've split `run` into private sub-fns, because it grew larger than clippy limit :)

atuin/src/shell/atuin.bash
```diff
@@ -1,5 +1,7 @@
 ATUIN_SESSION=$(atuin uuid)
+ATUIN_INTERPRETER=bash
 export ATUIN_SESSION
+export ATUIN_INTERPRETER
```
utterstep, line 4: All integrations now should (SHOULD, not MUST) include `ATUIN_INTERPRETER` variable

conradludgate:
Hi! I appreciate the PR, but this is far too much to review all at once. I respect that it's all split by commit, but I think it would be nice to have them split into multiple PRs.

Some preliminary feedback:
* `remove HistoryWithoutDelete` - I know this is an awkward one, but we explicitly want it to be awkward for now so we can fix it properly - so I'd ask to not make this change.

The rest of the changes seem fine, although I'm not sure how necessary they all are to this feature change. I like the builder pattern and the Dockerfile change, but again they should really be separate PRs.

Also, while the interpreter and the env storage changes are similar in scope, and I appreciate you making them 2 different migrations to track, I would also appreciate separate PRs.

I know this seems very awkward, but managing an open source project with quite a number of users does become quite a maintenance challenge. We really appreciate the enthusiasm to make atuin a better tool, and we will eventually get this all merged, but we need to be extra careful. I hope you understand. Feel free to message us on discord if there's anything you want clarifying

utterstep:
Hi, Conrad! No worries, I understand the difficulty and can split this PR in multiple ones.

Question about `HistoryWithoutDelete` – why do you need it?
The solution I propose still has compatibility `HistoryWithoutDelete` would give (_newer_ client + _older_ data), and it still highlights the problem of incompatibility of _older_ client with _newer_ data.
But it allows for adding fields without the need of `HistoryWithoutDeleteAndInterpreterAndContext` + `HistoryWithoutInterpreterAndContext` + `HistoryWithoutContext` in this case.
Is that ok with you?

We can have this conversation in a different PR, if that would be more convenient for you.

And thanks in any case for your attention :)

utterstep:
Created #931, #932 and #933 (updated PR summary to link which is which). I'll hold off `interpreter` and `context`/`env_vars` for now, as they are dependent on #933 :)

ellie:
> Hi, @conradludgate, @ellie – what do you think about that? I [created](https://github.com/ellie/atuin/pull/882#issuecomment-1532928655) some separate PRs and can do something else required :)
> 
> Would like to continue work on that feature

Hey! I'm sorry but I'm going to close this one for now. As a feature this would still be nice to have, however with the introduction of sync v2 and things like our key value store I think it's better suited to elsewhere. I'd also like to consider the ability to include the output of arbitrary shell commands as an attachment to history (so eg we could also store the current git branch, and this feature could just save `env`)

Apologies for the wasted time and energy here

utterstep:
Ok, not a big deal, things happen :)

Let me know if you'll need any help someday

dufferzafar:
@ellie What is the plan now for storing environment variables? I'd like to store the conda environment I'm in along with the commands so I can quickly filter based on that.

PR #894, opened 2023-04-18
Re-added package name to workspace.package

bdavj:
The package name was removed when `cargo.toml` was refactored to use `workspace.package`. This breaks the binary builds in actions, which depend on this key to generate a package name.

Re-add package name under `workspace.package`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,6 +2,7 @@
 members = ["atuin", "atuin-client", "atuin-server", "atuin-common"]
 
 [workspace.package]
+name = "atuin"
 version = "14.0.1"
 authors = ["Ellie Huxtable <ellie@elliehuxtable.com>"]
 rust-version = "1.59"


ellie:
Hey! Thanks for the contribution, but this has already been fixed in #885 

The issue we have now is triggering the _new_ workflow, but for an _old_ release. Which requires the workflow rewriting to take a SHA/release as input to build for, from workflow_dispatch

ellie:
...although actually, on second thoughts, it probably would be sensible to add this anyway 🤔

ellie (approved):
Thank you for this! And sorry for the open/close 😂 

If you fancy a contributors-only holographic sticker, then please fill out this form: https://notionforms.io/forms/contributors-stickers

Otherwise, do feel free to come hang out on our Discord! https://discord.gg/Fq8bJSKPHh

PR #905, opened 2023-04-22
Add command flag for `inline_height`

VuiMuich:
This PR adds the flag `--inline-height <INLINE_HEIGHT>` to the `search` subcommand.
It only applies for interactive mode, should this be added to the help text?
I suppose the "number of lines" is hint enough that it expects an integer value, right?

### DIFF
--- a/atuin/src/command/client/search.rs
+++ b/atuin/src/command/client/search.rs
@@ -96,6 +96,10 @@ pub struct Cmd {
     /// Example: --format "{time} - [{duration}] - {directory}$\t{command}"
     #[arg(long, short)]
     format: Option<String>,
+
+    /// Set the maximum number of lines Atuin's interface should take up.
+    #[arg(long = "inline-height")]
+    inline_height: Option<u16>,
 }
 
 impl Cmd {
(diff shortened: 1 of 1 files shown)


ellie (commented):
I think this should be self explanatory enough! Would you be able to add it to the docs please? Otherwise looks great, thanks for the work! 🥳

VuiMuich:
I also added two lines for `format` and `help` to be consistent with the actual output of `help`.

docs/docs/commands/search.md
```diff
@@ -9,21 +9,24 @@ Atuin search supports wildcards, with either the `*` or `%` character. By
...
+| `--interactive`/`-i` | Open the interactive search UI (default: false)                               |
+| `--human`            | Use human-readable formatting for the timestamp and duration (default: false) |
+| `--limit`            | Limit the number of results (default: none)                                   |
+| `--offset`           | Offset from the start of the results (default: none)                          |
+| `--delete`           | Delete history matching this query                                            |
+| `--delete-it-all`    | Delete all shell history                                                      |
+| `--reverse`          | Reverse order of search results, oldest first                                 |
+| `--format`/`-f`      | Available vareables: {command}, {directory}, {duration}, {user}, {host}, {time}, {exit} and {relativetime}. Example: --format "{time} - [{duration}] - {directory}$\t{command}" |
+| `--inline-height`    | Set the maximum number of lines Atuin's interface should take up              |
+| `--help`/`-h`        | Print help                                                                    |
```
conradludgate, line 29: starting to think we should have a separate subcommand (or maybe an arg group now) for interactive UI 😓

VuiMuich: :+1:  for the `atuin interactive` subcommand.

PR #907, opened 2023-04-22
Add exit mode `return-execute` including keybind

### LINKED ISSUE #78: Key binding to run the command immediately
It'd be nice if there were a key binding that somehow ran the selected command immediately instead of simply placing it on the command line. Having to hit enter twice to run a command I don't want to modify is a tad annoying.

VuiMuich:
## This PR adds a new `settings::ExitMode::ReturnExecute`.

In the match cases in `fn handle_key_input` we still return `RETURN_ORIGINAL` or `RETURN_QUERY` as we are not looking for other entries in the history, we just want to treat the output differently.
Therefore we check the value of `exit_mode` in `Cmd::run(...)` and decide if we execute or print.

I have not found a way I like to run the command though. `std::process::Command` would require us to parse the string into its command and argument parts just to reassemble them immediately, which is really ugly and it does not run it as child process of the shell but of atuin itself. So I decided to open as draft PR open to discussion.

Additionally I added `Shift+Enter` as a keybind to execute the result. Therefor we change the `exit_mode` on the fly with a new method `Settings::change_exit_mode(ExitMode)`.

Also up for discussion, how to proceed with `down arrow` action. For now we fall back to `RETURN_ORIGINAL` and switch on the fly to `ReturnOriginal` mode, as this is the least obtrusive.

closes #78


VuiMuich:
With the latest commit the new `Shift+Enter` keybind is working correct, afaict.
The `Esc` keybind does not work as expected, but I am to tired to see why at this point.
Will look at this at another time.

VuiMuich:
Fixed that the `ExitMode::ReturnExecute` works correct, when configured and `Esc` key is used.
Broke that `Enter` and `Shift+Enter` are distinct :facepalm:

VuiMuich:
Ok, for whatever reason this only works with `Alt+Enter`.
If anyone has some pointers as to why this is, I am willing to investigate how to get the desired keybind.

Besides that I guess this is review ready.

atuin/src/command/client/search.rs
```diff
@@ -123,7 +124,13 @@ impl Cmd {
 
         if self.interactive {
             let item = interactive::history(&self.query, settings, db).await?;
-            eprintln!("{item}");
+
+            if settings.exit_mode == ExitMode::Execute {
+                let mut cmd = shell(item);
```
ellie, line 129: I think we're going to need to ensure that Atuin executes the command as a child of the shell, and not of Atuin. We'll need to find some sort of structured data that can be returned to the shell plugin, parsed there, and then trigger execution or just insert with the line editor

That should also solve the aliases issue

VuiMuich: Oh, good point!
That would also explain, what I just noticed yesterday: those direct executions were not logged in the history.

I also started looking for such control character, but haven't found anything, yet.
I tried writing `\n` as byte directly to stdout, but that didn't work.

If I find something, it would probably be best to move this to its own `fn` so it's easier to `#[cfg(unix)]`/`#[cfg(windows)]`.

YummyOreo: You could make it exit w/ different exit codes catch that in the plugin and if exit code == x run the command

VuiMuich: As a temporary solution I implemented the fallback if the exit code is non-zero.
Little caveat atm is, that for some reason it looses colorinfo through the pipe.

And eventually I would prefer if we could run it as an actual child of the parent shell to the current atuin instance, but I need more time to research on this..

adamhp:
Any more thought or work being given to this @VuiMuich or @ellie?

ellie:
@adamhp hey! We discussed on discord and concluded we'd go in another direction 

tl;dr is that the command should not be executed as a child of atuin, and the responsibility lies with the shell plug-in to handle this

PR #910, opened 2023-04-23
Interactive delete

### LINKED ISSUE #854: Interactive deletion
``atuin search --delete -i`` should work as an interactive deletion.
Right now the ``--delete`` argument is ingored.

YummyOreo:
closes #854 

This implements if you press `Alt + d` on a entry. It will delete all of the entries matching that exactly.
> Note: this keybinding is subject to comments and change. 

To do this I added a `Strict` search mode. 

Todo:
- [ ] Add some sort of confirmation
- [ ] Only enable when deletion is enabled
- [ ] Improve ui (should not exit when after deletion)

### DIFF
--- a/atuin-client/src/settings.rs
+++ b/atuin-client/src/settings.rs
@@ -27,6 +27,9 @@ pub enum SearchMode {
     #[clap(aliases = &["fulltext"])]
     FullText,
 
+    #[serde(rename = "strict")]
+    Strict,
+
     #[serde(rename = "fuzzy")]
     Fuzzy,
 
@@ -39,6 +42,7 @@ impl SearchMode {
         match self {
             SearchMode::Prefix => "PREFIX",
             SearchMode::FullText => "FULLTXT",
+            SearchMode::Strict => "STRICT",
             SearchMode::Fuzzy => "FUZZY",
             SearchMode::Skim => "SKIM",
         }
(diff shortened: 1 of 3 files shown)


atuin/src/command/client/search/interactive.rs
```diff
@@ -574,15 +593,43 @@ pub async fn history(
...
+            while !entries.is_empty() {
+                for entry in &entries {
+                    println!("deleting {}", entry.command);
+                    db.delete(entry.clone()).await?;
+                }
+
+                entries =
+                    run_non_interactive(&settings, OptFilters::default(), query, &mut db).await?;
+            }
+            Ok(String::new())
```
YummyOreo, line 632: Pull this out into a function that both deletes can use

atuin-client/src/settings.rs
```diff
@@ -50,6 +54,7 @@ impl SearchMode {
             SearchMode::FullText if settings.search_mode == SearchMode::Skim => SearchMode::Skim,
             // otherwise fuzzy.
             SearchMode::FullText => SearchMode::Fuzzy,
+            SearchMode::Strict => SearchMode::Strict,
```
YummyOreo, line 57: Should not be the same

YummyOreo: Mby fuzzy/skim

ellie:
This is now part of the inspector! Closing 🙏

PR #913, opened 2023-04-24
Fix `--delete-it-all` and `--delete` commands

edeustua:
There are two issues with deleting:
1. If `--delete-it-all` is used, it atuin will enter an infinite loop, searches are performed on the whole history, including deleted items.
2. Blanking the command field, may violate the `unique(timestamp, cwd, command)` condition.

This PR adds a `deleted_at is null` condition to the search query to fix issue 1. And removes the blanking of the command to fix 2.


atuin-client/src/database.rs
```diff
@@ -515,7 +517,6 @@ impl Database for Sqlite {
     // but the time that the system marks it as deleted
     async fn delete(&self, mut h: History) -> Result<()> {
         let now = chrono::Utc::now();
-        h.command = String::from(""); // blank it
```
ellie, line 518: I totally agree it shouldn't really be blank, thanks for highlighting this. However would you be able to set it to some random data instead? This way, `strings history.db` won't still show whatever the deleted value is

ellie (approved):
Thank you for sorting this out!

PR #921, opened 2023-04-26
fix for zsh no-unset environments

ds-cbo:
When I run `atuin init zsh` in my zshrc, I am greeted with

```
_atuin_precmd:3: ATUIN_HISTORY_ID: parameter not set
```

any time I boot my shell, because I also have `set -o nounset` to tell me when I'm accidentally using variables that don't exist, for example because of a typo.

This makes sure that error doesn't happen


ellie (approved):
Ahhh nice catch, thank you!

ellie (approved):

PR #929, opened 2023-05-01
fix: nushell empty hooks

### LINKED ISSUE #928: Nushell init does not support undefined `config.hooks.pre_execution` and `config.hooks.pre_prompt`
If the array for any of the hooks in the title is not defined (does not have to be), the script produced by `atuin init nu` does not work.

I think this is the same as #841 (I get the same error).

A solution is to first assign empty arrays for those hooks and then adding elements to them:
```
let-env config = (
    $env.config | upsert hooks (
        $env.config.hooks
        | upsert pre_execution ($env.config.hooks | get -i pre_execution | default [] | append $_atuin_pre_execution)
        | upsert pre_prompt ($env.config.hooks | get -i pre_prompt | default [] | append $_atuin_pre_prompt)
    )
)
```

acidghost:
Fixes #928.

### DIFF
--- a/atuin/src/shell/atuin.nu
+++ b/atuin/src/shell/atuin.nu
@@ -35,10 +35,14 @@ def _atuin_search_cmd [...flags: string] {
     ] | str join "\n"
 }
 
-let-env config = (
+$env.config = ($env | default {} config).config
+$env.config = ($env.config | default {} hooks)
+$env.config = (
     $env.config | upsert hooks (
         $env.config.hooks
-        | upsert pre_execution ($env.config.hooks.pre_execution | append $_atuin_pre_execution)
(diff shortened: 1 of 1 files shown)


atuin/src/shell/atuin.nu
```diff
@@ -35,10 +35,14 @@ def _atuin_search_cmd [...flags: string] {
     ] | str join "\n"
 }
 
+let-env config = ($env | default {} config).config
```
ellie, line 38: let-env is going to be removed soon I'm afraid - see https://github.com/ellie/atuin/pull/1080

LeoniePhiline: This might be mergeable now. :)

PR #931, opened 2023-05-03
Move contributors list to top-level file

utterstep:
One part of #882

Extract content from `atuin/src/command/contributors.rs` to separate text file in root, so it would be more accessible to anyone browsing repo/editing this list. Top-level file seems more appropriate place to hold list of contributors, plus editing this list is now more obvious "edit top-level text file", rather than "find source code of a command and change it".

Also this PR adds myself to list :)


atuin/src/command/CONTRIBUTORS
conradludgate, line 1: can't have package files outside of the package directory, so have to symlink it in

ellie (approved):
This is great, thank you!

PR #932, opened 2023-05-03
clean apt cache in Dockerfile

utterstep:
One part of #882

Minor change, cleans apt cache in final image


Dockerfile
```diff
@@ -19,8 +19,9 @@ RUN cargo build --release --bin atuin
 FROM debian:bullseye-20230320-slim AS runtime
 
 RUN useradd -c 'atuin user' atuin && mkdir /config && chown atuin:atuin /config
-RUN apt update && apt install ca-certificates -y # so that webhooks work
-WORKDIR app
+# Install ca-certificates for webhooks to work
+RUN apt update && apt install ca-certificates -y && rm -rf /var/lib/apt/lists/*
```
ellie, line 23: any reason why not using `apt-get clean`?

utterstep: This are two different caches, AFAIKT: package list cache, created by `apt-get update` and packages cache, created by `apt-get install`.

And, as stated in [Docker Best Practices](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/), `apt-get clean` is running by default, so we only need to clean list cache here.

Screenshot_20230510-093858

conradludgate (approved):

conradludgate (approved):

PR #933, opened 2023-05-03
Builder interface for History objects

utterstep:
Also part of #882

Allow History to be created through typed builder interface, powered by [`typed-builder`](https://github.com/idanarye/rust-typed-builder) crate.

This allows to specify different required fields for a different scenarios:
1) entry, imported from shell history should have at least `command` and `timestamp` and _may_ have some other fields specified, if shell history stores them
2) entry, caught by hook – has `timestamp`, `command` and `cwd`. These are only fields accessible at hook's `start` phase, all the other ones (`duration`, `exit` and such) are added after command is executed, on an instance, queried from DB by id.
3) entry, queried from DB. In this case all fields are required, as we're querying the whole object as stored from DB.

I've also added documentation for `History` and provided builders, and added `_seal` field to the `History` to ensure that it could not be erroneously constructed from outside without going through the builder, which enforces required constraints.

### DIFF
--- a/atuin-client/src/history.rs
+++ b/atuin-client/src/history.rs
@@ -1,7 +1,6 @@
 use std::env;
 
 use chrono::Utc;
-use serde::{Deserialize, Serialize};
 
 use atuin_common::utils::uuid_v7;
 
@@ -23,7 +22,7 @@ pub(self) struct HistorySeal;
 /// ### Caution
 /// Any new fields MUST be `Optional<T>` and marked with `#[serde(default)]` to ensure backwards
 /// compatibility with older clients.
-#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, sqlx::FromRow)]
+#[derive(Debug, Clone, PartialEq, Eq, sqlx::FromRow)]
 pub struct History {
     /// A client-generated ID, used to identify the entry when syncing.
     ///
@@ -43,19 +42,8 @@ pub struct History {
     pub session: String,
     /// The hostname of the machine the command was run on.
     pub hostname: String,
-
     /// Timestamp, which is set when the entry is deleted, allowing a soft delete.
-    #[serde(default)]
     pub deleted_at: Option<chrono::DateTime<Utc>>,
-
-    /// Having this seal marker here we're ensuring that `History`
-    /// can only be constructed directly by the [`crate::history`] module.
-    ///
-    /// All users of `History` must use builders, such as
-    /// [`History::import()`], [`History::from_db()`] or [`History::capture()`].
-    #[doc(hidden)]
-    #[serde(skip)]
-    _seal: HistorySeal,
 }
 
 impl History {
@@ -91,7 +79,6 @@ impl History {
             session,
             hostname,
             deleted_at,
-            _seal: HistorySeal,
         }
     }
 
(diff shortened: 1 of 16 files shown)


utterstep:
I'll create PRs with other changes from #882 after this PR is resolved, because they depend on changes from this one

atuin-client/src/history.rs
```diff
@@ -5,38 +5,57 @@ use serde::{Deserialize, Serialize};
 
 use atuin_common::utils::uuid_v7;
 
-// Any new fields MUST be Optional<>!
+mod builder;
+
+/// A marker type used to seal the `History` struct, preventing it from being constructed directly.
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq)]
+pub(self) struct HistorySeal;
```
utterstep, line 12: Created `HistorySeal` ZST, which can be constructed from this module, but is unavailable for any external ones. This assures that now only builders can be used to created `History` struct, and these are type-safe, so incorrect `History` object cannot be produced.

atuin-client/src/history.rs
```diff
@@ -5,38 +5,62 @@ use serde::{Deserialize, Serialize};
...
+
+/// Client-side history entry.
+///
+/// Client stores data unencrypted, and only encrypts it before sending to the server.
+///
+/// To create a new history entry, use one of the builders:
+/// - [`History::import()`] to import an entry from the shell history file
+/// - [`History::capture()`] to capture an entry via hook
+/// - [`History::from_db()`] to create an instance from the database entry
```
utterstep, line 21: These are the new ways to construct `History` objects, suitable for different scenarios. As such, there are a less required (or possible) fields while capturing new entry, then there are fields for an imported one. And entry, queried from DB, should have all the fields.

atuin-client/src/history.rs
```diff
@@ -5,38 +5,62 @@ use serde::{Deserialize, Serialize};
...
+///
+/// ### Caution
+/// Any new fields MUST be `Optional<T>` and marked with `#[serde(default)]` to ensure backwards
```
utterstep, line 24: Details on what should be done to have compatible serialization without the need of smth like `HistoryWithoutDelete`

cc @conradludgate regarding removal of `HistoryWithoutDelete`

atuin-client/src/history.rs
```diff
@@ -5,38 +5,62 @@ use serde::{Deserialize, Serialize};
...
+    /// Timestamp, which is set when the entry is deleted, allowing a soft delete.
+    #[serde(default)]
     pub deleted_at: Option<chrono::DateTime<Utc>>,
-}
 
-// Forgive me, for I have sinned
-// I need to replace rmp with something that is more backwards-compatible.
-// Protobuf, or maybe just json
-#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, sqlx::FromRow)]
-pub struct HistoryWithoutDelete {
```
utterstep, line 26: We don't need such a struct with right serde attributes. Down below are the tests, confirming it.

cc @conradludgate regarding removal of HistoryWithoutDelete

atuin-client/src/history.rs
```diff
@@ -5,38 +5,62 @@ use serde::{Deserialize, Serialize};
...
+    /// [`History::import()`], [`History::from_db()`] or [`History::capture()`].
+    #[doc(hidden)]
+    #[serde(skip)]
+    _seal: HistorySeal,
 }
 
 impl History {
     #[allow(clippy::too_many_arguments)]
-    pub fn new(
+    fn new(
```
utterstep, line 63: `History::new` is private now, used only by builders, so it hurts less to add more parameters to it. :)

atuin-client/src/history.rs
```diff
@@ -62,10 +86,209 @@ impl History {
             session,
             hostname,
             deleted_at,
+            _seal: HistorySeal,
         }
     }
 
+    /// Builder for a history entry that is imported from shell history.
```
utterstep, line 93: here and below are new builders, with some documentation (and doctests) considering their scenarios and usage

atuin-client/src/history.rs
```diff
@@ -62,10 +86,209 @@ impl History {
             session,
             hostname,
             deleted_at,
+            _seal: HistorySeal,
```
utterstep, line 89: ZST that was mentioned earlier

atuin-client/src/history.rs
```diff
@@ -62,10 +86,209 @@ impl History {
...
+        pub duration: i64,
+        pub exit: i64,
+        pub command: String,
+        pub cwd: String,
+        pub session: String,
+        pub hostname: String,
+    }
+
+    #[test]
+    fn test_backwards_compatibility() {
```
utterstep, line 221: Here is the test showing that we don't need `HistoryWithoutDelete`

cc @conradludgate regarding removal of HistoryWithoutDelete

atuin-client/src/import/resh.rs
```diff
@@ -122,18 +122,17 @@ impl Importer for Resh {
...
-            .await?;
+            let imported = History::import()
+                .command(entry.cmd_line)
+                .timestamp(timestamp)
+                .duration(duration)
+                .exit(entry.exit_code)
+                .cwd(entry.pwd)
+                .hostname(entry.host)
+                // CHECK: should we add uuid here? It's not set in the other importers
+                .session(uuid_v7().as_simple().to_string());
```
utterstep, line 133: I left it as it was, but added the comment to highlight, that we don't set `session_id` in other importers.
So in this case all entries, imported from `resh` will have different `session_id`.

atuin/src/command/client/history.rs
```diff
@@ -184,110 +184,135 @@ fn parse_fmt(format: &str) -> ParsedFmt {
 }
 
 impl Cmd {
-    pub async fn run(&self, settings: &Settings, db: &mut impl Database) -> Result<()> {
-        let context = current_context();
+    async fn handle_start(
```
utterstep, line 187: I've split `Cmd::run` into private sub-fns, because it grew larger than clippy limit :)

atuin-client/src/history.rs
ellie: Hey! This is _intentionally_ horrible right now. We don't want to make it nice and easy to add new fields!

The reasoning here is that I'd like to change our serialization format to one that better supports versioning and migrations. By leaving it awkward, it'll force our hand in the future when we wish to add new fields. Making it "good enough" means that this will not happen.

utterstep: Ok, so in the initial PR we do want to add a new field :)

Would that be ok if I also change serialization format here, then?

utterstep: And in this PR I introduced failing test, so that we won't forget that migration is indeed broken in one direction: https://github.com/ellie/atuin/pull/933#discussion_r1183559996

conradludgate: I changed the mechanism to make how we encode histories for encryption to be explicit. This is the best of both worlds. `serde` is not designed for non self describing documents so it should not be used moving forward.

#1038 is using rmp-serde now too, so I will remove it from there in a separate PR.

conradludgate (approved):
Thanks, and sorry this took so long. I made some changes that would have been too complicated to try and explain.

I removed the HistorySeal as it's unnecessary pedantry, although I admire the effort for strict type safety

utterstep:
Thanks for your help and attention :)

PR #937, opened 2023-05-06
docs: fix "From source" `cd` command

rigrig:
Cloning does not automatically cd to the new directory, and we need to cd to "atuin" *inside* the cloned repository.

(Thanks to @YummyOreo for [pointing this out](https://github.com/ellie/atuin/issues/934#issuecomment-1535565743))


conradludgate (approved):
Thanks!

PR #942, opened 2023-05-06
respect ZSH's $ZDOTDIR environment variable

dennis-tra:
ZSH reads its configuration by default relative to a path set via $ZDOTDIR. If $ZDOTDIR is not set, then the value of $HOME is used; this is the usual case. Source:  https://zsh.sourceforge.io/Intro/intro_3.html#IDX11

Also, $HOME is safe to use because "The system shall initialize this variable at the time of login to be a pathname of the user's home directory". Source: https://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap08.html

Lastly, the shell parameter expansion syntax is safe to use either: https://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html#tag_02_06_02

I double-quoted the whole string in case there were spaces in the path.

--- 

You might want to double-check if similar considerations apply to bash.


ellie (approved):
Thanks for this! Sorry it took me to long to review

dennis-tra:
Thanks for merging! I just checked the change set again and saw that the line above should also have the same change. Will follow up with a second PR.

PR #974, opened 2023-05-11
Fix key regression

ellie:
Keys saved with the old code would load with the new code. Keys saved with the new code would not load with the new code.

The old code didn't save a slice, but the new code does!

This tries both methods of loading, one of which will work. Some people will have been running main already, so it's too late to just break things

I'd still like to ditch msgpack tbh.


conradludgate (approved):
Oops 🤔 but yes let's ditch msgpack

PR #980, opened 2023-05-15
feat: add delete account option (attempt 2)

yannickulrich:
This is implementing an account deletion option, similar to #266. I've made the following changes
 * added a `remove_user` function to the database that deletes the user from `users`, `sessions`, and `history`
 * added an API endpoint: currently this is the DELETE handler for the `/register` endpoint but could be moved. If this gets called by a logged in user, the account gets delete
 * added the `unregister` command to the CLI
```shell
$ atuin unregister
```

I'm wondering how we should prevent accidental account deletion. We could ask the user provide their password, email, or username
```shell
$ atuin unregister -e user@example.com
```
Any suggestions? 

PS: Thanks for looking into this and thanks for building this, it's a really cool tool :blush:

### DIFF
--- a/atuin-client/src/api_client.rs
+++ b/atuin-client/src/api_client.rs
@@ -222,10 +222,7 @@ impl<'a> Client<'a> {
         let url = format!("{}/register", self.sync_addr);
         let url = Url::parse(url.as_str())?;
 
-        let resp = self.client
-            .delete(url)
-            .send()
-            .await?;
+        let resp = self.client.delete(url).send().await?;
 
         if resp.status() == 403 {
             bail!("invalid login details");
--- a/atuin-client/src/api_client.rs
+++ b/atuin-client/src/api_client.rs
@@ -218,7 +218,7 @@ impl<'a> Client<'a> {
         Ok(())
     }
 
-    pub async fn unregister(&self) -> Result<()> {
+    pub async fn delete(&self) -> Result<()> {
         let url = format!("{}/register", self.sync_addr);
         let url = Url::parse(url.as_str())?;
 
--- a/atuin-client/src/api_client.rs
+++ b/atuin-client/src/api_client.rs
@@ -217,4 +217,22 @@ impl<'a> Client<'a> {
 
         Ok(())
     }
+
+    pub async fn unregister(&self) -> Result<()> {
+        let url = format!("{}/register", self.sync_addr);
+        let url = Url::parse(url.as_str())?;
+
+        let resp = self.client
+            .delete(url)
+            .send()
+            .await?;
+
+        if resp.status() == 403 {
+            bail!("invalid login details");
+        } else if resp.status() == 200 {
+            Ok(())
+        } else {
+            bail!("Unknown error");
+        }
+    }
 }
--- a/atuin-common/src/api.rs
+++ b/atuin-common/src/api.rs
@@ -20,6 +20,10 @@ pub struct RegisterResponse {
 }
 
 #[derive(Debug, Serialize, Deserialize)]
+pub struct UnregisterResponse {
+}
+
+#[derive(Debug, Serialize, Deserialize)]
 pub struct LoginRequest {
     pub username: String,
     pub password: String,
--- a/atuin-common/src/api.rs
+++ b/atuin-common/src/api.rs
@@ -20,7 +20,7 @@ pub struct RegisterResponse {
 }
 
 #[derive(Debug, Serialize, Deserialize)]
-pub struct UnregisterResponse {
+pub struct DeleteUserResponse {
 }
 
 #[derive(Debug, Serialize, Deserialize)]
--- a/atuin-common/src/api.rs
+++ b/atuin-common/src/api.rs
@@ -20,8 +20,7 @@ pub struct RegisterResponse {
 }
 
 #[derive(Debug, Serialize, Deserialize)]
-pub struct DeleteUserResponse {
-}
+pub struct DeleteUserResponse {}
 
 #[derive(Debug, Serialize, Deserialize)]
 pub struct LoginRequest {
--- a/atuin-server/src/database.rs
+++ b/atuin-server/src/database.rs
@@ -339,19 +339,19 @@ impl Database for Postgres {
 
     #[instrument(skip_all)]
     async fn delete_user(&self, u: &User) -> Result<()> {
-        sqlx::query_as::<_, Session>("delete from sessions where user_id = $1")
+        sqlx::query("delete from sessions where user_id = $1")
             .bind(u.id)
-            .fetch_all(&self.pool)
+            .execute(&self.pool)
             .await?;
 
-        sqlx::query_as::<_, Session>("delete from users where id = $1")
+        sqlx::query("delete from users where id = $1")
             .bind(u.id)
-            .fetch_all(&self.pool)
+            .execute(&self.pool)
             .await?;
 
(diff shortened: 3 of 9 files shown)


atuin-client/src/api_client.rs
```diff
@@ -217,4 +217,22 @@ impl<'a> Client<'a> {
 
         Ok(())
     }
+
+    pub async fn unregister(&self) -> Result<()> {
```
ellie, line 221: Could we call this delete please?

atuin-common/src/api.rs
```diff
@@ -19,6 +19,10 @@ pub struct RegisterResponse {
     pub session: String,
 }
 
+#[derive(Debug, Serialize, Deserialize)]
+pub struct UnregisterResponse {
```
ellie, line 23: Lets also call this one DeleteUserResponse

atuin-server/src/database.rs
```diff
@@ -27,6 +27,7 @@ pub trait Database {
     async fn get_user(&self, username: &str) -> Result<User>;
     async fn get_user_session(&self, u: &User) -> Result<Session>;
     async fn add_user(&self, user: &NewUser) -> Result<i64>;
+    async fn remove_user(&self, u: &User) -> Result<()>;
```
ellie, line 30: Again to keep naming consistent, could we call this delete_user please?

atuin-server/src/database.rs
```diff
@@ -336,6 +337,26 @@ impl Database for Postgres {
         Ok(())
     }
 
+    #[instrument(skip_all)]
+    async fn remove_user(&self, u: &User) -> Result<()> {
```
ellie, line 341: same here ;P

atuin-server/src/handlers/user.rs
```diff
@@ -138,6 +138,23 @@ pub async fn register<DB: Database>(
     }
 }
 
+#[instrument(skip_all, fields(user.id = user.id))]
+pub async fn unregister<DB: Database>(
```
ellie, line 142: and here 👀

atuin-server/src/handlers/user.rs
```diff
@@ -138,6 +138,23 @@ pub async fn register<DB: Database>(
...
+    user: User,
+    state: State<AppState<DB>>,
+) -> Result<Json<UnregisterResponse>, ErrorResponseStatus<'static>> {
+    debug!("request to remove user {}", user.id);
+
+    let db = &state.0.database;
+    if let Err(e) = db.remove_user(&user).await {
+        error!("failed to remove user: {}", e);
+
+        return Err(ErrorResponse::reply("failed to add history")
```
ellie, line 152:
```suggestion
        return Err(ErrorResponse::reply("failed to delete user")
```

atuin/src/command/client/sync/unregister.rs
```diff
@@ -0,0 +1,25 @@
+use std::path::PathBuf;
```
ellie, line 1: If we could also rename this file 🙌

docs/docs/commands/sync.md
```diff
@@ -32,6 +32,14 @@ notifications (security breaches, changes to service, etc).
 Upon success, you are also logged in :) Syncing should happen automatically from
 here!
 
+## Unregister
```
ellie, line 35:
```suggestion
## Delete
```

ellie (changes requested):
Hey! Thank you for your PR! Just a couple of comments

I was actually halfway through implementing this when you opened your PR, so also have a database migration that will make this really fast on the server. I can include my changes once this is in :D

yannickulrich:
Oh wow, that's such a quick response 🤩 I assume you still want to merge this, right? If you've already done a lot, I'm happy to close this..

Assuming you do, I rename it everywhere. Are we keeping the endpoint as it is?

atuin-server/src/router.rs
```diff
@@ -72,6 +72,7 @@ pub fn router<DB: Database + Clone + Send + Sync + 'static>(
         .route("/history", post(handlers::history::add))
         .route("/history", delete(handlers::history::delete))
         .route("/user/:username", get(handlers::user::get))
+        .route("/register", delete(handlers::user::unregister))
```
ellie, line 75:
```suggestion
        .route("/me", delete(handlers::user::unregister))
```

atuin-server/src/handlers/user.rs
yannickulrich: oops, sorry, copy & past 😶

ellie:
> Oh wow, that's such a quick response 🤩 I assume you still want to merge this, right? If you've already done a lot, I'm happy to close this..
> 
> Assuming you do, I rename it everywhere. Are we keeping the endpoint as it is?

Hahaha, I was looking at this anyway so all the context is there :D

Yup we can go with this! I'll probs just rebase what I have, as there are a couple of related changes I want to add. Thank you!

docs/docs/commands/sync.md
yannickulrich: What do we name the command, still `unregister` or `delete`?

atuin-server/src/router.rs
yannickulrich: Perfect, I like it 😁

yannickulrich: sorry, I've messed this up, hence the force push

atuin-server/src/database.rs
```diff
@@ -336,6 +337,26 @@ impl Database for Postgres {
...
+            .fetch_all(&self.pool)
+            .await?;
+
+        sqlx::query_as::<_, Session>("delete from history where user_id = $1")
+            .bind(u.id)
+            .fetch_all(&self.pool)
+            .await?;
+
+        Ok(())
+    }
```
ellie, line 358: Seeing as we don't need to fetch any data, we can actually just do this instead

```suggestion
    async fn delete_user(&self, u: &User) -> Result<()> {
        sqlx::query::("delete from sessions where user_id = $1")
            .bind(u.id)
            .execute(&self.pool)
            .await?;

        sqlx::query::("delete from users where id = $1")
            .bind(u.id)
            .execute(&self.pool)
            .await?;

        sqlx::query("delete from history where user_id = $1")
            .bind(u.id)
            .execute(&self.pool)
            .await?;

        Ok(())
    }
```

docs/docs/commands/sync.md
```diff
@@ -32,6 +32,14 @@ notifications (security breaches, changes to service, etc).
 Upon success, you are also logged in :) Syncing should happen automatically from
 here!
 
+## Delete
+
+You can delete your sync account with
+
+```
+atuin unregister
```
ellie, line 40: I've got some changes I'm making in this area, but I can make them once this is merged 😊

ellie (commented):
Just a couple more things then I think we are good to go! 🚀

docs/docs/commands/sync.md
ellie: I'll follow up with a change, I've restructured the commands a bit!

yannickulrich: Okay cool, in that case I leave it as it is  😊

yannickulrich:
I think I've changed everything you wanted, though I might've missed something somewhere (please let me know if there is)

ellie (approved):
Great! Thank you for working on this 🙏 

yannickulrich:
Thanks so much, I really love the idea of the stickers 🤩 Have just filled the form. I still worry about accidental deletion though. Would it not make sense to put some sort of safety in?

PR #984, opened 2023-05-16
Restructure account commands to account subcommand

ellie:
Also edit the history count trigger so it only runs on insert, and not on delete. This will make sure that deletes run super fast 🚀


atuin-server/migrations/20230515221038_trigger-delete-only.sql
```diff
@@ -0,0 +1,30 @@
...
+				new.user_id, 
+				(select count(1) from history where user_id = new.user_id)
+			);
+		end if;
+	end if;
+
+	return NEW; -- this is actually ignored for an after trigger, but oh well
+end;
+$func$
+language plpgsql volatile -- pldfplplpflh
```
conradludgate, line 24: Did you sneeze?

conradludgate (commented):
While this works as it did before, I realise that we don't count actual history delete requests here. Should we add a trigger on update? I think users might be confused if the remote count doesn't decrease when they delete

ellie:
> While this works as it did before, I realise that we don't count actual history delete requests here. Should we add a trigger on update? I think users might be confused if the remote count doesn't decrease when they delete

I wish :( If we actually update the remote count when rows are soft deleted, sync won't work. It's append-only, afterall, and expects counts to always increase

atuin-server/migrations/20230515221038_trigger-delete-only.sql
ellie: 🤧

ellie: No I just think the language statement is silly

ellie:
We should perhaps offset the deleted items from the remote count before displaying it though

conradludgate (approved):

PR #992, opened 2023-05-20
Allow ctrl modifier for number jumps

happenslol:
The modifier used to directly jump to a command in the results list is currently hardcoded to alt. I was looking for a different option here since I'm already using alt as a keybind for switching between terminal tabs.

Since ctrl+digits are usually interpreted as control sequences, I also had to enable keyboard enhancements in crossterm to make this work. The code for this is taken directly from the [crossterm event-read example](https://github.com/crossterm-rs/crossterm/blob/master/examples/event-read.rs). This works great with terminals that support advanced keyboard protocols like `wezterm` and `kitty`, and doesn't have any effect on other terminals.

One thing I'm unsure about here is whether we should notify the user if they've chosen ctrl as a modifier, but their terminal does not support keyboard enhancements. Maybe a note about this in the docs is enough, though.


happenslol:
Seems like this exact thing was implemented and accepted in #1106. Which is a bit sad since this has just been sitting around, but oh well :sweat_smile: 

On my terminal, I still need the enhancements for this to work, so I'm still having to use this branch to get the feature to work for me. Can anyone else confirm that the `ctrl + n` keybinds are working for them without keyboard enhancements enabled?

happenslol:
Ah, don't worry about it, I was just confused today when I brought my fork up to date and the functionality was duplicated.

I've maintained projects myself and I understand the workload can be immense. I'll keep in mind that adding documentation and option gates makes your job easier for the future 🙂

PR #993, opened 2023-05-20
Add option to completely disable help row

happenslol:
This adds an option to always hide the help row, independently from the mode or window height.


atuin/src/command/client/search/interactive.rs
```diff
@@ -536,7 +537,15 @@ pub async fn history(
             atuin_client::settings::Style::Compact => true,
             atuin_client::settings::Style::Full => false,
         };
-        terminal.draw(|f| app.draw(f, &results, compact, settings.show_preview))?;
+        terminal.draw(|f| {
+            app.draw(
+                f,
+                &results,
+                compact,
+                settings.show_preview,
```
ellie, line 545: I think at this point it is probably sensible to pass a reference to settings, rather than pass in some parts of it

ellie:
Hey! Thank you for contributing 😄 

Just the one comment but looks good otherwise. Would you also be able to update the docs please?

happenslol:
Sure thing, will do!

happenslol:
All done. Tell me if there's anything else you need!

PR #1004, opened 2023-05-23
new encryption scheme

conradludgate:
!! THIS IS UNTESTED DO NOT USE !!

This is a bunch of crypto changes and a tiny sync change

# History Entry Scheme

To allow future history encryption schemes than just our current crypto_box setup, this adds a scheme field to sync history items. It's backwards compatible (None -> legacy) and semi-forwards compatible by having built in support for unknown schemes.

# PasetoV4

PASETO (Platform Agnostic Security Tokens) is a scheme like JWT/JWE for signing and encrypting tokens for authentication. It is meant to be a lot more secure than JWT by design. While researching for the original design of this PR, I re-discovered paseto and read through their new V4 spec. It is quite similar to what I had invented myself (key derivation + authenticated encryption).

While I was confident with my solution being secure, I don't know what I don't know. Falling on the side of precaution, I've decided to switch to their scheme. Note, I don't want to be spec compliant with PASETO. It has more baggage than we need. However, taking inspiration is the goal here.

I've linked to the spec throughout the sync/pasetov4.rs file, and included a breakdown of the steps I am following.

---

⚠️ What follows is out of date, but leaving it here for comparison ⚠️

# XChaCha20Poly1305

Our current encryption is XSalsa20Poly1305. It uses the same key for every item but a random IV (note, not guaranteed to be unique).

The new scheme uses the more modern XChaCha20Poly1305, and makes use of the AEAD features by using the history ID as additional data to the algorithm. Additionally, we use a key derivation function (HKDF) to generate a new key for every entry. Combined with a random IV, this should be a lot more secure than the original setup. (I consulted some cryptography veterans to design this setup).

# Source

For the crypto research, I am basing it off a few sources

1. RustCrypto claims that ChaCha is just as fast as Salsa but with better encryption performance: https://github.com/RustCrypto/stream-ciphers/tree/master/chacha20
    + > It improves upon the previous [Salsa20](https://en.wikipedia.org/wiki/Salsa20) stream cipher with increased per-round diffusion at no cost to performance.
2. I was originally going to use pbkdf2, but it was suggested to me that with a high entropy key, you should use [hkdf](https://www.rfc-editor.org/rfc/rfc5869). It is similar, but works better for the key we already have.
3. The reason for using a KDF for each entry is to improve security. We don't guarantee that IV are unique. For stream ciphers like AESGCM or XChaCha20Poly1305, it's crucial to never [reuse the same IV with the same key](https://en.wikipedia.org/wiki/Stream_cipher_attacks). Generating a new key for each entry massively helps us to reduce that possibility. We now have about 448bits of entropy at our disposal, compared to 96bits we had before. A simple birthday paradox calculation tells us that we should expect a collision only after 2^224 history entries. That's good enough I think
4. Using a fixed salt for HKDF seemed weird, but it was suggested to me as better than the alternative. I later found this blog post https://soatok.blog/2021/11/17/understanding-hkdf/ which highlights the reasons why. Supposedly varying the salt parameters but not the info parameter _reduces_ security. Paseto V4 makes the same choice
    + > Instead of being used as a salt, however, it will be appended to the info tag


atuin-client/src/encryption.rs
```diff
@@ -8,130 +8,273 @@
...
-pub fn decrypt(mut encrypted_history: EncryptedHistory, key: &Key) -> Result<History> {
-    XSalsa20Poly1305::new(key)
-        .decrypt_in_place(
-            &encrypted_history.nonce,
-            &[],
-            &mut encrypted_history.ciphertext,
-        )
-        .map_err(|_| eyre!("could not encrypt"))?;
-    let plaintext = encrypted_history.ciphertext;
+        let mut plaintext = rmp_serde::to_vec(&HistoryPlaintext {
```
conradludgate, line 241: what to use instead of messagepack, hmmm

atuin-common/src/api.rs
```diff
@@ -56,6 +106,7 @@ pub struct SyncHistoryRequest {
 #[derive(Debug, Serialize, Deserialize)]
 pub struct SyncHistoryResponse {
     pub history: Vec<String>,
+    pub more_history: Vec<AddHistoryRequest>,
 }
```
conradludgate, line 111: TODO: use another struct

atuin-client/src/history.rs
```diff
@@ -19,21 +19,6 @@ pub struct History {
...
-    pub id: String,
-    pub timestamp: chrono::DateTime<Utc>,
-    pub duration: i64,
-    pub exit: i64,
-    pub command: String,
-    pub cwd: String,
-    pub session: String,
-    pub hostname: String,
-}
-
```
conradludgate, line 36: Moved as this encryption implementation detail

atuin-client/src/encryption.rs
conradludgate, line 1: Things to note for this file. I have made no changes to the existing key methods/xsalsa20poly1305 mechanisms, but I moved them to submodules for easier management.

The new xchacha20poly1305 module is the juicy one

atuin-client/src/api_client.rs
```diff
@@ -10,25 +9,18 @@ use reqwest::{
...
-    history::History,
-    sync::hash_str,
-};
+use crate::settings::Settings;
+use crate::{history::History, sync::hash_str};
 
 static APP_USER_AGENT: &str = concat!("atuin/", env!("CARGO_PKG_VERSION"),);
 
-// TODO: remove all references to the encryption key from this
-// It should be handled *elsewhere*
```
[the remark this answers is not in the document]
ellie: only took 2yrs, ty <3

conradludgate:
Superseded by #1058

PR #1005, opened 2023-05-24
lots of clippy chores

VuiMuich:
Was in the mood of doing some clippy chores.
But as I didn't want to deal with all the crono related stuff I based this on the `ellie:time` branch.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,7 +5,7 @@ members = ["atuin", "atuin-client", "atuin-server", "atuin-common"]
 name = "atuin"
 version = "14.0.1"
 authors = ["Ellie Huxtable <ellie@elliehuxtable.com>"]
-rust-version = "1.59"
+rust-version = "1.67"
 license = "MIT"
 homepage = "https://atuin.sh"
 repository = "https://github.com/ellie/atuin"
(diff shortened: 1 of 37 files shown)


atuin-client/src/settings.rs
```diff
@@ -132,6 +133,7 @@ pub enum WordJumpMode {
     Subl,
 }
 
+#[allow(clippy::struct_excessive_bools)]
```
VuiMuich, line 136: This is one of the clippies that are less fun to implement correctly...

atuin-client/src/database.rs
```diff
@@ -153,14 +153,15 @@ impl Sqlite {
         Ok(())
     }
 
+    #[allow(clippy::needless_pass_by_value)]
     fn query_history(row: SqliteRow) -> History {
```
VuiMuich, line 157: Clippy suggest to borrow as `&SqliteRow` but this leads to a whole lot of `.map(self.query_history)` fails, where I didn't see the fix.

atuin-client/src/import/resh.rs
```diff
@@ -11,6 +11,7 @@ use time::OffsetDateTime;
 use super::{get_histpath, unix_byte_lines, Importer, Loader};
 use crate::history::History;
 
+#[allow(clippy::module_name_repetitions)]
```
VuiMuich, line 14: If there is a good idea for a new name, this would be an easy fix to rename the symbol...

atuin-client/src/history.rs
```diff
@@ -24,6 +24,7 @@ pub struct History {
 // Forgive me, for I have sinned
 // I need to replace rmp with something that is more backwards-compatible.
 // Protobuf, or maybe just json
+#[allow(clippy::module_name_repetitions)]
```
VuiMuich, line 27: If there is a good idea for a new name, this would be an easy fix to rename the symbol...

atuin-client/src/encryption.rs
```diff
@@ -81,22 +81,22 @@ pub fn encode_key(key: &Key) -> Result<String> {
     Ok(buf)
 }
 
-pub fn decode_key(key: String) -> Result<Key> {
+#[allow(clippy::similar_names)]
```
VuiMuich, line 84: clippy suggests to rename `mbuf` to `buf` which might actually be a false suggested solution as this obviously has to fail. Any suggestions for `mbuf` instead?

Cargo.toml
```diff
@@ -2,7 +2,6 @@
 members = ["atuin", "atuin-client", "atuin-server", "atuin-common"]
 
 [workspace.package]
-name = "atuin"
```
VuiMuich, line 5: probably an artifact from before the libs were split off, right?

PR #1007, opened 2023-05-25
Add a configuration flag to control SQLite synchronous mode

mattico:
To help work around https://github.com/ellie/atuin/issues/952.

I don't love this solution but I can't use atuin without it.


mattico:
Oh. I started getting the lag with fsync=false.

I guess this isn't a solution!

PR #1016, opened 2023-05-30
Update Arch Linux links in README

eclairevoyant:
The [community] repository no longer exists and has been subsumed into [extra], hence updating links accordingly.


conradludgate (approved):
Thanks, interesting that this changed

eclairevoyant:
This had to do with [the recent migration from svn to git](https://archlinux.org/news/git-migration-announcement/) in the main repos, I'm guessing it was easier to make this change at the same time and simplify packaging overall.

PR #1021, opened 2023-06-01
Use alpine as base for Docker image

midzelis:
Alpine based images are 35mb while debian:bullseye-20230502-slim is 135mb. 

Docker-compose could use postgres:14-alpine for another reduction in size. 

Also added UID/GID to the docker-compose. This is a best practice. Otherwise the container will run with whatever user is running docker, which could be root. The Dockerfile was also modified to run with a user/group of 1000/1000 which also seems to be a common practice. (Ideally, the perms would be fixed as part of the entrypoint/cwd at runtime as well, to match the env vars, but I didn't do that)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,6 @@
-FROM lukemathwalker/cargo-chef:latest-rust-1.72.0-buster AS chef
+FROM rust:alpine as chef
+RUN apk add --no-cache musl-dev file make
+RUN cargo install cargo-chef 
 WORKDIR app
 
 FROM chef AS planner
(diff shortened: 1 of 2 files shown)


Sierra1011 (commented):
UID and GUID changes would be great, especially for running in Kubernetes with `securityContext` values.

Dockerfile
```diff
@@ -1,27 +1,25 @@
...
 COPY . .
 RUN cargo build --release --bin atuin
 
-FROM debian:bullseye-20230919-slim AS runtime
-
-RUN useradd -c 'atuin user' atuin && mkdir /config && chown atuin:atuin /config
-# Install ca-certificates for webhooks to work
-RUN apt update && apt install ca-certificates -y && rm -rf /var/lib/apt/lists/*
-WORKDIR app
+FROM alpine:3.12 as runner
```
Sierra1011, line 16: Using 3.18 would be a good up-to-date version with plenty of space left in the lifecycle.

PR #1035, opened 2023-06-08
Fix movement keys in inverted mode

majutsushi:
I noticed that when setting the new `invert` option several of the movement keys were upside-down. I've updated the keys to do the right thing depending on what the option is set to. I've also extracted the scrolling code into separate functions since the same few lines already had several copies.


conradludgate (approved):
Thanks!

PR #1036, opened 2023-06-08
refactor server to allow pluggable db and tracing

conradludgate:
Opens the way for third party servers to maintain and choose a database implementation themselves


ellie:
Looks good at a first pass, will test and 👀 again though to make sure

atuin-server/src/handlers/history.rs
```diff
@@ -140,8 +144,24 @@ pub async fn add<DB: Database>(
         })
         .collect();
 
-    let db = &state.0.database;
-    if let Err(e) = db.add_history(&history).await {
+    history.retain(|h| {
```
ellie, line 147: This is changing the behaviour right? So if someone inserts something too large, their client will continue trying to sync

conradludgate: Shouldn't be any change in functionality, I just move it from the database layer to the application layer https://github.com/ellie/atuin/blob/main/atuin-server/src/database.rs#L305-L318

ellie: Ahhh of course! Makes sense, thanks

ellie (approved):
:shipit:

PR #1038, opened 2023-06-08
Key values

ellie:
Super basic PoC, that adds a few things

1. Introduce the concept of a records, and a record store

This stores a linked list of... data. Actually just anything! This data could be shell history, but for now is simple key-values. The idea here is to store it locally in a format that works much better for effective syncing. Note that we store it per host. This is intentional, as when sync works the idea is that machines will write their own local data, but read from all

2. Introduce key value store

This is both a cool feature that we wanted, but also a demo of the sync. Rather than mess with history sync from day 0, we can do it with something lower stakes. 

This is not final. I still need to implement sync, and I still need to "build" the history records into an actual kv store. The idea of a record store is not that live queries will run on it, but that it will act as a source of truth

### DIFF
--- a/atuin-common/src/record.rs
+++ b/atuin-common/src/record.rs
@@ -1,0 +1,15 @@
+/// A single record stored inside of our local database
+pub struct Record {
+    pub id: i64,
+
+    pub host: String,
+
+    pub timestamp: u64,
+
+    /// The type of data we are storing here. It is probably useful to also
+    /// include some sort of version. For example, history.v2
+    pub tag: String,
+
+    /// Some data. This can be anything you wish to store. Use the tag field to know how to handle it.
+    pub data: Vec<u8>,
+}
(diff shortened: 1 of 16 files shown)


atuin-common/src/record.rs
```diff
@@ -0,0 +1,61 @@
...
+        data: Vec<u8>,
+    ) -> Record {
+        let id = crate::utils::uuid_v7().as_simple().to_string();
+        let timestamp = chrono::Utc::now();
+
+        Record {
+            id,
+            host,
+            parent,
+            timestamp: timestamp.timestamp_nanos() as u64,
```
conradludgate, line 45: Since we don't actually use the timestamp anymore for the design, should we use milli/microseconds instead? Still more than accurate enough for humans

ellie: Nah - why not use the most accurate we can? We still get 100s of years worth in a 64 bit int

Also, it _will_ still be used. When I finish fleshing this out with sync, we will need to be able to resolve conflicts if two hosts set the same key. First they will prefer the newer value, which means having the most accurate timestamp possible is valuable. Won't just be the timestamp, but yeah

PR #1040, opened 2023-06-10
Make Ctrl-d behaviour match other tools

majutsushi:
With this change Ctrl-d behaves differently depending on whether there is any input text available. If there is, it will delete the character to the right of the cursor if there is any. If there isn't it will instead quit interactive mode and leave the original shell command line unchanged.
This matches other line-based tools like bash and fzf.


ellie (approved):
Thank yoy!

PR #1053, opened 2023-06-15
Add workspace mode, enable if in git repo

ellie:
Introduces the concept of a workspace - shell history grouped by git repo

This will automatically filter by workspace if workspaces are enabled, which so far I have found more useful than just directory filtering

In the future
1. fancy sorting by workspace dir
2. kvs scoped by workspace?


conradludgate (approved):
Resisting my urge to nit pick

PR #1054, opened 2023-06-15
Enable multiple command stats to be shown

Shirtpantswallet:
These changes extend the `stats` subcommand with an `--ngram-size` argument, which supports providing statistics for combinations of commands. I have found this useful for identifying the most common combinations of commands as candidates for further automation.

### DIFF
--- a/atuin/src/command/client/stats.rs
+++ b/atuin/src/command/client/stats.rs
@@ -21,16 +21,69 @@ pub struct Cmd {
     /// How many top commands to list
     #[arg(long, short, default_value = "10")]
     count: usize,
+
+    /// The number of consecutive commands to consider
+    #[arg(long, short, default_value = "1")]
+    ngram_size: usize,
 }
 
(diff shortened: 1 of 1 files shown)


ellie:
Hey! Thank you for working on this 🙏 

Generally looks good, though I found an issue with unicode.

`chars` doesn't handle unicode characters so well, as per the [docs](https://doc.rust-lang.org/std/primitive.str.html#method.chars):

> It’s important to remember that [char](https://doc.rust-lang.org/std/primitive.char.html) represents a Unicode Scalar Value, and might not match your idea of what a ‘character’ is. Iteration over grapheme clusters may be what you actually want. This functionality is not provided by Rust’s standard library, check crates.io instead.

As mentioned there, we will probably need to use something like [unicode_segmentation](https://unicode-rs.github.io/unicode-segmentation/unicode_segmentation/index.html) instead

PR #1057, opened 2023-06-18
remove rmp-serde

conradludgate:
Messagepack is a nice serialisation format, but rmp-serde is not. serde is designed for self describing documents. Messagepack is not self describing, which makes it prone to breakages.

Common breakages include:
* reordering fields
* adding fields
* removing fields

By writing the encoding out explicitly, we can see clearer when the encode/decode change will cause a breakage. The record version is used to allow different encodings.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -155,7 +155,6 @@ dependencies = [
  "regex",
  "reqwest",
  "rmp",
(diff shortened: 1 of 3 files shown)


ellie (approved):
Nice one 👏

PR #1058, opened 2023-06-18
record encryption

conradludgate:
Using a combination of V4 PASETO/PASERK with randomized wrapped keys to encrypt the data of each record.

I've used a set of `DecryptedData`/`EncryptedData` types to ensure that we don't mistakenly write decrypted data to the database

TODO:
- [x] record tag+version -> PASETO version
- [x] consider migrating the database to use a string for the data row instead of bytes

### DIFF
--- a/atuin-client/src/record/encryption.rs
+++ b/atuin-client/src/record/encryption.rs
@@ -1,42 +1,35 @@
-use atuin_common::record::{DecryptedData, Record};
+use atuin_common::record::{DecryptedData, EncryptedData, Encryption};
 use base64::{engine::general_purpose, Engine};
 use eyre::{ensure, Context, ContextCompat, Result};
-use rusty_paseto::prelude::{
+use rusty_paserk::{
+    id::EncodeId,
+    wrap::{LocalWrapperExt, Pie},
+};
+use rusty_paseto::core::{
     Footer, Key, Local, Paseto, PasetoNonce, PasetoSymmetricKey, Payload, V4,
 };
 use serde::{Deserialize, Serialize};
 
-#[derive(Debug, Clone, PartialEq)]
-pub struct EncryptedData(pub Vec<u8>);
+/// Encryption using PASETO V4 with a random key
+#[allow(non_camel_case_types)]
+pub struct PASETO_V4_PIE;
 
-pub trait Encrypt {
-    type Output;
-    type Key;
+impl Encryption for PASETO_V4_PIE {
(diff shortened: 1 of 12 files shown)


atuin-client/src/record/encryption.rs
```diff
@@ -0,0 +1,363 @@
...
+This would be a lot more secure than using the master key directly.
+
+However, there's an established norm of using a random key. This scheme might be otherwise known as
+- client-side encryption
+- envelope encryption
+- key wrapping
+
+A HSM (Hardward security module) provider, eg AWS, Azure, GCP, or even a physical device like a yubikey
+will have some keys that they keep to themselves. These keys never leave their physical hardware.
+If they never leave the hardward, then encrypting large amounts of data means giving them the data and waiting.
```
mercxry, line 29: Small typo ;)
```suggestion
If they never leave the hardware, then encrypting large amounts of data means giving them the data and waiting.
```

atuin-client/src/record/encryption.rs
```diff
@@ -0,0 +1,363 @@
...
+Why do we use a random content-encryption key?
+Originally I was planning on using a derived key for encryption based on additional data.
+This would be a lot more secure than using the master key directly.
+
+However, there's an established norm of using a random key. This scheme might be otherwise known as
+- client-side encryption
+- envelope encryption
+- key wrapping
+
+A HSM (Hardward security module) provider, eg AWS, Azure, GCP, or even a physical device like a yubikey
```
mercxry, line 27: Wouldn't wanna leave for the Hacktoberfest 🤣
```suggestion
A HSM (Hardware Security Module) provider, eg: AWS, Azure, GCP, or even a physical device like a YubiKey
```

atuin-common/src/record.rs
```diff
@@ -35,17 +45,26 @@ pub struct Record {
     pub tag: String,
 
     /// Some data. This can be anything you wish to store. Use the tag field to know how to handle it.
-    pub data: Vec<u8>,
+    pub data: Data,
+}
+
+/// Extra data from the record that should be encoded in the data
+#[derive(Debug, Copy, Clone)]
+pub struct AdditonalData<'a> {
```
ellie, line 53:
```suggestion
pub struct AdditionalData<'a> {
```

sorry 😭

ellie (changes requested):
This is great, thank you! Just the spelling error and we're good

ellie (approved):

PR #1063, opened 2023-06-20
remove decryption from api-client

conradludgate:
I've tried making this change 2 times, but it was always part of a bigger PR that never got merged. Doing it separately now 👌

### DIFF
--- a/atuin-client/src/api_client.rs
+++ b/atuin-client/src/api_client.rs
@@ -1,5 +1,4 @@
 use std::collections::HashMap;
-use std::collections::HashSet;
 use std::env;
 
 use chrono::Utc;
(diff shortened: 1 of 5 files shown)


atuin-client/src/api_client.rs
```diff
@@ -14,22 +13,13 @@ use atuin_common::api::{
...
-use crate::{
-    encryption::{decode_key, decrypt},
-    history::History,
-    sync::hash_str,
-};
+use crate::{history::History, sync::hash_str};
 
 static APP_USER_AGENT: &str = concat!("atuin/", env!("CARGO_PKG_VERSION"),);
 
-// TODO: remove all references to the encryption key from this
```
ellie, line 27: Finally ☺️

ellie (approved):

PR #1080, opened 2023-06-30
Update Nushell support to use supported $env update syntax

jntrnr:
Greetings!

The Nushell team is working on removing `let-env FOO = ...` and moving code to use `$env.FOO = ...` as it's more honest about what's happening (rather than being scoped like a `let`). This PR should move atuin to use the supported syntax (assuming I didn't miss anything 😅).

### DIFF
--- a/atuin/src/command/init.rs
+++ b/atuin/src/command/init.rs
@@ -98,7 +98,7 @@ bind -M insert \e\[A _atuin_bind_up";
         println!("{full}");
 
         if std::env::var("ATUIN_NOBIND").is_err() {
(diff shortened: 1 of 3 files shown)


ellie (approved):
Hey! Thank you so much for letting us know proactively + also pushing out the fix 💖 

This totally makes sense - and should be everything that needs changing!

happysalada:
@ellie would it be possible to make a release with this fix ?
For package maintainers, we either have to apply this patch manually or wait for a new release. The latter is easier for us, so I thought I would ask. No worries if you can't.

PR #1085, opened 2023-07-04
Update install script

DynamicApproach:
Changed the way it targets a distro, and also added check for current shell. 

Work in progress, don't know if it's needed but thought I'd give it a shot


install.sh
```diff
@@ -161,28 +141,129 @@ __atuin_install_unsupported(){
...
+          "debian"* ) OS="debian" ;;
+          "fedora"* ) OS="fedora" ;;
+          # Add more distributions as needed
+          * ) OS="unknownwsl" ;;
+        esac
+      else
+        OS="unknownwsl"
+      fi
+    else
+      echo "Unknown"
```
ellie, line 185: We get to this case running on macOS

DynamicApproach: Small mistake with the get OS function, apologies, now fixed! 

```
  if [[ "$OSTYPE" == "darwin"* ]]; then
    echo "darwin"
  elif [ -f "/etc/os-release" ]; then
```

install.sh
```diff
@@ -161,28 +141,131 @@ __atuin_install_unsupported(){
...
+    OS="darwin"
+  elif [ -f "/etc/os-release" ]; then
+    OS="$(grep -Po '(?<=^ID=).*$' /etc/os-release | tr '[:upper:]' '[:lower:]')" 2>/dev/null
+  elif [ -f "/etc/debian_version" ]; then
+    OS="debian"
+  elif [ -f "/etc/centos-release" ]; then
+    OS="centos"
+  elif [ -f "/etc/manjaro-release" ]; then
+    OS="manjarolinux"
+  elif [ -f "/etc/fedora-release" ]; then
```
gergelypolonkai, line 165: I don’t know how relevant it could be and how wide you want to support Linux distros, but there is also `/etc/redhat-release`; however, be aware that it’s also present on Fedora systems. The same should be true with `ubuntu-release` and `debian-release` but i still have to check that. There might also be `mint-release` and who knows what… :roll_eyes:

ellie:
Thanks for the mac fix! I just tested this on my Ubuntu server, and got

> Unknown or unsupported OS

`/etc/os-release` definitely exists, and lists `NAME="Ubuntu"`. `/etc/debian_release` also exists

install.sh
DynamicApproach: I think the best way to resolve this would be to gather the actual dependencies needed and check for those but I don't know if that's worth it unless we start involving binaries. 

I can certainly edit it to incorporate other distros as needed though.

 Thoughts?

DynamicApproach:
> Thanks for the mac fix! I just tested this on my Ubuntu server, and got
> 
> > Unknown or unsupported OS
> 
> `/etc/os-release` definitely exists, and lists `ID=ubuntu`. `/etc/debian_release` also exists

I fixed that error (just a silly one) and made the whole thing more readable, leaving variables at the top and the actual non-functions at the bottom.

ellie:
Hey! I'm really sorry, but I'm not going to be merging this one. 

There was a few errors found early on, and it's a pretty large change. As it's not easy to test across platforms, and the risk of getting it wrong (users cannot install atuin easily) is high, I'd rather avoid this change for now.

PR #1088, opened 2023-07-06
Add support for `max_preview_height` setting

RichardDRJ:
This allows for more configuration around `show_preview`, specifically in the case where someone has shell commands spanning more than 4 lines in their history and need to differentiate between them with something other than the first few lines.

This change makes the following supporting changes:

- Update the preview renderer to split _lines_ when they go past the preview width, rather than splitting the whole string - this means that the preview is cleanly rendered even in cases where there are newlines
- Update the preview height size calculator to take into account existing newlines

A couple of other notes here:
- I've never written rust before, so very open to being told I've done something iffy here
- I think this strips out empty lines from the preview - I'm not sure the best way to fix it given my lack of Rust experience, but given that any more than 4 lines was cut off before, I'm not sure there'll be any visible regression

# Screenshots

```toml
style = "compact"
show_preview = true
max_preview_height = 6
```

## Short command with longer commands in history

image

## Long command, longer than `4` lines but shorter than `max_preview_height`

image

## Longest command in history, longer than `max_preview_height`

image

-----

```toml
style = "full"
show_preview = true
max_preview_height = 6
```

## Short command with longer commands in history

image

## Long command, longer than `4` lines but shorter than `max_preview_height`

image

## Longest command in history, longer than `max_preview_height`

image

### DIFF
--- a/atuin-client/src/settings.rs
+++ b/atuin-client/src/settings.rs
@@ -151,6 +151,7 @@ pub struct Settings {
     pub inline_height: u16,
     pub invert: bool,
     pub show_preview: bool,
+    pub max_preview_height: u16,
     pub show_help: bool,
     pub exit_mode: ExitMode,
     pub word_jump_mode: WordJumpMode,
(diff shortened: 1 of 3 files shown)


ellie (dismissed):
This makes sense and is a nice improvement, thank you! 

RichardDRJ:
Oops, thanks @ellie! Have addressed and pushed - hopefully it's all good :smile:

ellie (approved):

PR #1092, opened 2023-07-08
Update config.toml path default comments

### LINKED ISSUE #1091: Default OS X `db_path` does not match the comment in default `config.toml`
According to the default `config.toml`, on mac the default database will be stored in your "system data directory":
`~/Library/Application Support/com.elliehuxtable.atuin/history.db`
In this default config `db_path` is not set (only a comment), so one would expect to find the db located in this default location.

However, on my mac this is not true; with `db_path` unset, as per default config, I can see that the db is here: `~/.local/share/atuin/history.db`.
According to the default `config.toml` this is the default Linux location, but for me it's also where I find it on mac.

simon-b:
- Updates outdated mac db_path default
- Adds windows db_path default
- Adds windows, mac, linux defaults for key_path and session_path
- Changes example session_path to be different to example key_path

Closes #1091

### DIFF
--- a/atuin-client/config.toml
+++ b/atuin-client/config.toml
@@ -1,13 +1,17 @@
 ## where to store your database, default is your system data directory
-## mac: ~/Library/Application Support/com.elliehuxtable.atuin/history.db
-## linux: ~/.local/share/atuin/history.db
+## linux/mac: ~/.local/share/atuin/history.db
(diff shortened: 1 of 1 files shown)


ellie (approved):
Thanks for sorting this out! 🙌 

PR #1093, opened 2023-07-09
Add new sync

ellie:
So this has taken me a while, combined planning/implementation time

Improvements:

1. We always sync a linked list. It is always append-only, so we never insert into the past. This makes syncing things easy
2. We try and make the bare-min requests in order to sync. Get an index, compare with the local index, then do operations to get the state to match
3. We can sync... anything. Yeah so this actually just syncs the kv store, but would be easy to map history or other use cases in here

So yeah! Faster, more flexible, and far less wasteful.

This has not yet been fully tested, and may contain issues. Until released, migrations may be removed or deleted. Run at your own risk.


atuin-client/src/record/sqlite_store.rs
```diff
@@ -62,11 +63,11 @@ impl SqliteStore {
             "insert or ignore into records(id, host, tag, timestamp, parent, version, data, cek)
                 values(?1, ?2, ?3, ?4, ?5, ?6, ?7, ?8)",
         )
-        .bind(r.id.as_str())
-        .bind(r.host.as_str())
+        .bind(r.id.as_simple().to_string())
+        .bind(r.host.as_simple().to_string())
```
conradludgate, line 67: does sqlite not support 128 bit integers? :(

atuin-client/src/record/sqlite_store.rs
```diff
@@ -159,18 +168,47 @@ impl Store for SqliteStore {
...
+        )
+        .bind(tag)
+        .map(Self::query_row)
+        .fetch_all(&self.pool)
+        .await?;
+
+        Ok(res)
+    }
+
+    async fn tail_records(&self) -> Result<Vec<(Uuid, String, Uuid)>> {
```
conradludgate, line 196: return type 😭

atuin-common/src/record.rs
```diff
@@ -3,35 +3,38 @@ use std::collections::HashMap;
...
 pub struct DecryptedData(pub Vec<u8>);
 
-#[derive(Debug, Clone, PartialEq)]
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
 pub struct EncryptedData {
     pub data: String,
     pub content_encryption_key: String,
 }
 
+pub type Diff = Vec<(Uuid, String, Uuid)>;
```
conradludgate, line 17: oh no, it's also here 😭

atuin-server-database/src/lib.rs
```diff
@@ -55,6 +59,19 @@ pub trait Database: Sized + Clone + Send + Sync + 'static {
...
+        &self,
+        user: &User,
+        host: Uuid,
+        tag: String,
+        start: Option<Uuid>,
+        count: u64,
+    ) -> DbResult<Vec<Record<EncryptedData>>>;
+
+    // Return the tail record ID for each store, so (HostID, Tag, TailRecordID)
+    async fn tail_records(&self, user: &User) -> DbResult<Vec<(Uuid, String, Uuid)>>;
```
conradludgate, line 73: And another! 😭

atuin-server-postgres/migrations/20230623070418_records.sql
```diff
@@ -0,0 +1,15 @@
+-- Add migration script here
+create table records (
+	id uuid primary key,            -- remember to use uuidv7 for happy indices <3
+  client_id uuid not null,        -- I am too uncomfortable with the idea of a client-generated primary key
+	host uuid not null,             -- a unique identifier for the host
```
conradludgate, line 5:
```suggestion
create table records (
	id uuid primary key,            -- remember to use uuidv7 for happy indices <3
    client_id uuid not null,        -- I am too uncomfortable with the idea of a client-generated primary key
	host uuid not null,             -- a unique identifier for the host
```

atuin-server-postgres/src/lib.rs
```diff
@@ -1,14 +1,18 @@
+
+
 use async_trait::async_trait;
```
conradludgate, line 3:
```suggestion
use async_trait::async_trait;
```

you don't need to artificially inflate your line count

atuin-client/src/record/sqlite_store.rs
ellie: it does not :(

atuin/src/command/client/sync.rs
```diff
@@ -62,12 +71,26 @@ impl Cmd {
...
+    force: bool,
+    db: &mut impl Database,
+    store: &mut (impl Store + Send + Sync),
+) -> Result<()> {
+    let (diff, remote_index) = sync::diff(settings, store).await?;
+    let operations = sync::operations(diff, store).await?;
+    let (uploaded, downloaded) =
+        sync::sync_remote(operations, &remote_index, store, settings).await?;
+
+    println!("{uploaded}/{downloaded} up/down to record store");
```
conradludgate, line 85: records won't sync automatically here

ellie: intentional atm

ellie: I don't want to piggy back it on top of the old record sync triggers, would like to do something better

conradludgate (approved):

PR #1102, opened 2023-07-14
Add total history count to the index API

ellie:
Thought it would be fun to collect some cool stats, maybe put them on atuin.sh.

PR #1103, opened 2023-07-14
Delete the count cache row when a user is deleted

ellie:
Turns out we were not deleting from the count cache table!


ellie:
Discovered while working on #1102

conradludgate (approved):

PR #1106, opened 2023-07-16
use Ctrl-n instead of Alt-n on macOS

zygous:
To use `alt`-`n` modifiers on macOS the user must configure their terminal emulator to map `option` to `alt`. This isn't obvious -- hence issues have been raised here about those shortcuts not working -- but, moreover, doing so prevents `option`-`3` being used to type `#` on British Macs.

A KISS solution is to support using `ctrl`-`n` instead of `alt`-`n` on Macs.

This PR _only_ adds `ctrl`-`n` support, where `n` is a number: it does not affect the `alt`-`b` or `alt`-`f` shortcuts.

### DIFF
--- a/atuin/src/command/client/search/interactive.rs
+++ b/atuin/src/command/client/search/interactive.rs
@@ -1,4 +1,5 @@
 use std::{
+    env,
     io::{stdout, Write},
     time::Duration,
 };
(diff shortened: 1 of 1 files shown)


ellie (approved):
Thank you! LGTM.

PR #1123, opened 2023-07-28
Disable server tests in the nix build

patricksjackson:
The recent #1096 broke the Nix CI, this change fixes the nix build by disabling those tests. New tests that also use the postgres server will need to be excluded as well using additional '--skip' options or by keeping them all in the same excluded module. Adding a postgres server to make these pass is _technically_ possible, but probably not worth the maintenance burden for you.

For background, the `nix build` step runs the tests by default in an attempt to verify the correctness of the build. These tests are run in a sandbox with no network access, so it's fairly common that projects with complex test cases just won't work within it very easily.

Alternately we could just disable tests entirely in the nix build by using `doCheck = false;` instead of `checkFlags`. [See docs for guidance](https://github.com/NixOS/nixpkgs/blob/master/doc/languages-frameworks/rust.section.md#disabling-package-tests-disabling-package-tests). That may be preferable if you don't expect to get any value out of running the tests inside the nix sandbox.

(I noticed this because I do actually use the nix flake to run a bleeding edge version of the client on my own systems, so this flow is tested by at least 1 person :))

### DIFF
--- a/atuin.nix
+++ b/atuin.nix
@@ -35,6 +35,12 @@ rustPlatform.buildRustPackage {
       --zsh <($out/bin/atuin gen-completions -s zsh)
   '';
 
+  # Additional flags passed to the cargo test binary, see `cargo test -- --help`
+  checkFlags = [
+    # Registration tests require a postgres server
+    "--skip=registration"
+  ];
+
   meta = with lib; {
     description = "Replacement for a shell history which records additional commands context with optional encrypted synchronization between machines";
     homepage = "https://github.com/ellie/atuin";
--- a/flake.lock
+++ b/flake.lock
@@ -17,12 +17,15 @@
       }
     },
     "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
       "locked": {
-        "lastModified": 1676283394,
-        "narHash": "sha256-XX2f9c3iySLCw54rJ/CZs+ZK6IQy7GXNY4nSOyu2QG4=",
+        "lastModified": 1689068808,
+        "narHash": "sha256-6ixXo3wt24N/melDWjq70UuHQLxGV8jZvooRanIHXw0=",
         "owner": "numtide",
         "repo": "flake-utils",
-        "rev": "3db36a8b464d0c4532ba1c7dda728f4576d6d073",
+        "rev": "919d646de7be200f3bf08cb76ae1f09402b6f9b4",
         "type": "github"
       },
       "original": {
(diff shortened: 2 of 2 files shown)


atuin.nix
```diff
@@ -35,6 +35,12 @@ rustPlatform.buildRustPackage {
       --zsh <($out/bin/atuin gen-completions -s zsh)
   '';
 
+  # Additional flags passed to the cargo test binary, see `cargo test -- --help`
+  checkFlags = [
+    # Registration tests require a postgres server
+    "--skip=registration"
```
ellie, line 41:
```suggestion
    "--lib --bins"
```

Or similar may work better? If that's exactly what is passed to `cargo test`, then it will ensure we do not run any integration tests here

ellie (approved):
> These tests are run in a sandbox with no network access, so it's fairly common that projects with complex test cases just won't work within it very easily.

I see! That explains a lot, thank you 🙏 

> Alternately we could just disable tests entirely in the nix build by using doCheck = false; instead of checkFlags. [See docs for guidance](https://github.com/NixOS/nixpkgs/blob/master/doc/languages-frameworks/rust.section.md#disabling-package-tests-disabling-package-tests). That may be preferable if you don't expect to get any value out of running the tests inside the nix sandbox.

I'll leave the tests running as part of the build for now, but may disable them in future

atuin.nix
ellie: Merging as-is for now, I'll look at changing or updating this when we add more integration tests!

patricksjackson: The call to `cargo test` is [this invocation](https://github.com/NixOS/nixpkgs/blob/28482dfd4284560ccf20fb2d9535be326cac8d54/pkgs/build-support/rust/hooks/cargo-check-hook.sh#L36). The arguments passed to `checkFlags` are passed into the binary with `cargo test -- --lib --bins`, so that won't work, and it doesn't look like there's a way to pass those arguments without co-opting some other feature.

If that way of calling cargo makes more sense for the tests here, it's probably clearer to just override the `checkPhase` and write the testing script ourselves. It's just a bash script.

```
# Override the checkPhase script. This is run after building to verify correctness.
# Uses release mode to avoid building in debug mode just for testing
checkPhase = ''
  cargo test --release --lib --bins
'';
```

ellie: Ah I see, thanks for explaining! Will figure out which makes the most sense when we add more tests

PR #1131, opened 2023-07-28
fix(ubuntu/arm*): detect non amd64 ubuntu and handle

jinnko:
If we're running Ubuntu on non-amd64 we get an install error.  As there's no ARM package yet we can offer the user the option to install via cargo rather than install something that won't work.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -64,18 +64,21 @@ __atuin_install_arch(){
 }
 
 __atuin_install_ubuntu(){
-	echo "Ubuntu detected"
-	# TODO: select correct AARCH too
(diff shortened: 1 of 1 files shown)


ellie (approved):
Thank you!

ellie (dismissed):

jinnko:
@ellie - I've updated this PR to include the org name change that was made.

ellie (approved):
Thank you! Sorry it took me so long to circle back on the review 🙃 

I'd like to build debs for ARM at some point, seeing as we build binaries for ARM now

PR #1132, opened 2023-07-30
Move all references to the old repo

ellie:
The repo now lives on an org! Move all references to the old path.


hrfried:
Missed one on the website. :)
https://atuin.sh/docs/self-hosting/docker

PR #1143, opened 2023-08-02
Prepare release v16.0.0

ellie:


ellie:
Thank you so much @patricksjackson! I've done the same, on multiple machines. Looks good to go 😎

PR #1140, opened 2023-08-01
fix(docs): List all presently documented commands

LeoniePhiline:
This PR adds the remaining (presently existing) sub-pages with command descriptions to the menu / commands overview list at https://atuin.sh/docs/commands/ 

There's a whole bunch of commands not documented at yet, but this is a start! 

Fixes https://github.com/atuinsh/atuin/issues/1139

PR #1145, opened 2023-08-02
fix(docs): Correct command overview paths

LeoniePhiline:
This is a fix for my previous rushed mess-up. 😇

Context: https://github.com/atuinsh/atuin/pull/1140#issuecomment-1661288695


ellie (approved):
thank you!!! :shipit:

PR #1149, opened 2023-08-03
Fix invoking application mode after leaving it (Fixes #650)

### LINKED ISSUE #650: After using atuin interactively, home key stops working in the shell
When  I first open my terminal (terminator, gnome terminal, xterm), home key works as intended. However, after going interactive on Atuin and exit out, home key stop working (the cursor doesn't go to the beginning of the line)

JaCoB1123:


tpoliaw:
This is what I tested locally and it solved the problem for me but I wasn't sure if it whether it'd cause issues elsewhere. The comment `Switch to cursor mode, then back to application` suggests that the current way around was intentional and that it should be left in application mode.

Either way, if this PR is ok, the comment should be removed/updated so they're not out of sync.

JaCoB1123:
I just updated the comments to match the code.

@ellie I saw you added this in #31 / https://github.com/atuinsh/atuin/pull/31/commits/4b4052782772b8b786908b519759b4e28c1b3745. Do you by any chance still know why it was added?

ellie:
> I just updated the comments to match the code.
> 
> @ellie I saw you added this in #31 / [4b40527](https://github.com/atuinsh/atuin/commit/4b4052782772b8b786908b519759b4e28c1b3745). Do you by any chance still know why it was added?

I'll have to double check before merging this. I remember there were issues with certain keycodes not being recognized in application mode, but that's probably been resolved by now.

(also... can't believe it's been >2yrs 😭)

PR #1155, opened 2023-08-07
Fix client-only builds

ellie:


lespea:
Just curious if there's going to be a patch release with this fix soon?  Not a huge deal if not obviously.

PR #1156, opened 2023-08-07
Display hostname in interactive search

flifloo:
Hi o/

I'm using multiples computer and servers with Atuin and like to see where I used a command.
So this is a proposal to display the user and host in the interactive search :

image

This could be disabled by config/arg and when in host/session filter mode.
I'm not trying this because it covers more code/files that I'm comfortable with.

It's my first time working with rust, not sure how good this PR is.

### DIFF
--- a/atuin/src/command/client/search/history_list.rs
+++ b/atuin/src/command/client/search/history_list.rs
@@ -1,4 +1,5 @@
 use std::time::Duration;
+use std::env;
 
 use atuin_client::history::History;
 use ratatui::{
@@ -69,6 +70,7 @@ impl<'a> StatefulWidget for HistoryList<'a> {
             s.index();
             s.duration(item);
             s.time(item);
+            s.hostname(item);
             s.command(item);
 
             // reset line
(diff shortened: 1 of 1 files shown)


ellie:
Hey! Thank you for the work here, but we would like to avoid any additional attributes being shown in the search UI as-is. More information leads to more visual clutter, which makes it harder to quickly find what you are looking for.

If anything, I'd like to rebase and test #861 - which means we can show more information without so much clutter

ellie:
I think if anything I'd be open to a more generalised configuration of what we display in the results table, but I would like to avoid special-casing it for each individual thing.

papertigers:
> I think if anything I'd be open to a more generalised configuration of what we display in the results table, but I would like to avoid special-casing it for each individual thing.

I came here looking for this feature before I started working on a PR. I think there's value to seeing the hostname in the search UI, or at least having it be togglable via hot keys.

Rather than having a plethora of individual config options to enable disable features we could opt for an array of elements the user wants visible.  Take for example [this config from helix](https://github.com/helix-editor/helix/blob/master/book/src/configuration.md#editorstatusline-section).

That would allow us to have the default be something like `[execution_time, age, command]` allowing a user to specify something like `[age, hostname, command]`, or whatever data we may want to expose.

ellie:
> Rather than having a plethora of individual config options to enable disable features we could opt for an array of elements the user wants visible. Take for example [this config from helix](https://github.com/helix-editor/helix/blob/e4c95f65a64a0553e1dafa365a40dcd5c328fa69/book/src/configuration.md#editorstatusline-section).

I'd be open to this approach, if anyone fancies making a PR. I would still like to have a statusline though.

I'll be closing this PR for now though, due to lack of response

Issue #1158, opened 2023-08-09
SQLite as the default server database


ellie:
Ever since Atuin first released, we have had requests to support SQLite on the server.

For a very long time, I have resisted this. SQLite might work for the average self-hoster (my assumption is that most people self host just for themselves, or perhaps for them and a couple of friends), but it does not work for api.atuin.sh - at time of writing, we are approaching 2500 users and are growing pretty rapidly. I did not want to test each release for two database backends.

@conradludgate implemented an "unofficial" SQLite server over [here](https://github.com/conradludgate/atuin-server-sqlite), which somewhat solves the problem. Lately I've been thinking about ensuring api.atuin.sh continues to scale well, but this may involve making more changes that make self hosting awkward.

So the proposal here is to switch this around. The in-tree database implementation should be SQLite, and we can maintain a separate implementation for api.atuin.sh. This should ensure we are free to handle scaling issues, but also means that the average self-hoster can very easily get started. Integration tests are much easier to write with sqlite, as we can just run against `:memory:`.

What I would appreciate is input from anyone self hosting

1. How many users do you host a server for?
2. Would SQLite be a "better" or "easier" solution for you?
3. Would you be ok with some manual steps to migrate from postgres -> sqlite in a possible v17?
4. Any other thoughts?

I'm also interested in hearing from anyone hosting Atuin for their company, and what we can do to support you better.

sberrevoets:
I tried setting up self-hosting on a raspberry pi the other day, and the dependency on Postgres 14 was a showstopper since that wasn't available as a pre-compiled package so I had to compile it myself which was too much work.

I looked at the SQLite option but since it wasn't official I didn't want to rely on it.

I currently don't sync at all, but an RPi + SQLite setup for me personally would be great :)

CygnusHyoga:
What I would appreciate is input from anyone self hosting

How many users do you host a server for?

6 currently

Would SQLite be a "better" or "easier" solution for you?

No. I think SQLite it's good for low resources devices, and for test the software as an optionally alternative...I prefer PostgreSQL or Mysql/MariaDB...I don't trust SQLite for example in nfs pvc in k8s cluster, sometimes works great, sometimes  gave corruption error.

Would you be ok with some manual steps to migrate from postgres -> sqlite in a possible v17?

I think as an alternative solution, make them both available, for people to chose, what solution is best for their uses cases. 

Any other thoughts?

[...]

PR #1160, opened 2023-08-10
Adds support for sqlite-backed server; works, not tested.

nebkor:
Neither the docs nor the tests are updated, but I'm currently running the server with sqlite:

```
$ ~/git/atuin » sqlite3 ~/.atuin-server.db
SQLite version 3.39.2 2022-07-21 15:24:47
Enter ".help" for usage hints.
sqlite> select * from total_history_count_user;
1|1|17
sqlite> select * from users;
1|joe|joe@mail|<redacted>|1691707104
sqlite> .quit

$ ~/git/atuin 101 » bat ~/.config/atuin/config.toml
sync_address = "http://127.0.0.1:8888"
sync_frequency = "30s"
filter_mode = "host"
```

Marking this as a draft until I update the tests and note in the docs that you can run it without postgres or docker, and how to do so (build it with `--no-default-features --features=sqlite-server,client,sync`).


ellie:
Hey! Just closing this before you put any further effort in - we are currently discussing what needs to happen with SQLite in https://github.com/atuinsh/atuin/issues/1158, and already have an implementation of this. Just out of tree

Sorry for the hassle here, but in future please open an issue before working on a large feature

ellie:
The issue so far with SQLite has not been lack of implementation, but being averse to supporting another database backend. Without properly thinking it through, it doubles the amount of support and testing we need to do. While work like this is appreciated, generally contributors are not willing to test the feature they added for every future release of Atuin - so if we accept something like this, we have to be sure that it's needed enough to be worth the support load

nebkor:
> Hey! Just closing this before you put any further effort in - we are currently discussing what needs to happen with SQLite in #1158, and [already have an implementation of this](https://github.com/conradludgate/atuin-server-sqlite). Just out of tree
> 
> Sorry for the hassle here, but in future please open an issue before working on a large feature

Hey, no worries! I might have spared the effort myself if I knew there was an existing impl available, but it was only a few hours and I don't mind throwing it away :) Besides, I wanted it for myself, since I don't want to run docker or postgres.

Anyway, thank you for making something so cool!

PR #1165, opened 2023-08-13
feat(k8s): add kubernetes native manifests for external usage

davinkevin:
Like proposed months ago, I have made a PR with kubernetes native / kustomize manifests for installation purpose. 

In this PR, we have:

* `kustomize/base` which is the root of everything, with the common core for every overlays and components for "off the shelf" customisations.
* `kustomize/overlays/all-in-one`, the "alternative" to the current Kompose solution provided, with a dedicated PG instance in its own pod/service. The is no exposition for it at the moment in this one.
* `kustomize/overlays/core` for people who wants to use another PG instance (externally provided). Just the minimum. It's very useful when you integrate it in an advanced k8s env
* `kustomize/overlays/example/ingress` contains an example using `ingress` to deploy securely atuin using a custom domain name and certificate. Certificate has been generated with `mkcert`, so we can share them, it's just to share an example to end users, with `ingress` API, which is the most common k8s api for the moment (even if it's a pain to manage, the new GatewayAPI will provide a simpler solution)

If you have any question, I'm free to discuss this in this PR or in a sync call 😇


davinkevin:
Is there something I can do to ease the review?

ellie:
> Is there something I can do to ease the review?

Hey! Nothing in particular - I'm feeling a bit drained at the moment so I'm taking it slow for a few weeks to avoid burning out, will be back to reviewing bigger PRs soon!

davinkevin:
No problem, rest well and come back fresh! 

If you need some help to review this PR, I can present you some part of it in a call, as you want! 😇

davinkevin:
I've maintained the branch and introduced metrics `component` to leverage `/metrics` endpoint introduced in v17.1.0. 

Still available for explanations or else if required on this aspect 😇. 

FYI, I use it in my prod cluster without any problems!

ellie:
I'm really sorry for the back and forth and slowness here, but I don't think I'm going to be going ahead with this one. I appreciate the effort + usefulness in this PR, however if I am going to be merging + maintaining anything then it will be a helm chart

davinkevin:
😓 and good luck with `helm` chart and especially with `go-teamplate` 😇. 

I proposed this solution because it's the official approach, and the only supported and maintained by Kubernetes directly.

If you have second thoughts, feel free to ping me 😇

ellie:
> 😓 and good luck with helm chart and especially with go-teamplate 😇.

hah yeah I'm not a fan of how they use go-template, but I've maintained + used them professionally for several years now, and most software is distributed on k8s via helm at this point

> If you have second thoughts, feel free to ping me 😇

thanks! I will do

davinkevin:
> hah yeah I'm not a fan of how they use go-template, but I've maintained + used them professionally for several years now, and most software is distributed on k8s via helm at this point

Yeah, I do the same professionally and so many clients are doing that because `helm` is the only solution too 😅. Our customers are doing `helm template` + `kustomization` because they want to do a lot of things…

ellie:
> Yeah, I do the same professionally and so many clients are doing that because `helm` is the only solution too 😅. Our customers are doing `helm template` + `kustomization` because they want to do a lot of things…

Yeah, if it were internal only kustomize would be cool but for distribution + user install people will want a helm chart. If I'm going to maintain 1 thing only, it'll just have to be helm 😞 can always be kustomize-d on top

davinkevin:
For what it's worth and because I forgot to mentioned it, if you want I can maintain the `kustomize` part 😇

davinkevin:
👋 

With you working all time on it, is your position for this PR changed? It worth asking! 😇
Like I said, it would be a pleasure to help maintain that part if required.

ellie:
Hey! Thanks for asking, but I'm currently only interested in maintaining the helm chart here: https://github.com/atuinsh/helm-charts. I'm going to be removing the k8s config from here soon, and don't want to support 2 methods of installing on Kubernetes

I appreciate the offer of help though!

PR #1167, opened 2023-08-14
Run check for client-only feature set

tobiasge:
Checks if client-only builds on PRs


ellie (approved):
Looks good to me! Been meaning to add this since we broke client builds, thank you!

PR #1177, opened 2023-08-16
WIP: Use tmux display-popup when available

expipiplus1:
I don't have the availability to finish this, tests/config/proper-escaping, so I suggest you close this PR. I'm just opening it so that if someone happens to want something similar they can find this WIP if they search "atuin tmux popup" or something.


immae:
Just a note: I’m working on implementing that feature. Due to the evolution of the software the diff looks more like this for the moment:

```
diff --git a/crates/atuin/src/shell/atuin.bash b/crates/atuin/src/shell/atuin.bash
index cd5a8eff..9a95f6c9 100644
--- a/crates/atuin/src/shell/atuin.bash
+++ b/crates/atuin/src/shell/atuin.bash
@@ -254,7 +254,21 @@ __atuin_history() {
         local READLINE_LINE="" READLINE_POINT=0
 
     local __atuin_output
-    __atuin_output=$(ATUIN_SHELL_BASH=t ATUIN_LOG=error ATUIN_QUERY="$READLINE_LINE" atuin search "$@" -i 3>&1 1>&2 2>&3)
+    if [ -z "${TMUX:-}" ]; then
+      __atuin_output=$(ATUIN_SHELL_BASH=t ATUIN_LOG=error ATUIN_QUERY="$READLINE_LINE" atuin search "$@" -i 3>&1 1>&2 2>&3)
+    else
+      tmpdir=$(mktemp -d)
+      mkfifo "$tmpdir/pipe"
+      trap "rm -rf '$tmpdir'" EXIT HUP INT TERM
+      __atuin_output=$(
+        cat "$tmpdir/pipe" &
+        # tmux display-popup uses the env from the tmux server by
+        # default, we need to pass them.
+        # FIXME: find a better way to pass the env from the bash
+        tmux display-popup -d "$(pwd)" -E -h 80% -w 80% -- "$(printf "%q " ATUIN_SESSION="$ATUIN_SESSION" ATUIN_HISTORY_ID="$ATUIN_HISTORY_ID" ATUIN_PREEXEC_BACKEND="$ATUIN_PREEXEC_BACKEND" ATUIN_SHELL_BASH=t ATUIN_LOG=error ATUIN_QUERY="$READLINE_LINE" atuin search "$@" -i) 1>&2 2>$tmpdir/pipe"
+        )
+      rm -rf "$tmpdir"
+    fi
```

note the environment passing issue which may have been overlooked in the initial draft (I’ll provide update once things settle)

lmBored:
@immae are you still working on this? I like this feature and would be happy to take over the PR, just wanted to check in first so we don’t double up on the work

immae:
> @immae are you still working on this? I like this feature and would be happy to take over the PR, just wanted to check in first so we don’t double up on the work

I have a fix that is trivial and that works like a charm, but saddly it is not exactly mergeable "as is" for now (because the solution is quite specific to my setup), so that’s why I never followed up on this work, sorry.

I’m sharing it below in case you want to take inspiration from it

```
diff --git a/crates/atuin/src/shell/atuin.bash b/crates/atuin/src/shell/atuin.bash
index cd5a8eff..eeefbeff 100644
--- a/crates/atuin/src/shell/atuin.bash
+++ b/crates/atuin/src/shell/atuin.bash
@@ -254,7 +254,7 @@ __atuin_history() {
         local READLINE_LINE="" READLINE_POINT=0
 
     local __atuin_output
-    __atuin_output=$(ATUIN_SHELL=bash ATUIN_LOG=error ATUIN_QUERY="$READLINE_LINE" atuin search "$@" -i 3>&1 1>&2 2>&3)
+    __atuin_output=$(ATUIN_SHELL=bash ATUIN_LOG=error ATUIN_QUERY="$READLINE_LINE" cmd-tmux -p80%,60% -- bash -c "$(printf "%q " atuin search "$@" -i) 3>&1 1>&2 2>&3")
 
     # We do nothing when the search is canceled.
     [[ $__atuin_output ]] || return 0
```

and cmd-tmux is a script (attached) that is heavily inspired from https://github.com/junegunn/fzf/blob/master/bin/fzf-tmux but that I made independent from fzf
[cmd-tmux.sh](https://github.com/user-attachments/files/24593146/cmd-tmux.sh)
(very useful not only in the context of atuin)

Feel free to take over and good luck with your implementation :-)

PR #1182, opened 2023-08-19
Automatically filter out secrets

ellie:
I'd like to extend the regex list here very soon, but start off by automatically filtering out secrets. Do not store them in history!

I've included regex for:

1. AWS key id
2. Github pat (old and new)
3. Slack oauth tokens (bot, user)
4. Slack webhooks
5. Stripe live/test keys

#806 will need rebasing

### DIFF
--- a/atuin-client/src/history.rs
+++ b/atuin-client/src/history.rs
@@ -3,6 +3,9 @@ use std::env;
 use chrono::Utc;
 
 use atuin_common::utils::uuid_v7;
+use regex::RegexSet;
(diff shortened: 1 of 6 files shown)


conradludgate:
Maybe add a settings entry for custom exclusions?

ellie:
> Maybe add a settings entry for custom exclusions?

We already have that haha, it's history_filter 

This is a set of defaults you can opt out of, regardless of your history filter settings

PR #1200, opened 2023-08-29
Use `case` for Linux distro choice in `install.sh`

mentalisttraceur:
Just a small stylistic cleanup. I was looking inside `install.sh` and spotted this `if` which can be less noisy as a `case`.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -92,14 +92,15 @@ __atuin_install_linux(){
   else
     OS=$(lsb_release -i | awk '{ print $3 }' | tr '[:upper:]' '[:lower:]')
   fi
-	if [ "$OS" == "arch" ] || [ "$OS" == "manjarolinux" ] || [ "$OS" == "endeavouros" ]; then
-		__atuin_install_arch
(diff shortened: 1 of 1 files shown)


mentalisttraceur:
I imitated the formatting used in the other two `case` statements in this file as much as reasonable (the only thing I did differently was add a line break between the patterns and their commands, since that on-the-same-line condensed form stops making sense when the case patterns and/or the commands are too long), but I recommend having case patterns on the same intendation level as the `case` keyword/block. I don't think it adds any clarity or readability to have two indent levels for every `case` when a single one contains the same information unambiguously.
 
I also kept the quoting the same, but in shell I recommend the habit/practice of using `'` instead of `"` whenever you don't need the special interpretation/expansion that `"` has.

mentalisttraceur:
I don't know why 5/6 of the PR's automated workflows are in "Expected - Waiting for status ..." state. Anyone have any ideas? Should I just push a dummy empty commit to force it to re-run or something?

ellie (approved):
Thank you! That's much neater.

The CI wasn't running as I need to approve it

ellie:
I'll merge regardless of the failure, as it's unrelated to this diff. I suspect a rebase would resolve it.

robinshen36:
@algora-pbc attempt

This looks interesting - wanted to throw my hat in the ring.

Just wrapped up a project that had some overlap with this.

My approach would be to start by understanding the encryption, then iterate from there.

Budget-wise, $190-$330 is what I had in mind. What do you think?

Hit me up if you want to chat through details.

Issue #1204, opened 2023-08-31
Latest docker image fails with glibc errors


Artanicus:
On amd64:

```
%> docker run --rm -it ghcr.io/atuinsh/atuin:main
Unable to find image 'ghcr.io/atuinsh/atuin:main' locally
main: Pulling from atuinsh/atuin
14726c8f7834: Pull complete 
69bdb7f9d6f1: Pull complete 
d5c9d835c8d1: Pull complete 
05fe3086f264: Pull complete 
d7513519db61: Pull complete 
Digest: sha256:fc0fb82d642b3231fdd76624b15e021de222eb45eab2cbf97bd2c30b97bb1afe
Status: Downloaded newer image for ghcr.io/atuinsh/atuin:main
/usr/local/bin/atuin: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by /usr/local/bin/atuin)
/usr/local/bin/atuin: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by /usr/local/bin/atuin)
/usr/local/bin/atuin: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by /usr/local/bin/atuin)
```

Running from `ghcr.io/atuinsh/atuin:main`,  `sha256:fc0fb82d642b3231fdd76624b15e021de222eb45eab2cbf97bd2c30b97bb1afe`

Artanicus:
I've confirmed 14f07da741 is the commit responsible. My guess is the default changed from buster to bookworm and that causes this.

Works: lukemathwalker/cargo-chef:latest-rust-1.72.0-buster
Doesn't work: lukemathwalker/cargo-chef:latest-rust-1.72.0-bookworm

Will create a PR for using buster explicitly, but probably something needs to be changed upstream or in atuin to fix the real issue of builds on bookworm.

PR #1205, opened 2023-08-31
Explicitly use buster image for cargo-chef, mitigates #1204

Artanicus:
This mitigates #1204 but isn't really a permanent fix.


ellie (approved):
Thank you! We do definitely need to look into what's going on with libc, but I appreciate the patch 💖 

PR #1209, opened 2023-09-03
feat: mouse selection support

### LINKED ISSUE #160: Select text
I can't use my mouse to select (and then copy) text in a command - sometimes I don't want to run a command, or get the whole thing, I just want to copy a hash or id or what have you from it.

Right now I do that by finding the command I want, hitting 'enter' to put it in the terminal, and then I can copy/paste/edit/whatever there instead of just hitting enter again to run as-is.  But it'd be less cognitive friction if I could copy-paste from the tui like I expect when I see text on a screen.

YummyOreo:
Adds mouse selection support.

closes #160

### DIFF
--- a/atuin/src/command/client/search/interactive.rs
+++ b/atuin/src/command/client/search/interactive.rs
@@ -1,10 +1,13 @@
 use std::{
-    io::{stdout, Write},
+    io::{self, stdout, Write},
     time::Duration,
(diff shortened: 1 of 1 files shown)


ellie (approved):
Nice one, thank you! This has been an issue for a while 🙏

PR #1224, opened 2023-09-13
feat(keyup): Configure SearchMode for KeyUp invocation #1216

deicon:
Adding another config option to setup Search Mode on KeyUp invocations.
Fixing #1216

PR #1235, opened 2023-09-16
Add search_mode_shell_up_key_binding

bretthoerner:
Hopefully self explanatory. :) I guess I'm weird but I have years of using *up* for prefix search, and then prefer Control-R for fuzzy.


deicon:
Have you seen my PR #1224 ? 

Looks like this is tackling the same issue.

bretthoerner:
Ha, no I hadn't even looked at PRs. I'll leave a comment there, thanks.

ellie:
Thank you for the work here either way 🙏

PR #1241, opened 2023-09-19
Add commands to print the default configuration

tobiasge:
When updating a software I often want to compare my configuration with the configuration of the new version. To make this possible atuin can now print the default configuration.
This also updates the example files with the actual values used as default in the settings.rs files.


atuin/src/command/server.rs
```diff
@@ -19,6 +19,9 @@ pub enum Cmd {
         #[clap(long, short)]
         port: Option<u16>,
     },
+
+    /// Print server example configuration
+    Config,
```
ellie, line 24: maybe let's adjust this so the usage is `default-config`, that way it's clear that it's going to print the default config. Otherwise, the user may think they are able to get/set current config values via the cli

atuin/src/command/client.rs
```diff
@@ -44,6 +45,10 @@ pub enum Cmd {
 
     #[command(subcommand)]
     Kv(kv::Cmd),
+
+    /// Print example configuration
+    #[command()]
+    Config,
```
ellie, line 51: same here!

ellie (dismissed):

ellie:
Thank you! 🚀

ellie (approved):

conradludgate (approved):

PR #1249, opened 2023-09-23
feat: copy to clipboard

### LINKED ISSUE #130: Add shortcut to copy current selection to clipboard
Often I'm browsing my history to copy paste a command to someone else. Right now, I've to select it, hit enter, then copy paste it with my mouse. Instead, would be great to copy it directly with a keyboard shortcut

YummyOreo:
closes #130


ellie:
Looks great, thank you! Would you also be able to update the docs please? 🙏

ellie (approved):
yay thank you for working on this! 🚀

Issue #1239, opened 2023-09-18
Allow something like: `atuin history list --reverse`


Janik-Haag:
**Is your feature request related to a problem? Please describe.**
Currently `atuin history list` shows `old -> new` and I would like to reverse that order because I'm using that mostly with a custom fzf script for theming and currently reverse the list after atuin is down outputting which gives quite a bit of delay with a large history limit. 

**Describe the solution you'd like**
The same as with `atuin history list` but with a reversed order from `new -> old`

kiran-4444:
I'd like to work on this issue. Can someone help me by giving me some pointers to get started? Thanks!

deicon:
Starting from the List Command Line Argument of the atuin client, this seems to end up here https://github.com/atuinsh/atuin/blob/cea68660fd43ecdefa735f027199000cc6a60aad/atuin-client/src/database.rs#L255
beeing a matter of setting the orderBy statement appropriately. 
But be aware, that this list method is called not only to list but also inside the engine and stats.

deicon:
@ellie can be closed as fixed with #1252

ellie:
Thanks!

PR #1252, opened 2023-09-25
add --reverse to history list

kiran-4444:
Tried working on this issue: https://github.com/atuinsh/atuin/issues/1239
This is a WIP and a very minimal (and brute force) implementation I did to get a hang of the repo. Seeking feedback on this. Thanks!

### DIFF
--- a/atuin/src/command/client/history.rs
+++ b/atuin/src/command/client/history.rs
@@ -52,6 +52,10 @@ pub enum Cmd {
         cmd_only: bool,
 
         #[arg(long, short, default_value = "true")]
+        // accept no value
+        #[arg(num_args(0..=1), default_missing_value("true"))]
+        // accept a value
+        #[arg(action = clap::ArgAction::Set)]
         reverse: bool,
 
         /// Available variables: {command}, {directory}, {duration}, {user}, {host}, {exit} and {time}.
--- a/atuin/src/command/client/history.rs
+++ b/atuin/src/command/client/history.rs
@@ -287,7 +287,7 @@ impl Cmd {
             None
         };
 
-        let history = match (session, cwd) {
+        let mut history = match (session, cwd) {
             (None, None) => db.list(settings.filter_mode, &context, None, false).await?,
             (None, Some(cwd)) => {
                 let query = format!("select * from history where cwd = '{cwd}';");
@@ -305,13 +305,10 @@ impl Cmd {
             }
         };
 
-        print_list(&history, mode, format.as_deref());
-
         if reverse {
-            let mut history = history;
             history.reverse();
-            print_list(&history, mode, format.as_deref());
         }
+        print_list(&history, mode, format.as_deref());
 
         Ok(())
     }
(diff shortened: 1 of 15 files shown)


deicon:
First off, please change this to be a Draft PR.

atuin/src/command/client/history.rs
```diff
@@ -303,6 +307,12 @@ impl Cmd {
 
         print_list(&history, mode, format.as_deref());
```
deicon, line 308: Looks like this is printing the list twice if reverse == true.

atuin/src/command/client/history.rs
```diff
@@ -303,6 +307,12 @@ impl Cmd {
...
 
+        if reverse {
+            let mut history = history;
```
deicon, line 311: This can be achieved by changing the history to mutable and remove this temp variable here. 
Change the declaration of history above like this should be enough and more explicit
```
let mut history = match (session, cwd) {
```

atuin/src/command/client/history.rs
```diff
@@ -271,6 +274,7 @@ impl Cmd {
         cwd: bool,
         mode: ListMode,
         format: Option<String>,
+        reverse: bool,
```
deicon, line 277: Clippy (Rust linter) complains about too many arguments now. 
https://rust-lang.github.io/rust-clippy/master/index.html#/too_many_arguments

deicon:
Documentation is missing for the new parameter
https://github.com/atuinsh/atuin/blob/main/docs/docs/commands/list.md

deicon:
One last question here would be, if it would be better to get the result sorted from the DB instead of reverse the vec afterwards. I would argue, in this case here, reverse should be fine, but obviously not the most optimal way of doing things.

atuin/src/command/client/history.rs
kiran-4444: My bad, just realised it. Changing this..

deicon:
I asked for Draft, as it was titled WIP. But it looks like its more than WIP.

kiran-4444:
> One last question here would be, if it would be better to get the result sorted from the DB instead of reverse the vec afterwards. I would argue, in this case here, reverse should be fine, but obviously not the most optimal way of doing things.

I'm intimidated to change that `Database`'s `list` function, I'll now try to do this at DB level.

conradludgate:
> One last question here would be, if it would be better to get the result sorted from the DB instead of reverse the vec afterwards. I would argue, in this case here, reverse should be fine, but obviously not the most optimal way of doing things.

We can avoid any reversing in sql or in code.

In `print_list` we iterate over the history items already in reverse.

https://github.com/atuinsh/atuin/blob/fbed2862fda127b747718e4ae6f5f36a56f29a51/atuin/src/command/client/history.rs#L116

We can choose whether we want to iterate over `h.iter()` or `h.iter().rev()` depending.

Let me know if you want some more help navigating the rust side of implementing this

kiran-4444:
> Let me know if you want some more help navigating the rust side of implementing this

Yeah, I need some help with that.

deicon:
@kiran-4444 you might want to consoder to remove the WIP and make it an active PR

deicon:
If no longer wip just mark as active :)

conradludgate (approved):
Thanks :)

kiran-4444:
Are there any issues I can work on?

PR #1253, opened 2023-09-25
handle missing entries (fixes #1236)

### LINKED ISSUE #1236: Setting db_path errors "no rows"
**Describe the bug**
In `config.toml` whenever I set `db_path`, such as

```
db_path = "~/.local/share/atuin/history-nu.db
```

I get errors:

```
Error: no rows returned by a query that expected to return at least one row
```

**To Reproduce**
1. Customize `db_path` in `config.toml`

**Expected behavior**
No error ;)

**Error logs**
n/a

 - OS: Linux
 - Shell n/a
 - Version : 16.0.0

**Additional context**
I am trying to keep separate histories for `zsh` and `nushell` as I am trying to migrate to the latter - because they are very different.  ;)  I use `ATUIN_CONFIG_DIR` to point the ` nushell` instance to a dedicated config file.

I understand that to maintain separate histories for both shells, I most likely also need to customize `key_path` and `session_path` and create a separate registrations to get sync across machines to work.

conradludgate:
`atuin history end` errors if the ID doesn't exist. This can happen if the user changes their database setting mid-command.

Fix: ignore these errors, and log a warning.
Additional: Removes all potential missing entry database queries and handles the options appropriately

### DIFF
--- a/atuin-client/src/database.rs
+++ b/atuin-client/src/database.rs
@@ -73,7 +73,7 @@ pub trait Database: Send + Sync + 'static {
     async fn save(&mut self, h: &History) -> Result<()>;
     async fn save_bulk(&mut self, h: &[History]) -> Result<()>;
 
-    async fn load(&self, id: &str) -> Result<History>;
+    async fn load(&self, id: &str) -> Result<Option<History>>;
     async fn list(
         &self,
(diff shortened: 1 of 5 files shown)


ellie (approved):
🎉

PR #1254, opened 2023-09-25
better sync error messages

### LINKED ISSUE #1233: Unclear error messages on sync failure
**Describe the bug**
When running `atuin sync` an error occurs. Ideally we fix this, but at least I should get a meaningful error describing the problem.

**To Reproduce**

```

∴atuin sync -f
Error: error decoding response body: EOF while parsing a value at line 1 column 0

Caused by:
    EOF while parsing a value at line 1 column 0

Location:
    atuin-client/src/api_client.rs:246:21
```

 - OS: macOS 13.5.2 (22G91)
 - Shell zsh 5.9 (x86_64-apple-darwin22.1.0)

**Additional context**

It is possible I forgot my username or password.

conradludgate:
fixes #1233

### DIFF
--- a/atuin-client/src/api_client.rs
+++ b/atuin-client/src/api_client.rs
@@ -181,9 +181,19 @@ impl<'a> Client<'a> {
 
         let resp = self.client.get(url).send().await?;
 
-        let history = resp.json::<SyncHistoryResponse>().await?;
-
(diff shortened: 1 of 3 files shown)


ellie (approved):
🚀

PR #1259, opened 2023-09-26
calendar timezones

conradludgate:
no idea what this endpoint is for tbh, but I've fixed up the timezone support and simplified the logic


ellie (approved):
Probably the activity chart uses it, but I'll check

PR #1263, opened 2023-09-28
clear history id

### LINKED ISSUE #1261: Time of last-run command can be wrong (now instead of actual)
**Describe the bug**
The time for the most-recent command is often wrong. It shows basically now rather than the actual time. 

**To Reproduce**
1. Start a shell, run some commands; maybe use `date` as the last.
2. leave the shell idle for a while
3. hit return (an empty command)
4. hit up-arrow or ctrl-R to see command history
5. last command shows as Xms then 1s ago, counting up

**Expected behavior**
Time of commands should be the time they were run and entered into history

**Error logs**

 - OS: NixOS-unstable
 - Shell: bash
 - Version: `16.0.0`

conradludgate:
1. `s/RUST_LOG/ATUIN_LOG/` for all shells (cc #642) 
2. unset `ATUIN_HISTORY_ID` for all shells (fixes #1261)
3. Add trace log for the running command

### DIFF
--- a/atuin-client/src/database.rs
+++ b/atuin-client/src/database.rs
@@ -70,8 +70,8 @@ pub fn current_context() -> Context {
 
 #[async_trait]
 pub trait Database: Send + Sync + 'static {
-    async fn save(&mut self, h: &History) -> Result<()>;
-    async fn save_bulk(&mut self, h: &[History]) -> Result<()>;
+    async fn save(&self, h: &History) -> Result<()>;
+    async fn save_bulk(&self, h: &[History]) -> Result<()>;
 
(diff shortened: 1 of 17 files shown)


conradludgate:
tested in bash, need to test all the other shells that the syntax works.

It seems that in bash the preexec doesn't run if there's no command, but the precmd _does_ run. Fun!

ellie (dismissed):

ellie:
@conradludgate are we good to merge this one soon?

conradludgate:
Ah, I was meaning to test other shells

ellie (approved):

PR #1272, opened 2023-09-29
Fix/1207 deleted entries shown in interactive search

deicon:
* Deleted entries have not been excluded from the db queries
* Using deleted_at field to include/exclude deleted items


atuin/src/command/client/history.rs
```diff
@@ -266,6 +270,7 @@ impl Cmd {
         Ok(())
     }
 
+    #[allow(clippy::too_many_arguments)]
```
deicon, line 273: @ellie @conradludgate Not happy about this one, but seems to be a decision of yours, to not use subcommand structs to pass arguments into methods.

conradludgate: Not really a thought-out decision, It's just the easiest thing to do over time. This isn't a library so the chances of making mistakes with high arg count functions is low.

deicon: Totally agree on this one. Having structs with many booleans is also not much better and will also break clippy. So I would consider this a reasonable decision. Even if it not a decision :)

atuin/src/command/client/history.rs
```diff
@@ -51,6 +51,10 @@ pub enum Cmd {
         #[arg(long)]
         cmd_only: bool,
 
+        /// explicitely include deleted
+        #[arg(long)]
+        include_deleted: bool,
```
ellie, line 56: I don't _think_ there is ever a time when the user would want to include deleted records in outputs, so I think we're safe to only "include deleted" from sync functions - elsewhere, just default to false.

ellie (approved):
thank you! 🙌

deicon:
@ellie reverted your auto merge as cargo fmt was missing. my bad

deicon:
Anything missing still here ?

deicon:
@ellie Alright. Missed the print0 PR as it came later. I added the missing documentation of the print0 parameter as well after fixing the conflict

ellie (dismissed):

deicon:
@ellie I keep forgetting cargo fmt. need to update my config

ellie (approved):

PR #1273, opened 2023-09-29
fix(1220): Workspace Filtermode not handled in skim engine

### LINKED ISSUE #1220: Search mode "skim" does not work with "workspace" or "session" filter modes
I've been trying to use the "workspace" filter mode, but setting it up does not work as expected.

```toml
search_mode = "skim"
filter_mode = "workspace"
```

With this configuration, the workspace filter mode does come up and does show a list of workspace history. However searching it for *anything* results in a blank list. Changing the search mode to "fuzzy" or anything else works fine and searches the list as expected. The "skim" search mode works as expected for the other filter modes "host,global,directory".

 - OS: Linux
 - Shell zsh
 - Version 16.0.0

**Self hosted**
No

deicon:
This should Fix #1220, as the FilterMode::Workspace was not handled in skim engine

### DIFF
--- a/atuin/src/command/client/search/engines/skim.rs
+++ b/atuin/src/command/client/search/engines/skim.rs
@@ -54,10 +54,11 @@ async fn fuzzy_search(
             yield_now().await;
         }
         let context = &state.context;
-        let git_root = context.git_root.clone().map_or_else(
-            || context.cwd.clone(),
-            |git_root| git_root.to_str().unwrap_or("/").to_string(),
-        );
+        let git_root = context
+            .git_root
+            .as_ref()
+            .and_then(|git_root| git_root.to_str())
+            .unwrap_or(&context.cwd);
         match state.filter_mode {
             FilterMode::Global => {}
             // we aggregate host by ',' separating them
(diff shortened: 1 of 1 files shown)


deicon:
@conradludgate could you have a look here ? Not sure if this was all what's missing. Looks like the match was missing the workspace case. Or was this intentional ?

atuin/src/command/client/search/engines/skim.rs
```diff
@@ -54,6 +54,10 @@ async fn fuzzy_search(
             yield_now().await;
         }
         let context = &state.context;
+        let git_root = context.git_root.clone().map_or_else(
+            || context.cwd.clone(),
+            |git_root| git_root.to_str().unwrap_or("/").to_string(),
+        );
```
conradludgate, line 60: I'd like to avoid allocations in the search loop.

I believe this should compile
```suggestion
        let git_root = context.git_root.as_ref().and_then(|git_root| git_root.to_str()).unwrap_or(&context.cwd);
```

conradludgate (commented):
> Looks like the match was missing the workspace case. Or was this intentional ?
Not intentional, just forgotten

ellie (approved):
Thank you for the fix 🙏

PR #1274, opened 2023-09-29
Add --print0 to `history list`

offbyone:
Adding this enables tools like fzf to handle multiline commands, in the
same way that `find -print0 | fzf --read0` does.

I'm happy to add documentation and changelog entries to this as suits you; just
let me know where to put them.

### DIFF
--- a/atuin/src/command/client/history.rs
+++ b/atuin/src/command/client/history.rs
@@ -51,6 +51,10 @@ pub enum Cmd {
         #[arg(long)]
         cmd_only: bool,
 
+        /// Terminate the output with a null, for better multiline support
+        #[arg(long)]
+        print0: bool,
+
         #[arg(long, short, default_value = "true")]
         // accept no value
         #[arg(num_args(0..=1), default_missing_value("true"))]
@@ -100,7 +104,13 @@ impl ListMode {
 }
 
 #[allow(clippy::cast_sign_loss)]
-pub fn print_list(h: &[History], list_mode: ListMode, format: Option<&str>, reverse: bool) {
+pub fn print_list(
+    h: &[History],
+    list_mode: ListMode,
+    format: Option<&str>,
+    print0: bool,
+    reverse: bool,
+) {
     let w = std::io::stdout();
     let mut w = w.lock();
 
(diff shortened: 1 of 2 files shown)


atuin/src/command/client/history.rs
```diff
@@ -126,8 +136,16 @@ pub fn print_list(h: &[History], list_mode: ListMode, format: Option<&str>, reve
         Box::new(h.iter()) as Box<dyn Iterator<Item = &History>>
     };
 
+    let entry_terminator = if print0 { "\0" } else { "\n" };
```
conradludgate, line 139: No newline with nulls seems weird. Is that actually how the other tools work?

offbyone: Yes; the point of this is that if the string might contain newlines -- say, if the command does -- then this allows linewise processors to know where the boundaries of the command are.

The common pattern here is `find -print0 | xargs -0` which allows for filenames with all sorts of bonkers delimiter-like characters.

offbyone: I _do_ have one rust-ish question on this, though; I'm used to IO that flushes on newlines, which is why I am only calling `flush` when the delimiter is null. However, I'm not entirely sure if the output flushing for `write!` will behave that way, or if I'd need to flush when calling `write!(...something with a newline)` manually.

conradludgate: You are correct. write for std::io::stdout will flush on  newlines automatically. What you are doing here is equivalent

offbyone: Perfect. Well, let me know if you want any other details around this change; I'd love to get multiline history search in my shell again, and this is a key step)

ellie (approved):
This is great, thank you!

(also - nice work on your mastodon instance!)

atuin/src/command/client/history.rs
ellie: > I'd love to get multiline history search in my shell again

Mind if I ask where we're lacking in that regard?

offbyone: Honestly... I don't like the built-in search. I find the search tools fzf offers more enjoyable from my side, so finding a way to get atuin and fzf to play nice is my current goal. 

So, I've been working on history search functions that combine the output of `atuin`, a bit of null-terminated `gawk` scripting, and `fzf` to do the search.

conradludgate (approved):

offbyone:
> (also - nice work on your mastodon instance!)

Thanks on that last! It kind of fell into my life; I didn't start it, but I am keeping it afloat.

offbyone:
Will this be in 17.0? When (roughly) will that be released? I'm gonna be happy to stop running a homebuild version :D

ellie:
> Will this be in 17.0? 

Absolutely!

> When (roughly) will that be released?

Sometime in the next couple of weeks, time depending. There were some features (sync changes) I wanted to try and get in, but between myself + Conrad having some time off lately that didn't happen

PR #616, opened 2022-11-17
Re-enable arm docker builds

patricksjackson:
The issues from the linked issue seemed to be closed, so I re-enabled arm docker builds to check.

[The build succeeded](https://github.com/patricksjackson/atuin/actions/runs/3485248571), and you can find the packages [here](https://github.com/patricksjackson/atuin/pkgs/container/atuin). I haven't had the chance to manually test it yet.

Closes #439

PR #1276, opened 2023-10-02
Re-enable `linux/arm64` platform in CI docker build

rriski:
This was previously enabled in #616 and then reverted in #730.

The `ghcr.io/atuinsh/atuin:main` docker image doesn't provide an arm variant, making it harder to self-host on arm using docker.


ellie (approved):
I'm happy to give this a try, but until we can run it on native ARM it probably won't be great. Historically it has been pretty unreliable and slow

ellie:
If it's slow and unreliable I will likely disable it until github offer ARM or we just run out own. I know that's not ideal, sorry

PR #1278, opened 2023-10-03
Revert "Re-enable `linux/arm64` platform in CI docker build"

ellie:
Reverts atuinsh/atuin#1276

Sorry, this is taking more than an hour to build. I'll pay for + setup a ARM machine for the next release, but it's pretty untenable as is


rriski:
Woah that is extremely slow compared to the sub 3min with layer caching, e.g. https://github.com/atuinsh/atuin/actions/runs/6388325157/job/17337969068. I wonder how fast it would be with the arm layers cached as well? Anyway, thanks for taking a look at the first PR and handling reverting.

Issue #1199, opened 2023-08-29
atuin sync: thread 'main' panicked at 'failed to decrypt history! check your key: could not encrypt


sheeeng:
Experienced the following error when I run `atuin sync`.

```console
➜  ~ atuin sync      
0/0 up/down to record store
thread 'main' panicked at 'failed to decrypt history! check your key: could not encrypt
```

I have double checked the following values.

- `cat ~/.local/share/atuin/session` has the same value between two separate system.
- `echo $ATUIN_SESSION` has different value between two separate system.

I have tried `atuin logout` and `atuin login -u "${ATUIN_USERNAME}" -p "${ATUIN_PASSWORD}" -k "${ATUIN_KEY}"` again on the same that has this issue, but to no avail.

How we can further troubleshooting this issue?

LiamAEdwards:
I am having the same issue.

I have also double checked the following values.

    cat ~/.local/share/atuin/session has the same value between two separate system.
    echo $ATUIN_SESSION has different value between two separate system.

As well as logging out and manually logging in using the same details on my local machine (which works and is running Arch) and the remote machine.

I'm trying to run atuin on 

20.04.6 LTS (Focal Fossa)
In AWS EC2 with full ingress and egress.

Here is a backtrace.

```
 RUST_BACKTRACE=full atuin sync
0/0 up/down to record store
thread 'main' panicked at 'failed to decrypt history! check your key: could not encrypt

Location:
    /root/.cargo/registry/src/index.crates.io-6f17d22bba15001f/atuin-client-16.0.0/src/encryption.rs:132:22', /root/.cargo/registry/src/index.crates.io-6f17d22bba15001f/atuin-client-16.0.0/src/sync.rs:74:38
stack backtrace:
   0:     0x55709be4e051 - std::backtrace_rs::backtrace::libunwind::trace::he648b5c8dd376705
                               at /rustc/5680fa18feaa87f3ff04063800aec256c3d4b4be/library/std/src/../../backtrace/src/backtrace/libunwind.rs:93:5
   1:     0x55709be4e051 - std::backtrace_rs::backtrace::trace_unsynchronized::h5da3e203eef39e9f
                               at /rustc/5680fa18feaa87f3ff04063800aec256c3d4b4be/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:     0x55709be4e051 - std::sys_common::backtrace::_print_fmt::h8d28d3f20588ae4c
[...]

PR #1283, opened 2023-10-04
Fix bogus error message

alerque:
This PR is related to #1199, but does *not* fix it. It only corrects the error message: the operation is an attempt to decrypt and is failing, but the error thrown says something about encrypting. Whatever is causing the error remanis unresolved, but at least the confusing/contradictory message can be fixed.


tessus:
This PR needs an update. The files are now under the crates dir and the line has changed to 143

tessus:
LGTM.

@ellie sorry to bug you. this PR might have been forgotten, but is a low hanging fruit to reduce the number of open PRs.

ellie (approved):
Thanks @tessus! This is for sync v1 though, so will be gone fairly soon. But good to merge anyway

PR #1311, opened 2023-10-19
Add enter_accept to immediately execute an accepted command

ellie:
`enter_accept` will make Atuin immediately accept an execute a command
when selected. It defaults to false in our binary, but the default
config enables it.

This means that users who already use atuin will not default to the new
behaviour unless they opt in, but new users will have it by default.

Thanks to @davidhewitt for the patch and bulk of this implementation!
Currently we have it just for zsh, but I'll follow up with other shells
(unless anyone beats me to it :D)

This also uses "tab" to select and edit a command

### DIFF
--- a/atuin/src/command/client/search/interactive.rs
+++ b/atuin/src/command/client/search/interactive.rs
@@ -47,6 +47,7 @@ struct State {
     switched_search_mode: bool,
     search_mode: SearchMode,
     results_len: usize,
+    accept: bool,
 
     search: SearchState,
     engine: Box<dyn SearchEngine>,
@@ -130,7 +131,11 @@ impl State {
                     ExitMode::ReturnQuery => RETURN_QUERY,
                 })
             }
+            KeyCode::Tab => {
+                return Some(self.results_state.selected());
+            }
             KeyCode::Enter => {
+                self.accept = true;
                 return Some(self.results_state.selected());
             }
             KeyCode::Char('y') if ctrl => {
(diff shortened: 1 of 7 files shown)


ellie:
Ref: https://github.com/atuinsh/atuin/discussions/817#discussioncomment-7323475

atuin/src/command/client/search/interactive.rs
```diff
@@ -690,8 +701,12 @@ pub async fn history(
     }
 
     if index < results.len() {
+        let mut command = results.swap_remove(index).command;
+        if accept {
+            command = String::from("__atuin_accept__:") + &command;
```
conradludgate, line 706: For non zsh shells, this will still be output, right? I think we should use a secondary gate on each supported shell `ATUIN__SUPPORTS_ACCEPT` or something

ellie: oh oof that's true. yeah I'll do something like that, thank you

ellie: ended up just checking for existence of `ZSH_VERSION`

docs/docs/config/config.md
```diff
@@ -305,3 +305,16 @@ Default: 5
 
 The max time (in seconds) we wait for a connection to become established with a
 remote sync server. Any longer than this and the request will fail.
+
+## enter_accept
+Default: false
+
+Only supported on Bash.
```
conradludgate, line 312:
```suggestion
Only supported on Zsh.
```

conradludgate (approved):

docs/docs/config/config.md
ellie: 🤦‍♀️

ellie:
@davidhewitt thanks again for this! If you fancy a shiny contributor-only sticker in the post, please fill out this [form](https://notionforms.io/forms/contributors-stickers)!

davidhewitt:
Thanks! Not sure I did a huge amount, but I filled out the form anyway 👀

PR #1316, opened 2023-10-20
Add bash support to `enter_accept`

ellie:
It's a little less smooth than ZSH or Fish. The actual text to execute isn't entered before being eval-d like on the other shells, and if I try and set it then the command is inserted _after_ it is evaluated. ew.

Maybe explore BLE or similar in the future


conradludgate (approved):
Meh, fine

ellie:
> Meh, fine

Honestly don't think we will get bash to ever _feel_ as nice without layers of hacks 🙃

PR #1318, opened 2023-10-21
allow binding server to hostname

conradludgate:
I was sent a message with:
> `host = "mithril"`: `Error: invalid IP address syntax` why's it called host if it wants an IP

This makes the host field support more than just IPs, doing a getaddrinfo lookup if necessary


ellie:
> why's it called host if it wants an IP

Bind hosts do generally mean IP 😅

ellie (approved):

PR #1319, opened 2023-10-21
Fix no-exec commands not making it into the history 

### LINKED ISSUE #467: Add option to include commands that start with `#` in history
A common use-case for me is to

* write a command
* go to the beginning of the line
* add a `#`
* then doing some other command first to retrieve some data
* go into the shell history, select the commented out command, paste the extracted data, remove the `#`

Very basic example to illustrated what I'm doing also recorded on [asciinema](https://asciinema.org/a/jJ2ANfkVVqSegQIBMv5WUxsZc)

This appears not to be compatible with atuin right now, as adding a `#` is the equivalent of adding a ` ` at the beginning of the line.

norpol:
Fixes https://github.com/atuinsh/atuin/issues/467
Uses the same as zsh-histdb.
This will fix not adding to the history:

* when you start a command with `#` so `#echo asdf` will be stored in the history, even if you are using zsh `setopt interactivecomments`
* when you have a typo in your CLI, it should also now end up in the history, before atuin would only successfully store anything that could be executed

https://github.com/larkery/zsh-histdb/blob/30797f0c50c31c8d8de32386970c5d480e5ab35d/sqlite-history.zsh#L175C14-L175C27

Link to official zsh docs on this

https://zsh.sourceforge.io/Doc/Release/Functions.html#:~:text=is%20being%20executed.-,zshaddhistory,-Executed%20when%20a

### DIFF
--- a/atuin/src/shell/atuin.zsh
+++ b/atuin/src/shell/atuin.zsh
@@ -28,11 +28,13 @@ export ATUIN_SESSION=$(atuin uuid)
 ATUIN_HISTORY_ID=""
 
 _atuin_preexec() {
+    # remove ^J from end of $1 as it's passed by zshaddhistory
+    local cmd="${1[0, -2]}"
     # skip history for empty calls
     # remove space, \t, \n, and literal '\'
     [[ -n "${${1//[[:space:]]/}//\\/}" ]] || return
     local id
-    id=$(atuin history start -- "$1")
+    id=$(atuin history start -- "$cmd")
     export ATUIN_HISTORY_ID="$id"
     __atuin_preexec_time=${EPOCHREALTIME-}
 }
--- a/atuin/src/shell/atuin.zsh
+++ b/atuin/src/shell/atuin.zsh
@@ -1,3 +1,4 @@
+# shellcheck shell=bash # doesn't know about zsh
 # shellcheck disable=SC2034,SC2153,SC2086,SC2155
 
 # Above line is because shellcheck doesn't support zsh, per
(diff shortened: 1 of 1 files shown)


ellie:
Totally happy to merge this - however I would like to wait until after v17, which I'm releasing this week (now that I am **properly** home). It's a small change but has a large potential for unknown issues, so I'd like to run it for a while before release.

norpol:
For the time being one can add this snippet to your `.zshrc` as a workaround:

```bash
if echo "${preexec_functions}" | grep -qs atuin; then
	add-zsh-hook -d preexec _atuin_preexec
	add-zsh-hook zshaddhistory _atuin_preexec
else
	echo "remove atuin zshaddhistory hook workaround"
fi
```

norpol:
So after running it for two weeks now, the only issue I encounter is that occasionally empty history items make it into the history.

norpol:
I've added the check

`[[ -n "${${1//[[:space:]]/}//\\/}" ]] || return` which would ensure no calls to `atuin history start` when you'd accidentally type in spaces and empty new-lines.

atuin/src/shell/atuin.zsh
```diff
@@ -12,6 +12,9 @@ autoload -U add-zsh-hook
 export ATUIN_SESSION=$(atuin uuid)
 
 _atuin_preexec() {
+    # skip history for empty calls
+    # remove space, \t, \n, and literal '\'
+    [[ -n "${${1//[[:space:]]/}//\\/}" ]] || return
```
ellie, line 17: I appreciate why this is here, but would it be possible to implement it rust-side instead? we could then add some tests as well and ensure it doesn't happen for any shell

norpol: Would this be the right place to introduce such a check?

https://github.com/atuinsh/atuin/blob/1a432b61ee4f52f3ff0da1c991aa5d605f700869/atuin/src/command/client/history.rs#L293

markovejnovic:
This should be fixed in 86f38f17c9b36439cfdb13d35bfe6d2c825489d5 and a24e7571e9cacc8ca46d04df6811288d926a99b9.

Thank you all

PR #1320, opened 2023-10-23
fix(nushell): Ignore stderr messages

### LINKED ISSUE #1214: Working remotely
Thanks for creating a useful tool.

Perhaps my use case is a little odd, but when I am working remotely, my self-hosted server endpoint is not available. This means that when the tool attempts to periodically sync, I see panics:

`
Error: error sending request for url (https://domain.com/sync/status): error trying to connect: dns error: failed to lookup address information: nodename nor servname provided, or not known
                  Caused by:
                               0: error trying to connect: dns error: failed to lookup address information: nodename nor servname provided, or not known
                               1: dns error: failed to lookup address information: nodename nor servname provided, or not known
                               2: failed to lookup address information: nodename nor servname provided, or not known
                               Location:
                               /private/tmp/atuin-20230807-5737-1yn4qa1/atuin-16.0.0/atuin-client/src/api_client.rs:140:20
`

Not sure what the best way of handling this is, since ordinarily this would be something most users might want to know about, but in my case, this is totally expected and not something that causes concern.

Would it be possible, in the future, to have a way to disable such warnings/panics and perhaps only report the issue if the connection cannot be made within a certain configurable period of time ? For example, if I am offline and the server is unreachable for longer than a few days, then begin to notify the user of a connection issue?

arcuru:
Redirects stderr to null in nushell, following instructions at https://www.nushell.sh/book/stdout_stderr_exit_codes.html#stderr

The other shell scripts already redirect stderr for the `atuin history end` call, so this is matching their functionality.

I've tested this locally but I have not written nushell before, so this may not be idiomatic.

Closes #1214

### DIFF
--- a/atuin/src/shell/atuin.nu
+++ b/atuin/src/shell/atuin.nu
@@ -20,7 +20,8 @@ let _atuin_pre_prompt = {||
         return
     }
     with-env { ATUIN_LOG: error } {
-        atuin history end $'--exit=($last_exit)' -- $env.ATUIN_HISTORY_ID | null
+        do { atuin history end $'--exit=($last_exit)' -- $env.ATUIN_HISTORY_ID | null } | null
+
     }
     hide-env ATUIN_HISTORY_ID
 }


ellie (approved):
Thank you! 🚀

PR #1325, opened 2023-10-25
Revert "Revert "Revert "Use github runners for unit tests (for now)"""

ellie:
Reverts atuinsh/atuin#1295

Should be sorted THIS TIME. I just plugged it in and sorted it out, reconfigured so this shouldn't happen again.


ellie:
🎵 revert revert revert 🎵

Issue #1025, opened 2023-06-02
nushell: atuin should not always trigger on ⬆️


freijon:
With the nushell integration the "Arrow up" key is used to browse the history with `atuin`.
However, I don't always want that. For example when I'm in "picker mode" of nushell.

Example:
Type "math" and press <TAB>. A menu with available math commands is shown that (normally) can be navigated with ⬆/⬇. However with the `atuin` integration, the ⬆ key closes the pick menu and shows the history with `atuin`.

The history should only show up when in normal prompt mode. Not when a menu is shown.

arcuru:
This issue leads to a pretty terrible experience, should we disable the up arrow by default for Nushell until this can be fixed?

ellie:
> This issue leads to a pretty terrible experience, should we disable the up arrow by default for Nushell until this can be fixed?

I think that would probably be best, though hopefully their keybindings become more flexible soon 🙏  Are you happy to make the PR?

from-nibly:
FYI the default keybinding in nu is this

```nu
    {
        name: move_up
        modifier: none
        keycode: up
        mode: [emacs, vi_normal, vi_insert]
        event: {
            until: [
                {send: historyhintcomplete }
                {send: menuup}
                {send: up}
            ]
        }
    }
```

I think the "until" is how you leverage "doing the right thing at the right time".  I don't know any more details than that.

remmycat:
I'm not sure where you copied this from, but the [default config](https://github.com/nushell/nushell/blob/c2992d5d8b8541b64e9fc1440ed2917fc9b757f7/crates/nu-utils/src/sample_config/default_config.nu#L457) seems to be this and hasn't changed in a while, except for formatting:

```nushell
        {
            name: move_up
            modifier: none
            keycode: up
            mode: [emacs, vi_normal, vi_insert]
            event: {
                until: [
                    {send: menuup}
                    {send: up}
                ]
            }
        }
```

You can find the `until` reference here: https://www.nushell.sh/book/line_editor.html#until-type

Based on this, I modified the commented out part in the init script
```nushell
$env.config = (
    $env.config | upsert keybindings (
        $env.config.keybindings
        | append {
            name: atuin
            modifier: none
            keycode: up
            mode: [emacs, vi_normal, vi_insert]
            event: {
                until: [
                    {send: menuup}
                    {send: executehostcommand cmd: (_atuin_search_cmd '--shell-up-key-binding') }
                    {send: up}
                ]
            }
        }
[...]

PR #1329, opened 2023-10-26
fix(nu): disable the up-arrow keybinding for Nushell

arcuru:
The up-arrow keybinding triggers anytime the key is pressed, including in menus or multi-line commands. Disabling it by default for new users. See #1025 for additional context.

This leaves the binding in the init script but commented out, so users can enable it easily if needed.


ellie:
Thank you! 🙏

ellie (approved):

PR #1336, opened 2023-10-26
Silence fish string match

### LINKED ISSUE #1335: Spurious `__atuin_accept__:` when using `enter_accept` with fish
**Describe the bug**

When setting `enter_accept` to `true`, and pressing enter in atuin, the command printed in the shell starts with `__atuin_accept__:`.
It does execute correctly, but it looks a bit weird.

**To Reproduce**
1. Use fish
1. Enable `enter_accept`
2. Press enter on a history item and look at the command that was printed

**Expected behavior**
Just the command should be printed, without the extra prefix.

**Error logs**
None

 - OS: Linux, NixOS
 - Shell: fish 3.6.1
 - Version: 17.0.0

ellie:
Oops. `string match` is a bit too noisy, make it quiet. Also print the match before executing it.

Resolve #1335

### DIFF
--- a/atuin/src/shell/atuin.fish
+++ b/atuin/src/shell/atuin.fish
@@ -19,15 +19,18 @@ end
 
 function _atuin_search
     set h (ATUIN_SHELL_FISH=t ATUIN_LOG=error atuin search $argv -i -- (commandline -b) 3>&1 1>&2 2>&3)
-    commandline -f repaint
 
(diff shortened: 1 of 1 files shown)


conradludgate (approved):

arcuru:
LGTM :+1: This fixes the problem for me.

PR #1340, opened 2023-10-26
update support template

ellie:


ellie:
thanks for the catch @oliviacrain 🙌

PR #1341, opened 2023-10-27
fix(fish): improve output for `enter_accept`

arcuru:
Fixes 2 issues with the fish shell prompt when using `enter_accept`
1. Runs pre/post exec so the command is added to the atuin history.
2. Adds padding so that the repainting of the shell prompt doesn't overwrite the output.

This adds 2 lines of padding to account for prompts up to 2 lines tall, larger prompts will still cause repainting problems and smaller prompts will be getting an extra line. I don't think there's a way to know how tall the prompt is.

I haven't looked at whether zsh/bash have similar problems.


ellie (approved):
totally makes sense - thank you!

PR #1346, opened 2023-10-27
Add deb link download status check

HHPLow:
Sometimes, the deb package generates slower than other packages.
When we see the GitHub release news and use install.sh to install the latest deb will fail, and there is no friendly notice.


ellie:
Hey! This only happened this time because the deb build was actually failing. Normally, the deb file is built around the same time as the other binaries - which is only a few minutes after the release is out

Thank you for the contribution, but I'd rather leave it as-is at the moment

Issue #952, opened 2023-05-07
Frequent delays of up to 5s on cli commands


happenslol:
I've just begun using atuin, and I absolutely love it so far. However, there's been a recurring issue for me, which I've found hard to diagnose:

My prompt regularly blocks for between 500ms to 5s whenever I run a command. I've narrowed this down to the `_atuin_preexec` function, by manually importing the shell hook generated from `atuin init zsh` and annotating it with logging and `time` calls. Here's a sample time call from a time where it hang:

```
Running pre-exec for cd ~

0.00user 0.00system 0:04.93elapsed 0%CPU (0avgtext+0avgdata 8192maxresident)k
52036inputs+1064outputs (15major+512minor)pagefaults 0swaps

Pre-exec done for cd ~
```

Here's how I modified the hook to get the result:

```bash
_atuin_preexec() {
    log "Running pre-exec for $1\n" >> /tmp/atuin.log
    local id
    id=$(/usr/bin/time -a -o /tmp/atuin.log atuin history start -- "$1")
    export ATUIN_HISTORY_ID="$id"
    echo "\nPre-exec done for $1" >> /tmp/atuin.log
}
```

I've tried to replicate the behavior in cli use outside of the hook using `hyperfine`, and was successful:

```
» hyperfine -r 1000 "atuin search --limit 5"
Benchmark 1: atuin search --limit 5
  Time (mean ± σ):      18.3 ms ± 114.8 ms    [User: 4.9 ms, System: 8.2 ms]
  Range (min … max):    12.5 ms … 2587.9 ms    1000 runs
```

This does not happen on every benchmark, even with 1000 runs. My initial thought was that this has to be contention on the database file, but I saw that you're already using WAL, so concurrent writes/reads should not be a problem. I can also trigger the delay by repeatedly opening the search widget, which should not even be doing writes to the database, which confuses me even more.

Do you have any idea on how I could gather further data on this?

conradludgate:
Do you use the sync service? It's possible it's getting stuck and timing out (we run it occasionally in the pre-exec). Try `ATUIN_LOG=debug atuin sync` and see if there's anything interesting

happenslol:
Yeah, I'm syncing with my own server. What exactly am I looking for in the debug output? It seems to be running fine for me, does a whole lot of queries and seems to consistently complete in around 2.5 seconds.

[...]

PR #1348, opened 2023-10-27
fix(perf): Move DB writes off the critical path

arcuru:
**Status**: The code is definitely not ready here (fork crate only supports x86 linux/mac?), but opening this in case there's feedback. I've started dogfooding this on my own machines as it seems to be functional.

This moves the DB writes that occur on the critical path off into their own child processes. This will help in cases like Nushell which doesn't support backgrounding tasks, and in cases like #952 where writes to the filesystem are delayed for some other reason. With this new architecture the user won't be waiting on any filesystem writes to complete during normal operation.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -159,6 +159,7 @@ dependencies = [
  "directories",
  "env_logger",
  "eyre",
+ "fork",
  "fs-err",
  "futures-util",
  "fuzzy-matcher",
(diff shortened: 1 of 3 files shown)


conradludgate:
hmmmmm

ellie:
I'll be honest and say I'm pretty tentative to merge or consider this

1) `history end` is _already_ forked on all shells that are not NuShell, by the shell integration. It may make more sense for us to use pueue than to do this
2) `history start` in essentially all cases is incredibly fast - if it is not, there are undoubtedly filesystem or other serious errors occuring. 
3) If `start` is slow, and forked, we could end up in a race condition where start/end are trying to run at the same time. `end` will fail if `start` has not completed

It's not ideal, but in many cases users have only become aware of issues with their filesystem because Atuin blocks for a second or two (eg, zfs)

arcuru:
Yea I think I'm going to drop this. I've added workaround info to the ZFS thread so they can fix it themselves, and Nushell has known issues because of the background tasks. It's probably best if Atuin just stays out of trying to do this for now.

PR #1350, opened 2023-10-27
fix(enter_accept): clear old cmd snippet

arcuru:
Followup for #1341 and #1342. This clears the prompt of any command snippet the user entered before opening the atuin history search.

Both prompts still have issues with how the fake command looks after we've run it. Bash overwrite the whole prompt line with the command (See #1342), and I wasn't able to figure out how to clear the pre-existing cmd snippet from fish.

For the fish problem, as an example if you type `ech`, then hit ctrl+r and select `echo here` it will appear like this:

```
$ echecho here
here

$ 
```

I'm not sure how solid these fixes actually are, both shell's will cause a bit of confusion for users but I think it's the best we can do at the moment :shrug:

### DIFF
--- a/atuin/src/shell/atuin.bash
+++ b/atuin/src/shell/atuin.bash
@@ -29,6 +29,8 @@ __atuin_history() {
       eval "$HISTORY"
       _atuin_precmd
       echo
+      READLINE_LINE=""
+      READLINE_POINT=${#READLINE_LINE}
     else
       READLINE_LINE=${HISTORY}
       READLINE_POINT=${#READLINE_LINE}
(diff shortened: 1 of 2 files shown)


ellie (approved):
Thank you! 

I'll try and iterate on this a bit but will consider disabling it by default for those shells. Very frustrating.

arcuru:
If you're going to touch this yourself, I will say that I did try several different ways of fixing the fish problem. It appears that all the calls to `commandline` didn't take effect until after this function returned. They seemed to be queued up commands that fish would handle at the end.

So these 3 ways of fake replacing the original command didn't work
```
set hist (....)
commandline -r "$hist"
commandline -f repaint
```
```
set hist (....)
commandline -r ""
commandline -f repaint
echo $hist
```
```
set hist (....)
commandline --cursor 0 # <-- moves cursor to position 0
commandline -f repaint
echo $hist
```

PR #1352, opened 2023-10-28
Fix for breaking change in ratatui 0.24.0

tobiasge:
See here: https://github.com/ratatui-org/ratatui/pull/530
Replaces #1330


conradludgate (approved):
Thanks

PR #1357, opened 2023-10-29
feat(nix): Add a nixpkgs overlay

arcuru:
This adds a [Nix overlay](https://nixos.wiki/wiki/Overlays) for the Atuin flake.

This makes it easier for a consumer to directly use the repo instead of the packaged version in nixpkgs, and it is fairly standard for this setup. The simple explanation is it lets you overlay the git version of Atuin on top of the standard list of packages.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -31,5 +31,10 @@
           ];
         RUST_SRC_PATH = "${pkgs.rustPlatform.rustLibSrc}";
       });
-    });
+    })
+    // {
+      overlays.default = final: prev: {
+        inherit (self.packages.${final.system}) atuin;
+      };
+    };
 }


ellie (approved):
Thank you!

DoomHammer:
And how should one use it?

arcuru:
You apply the overlay to nixpkgs, though you'll need to be using Nix flakes.

Using it depends on exactly how you are using nixpkgs but there are examples here which probably apply: https://nixos.wiki/wiki/Overlays#Using_overlays

The overlay to use is generally going to be `atuin.overlays.default`.

It doesn't look like you are using flakes (based on your dotfiles), so you probably need to define the overlay yourself. Just using overrideAttrs to override the version should be enough: See https://nixos.wiki/wiki/Overlays#Overriding_a_version for some examples. For the `rev` you can set the tag that you want, e.g. "v17.0.1", or the git commit hash.

PR #1358, opened 2023-10-30
fix(nix): Add Appkit to the package build

arcuru:
Atuin v17 relies on Appkit to build on Mac. This was pointed out in discord and found in the [Nixpkgs PR](https://github.com/NixOS/nixpkgs/pull/263530)


conradludgate:
That's interesting. What changed to require appkit🤔

ellie (approved):

PR #1359, opened 2023-10-30
fix(zsh): Only trigger up-arrow on first line

### LINKED ISSUE #1351: Up-arrow keybinding for ZSH triggers for multiline commands
**Describe the bug**
If you write a multiline command in zsh, and try to move up a line, the atuin search will always trigger.

**To Reproduce**
1. Type any multiline command, e.g.

```
$ for x in *; do
echo $x
done
```
2. With your cursor on `done`, press up.

**Expected behavior**
The cursor moves to the line with `echo $x`. Instead the Atuin interactive search is triggered.

 - OS: Linux (NixOS)
 - Shell: zsh
 - Version: 17.0.0

**Additional context**
Fish does this binding by only triggering if it's on the top line of the command, and I was expecting that to be the case.

arcuru:
Fixes #1351

This adds a check so that the up-arrow keybinding only triggers if it is on the first line of a buffer. This fixes the case where we are moving in a multiline command.

Alternatively: ZSH's builtin command for this, `up-line-or-search`, disables history entirely when the command is multiple lines. We could simply match that behavior instead.

### DIFF
--- a/atuin/src/shell/atuin.zsh
+++ b/atuin/src/shell/atuin.zsh
@@ -50,8 +50,8 @@ _atuin_search() {
 }
 
 _atuin_up_search() {
-    # Only trigger on the first line of the buffer
-    if [[ ! $LBUFFER == *$'\n'* ]]; then
+    # Only trigger if the buffer is a single line
+    if [[ ! $BUFFER == *$'\n'* ]]; then
         _atuin_search --shell-up-key-binding
     else
         zle up-line
--- a/atuin/src/shell/atuin.zsh
+++ b/atuin/src/shell/atuin.zsh
@@ -50,7 +50,12 @@ _atuin_search() {
 }
 
 _atuin_up_search() {
-    _atuin_search --shell-up-key-binding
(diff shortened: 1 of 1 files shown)


ellie:
Thanks for the fix!

> Alternatively: ZSH's builtin command for this, up-line-or-search, disables history entirely when the command is multiple lines. We could simply match that behavior instead.

I think this behaviour might be the best one for us to match, though either solves the problem

arcuru:
Hmm, actually the true default zsh, `zsh -d -f -i`, does seem to trigger on the first line and not disable fully on a multiline command, the same as my first commit here. I've only been playing around with ZSH the last few days, there may be something in my config making the opposite decision.

Both seem fairly reasonable to me though. I'm happy to go match the behavior of whatever you choose in the fish config if you want.

I think it makes the most sense to trigger on the first line, but I don't have any strong feelings. The new commit here will disable entirely in multiline commands.

ellie (approved):
Thank you so much! Really appreciate you working on our shell integrations lately 🙏

PR #1360, opened 2023-10-30
fix(zsh): bind in the most popular modes

arcuru:
Improves the zsh keybindings so that we bind to the most common modes in ZSH.

Without this change we are only binding in the default keymap, and zsh's builtin history search would trigger in the others.

This also binds to 'k' in the vicmd map, since 'k' is the up key in that mode (and already bound to zsh's builtin `up-line-or-search` command).


ellie (approved):
Thank you!

PR #1362, opened 2023-10-30
Update nix flake to specify version

plttn:
Currently, the flake doesn't specify a version, resulting in `home-manager packages` reporting a blank version. 

Stretch goal for this would be to automatically bump the flake version number on a release, but that would be release workflow and I didn't want to touch that too much.


plttn:
Nevermind, realized this doesn't do what I expected it would do.

PR #1370, opened 2023-11-02
fix(fish): clean up the fish script options

### LINKED ISSUE #1368: atuin reports an unknown option "-qg" for string match
**Describe the bug**
When I use atuin to lookup a recent command, atuin reports the error as detailed in the _error logs_ section.
Note that atuin still succeeds in retrieving the command.

I recently upgraded to atuin version 17.0.1 and am using the fish shell

**To Reproduce**
1. use the fish shell
2. have `atuin init fish | source` in your fish configuration
3. have atuin version 17.0.1 installed
4. Find a recent command by pressing arrow up and selecting a command

**Expected behavior**
I would expect that atuin would succeed without reporting an unknown option

**Error logs**

```raw
 string match: Unknown option “-qg”                                                                                 

- (line 24): 
        if string match -qg '__atuin_accept__:*' "$h"
           ^
in function '_atuin_search' with arguments '--shell-up-key-binding'
	called on line 55 of file -
in function '_atuin_bind_up'

(Type 'help string' for related documentation)
```
 - OS: Ubuntu 20.04
 - Shell: fish
 - Version: 17.0.1

**Self hosted**
If am am not self hosting an Atuin server

**Additional context**

arcuru:
Closes #1368 

I reviewed and updated the options we use in the fish script, fixing several issues:

1. Removing the `--groups` option as reported in #1368, it's not needed if we're only testing whether the match exists.
2. Setting the correct scoping for variables. Setting all local variables to be local, `-l`, and removing the `-x` (export) setting for ATUIN_HISTORY_ID as it's unnecessary. The non-local variables (`s`, `h`, `hist`) were clobbering any user-defined variables with the same name.
3. Renaming `h` and `hist` to have an `ATUIN_` prefix, as those variables are accessible from user-defined code. Without this, using `enter_accept` with the history item `echo $h` would have given the user the script value of `$h` instead of their own.

### DIFF
--- a/atuin/src/shell/atuin.fish
+++ b/atuin/src/shell/atuin.fish
@@ -2,12 +2,12 @@ set -gx ATUIN_SESSION (atuin uuid)
 
 function _atuin_preexec --on-event fish_preexec
     if not test -n "$fish_private_mode"
-        set -gx ATUIN_HISTORY_ID (atuin history start -- "$argv[1]")
+        set -g ATUIN_HISTORY_ID (atuin history start -- "$argv[1]")
     end
 end
 
 function _atuin_postexec --on-event fish_postexec
-    set s $status
+    set -l s $status
 
     if test -n "$ATUIN_HISTORY_ID"
         ATUIN_LOG=error atuin history end --exit $s -- $ATUIN_HISTORY_ID &>/dev/null &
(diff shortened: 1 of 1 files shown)


ellie (approved):
Thank you! 🙏

PR #1373, opened 2023-11-03
fix(fish): use fish builtins for `enter_accept`

arcuru:
After checking again, I found fish's equivalent of zsh's `zle accept-line`. This change correctly handles the execute without resorting to any creative workarounds.

Note that the `repaint` in the middle here is needed, otherwise the prompt that contains the accepted command won't redraw.


ellie (approved):
Much much nicer, ty 🥳

PR #1375, opened 2023-11-05
Add xonsh support

Matthieu-LAURENT39:
This PR aims to add support for Xonsh to atuin.


conradludgate (commented):
That was quick. I guess xonsh has no mechanism for altering the ctrl-R keybindings?

Matthieu-LAURENT39:
This PR is blocked while i wait for a reply from the xonsh community.
It seems that xonsh doesn't have a way to source the output of commads, as it doesn't have an `eval` command (and the `eval` and `evalx` functions and the `@$()` syntax don't work). 
Therefore, the output of `atuin init xonsh` would have to be written to a temporary file, then sourced, which is far from ideal.
I'm waiting to get an answer on whether or not before there is a workaround before continuing, which might take a while as xonsh's matrix channel seems quite inactive.

arcuru:
> Therefore, the output of atuin init xonsh would have to be written to a temporary file, then sourced, which is far from ideal.

We already need to do that for nushell, so it's annoying but not that odd - https://atuin.sh/docs/advanced-install#nushell

I think the keybindings can be set using the `prompt_toolkit`? - https://xon.sh/tutorial_ptk.html

Matthieu-LAURENT39:
In that case, i'll use a temp file for now.
As for the prompt toolkit, it does look like it's the proper thing to use, thanks for the info!

jfmontanaro:
What's the status of this PR? I have been working getting atuin to work with my xonsh install and wound up with something fairly similar, except that I'm using `subprocess` directly to run the atuin search because doing it as a regular xonsh shell command results in a broken buffer state after it returns (e.g. for whatever reason you can't navigate left/right with arrow keys until you've either run the command or ctrl-c'd out.)

I'd be happy to contribute changes to this PR, or submit a new one if @Matthieu-LAURENT39 would prefer.

Matthieu-LAURENT39:
I didn't have time to work on this PR lately, so feel free to add improvements to it

jfmontanaro:
Followup question for the atuin maintainers: I'm working on history importers for xonsh to go with this (like zsh and nushell, xonsh has an alternate mode that stores history in a sqlite db.) I have two questions:

1. Xonsh has a concept of a "session id" much like atuin does, except that xonsh uses UUIDv4's while atuin uses UUIDv7's. Can I import history entries with the xonsh session id, or does atuin rely on e.g. the orderability of UUIDv7's for something?
2. Xonsh doesn't store the hostname of where a command was run, but it also doesn't have any kind of history-syncing built in. Is it reasonable to just assume that all commands in the history were run on the current device? Or is it too likely that e.g. people have set up their own syncing solutions that would break that assumption? (I notice none of the existing importers do this for hostnames, so maybe it's already been determined that it's preferable to leave it blank in this situation.)

ellie:
> Can I import history entries with the xonsh session id, or does atuin rely on e.g. the orderability of UUIDv7's for something?

We don't rely on it! It's purely for index performance. I'd prefer if you could generate a new UUIDv7 if possible, though if not the Xonsh ID is also OK

> Is it reasonable to just assume that all commands in the history were run on the current device?

This would be reasonable! If we don't have the information then there is little that can be done.

Matthieu-LAURENT39:
@jfmontanaro do you have anything else you want to add, or should i mark the PR as ready for review?

jfmontanaro:
I think you can go ahead and mark it ready. I am working on functionality for importing xonsh history, but I think I'll just open another PR for that later.

Matthieu-LAURENT39:
I'm adding support for `ATUIN_NOBIND`, but there's one thing i'm not sure about.
Do we check for that variable every time ctrl+r or up arrow is pressed, or only once during generation with `atuin init xonsh`?

Matthieu-LAURENT39:
In the end, i went with what it seems to do for bash, aka only check once during the initial sourcing, and not checking again afterwards.

Anyways, that's all i wanted to add, this should be ready for review!

ellie (approved):
From what I understand of Xonsh (not much), looks great! Thank you both 🙏 

Would one of you also be able to update the [docs](https://github.com/atuinsh/docs) please? 🙏 

Matthieu-LAURENT39:
I have some free time, i'll make a new PR to add some documentation :+1:

Issue #1003, opened 2023-05-23
Some commands not showing up and inaccurate command duration


gvlassis:
I am using freshly installed Atuin v14.0.1 through Homebrew on bash+macOS.

I have encounterd two weird issues that might or might not be related:
1) Sometimes a command does not show up in history (see "fastfetch" command in first screenshot). I have not noticed any pattern, even though there might be one.
2) Sometimes the duration of a command is inaccurate (see "which atuin" in second screenshot). Also, it sometimes shows up as 0 on the first invocation of the interactive history search and then as something else on the following invocations (compare "which atuin" in first and second screenshots). 
![Screenshot 2023-05-23 at 12 49 22 PM](https://github.com/ellie/atuin/assets/74119653/69a5a61a-92be-46c4-b0c0-6c5d8947e248)
![Screenshot 2023-05-23 at 12 49 49 PM](https://github.com/ellie/atuin/assets/74119653/10f83f4c-3be4-496d-8273-fa77a6129683)

conradludgate:
Hmm. That's very weird. What shell do you use?

danwilliams:
I have a brand-new installation on Bash, and I am finding that lots of my commands do not end up in history. Very odd.

I will run a command, hit up arrow to run it again - and it's not there. I have up arrow disabled for Atuin at present, too, but the command also doesn't show int he Atuin interface with Ctrl-R.

There is no pattern to this and if I run it again the command will generally then show up.

The problem is it leads to very unreliable behaviour and slows down shell interaction as you have to check what's about to be edited and not rely on your memory.

danwilliams:
I think I have found by usage an observation that the problem happens consistently when using Ctrl-R - the command chosen by that process never seems to be added afresh to the current history context, meaning that it's not available when you press Up Arrow. I'm not sure if it happens in other ways as well, but I will continue to observe.

arcuru:
@danwilliams You haven't said, but I suspect that you are using [`enter_accept`](https://atuin.sh/docs/config/#enter_accept) and you are using Atuin v17.0.0.

v17.0.1 has #1342, which will ensure that the command you select using `enter_accept` from the Atuin interactive search will show up in the atuin history.

I've opened #1383 to make sure that those commands will be added to the bash history as well so that your up-arrow usage will also work.

You can use both these fixes today by putting the code of `atuin init bash --disable-up-arrow` into .bashrc and editing the `__atuin_history` function to the version in the #1383 PR.

[...]

PR #1383, opened 2023-11-10
fix(bash): Add the `enter_accept` command to the bash history

arcuru:
Continues evolving the `enter_accept` bash prompt by adding the command to the bash history itself.

A good number of users disable the up-arrow key and rely on their shell history for that functionality. This leads to surprising behavior when using `enter_accept`. See  https://github.com/atuinsh/atuin/issues/1003#issuecomment-1805224058


arcuru:
Closing this for a moment, will open a new one that is more extensive.

PR #1384, opened 2023-11-10
Bash `enter_accept` best effort fixes

### LINKED ISSUE #1372: Commands executed via enter_accept from history can't take input
**Describe the bug**

When I complete a command from history, and that command asks for input from stdin, I can't actually give that input. Nothing happens when I type and the program receives no input.

**To Reproduce**

Assuming you have Python installed, this is a short and easy demonstration of the problem:

1. If you are an existing user, enable `enter_accept` first
2. Start a fresh shell
3. Execute the following short line of bash: `bash -c 'read -p "Say something? " tosay; echo "Say: $tosay"'`, enter something to be echoed back.
4. Recall the command, hit `<TAB>` and then enter to execute it again. It still works the same.
5. Recall the command and hit enter to execute it immediately. The `Say something? ` prompt runs but no input can be given any more.

If you remove the `bash -c '...'` wrapper and execute the commands directly in the current shell, then the above reproducer can't be interrupted either. You'll have to kill that shell from another session.

**Expected behavior**

`enter_accept` should not leave the environment in a broken state where input isn't possible.

 - OS: Linux
 - Shell bash
 - Version v17.0.1

arcuru:
This is a fairly large rewrite of the bash integration using `enter_accept`. This improves the feature in bash, redrawing the prompt correctly, running the preexec/precmd functions, and adding the command to the bash history.

**Note that this does not fix #1372, which is not fixable in bash.** It would only be fixable if we can find an equivalent of `zle accept-line` in ble.sh. I worked on this change hoping that I would be able to find a fix but I was unable to find one.

Despite putting in a bit of effort to clean this up, I think we should remove the `enter_accept` feature on bash, or at least disable it by default, until we can find a solution to #1372. It is very surprising behavior and will confuse anyone using this in bash. @ellie?

I did identify a separate issue with bash-preexec in my testing that is also not fixed by this change, though this change does not make it worse. When pressing ctrl+r, bash-preexec will execute the preexec when you press ctrl+r using the previous command entered, and it will not execute preexec again for that prompt. This means we're getting extra `atuin history start` commands for the same entry, which can end up causing problems with our recorded history. I'm not sure how fixable that bug is as it is some problem with the bash-preexec hooks.

Docs have also been updated to point users towards using ble.sh and warn of that bug, as I didn't find any of the same problems using that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -248,16 +248,30 @@ antigen bundle atuinsh/atuin@main
 
 ### bash
 
-We need to setup some hooks, so first install bash-preexec:
+#### [ble.sh](https://github.com/akinomyoga/ble.sh)
 
+Atuin works best in bash when using [ble.sh](https://github.com/akinomyoga/ble.sh).
+
+With ble.sh installed, just add atuin to your .bashrc
+
+```bash
+echo 'eval "$(atuin init bash)"' >> ~/.bashrc
 ```
+
+#### [bash-preexec](https://github.com/rcaloras/bash-preexec)
+
+[Bash-preexec](https://github.com/rcaloras/bash-preexec) can also be used, but you may experience some minor problems with the recorded duration and exit status of some commands.
+
+To use bash-preexec, download and initialize it
+
+```bash
 curl https://raw.githubusercontent.com/rcaloras/bash-preexec/master/bash-preexec.sh -o ~/.bash-preexec.sh
 echo '[[ -f ~/.bash-preexec.sh ]] && source ~/.bash-preexec.sh' >> ~/.bashrc
 ```
 
 Then setup Atuin
 
-```
+```bash
 echo 'eval "$(atuin init bash)"' >> ~/.bashrc
 ```
 
--- a/atuin/src/shell/atuin.bash
+++ b/atuin/src/shell/atuin.bash
@@ -1,19 +1,23 @@
 ATUIN_SESSION=$(atuin uuid)
 export ATUIN_SESSION
 
-_atuin_preexec() {
+__atuin_preexec() {
     local id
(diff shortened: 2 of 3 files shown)


ellie:
Thank you for all the effort!

> It would only be fixable if we can find an equivalent of zle accept-line in ble.sh. I worked on this change hoping that I would be able to find a fix but I was unable to find one

That's a shame. I agree with the suggestion here, we should disable enter_accept support in Bash until #1372 can be resolved

ellie (approved):
Thank you!

Maybe worth thinking about changing the bash install script too, although ble.sh isn't going to be a straight swap for bash-preexec and I don't think we can install something that large without asking the user.

arcuru:
> Maybe worth thinking about changing the bash install script too, although ble.sh isn't going to be a straight swap for bash-preexec and I don't think we can install something that large without asking the user.

I think ble.sh is a pretty big thing to install for them, since it takes over the entire command prompt. I think the current setup is pretty good, stating in the docs that it works better with ble.sh but not trying to install it for them in the install script. bash-preexec doesn't have nearly as many side effects.

akinomyoga:
> It would only be fixable if we can find an equivalent of `zle accept-line` in ble.sh. I worked on this change hoping that I would be able to find a fix but I was unable to find one.

You can simply call `ble/widget/accept-line`. I submitted a PR #1465. [These two lines](https://github.com/atuinsh/atuin/pull/1465/commits/6b956bb065dfd97477b0b4c4aaaeec2f46e0d627#diff-3f807152658ad715b684e8c72f0849926702cfb5b9e4d5966847909b822ce0e1R33-R34) sets the content of the command line and sets up the execution of the command.

PR #1385, opened 2023-11-11
Disable enter-accept for bash

ellie:
Ref #1384, #1372

It doesn't look like we will be able to get this to a point where it works well on bash. If this happens, we can enable it again


ellie:
failing test is upstream docker hub issues

PR #1393, opened 2023-11-16
Make `atuin account delete` void session + key

shymega:
When I deleted my Atuin.sh account, I found the command did not delete the key or session files.

This caused `bail!`s from Atuin when querying the session status, or synchronising.

Of course, one shouldn't expect Atuin to be able to synchronise or query with the server when the account is deleted, but the relevant files were still present, so the behaviour was different to what we'd expect.

I discussed this with @ellie on Discord, and we came to the conclusion that I should open a PR, and submit a patch.

I'm not sure how well this PR fits in with the borrow checker, but I've run tests, a Nix build, and a Cargo build - all goes well. I have not tested on macOS or Windows; only NixOS.

### DIFF
--- a/atuin/src/command/client/account/delete.rs
+++ b/atuin/src/command/client/account/delete.rs
@@ -1,9 +1,11 @@
 use atuin_client::{api_client, settings::Settings};
 use eyre::{bail, Result};
+use std::fs::remove_file;
 use std::path::PathBuf;
 
 pub async fn run(settings: &Settings) -> Result<()> {
     let session_path = settings.session_path.as_str();
+    let key_path = settings.key_path.as_str();
 
     if !PathBuf::from(session_path).exists() {
         bail!("You are not logged in");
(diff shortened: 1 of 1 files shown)


ellie (approved):
Thank you! 🙌 

ellie:
There's a few clippy issues + the formatter needs running - if you could sort that please, we're good to merge 😊

shymega:
Are you OK if I force-push and amend the commit, or would you prefer a separate commit for each fix, which could be squashed?

ellie:
Whatever you're happiest with, we squash and merge all PRs anyway!

shymega:
Decided to amend the commit and force-push. Made sense to me, as it's only formatting changes, and if you're squashing+merging - not much in it to separate each commit!

ellie:
Remaining clippy errors are unrelated to this PR, I'll fix separately

PR #1394, opened 2023-11-16
feat: add metrics server and http metrics

ellie:
I've setup a metrics server that serves prometheus metrics, when enabled. It runs on a totally separate server, to make it easier to not serve up public metrics.

I'll follow up with some more metrics, but the base http ones are pretty useful as-is. I won't need to scrape them from the reverse proxy any more

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -256,6 +256,8 @@ dependencies = [
  "eyre",
  "fs-err",
  "http",
+ "metrics",
+ "metrics-exporter-prometheus",
  "rand 0.8.5",
  "reqwest",
  "semver",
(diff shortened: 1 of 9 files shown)


mloskot:
I may be asking a silly question, but what is the rationale and purpose of having those servers in Atuin? 😊

ellie:
> I may be asking a silly question, but "what for?", what is the rationale and purpose of having those servers in Atuin? 😊 

So I can monitor api.atuin.sh and ensure it's healthy for users, rather than only knowing if it's online or not 

It's probably not useful if you're only running a small server

mloskot:
Hi @ellie, my apologies, but I've just deleted the comment with my question as I have realised after the fact (and after double-checking the https://atuin.sh/docs) what purpose the servers actually serve :)  Thanks anyway!

ellie:
Ahh no worries! Thank you

PR #1400, opened 2023-11-19
[draft] feat: history sync on top of records

ellie:
Using the new sync I introduced for records, sync history!

I'll need to test this a _tonne_, but so far records are

1. Appended to the local store
2. Synced up and down with `atuin sync`

Next: take records and append them to the atuin history db when received, and write a full history "rebuild" from the record store. Also deletions 🙃


ellie:
So I actually found a few annoyances (not huge issues) with the new sync while implementing this.

Nothing HUGE, but far far easier to fix now than later. I'm replacing the linked-list approach with more of an "array". So each (host, tag) store is an array, and the maths on indices is far easier than with the list.

Will just need to make sure we double check that each store has only a single writer

ellie:
will open a new PR for this as it's all old and a mess. basically the same code

Issue #1379, opened 2023-11-06
Can't take input in ssh


AlexeyMatskevich:
**Describe the bug**
After connecting via ssh to a remote server, I can't enter commands in shell. Sometimes commands are entered, but are not displayed until I press enter.

**To Reproduce**
1. cargo install atuin
2. curl https://raw.githubusercontent.com/rcaloras/bash-preexec/master/bash-preexec.sh -o ~/.bash-preexec.sh
3. echo '[[ -f ~/.bash-preexec.sh ]] && source ~/.bash-preexec.sh' >> ~/.bashrc
4. echo 'eval "$(atuin init bash)"' >> ~/.bashrc
5. source ~/.bashrc and reopen console
6. atuin import auto
7. test that atuin work (success for me)
8. ssh domain.example.com -p 9999

**Expected behavior**
It is expected that input is displayed and entered as normal.

**Error logs**
Please try and provide error logs where possible

**Device (please complete the following information):**
 - OS: Ubuntu 22.04 WSL2 (Windows 11)
 - Shell bash
 - Version 17.0.1

**Additional context**
If I disable atuin, all work
My bashrc
```sh
...
# Some default settings and commands from the system installation are presumed to be bash.
eval "$(~/.rbenv/bin/rbenv init - bash)"

export NVM_DIR="$HOME/.nvm"
[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"  # This loads nvm
[ -s "$NVM_DIR/bash_completion" ] && \. "$NVM_DIR/bash_completion"  # This loads nvm bash_completion
. "$HOME/.cargo/env"

# proto
export PROTO_HOME="$HOME/.proto"
export PATH="$PROTO_HOME/bin:$PATH"
export PATH="$HOME/.moon/bin:$PATH"
# [[ -f ~/.bash-preexec.sh ]] && source ~/.bash-preexec.sh
# eval "$(atuin init bash)"
```

ellie:
Unfortunately we've had issues with our bash implementation, see #1385, #1384, #1372

The solution so far is disabling "enter_accept". In the next version of Atuin this will be disabled for everyone. So

```
enter_accept = false
```

In your Atuin config. If this functionality is something you'd like, it's currently well supported in both ZSH and Fish

PR #1408, opened 2023-11-21
fix: reenable enter_accept for bash

arcuru:
I was looking through old issues, and found this: https://github.com/atuinsh/atuin/issues/78#issuecomment-1555949322. It appears that @yannickulrich may have already implemented a fix for our bash `enter_accept` input issue back in their initial suggestion. 

It does work for me for some of the previously broken commands. `ssh` works fine, as does `read -p "Say something?" tosay; echo "Say: $tosay"`. I do not know if there are limitations.

I have intentionally made this a draft PR because I think it should be tested a bit more first (and there's probably some doc updates that need to happen). I don't know what issues there might be and we should probably test this out a bit before we re-enable it for everyone. We don't want to be causing further confusion if we can help it.

List of the issues I could find, might be a good idea to ping some threads if this turns out to be a real fix: #1372, #1390, #1379

### DIFF
--- a/atuin/src/command/client/search/interactive.rs
+++ b/atuin/src/command/client/search/interactive.rs
@@ -705,7 +705,7 @@ pub async fn history(
 
     if index < results.len() {
         let mut command = results.swap_remove(index).command;
-        if accept && (utils::is_zsh() || utils::is_fish()) {
+        if accept && (utils::is_zsh() || utils::is_fish() || utils::is_bash()) {
             command = String::from("__atuin_accept__:") + &command;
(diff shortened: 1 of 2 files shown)


ellie:
ohhh fantastic! I agree, let's get it tested and then see 🤞

ellie:
Did you get much of a chance to test this? I'd like to release again fairly soon with all the fixes that have landed lately. No worries if not, I can dig in soon!

ellie (approved):
tysm ✨

PR #1413, opened 2023-11-27
Add support for installing binaries

crodjer:
As noted in a existing TODO, this installs the `atuin` binary directly at `/usr/local/bin` for operating systems that don't have an official method. For example: Raspberry Pi / MacOS (without homebrew).

Steps:
1. Get latest release info to find out the version.
2. Get the release archive and extract it to a temporary directory.
3. Move the `atuin` binary to PATH.

> This is based on the [installation script](https://github.com/crodjer/sysit/blob/main/scripts/install.sh)
> for `sysit` - a small utility I made.


install.sh
```diff
@@ -154,6 +153,45 @@ __atuin_install_cargo(){
 	cargo install atuin
 }
 
+__atuin_install_binary() {
+    local target
+    local arch=$(uname -sm)
+    local bin_dir=/usr/local/bin
+
+    # todo: Can make it an optional argument to install script as well,
```
ellie, line 161: could we do this now? would be really great to have!

install.sh
```diff
@@ -154,6 +153,45 @@ __atuin_install_cargo(){
...
+        return
+        ;;
+    esac
+
+    local latest_version=$(curl -s https://api.github.com/repos/atuinsh/atuin/releases/latest | grep tag_name  | grep -Eo "v[0-9]+\.[0-9]+\.[0-9]+")
+    local archive_name=atuin-$latest_version-$target
+    local binary_url="https://github.com/atuinsh/atuin/releases/download/v17.0.1/$archive_name.tar.gz"
+    download_dir=$(mktemp -d)
+    curl -s -L $binary_url | tar -xz -C $download_dir -f -
+    sudo mkdir -p $bin_dir
```
ellie, line 190: I think it would be good to 

1) Let the user know what's going on (eg tell them we couldn't find anything + will install a binary from releases)
2) Tell them where it's being installed, and ask for y/n to go ahead

I don't think we should go putting files in the users filesystem without getting their confirmation that is OK, especially when they'll be getting a `sudo` prompt without explanation

ellie (changes requested):
Thanks! Just a couple of comments

install.sh
crodjer: True. I think I can add logging here, with a prompt.

crodjer:
@ellie, made the suggested changes.

install.sh
```diff
@@ -34,6 +34,8 @@ LATEST_RELEASE=$(curl -L -s -H 'Accept: application/json' https://github.com/atu
 # Allow sed; sometimes it's more readable than ${variable//search/replace}
 # shellcheck disable=SC2001
 LATEST_VERSION=$(echo "$LATEST_RELEASE" | sed -e 's/.*"tag_name":"\([^"]*\)".*/\1/')
+# For binary install, use the first argument as the installation directory.
+INSTALL_DIRECTORY=${1:-/usr/local/bin}
```
ellie, line 38: I'd really rather we just made this an env var input. That way we don't change the interface _at all_, but still allow customization.

install.sh
```diff
@@ -154,6 +155,42 @@ __atuin_install_cargo(){
...
+        ;;
+    *)
+        __atuin_install_unsupported
+        return
+        ;;
+    esac
+
+    local latest_version=$(curl -s https://api.github.com/repos/atuinsh/atuin/releases/latest | grep tag_name  | grep -Eo "v[0-9]+\.[0-9]+\.[0-9]+")
+    local archive_name=atuin-$latest_version-$target
+    local binary_url="https://github.com/atuinsh/atuin/releases/download/v17.0.1/$archive_name.tar.gz"
```
ellie, line 183: This hardcodes v17.0.1

ellie (changes requested):
Just a couple more things

install.sh
crodjer: This makes sense.

install.sh
crodjer: Made a few updates for the review, shellcheck and a small fix for darwin's `stat` utility.

crodjer: @ellie , did you get a chance to see the changes?

ellie:
@crodjer looks like the prompt I mentioned in a previous review isn't here? 

> I think it would be good to
>
>   Let the user know what's going on (eg tell them we couldn't find anything + will install a binary from releases)
 >   Tell them where it's being installed, and **ask for y/n to go ahead**

I really don't want to go installing files on the users system without confirmation. It would be good to _tell_ them what files are going to be downloaded where, and also explain how they can change that.

Also - the real fix for Ubuntu is to also build a deb for aarch64 and install with the proper tooling, not to install a binary manually. This should be a last resort.

crodjer:
> looks like the prompt I mentioned in a previous review isn't here?

Yes, I figured we could rely on the `sudo` call. But you are right, users may miss out on seeing the prompt if sudo authorization is cached. So, added an explicit prompt as well. It'll behave like this:

```
Will copy atuin binary to /usr/local/bin...
Okay? [Y/n]: Y
Downloading atuin v17.1.0 from the latest github release...
Copying atuin binary to /usr/local/bin ...
```

I didn't want an explicit prompt because it'll then affect my scripts. For that, I have added another environment variable: `ATUIN_NO_PROMPT`. My goal toward having a binary install support was primarily to ensure my automation works on all my machines.

crodjer:
I understand. Lately I haven't been using `atuin`, so I didn't ping further for review.

I discovered https://github.com/cargo-bins/cargo-binstall, so `cargo binstall atuin` should be able to serve a similar use-case.
You might consider adding that as an option to the doc.

PR #1414, opened 2023-11-27
feat: allow spaces in stats prefixes

arcuru:
This change allows users to use spaces in their custom prefixes/subcommands.

A user could, for example, use both `cargo` and `cargo build` as subcommands and it would work as expected, with `cargo build foo` showing up as its own command.

I hit this because I use [script-directory](https://github.com/ianthehenry/sd) to organize my personal scripts, so I end up with heirarchically named shell scripts, and it makes sense to be able to separate them out.


ellie (approved):
lgtm, thank you! 🙌

PR #1417, opened 2023-11-29
Allow activating workspaces without making them the default everywhere

hco:
Hey there 👋 
While I love the workspaces feature, I don’t want it to always be the default.
This way you can disable that workspaces is the default filter if a workspace is found.

This is the first time ever I wrote rust code, so this might be really wrong. Let me know if I can improve something!

PS: Thanks for Atuin and all the great work around it, I really like it.

### DIFF
--- a/atuin-client/config.toml
+++ b/atuin-client/config.toml
@@ -42,10 +42,17 @@
 ## in any directory within a git repositiry tree (default: false)
 # workspaces = false
 
+## If a workspace was detected, Atuin will default to the workspace filter
+# filter_mode_prefer_workspaces = true
+
 ## which filter mode to use when atuin is invoked from a shell up-key binding
 ## the accepted values are identical to those of "filter_mode"
 ## leave unspecified to use same mode set in "filter_mode"
 # filter_mode_shell_up_key_binding = "global"
+
+## If a workspace was detected, Atuin will default to the workspace filter 
+## from the shell up-key binding
+# filter_mode_shell_up_key_binding_prefer_workspaces = true
 
 ## which search mode to use when atuin is invoked from a shell up-key binding
 ## the accepted values are identical to those of "search_mode"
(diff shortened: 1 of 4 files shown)


arcuru:
I think this is a good idea, but I'd suggest an alternate approach.

`workspaces` is just a filter mode that doesn't always apply, so we should expose it to the user as a filter mode. It should always show up in the filter list if it is a valid filter (i.e. if we're in a workspace), but not as the default unless the user has set it.

We should just have `filter_mode` take a list of filter modes. So if a user wanted to use workspaces they should set that to "workspace, global", or even a list ["workspace" "global"]. That way it will be easier to add more optional filter modes in the future as well.

Even better, maybe just allow the user to set a list of all the filter modes that they want to see, and in what order.

hco:
I love your idea. I can have a look if I can make this work.

That means you'd rather not merge this PR the way it is, right?

Apart from the feature going in the wrong direction: Have you looked if the code if that's okay in general? As this is my first time writing rust code anything I should do differently in a next iteration would be great to hear as early as possible 😊

arcuru:
First, please don't make major changes until you get an opinion from ellie or conrad. I've contributed here but am not a maintainer, so I was just expressing my opinion :) I was just a little worried about adding several more flags to the config file so I wanted to propose something a little cleaner.

Second, I think my suggestion would require a larger change than this but it shouldn't be too bad. It'd mainly require some changes in how `filter_mode` itself is handled/parsed. It does depend on what they decide to do for this feature but I'm sure you could handle it.

The structure of your code here looks fine to me, and it seems like you understand how this code is working as well. I'm confident you'd have no trouble implement a different approach.

hco:
Thanks for the input. So the best thing right now would be to wait for feedback from ellie or conrad?

ellie:
> We should just have filter_mode take a list of filter modes. So if a user wanted to use workspaces they should set that to "workspace, global", or even a list ["workspace" "global"]. That way it will be easier to add more optional filter modes in the future as well.

Sounds like a good idea to me! We'd need to ensure that there's a default in the case where `workspaces = true`, but the filter modes haven't been customized just yet, but otherwise I think this would be the best approach

Thanks for the work here @hco, do let us know if you need any help! And thanks @arcuru for your feedback and suggestions 🙏

PR #1418, opened 2023-11-29
fix(fish): accept multiline commands

### LINKED ISSUE #1404: Multiline command in fish becomes a single line when retrieved.
**Describe the bug**
Multiline command in fish becomes a single line when retrieved (even though previewed correctly).

**To Reproduce**
1. Run a multiline command
2. Invoke atuin (up arrow or ctrl R, doesn't matter)
3. Select the command with tab or enter (enter_select mode doesn't matter)

**Expected behavior**
Expect the multi-line command to be populated in the prompt.
Instead, the lines are smushed into one.

**Error logs**
N/A

 - OS: WSL2 v2 (ubuntu)
 - Shell fish
 - Version 17.0.1

**Self hosted**
N/A

**Additional context**
Noticed in all terminal applications.

arcuru:
Closes #1404

This fixes the multiline handling in fish, and as a side effect it also removes the dependency on awk. We need to directly capture the command substitution into a string to preserve the newlines instead of setting it to a variable directly.

### DIFF
--- a/atuin/src/shell/atuin.fish
+++ b/atuin/src/shell/atuin.fish
@@ -18,11 +18,11 @@ function _atuin_postexec --on-event fish_postexec
 end
 
 function _atuin_search
-    set -l ATUIN_H (ATUIN_SHELL_FISH=t ATUIN_LOG=error atuin search $argv -i -- (commandline -b) 3>&1 1>&2 2>&3)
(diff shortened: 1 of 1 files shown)


ellie (approved):
Thank you!

PR #1437, opened 2023-12-11
fix: don't require all fields under [stats]

r-vdp:
Before this change, when configuring only `common_subcommands` and not `common_prefix` (so it would take its default value), atuin produces an error message:
```
Error: could not load client settings  Caused by:     failed to deserialize: missing field `common_prefix`  Location:     atuin-client/src/settings.rs:456:26
Error: could not load client settings

Caused by:
    failed to deserialize: missing field `common_prefix`

Location:
    atuin-client/src/settings.rs:456:26
Error:: command not found
```

With this change, the fields can be specified separately and missing fields will take their default values.

### DIFF
--- a/atuin-client/src/settings.rs
+++ b/atuin-client/src/settings.rs
@@ -145,18 +145,30 @@ pub enum WordJumpMode {
 
 #[derive(Clone, Debug, Deserialize)]
 pub struct Stats {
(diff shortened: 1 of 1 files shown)


ellie (approved):
Nice fix! Thank you 🙏 

PR #1441, opened 2023-12-12
fix: respect ZSH's $ZDOTDIR environment variable

dennis-tra:
Follow up of #942


ellie (approved):
Thanks for following up! 🙏

PR #1445, opened 2023-12-14
docs: add Void Linux install instruction

tranzystorekk:
🎉🎉🎉


ellie (approved):
🥳 🥳 🥳 Thank you so much for this! 

If you would like a holographic contributors-only Atuin sticker, then please fill out this [form!](https://notionforms.io/forms/contributors-stickers)

PR #1456, opened 2023-12-19
feat: add semver checking to client requests

ellie:
This ensured that the client cannot try and sync while running a greater major version than the server

So far, servers have been backwards compatible with _all_ clients. Clients have not been forwards compatible with all servers. The former is much easier to achieve than the latter.

Instead of clients reporting weird errors when servers don't respond as they should, ensure that servers are kept up to date.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -219,6 +219,7 @@ dependencies = [
  "rand",
  "regex",
  "reqwest",
+ "reqwest-middleware",
  "rmp",
  "rusty_paserk",
  "rusty_paseto",
@@ -243,8 +244,10 @@ name = "atuin-common"
 version = "17.1.0"
 dependencies = [
  "eyre",
+ "lazy_static",
  "pretty_assertions",
  "rand",
+ "semver",
  "serde",
  "sqlx",
  "time",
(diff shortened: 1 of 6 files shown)


ellie:
will make #1400 much easier to roll out 🙏

conradludgate:
Should is also error if client version > server version? I recall we had problems before with self-hosters not updating the server and having obscure errors

ellie:
> Should is also error if client version > server version? I recall we had problems before with self-hosters not updating the server and having obscure errors

Yep - following up with that + server side checks, hence it's still a draft ;)

arcuru:
I'm not sure this is a good idea. When you bump the main sync server every user will be unable to sync until they update, and for many users that means waiting for their package manager to support the new version.

Is it possible to version the Sync API and have the client/server negotiate the version? Or is that too much work?

ellie:
> When you bump the main sync server every user will be unable to sync until they update, and for many users that means waiting for their package manager to support the new version.

This is pretty intentional. I don't plan on bumping the major version like we currently do - historically our version number hasn't meant anything at all and is essentially just a build/release number. v0.5.0 still syncs, and (very frustratingly) still has users.

We have no way of blocking older clients _and_ telling them that their version is no longer supported. Without this, we're stuck supporting every version forever or breaking old clients in a way that isn't informative to them

> Is it possible to version the Sync API and have the client/server negotiate the version? Or is that too much work?

This can be done pretty easily (and is done in some places) by branching on client user agent/version. But again, I'd like to be able to have a hard stop, and this currently isn't doable. 

I could be convinced to change this to have a sync protocol version independent of the rest, but I'm not sure that layer of indirection is necessary

ellie:
Been thinking, and the issue @arcuru raised is worth more thought. I'd like to revisit this, but maybe not so strongly as initially proposed here.

For now, I've set it so that we only error if the server is running a lower major version than the client. It's reasonable to expect server operators to ensure things are up to date, imo.

PR #1457, opened 2023-12-25
Add TLS to atuin-server

drbrain:
atuin as a project already includes most of the dependencies necessary for server-side TLS.  This allows `atuin server start` to use a TLS certificate when self-hosting in order to avoid the complication of wrapping it in a TLS-aware proxy server.

Configuration is handled similar to the metrics server with its own struct and currently accepts only the private key and certificate file paths:

```toml
# …

[tls]
enable = true
cert_path = "/path/to/letsencrypt/live/FQDN/fullchain.pem"
pkey_path = "/path/to/letsencrypt/live/FQDN/privkey.pem"
```

Starting a TLS server and a TCP server are divergent because the tests need to bind to an arbitrary port to avoid collisions across tests.  The API to accomplish this for a TLS server is much more verbose.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -110,6 +110,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a4668cab20f66d8d020e1fbc0ebe47217433c1b6c8f2040faf858554e394ace6"
 
 [[package]]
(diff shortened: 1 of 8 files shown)


ellie (approved):
Thank you!

PR #1465, opened 2023-12-28
fix(bash): improve the support for `enter_accept` with `ble.sh`

akinomyoga:
There are three commits: 1) the check for the ble.sh version, 2) the ble.sh's adapter to the bash-preexec interface, and 3) the `accept-line` widget to run the accepted command. A detailed description can be found in each commit message.

### DIFF
--- a/atuin/src/shell/atuin.bash
+++ b/atuin/src/shell/atuin.bash
@@ -94,9 +94,7 @@ __atuin_history() {
 
 # shellcheck disable=SC2154
 if [[ -n "${BLE_VERSION-}" ]] && ((_ble_version >= 400)); then
-    blehook PRECMD-+=__atuin_precmd
-    blehook PREEXEC-+=__atuin_preexec
-else
-    precmd_functions+=(__atuin_precmd)
-    preexec_functions+=(__atuin_preexec)
+    ble-import contrib/integration/bash-preexec
 fi
+precmd_functions+=(__atuin_precmd)
+preexec_functions+=(__atuin_preexec)
--- a/atuin/src/shell/atuin.bash
+++ b/atuin/src/shell/atuin.bash
@@ -92,7 +92,8 @@ __atuin_history() {
     fi
 }
 
-if [[ -n "${BLE_VERSION-}" ]]; then
+# shellcheck disable=SC2154
+if [[ -n "${BLE_VERSION-}" ]] && ((_ble_version >= 400)); then
     blehook PRECMD-+=__atuin_precmd
     blehook PREEXEC-+=__atuin_preexec
 else
--- a/atuin/src/shell/atuin.bash
+++ b/atuin/src/shell/atuin.bash
@@ -28,17 +28,19 @@ __atuin_history() {
     if [[ $HISTORY == __atuin_accept__:* ]]
     then
       HISTORY=${HISTORY#__atuin_accept__:}
-      # Reprint the prompt, accounting for multiple lines
-      local __atuin_prompt_offset
-      __atuin_prompt_offset=$(echo -n "${PS1@P}" | tr -cd '\n' | wc -c)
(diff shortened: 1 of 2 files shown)


atuin/src/shell/atuin.bash
```diff
@@ -28,25 +28,27 @@ __atuin_history() {
...
-      if ((__atuin_prompt_offset > 0)); then
-        tput cuu "$__atuin_prompt_offset"
-      fi
-      echo "${PS1@P}$HISTORY"
 
-      if [[ -n "${BLE_VERSION-}" ]]; then
-        blehook/invoke PREEXEC "$HISTORY"
+      if [[ -n "${BLE_ATTACHED-}" ]]; then
+        ble-edit/content/reset-and-check-dirty "$HISTORY"
+        ble/widget/accept-line
```
ellie, line 34: this is so nice 🙏

ellie (approved):
Thank you! I'd be interested in replacing bash-preexec with ble.sh someday, but would like to get your thoughts on that. What sort of side-effects might a user notice?

Could potentially be optional in the installer.

akinomyoga:
>  I'd be interested in replacing bash-preexec with ble.sh someday,

Thank you for thinking positively about adopting ble.sh!

However, after thinking about it a bit, I kind of have the same opinion as @arcuru in https://github.com/atuinsh/atuin/pull/1384#issuecomment-1811600600. While bash-preexec is a small script and can be a small helper for atuin, ble.sh is an independent line editor that significantly changes the behavior including syntax highlighting and autosuggestions. This could be annoying for users who just want the history manipulation feature of atuin but not any other extra features like syntax highlighting and autosuggestions.

I think it would be good to keep atuin to work without forcing users to have syntax highlighting, autosuggestions, and all the extra stuff of ble.sh. Nevertheless, ble.sh can be an option that the user can choose on the installation phase.

> What sort of side-effects might a user notice?

I try to eliminate the side effects of using ble.sh as much as possible, but there are still some inevitable drawbacks.

- One is related to the behavioral change. The aim of ble.sh is not to perfectly reproduce the behavior of GNU Readline (the default line editor of Bash), but *to add features*. For this reason, the detailed behavior is intensionally different from that of GNU Readline. However, some users might not want the new behavior or would be confused by the behavior difference. I try to provide configurable settings for such behavior differences, but typical users do not want to look for the settings but instead expect *the OOTB experience*.
- Also, this behavior difference can conflict with the other Bash configuration that relies on the detailed behavior of GNU Readline. For example, fzf uses the programmable completion of Bash in an unusual way assuming the detailed behavior of the completion. However, since ble.sh extends the completion system, the assumption by fzf is broken. Currently, the conflict is resolved by dynamically patching the settings by fzf when ble.sh detects it. I believe the most Bash configurations can be used with ble.sh without additional adjustments, but the configurations that hacks the detailed behavior of Bash can still be broken when used with ble.sh.
- One typical deal breaker for ble.sh seems to be the performance. Although I do not see any noticeable delay as far as I test it in GNU/Linux, *BSD, some people report that the response becomes slower. I do not have a perfect list of what can cause the delay, but I know some of them.
  - For example, in WSL2/Windows, the slow performance of the bridge between the filesystems in the Windows and Linux subsystems also affects the performance of ble.sh.
  - ble.sh is also slightly slow in Cygwin and MinGW/MSYS2 on Windows probably because of the slow TTY and process creation in the Windows subsystem.
  - ble.sh becomes extremely slow when used with an older version of Bash <= 3.2, or when used with the devel version of Bash where the debugging code is enabled.
  - ble.sh becomes slow also In a small devices with smaller CPU performance and smaller RAM. For example, there was a comment that ble.sh makes Bash slow in Termux, which is a Linux emulation in Android.
- Finally, as mentioned first, the user might not want extra features. Also, some users simply do not like the fact that ble.sh replaces the line editor.

PR #1467, opened 2023-12-29
fix(bash): fix small issues of `enter_accept` for the plain Bash

akinomyoga:
This PR resolves small issues of `enter_accept` for Bash without ble.sh. There are five different fixes, and again the details are in the commit messages. The original author @arcuru might have comments.

I plan to submit another PR for refactoring and a small performance improvement (which contains again 5 commits). If I should separate the PR into smaller chunks, please let me know. In such a case, I can re-submit those commits in separate PRs.

### DIFF
--- a/atuin/src/shell/atuin.bash
+++ b/atuin/src/shell/atuin.bash
@@ -67,14 +67,14 @@ __atuin_history() {
         stty_backup=$(stty -g)
         stty "$ATUIN_STTY"
 
-        eval "$HISTORY"
+        eval -- "$HISTORY"
         exit_status=$?
 
         stty "$stty_backup"
 
         # Execute preprompt commands
         __atuin_set_ret_value "$exit_status" "$HISTORY"
-        eval "$PROMPT_COMMAND"
+        eval -- "$PROMPT_COMMAND"
         # Bash will redraw only the line with the prompt after we finish,
         # so to work for a multiline prompt we need to print it ourselves,
         # then move up a line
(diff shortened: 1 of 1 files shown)


akinomyoga:
Not directly related to the fixes in this PR, but we might separate the related code into a new function `__atuin_accept_line` for the refactoring because most lines of `__atuin_history` are somehow the `accept-line` emulation. What do you think?

ellie (approved):
Thank you for all the fixes! And I appreciate how nicely broken down the commits here are 🙏 

> but we might separate the related code into a new function __atuin_accept_line for the refactoring because most lines of __atuin_history are somehow the accept-line emulation despite the function name. What do you think?

I think that would make sense. I don't feel strongly enough that it should block this PR, but it would be nice

akinomyoga:
> I think that would make sense. I don't feel strongly enough that it should block this PR, but it would be nice

Thank you for your reply! Then, should I submit a PR for `__atuin_accept_line`?

As I mentioned in [the first comment](https://github.com/atuinsh/atuin/pull/1467#issue-2058956778), I'll submit a PR for additional refactoring, but I suggest separating the `__atuin_accept_line` PR from the refactoring PR because the history tracking of `git blame` would be discontinued if we mix the code movement and changes in a single commit.

ellie:
Yes please! Sounds good to me 😊

akinomyoga:
Thanks!

akinomyoga:
I created PR #1476 for `__atuin_accept_line`.

PR #1469, opened 2023-12-30
feat(zsh): try to go to the position in zsh's history

lilydjwg:
by using infer-next-history and then up-history.

This is very helpful to execute consecutive commands with accept-line-and-down-history.

### DIFF
--- a/crates/atuin/src/shell/atuin.zsh
+++ b/crates/atuin/src/shell/atuin.zsh
@@ -68,6 +68,8 @@ _atuin_search() {
         then
             LBUFFER=${LBUFFER#__atuin_accept__:}
             zle accept-line
+        else
+            zle infer-next-history && zle up-history
         fi
     fi
 }


ellie:
Sorry, I totally missed this one and let it sit for a long time. I'd be totally happy to merge it - would you be able to resolve the conflicts please?

lilydjwg:
Hi, in case you didn't get the notification, I've resolved the conflict.

ellie:
Thank you! I did not get a notification :/ either that or it was lost amongst all the others. thanks for the ping!

ellie (approved):

AnimiVulpis:
For some reason this leads to weird behaviour sometimes in conjunction with the zsh-autosuggestion zsh plugin:

For example **before** this change it would "play out" like this:
```bash
❯ ▌ # CTRL+R to enter atuin
# Select e.g. cargo install --path .
❯ cargo install --path .▌
```

Now **after** this change it would "play out" like this:
```bash
❯ ▌ # CTRL+R to enter atuin
# Select e.g. cargo install --path .
❯ cargo install --path .⌷cat some.file
#                        ^^^^^^^^^^^^^ <- This part is grey like a suggested completion by zsh-autocomplete
```

I assume that this change (for some reason) leads to another suggestion being triggered after the atuin selection.

Any idea what might be causing this and how to fix this?
(For now I'll keep the 2 lines commented out)

If it help you can check my zsh config [here](https://github.com/AnimiVulpis/dotfiles/blob/main/stow-package/.zshrc)

AnimiVulpis:
@lilydjwg I finally managed to create a minimal reproduction setup.
See [here](https://github.com/AnimiVulpis/atuin-zsh-bug-reproduction).
It includes the steps necessary to reproduce the error locally.

What I found out so far is that this is a combination of zsh-autosuggestions and atuin.
It was hard to reproduce because zsh-autosuggestions needs to be configured with a `BUFFER_MAX_SIZE` and `USE_ASYNC`.
Then the command you select has to be longer than the buffer and the command following it in history (retrieved by `infer-next-history` I suspect) must be short enough to fit into the buffer.

I will do what I can to help you figure out how to fix this.

Does this help (at least making it possible to reproduce the error)?

lilydjwg:
@AnimiVulpis I reproduced it.

The newly-added two zle calls are captured by autosuggest: one short, one long. autosuggest ignores the long line but it also no longer cancels the previous request, which arrives later and accepted by autosuggest:

```
43.910341240 _zsh_autosuggest_modify buffer [echo 'a short command'], original widget autosuggest-orig-1-infer-next-history
43.913608334 _zsh_autosuggest_modify calls _zsh_autosuggest_fetch
43.915730511 _zsh_autosuggest_fetch echo 'a short command'
43.919389880 _zsh_autosuggest_modify buffer [echo 'a command so long it does not fit into the buffer of size 30'], original widget autosuggest-orig-1-up-history
43.921963316 _zsh_autosuggest_modify buffer [echo 'a command so long it does not fit into the buffer of size 30'], original widget autosuggest-orig-1-atuin-search
43.931162929 _zsh_autosuggest_async_response echo 'a short command'
```

I'll make a pull request to avoid this, but autosuggest should care to cancel previous requests even when the current one is going to be ignored.

PR #1477, opened 2024-01-01
fix(stats): time now_local not working on Linux

conradludgate:
On many platforms, getting local timezone requires a call to get_env which is not thread safe. This means it conservatively doesn't work if a thread has been spawned (I believe sqlite is using a separate thread here).

Attempt to fix by acquiring the local tz as early as possible.

### DIFF
--- a/atuin-client/src/settings.rs
+++ b/atuin-client/src/settings.rs
@@ -219,7 +219,13 @@ pub struct Settings {
 
     // This is automatically loaded when settings is created. Do not set in
     // config! Keep secrets and settings apart.
+    #[serde(skip)]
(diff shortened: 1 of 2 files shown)


ellie (approved):
🥳

conradludgate:
Let me verify that it actually works

ellie:
ah yes, draft 🤦‍♀️

conradludgate:
> I believe sqlite is using a separate thread here

Confirmed. I pulled in the `num_threads` crate that time uses to verify, and before the sqlite dbs are opened it is single threaded, after they are opened it has 3 threads. This change should indeed work

PR #1478, opened 2024-01-01
feat: rework record sync for improved reliability

ellie:
So, to tell a story

1. We introduced the record sync, intended to be the new algorithm to sync history.
2. On top of this, I added the KV store. This was intended as a simple test of the record sync, and to see if people wanted that sort of functionality
3. History remained syncing via the old means, as while it had issues it worked more-or-less OK. And we are aware of its flaws
4. If KV syncing worked ok, history would be moved across

KV syncing ran ok for 6mo or so, so I started to move across history. For several weeks, I ran a local fork of Atuin + the server that synced via records instead.

The record store maintained ordering via a linked list, which was a mistake. It performed well in testing, but was really difficult to debug and reason about. So when a few small sync issues occured, they took an extremely long time to debug.

This PR is huge, which I regret. It involves replacing the "parent" relationship that records once had (pointing to the previous record) with a simple index (generally referred to as idx). This also means we had to change the recordindex, which referenced "tails". Tails were the last item in the chain.

Now that we use an "array" vs linked list, that logic was also replaced. And is much simpler :D

Same for the queries that act on this data.

----

This isn't final - we still need to add

1. Proper server/client error handling, which has been lacking for a while
2. History store building. History records are pushed and synced, they just are not actually used for anything at all. 

The _only_ caveat here is that we basically lose data synced via the old record store. This is the KV data from before.

It hasn't been deleted or anything, just no longer hooked up. So it's totally possible to write a migration script. I just need to do that.


atuin-client/src/record/sqlite_store.rs
```diff
@@ -122,29 +123,75 @@ impl Store for SqliteStore {
...
+
+    async fn len(&self, host: HostId, tag: &str) -> Result<Option<u64>> {
+        let last = self.last(host, tag).await?;
+
+        if let Some(last) = last {
+            return Ok(Some(last.idx + 1));
+        }
+
+        return Ok(None);
+    }
```
conradludgate, line 163: a bit weird that empty doesn't return 0

atuin-client/src/record/sqlite_store.rs
```diff
@@ -122,29 +123,75 @@ impl Store for SqliteStore {
...
+
+    async fn next(
+        &self,
+        host: HostId,
+        tag: &str,
+        idx: RecordIdx,
+        limit: u64,
+    ) -> Result<Vec<Record<EncryptedData>>> {
+        let res =
+            sqlx::query("select * from store where idx >= ?1 and host = ?2 and tag = ?3 limit ?4")
```
conradludgate, line 173: Seems strange to me that `next` would include the current idx field. I assume you add 1 elsewhere

atuin-common/src/record.rs
```diff
@@ -162,21 +178,30 @@ impl RecordIndex {
...
 
-        ret.sort_by(|a, b| (a.host, a.tag.clone(), a.tail).cmp(&(b.host, b.tag.clone(), b.tail)));
+        // Stability is a nice property to have
+        ret.sort_by(|a, b| {
+            (a.host, a.tag.clone(), a.local, a.remote).cmp(&(
+                b.host,
+                b.tag.clone(),
+                b.local,
+                b.remote,
+            ))
```
conradludgate, line 203:
```suggestion
            std::cmp::Ord::cmp(
                &(a.host, &a.tag, a.local, a.remote),
                &(b.host, &b.tag, b.local, b.remote),
            )
```

alternatively, `#[derive(PartialEq, PartialOrd, Eq, Ord)]` on Diff and just `ret.sort()`.

atuin-client/src/history/store.rs
```diff
@@ -0,0 +1,219 @@
...
+        let record_type = decode::read_u8(&mut bytes).map_err(error_report)?;
+
+        match record_type {
+            // 0 -> HistoryRecord::Create
+            0 => {
+                // not super useful to us atm, but perhaps in the future
+                // written by write_bin above
+                let _ = decode::read_bin_len(&mut bytes).map_err(error_report)?;
+
+                let record = History::deserialize(bytes.remaining_slice(), HISTORY_VERSION)?;
```
conradludgate, line 79: this version is supposed to come from the encoded record version, so if a new client pushes a new record version, we can decode it properly or error

atuin/src/command/client/history.rs
```diff
@@ -367,13 +386,56 @@ impl Cmd {
...
+        self,
+        settings: &Settings,
+        db: &impl Database,
+        store: SqliteStore,
+    ) -> Result<()> {
         let context = current_context();
 
+        let encryption_key: [u8; 32] = encryption::load_key(settings)
+            .context("could not load encryption key")?
+            .into();
```
conradludgate, line 430: will break for non-sync users, right?

atuin/src/command/client/history.rs
```diff
@@ -300,10 +309,20 @@ impl Cmd {
         };
 
         db.update(&h).await?;
+        history_store.push(h).await?;
```
conradludgate, line 312: it's a bit unfortunate that this can run before InitStore, causing the indexes to be out of order. I guess it doesn't matter for history since it's ordered by the internally stored timestamp anyway

atuin-client/src/record/sqlite_store.rs
conradludgate: ok, I see how this is used during sync_upload. I think I might suggest a different name, or at least a different doc comment because `/// Get the record that follows this record` is wrong.

Maybe `get_page` and `/// Get a record page starting at the current record index`

atuin-client/src/record/sqlite_store.rs
ellie: true! At the time I was thinking absense of store vs length 0, but should totally simplify that

atuin/src/command/client/history.rs
ellie: Nope - `load_key` will generate a key if one doesn't exist. Regardless of sync, the store should save all data as encrypted

PR #1479, opened 2024-01-01
feat: integrate with zsh-autosuggestions

ellie:


atuin/src/shell/atuin.zsh
```diff
@@ -9,6 +9,15 @@
...
 autoload -U add-zsh-hook
 
+
+# If zsh-autosuggestions is installed, configure it to use Atuin's search. If
+# you'd like to override this, then add your config after the $(atuin init zsh)
+# in your .zshrc
+_zsh_autosuggest_strategy_atuin_top() {
+    suggestion=$(atuin search --cmd-only --limit 1 --search-mode prefix $1)
+}
+ZSH_AUTOSUGGEST_STRATEGY=atuin_top
```
arcuru, line 19: It looks like this is actually an array, we should probably append to the front: https://github.com/zsh-users/zsh-autosuggestions/blob/c3d4e576c9c86eac62884bd47c01f6faed043fc5/src/config.zsh#L18

We may also want to change the name to just `atuin` instead of `atuin_top`?

```suggestion
ZSH_AUTOSUGGEST_STRATEGY=("atuin_top" "${ZSH_AUTOSUGGEST_STRATEGY[@]}")
```

atuin/src/shell/atuin.zsh
```diff
@@ -9,6 +9,15 @@
 # Source this in your ~/.zshrc
 autoload -U add-zsh-hook
 
+
+# If zsh-autosuggestions is installed, configure it to use Atuin's search. If
+# you'd like to override this, then add your config after the $(atuin init zsh)
+# in your .zshrc
+_zsh_autosuggest_strategy_atuin_top() {
+    suggestion=$(atuin search --cmd-only --limit 1 --search-mode prefix $1)
```
arcuru, line 17: Do we need to quote this in zsh?
```suggestion
    suggestion=$(atuin search --cmd-only --limit 1 --search-mode prefix "$1")
```

ellie: It can't hurt

arcuru:
I double checked that this final version works on my machine :+1:

ellie:
Works great for me too :D

PR #1484, opened 2024-01-02
feat(shell): support high-resolution duration if available

akinomyoga:
This is a suggestion for a feature. If this is something you don't want, please don't hesitate to reject the PR.

ble.sh internally measures the command execution time using the builtin features of Bash. In particular, in bash >= 5.0, ble.sh uses `$EPOCHREALTIME` to measure the high-resolution execution time. The measured result is available in the shell variable `$_ble_exec_time_ata` in the `PRECMD` hook. `_ble_exec_time_ata` contains the execution time in the unit of microseconds.

In this patch, I suggest using this information for atuin to measure the command execution time, since the current duration measurement of atuin includes the fork time of atuin itself. Also, I added the units (us and ns) to display short duration in the search view.

In the second commit, I also implement support for the high-resolution execution time measurement for zsh using `EPOCHREALTIME` provided by module `zsh/datetime`.

### DIFF
--- a/atuin/src/command/client/history.rs
+++ b/atuin/src/command/client/history.rs
@@ -34,6 +34,8 @@ pub enum Cmd {
         id: String,
         #[arg(long, short)]
         exit: i64,
+        #[arg(long, short, default_value = "-1")]
+        duration: i64,
     },
 
     /// List all items in history
@@ -272,6 +274,7 @@ impl Cmd {
         settings: &Settings,
         id: &str,
         exit: i64,
+        duration: i64,
     ) -> Result<()> {
         if id.trim() == "" {
             return Ok(());
@@ -290,8 +293,12 @@ impl Cmd {
         }
 
         h.exit = exit;
-        h.duration = i64::try_from((OffsetDateTime::now_utc() - h.timestamp).whole_nanoseconds())
-            .context("command took over 292 years")?;
+        h.duration = if duration >= 0 {
+            duration
+        } else {
+            i64::try_from((OffsetDateTime::now_utc() - h.timestamp).whole_nanoseconds())
+                .context("command took over 292 years")?
+        };
 
         db.update(&h).await?;
 
@@ -366,7 +373,7 @@ impl Cmd {
 
         match self {
             Self::Start { command } => Self::handle_start(db, settings, &command).await,
-            Self::End { id, exit } => Self::handle_end(db, settings, &id, exit).await,
+            Self::End { id, exit, duration } => Self::handle_end(db, settings, &id, exit, duration).await,
             Self::List {
                 session,
                 cwd,
--- a/atuin/src/command/client/history.rs
+++ b/atuin/src/command/client/history.rs
@@ -34,8 +34,8 @@ pub enum Cmd {
         id: String,
         #[arg(long, short)]
         exit: i64,
-        #[arg(long, short, default_value = "-1")]
-        duration: i64,
+        #[arg(long, short)]
+        duration: Option<i64>,
     },
 
     /// List all items in history
@@ -274,7 +274,7 @@ impl Cmd {
         settings: &Settings,
         id: &str,
         exit: i64,
-        duration: i64,
+        duration: Option<i64>,
     ) -> Result<()> {
         if id.trim() == "" {
             return Ok(());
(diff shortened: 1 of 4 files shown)


atuin/src/command/client/history.rs
```diff
@@ -34,6 +34,8 @@ pub enum Cmd {
         id: String,
         #[arg(long, short)]
         exit: i64,
+        #[arg(long, short, default_value = "-1")]
+        duration: i64,
```
ellie, line 38: I'd rather we make this an option, it avoids the -1 special casing and ensures this is properly represented in the type system. We can then make it a u64

```suggestion
        #[arg(long, short)]
        duration: Option<u64>,
```

atuin/src/command/client/history.rs
```diff
@@ -272,6 +274,7 @@ impl Cmd {
         settings: &Settings,
         id: &str,
         exit: i64,
+        duration: i64,
```
ellie, line 277:
```suggestion
        duration: u64,
```

atuin/src/command/client/history.rs
```diff
@@ -290,8 +293,12 @@ impl Cmd {
         }
 
         h.exit = exit;
-        h.duration = i64::try_from((OffsetDateTime::now_utc() - h.timestamp).whole_nanoseconds())
-            .context("command took over 292 years")?;
+        h.duration = if duration >= 0 {
```
ellie, line 296: will also need changing to handle the optional

ellie (changes requested):
Generally happy with this - thanks! I didn't know about `$EPOCHREALTIME`

I've made some suggestions though - instread of i64 for duration, I think it would be better if we made it a `Option<u64>`. We can represent lack of value in the type system that way, rather than using a significant value

akinomyoga:
Thank you for those suggestions! I pushed commit 4478021, but I'm not familiar with Rust.

### `i64` vs `u64`?

First of all, I tried to change the type to the unsigned one, but it causes extra type-casting later. Member `duration` of `History` seems defined with the type `i64`:

https://github.com/atuinsh/atuin/blob/7f443588cf591b53ff77614d42ddc7823f20394f/atuin-client/src/history.rs#L25-L33

For this reason, typing the command-line argument `duration` as `Option<u64>` would cause inconsistency later. We might cast the type when we assign the result to `h.member`, but I kept it to be signed here to make the code cleaner. If the `duration` of the command-line argument still wants to be `u64` regardless of the definition in `struct History`, I can again adjust the PR. Please let me know.

### `Option<T>::unwrap_or_else` and `Result<T, E>` in closure

I first tried to write something like `duration.unwrap_or_else(|| i64::try_from(...)? )`, but the `?` operator cannot be used inside a closure that is supposed to return `i64`. What is the common practice in Rust in this case? I can instead write it like `duration.unwrap_or(i64::try_from(...)?)`, but this performs the retrieval of the current time and the duration calculation even when it is not needed. I'm currently using the `match` operator.

akinomyoga:
I pushed a new commit 81dce9a4. Maybe this could be an answer to the questions in my previous reply. If it's not, I can drop the commit any time.

ellie:
Great, thank you! In answer to your questions

> First of all, I tried to change the type to the unsigned one, but it causes extra type-casting later.

I see that you've sorted this anyway, but for clarity's sake; the history record is an abstraction around something stored in SQLite. SQLite doesn't support unsigned integers, hence the usage of i64 here. We probably shouldn't be leaking this implementation detail, but it's not a high priority to change.

> What is the common practice in Rust in this case? I can instead write it like duration.unwrap_or(i64::try_from(...)?), but this performs the retrieval of the current time and the duration calculation even when it is not needed. I'm currently using the match operator.

`match` is nice and clear, though you could also use [`unwrap_or_else`](https://doc.rust-lang.org/std/option/enum.Option.html#method.unwrap_or_else). This takes a closure, which will only be executed if required.

ellie (approved):
This is great, thank you!

In testing, looks like we're getting much more accurate timings for short-lived commands 🙏

akinomyoga:
> I see that you've sorted this anyway, but for clarity's sake; the history record is an abstraction around something stored in SQLite. SQLite doesn't support unsigned integers, hence the usage of i64 here.

Oh, I see. Thank you for your clarification on the background of `i64`!

Thank you for reviewing!

ellie:
> Thank you for reviewing!

Not a problem! Thank you for putting so much effort into contributing 🙏

akinomyoga:
@gvlassis Can you check the change in #1940?

akinomyoga:
Thanks!

PR #1486, opened 2024-01-02
fix zsh_autosuggest_strategy for no-unset environment

ds-cbo:
Fix some changes introduced in https://github.com/atuinsh/atuin/pull/1479 for environments with [`set -o nounset`](https://zsh-manual.netlify.app/options#:~:text=Treat%20unset%20parameters) leading to the following error:

```
(eval):19: ZSH_AUTOSUGGEST_STRATEGY[@]: parameter not set
```

Somewhat related to https://github.com/atuinsh/atuin/pull/921 ... maybe we should add a test for this environment, though I'm unsure how easy that would be

(Should probably squash this merge, I had some issues bringing my fork back up to origin)


ellie (approved):
Thank you! 🙏 

> maybe we should add a test for this environment

Absolutely. I'm not sure how easy that would be to do either, but for now I've just set my env so that I'll catch it in future. Not "scalable", but good enough for now

PR #1488, opened 2024-01-02
fix(bash): fix error by the use of `${PS1@P}` in bash < 4.4

akinomyoga:
The parameter expansions for the prompt strings, `${PS1@P}`, is only available in bash >= 4.4.  In bash <= 4.3 with bash-preexec, the current implementation produces error messages.  There is no way to evaluate PS1 with bash < 4.4, so we give up the adjustments for multiline prompts in bash < 4.4 in this patch.


ellie (approved):
🥳

akinomyoga:
Thanks!

Issue #193, opened 2021-10-20
Custom key bindings


mightyiam:
I'm using this and liking it. Thank you!

Just for the record, I'd like to use different key bindings than the default arrows.

dceddia:
I just found this and it seems awesome, but I wanted to customize the bindings too. I'm using zsh, and in the zsh setup the section that binds the keys is guarded by the ATUIN_NOBIND env var [here](https://github.com/ellie/atuin/blob/main/src/shell/atuin.zsh#L52).

For my case, I just wanted to disable the arrows but still use Ctrl-R, so I put this in my `.zshrc` and it did the trick!

```
ATUIN_NOBIND=1 eval "$(atuin init zsh)"
bindkey '^r' _atuin_search_widget
```

ellie:
I'm using this issue as the main issue for all "custom key bindings", whether that is inside the search TUI or from the shell

elthariel:
In addition to switching the custom key bindings, I'd really like to edit the key binding of the UI. More specifially, I'd like to have the `enter` key behave like the `tab` key, all the time. 
I mostly search command to edit previous long commands, so most of the time I don't want to run it before inspecting it

nghduc97:
> enter_ac



> In addition to switching the custom key bindings, I'd really like to edit the key binding of the UI. More specifially, I'd like to have the `enter` key behave like the `tab` key, all the time. I mostly search command to edit previous long commands, so most of the time I don't want to run it before inspecting it
> 
> Edit: found the `enter_accept` setting. My comment is way less relevant :)

Thanks man, I almost thought there is no way to change this default behavior. Made a pull request to put it in a more visible place [PR](https://github.com/atuinsh/docs/pull/57/files).

ShalokShalom:
So this can be closed?

Leandros:
I hate to ask, but any progress on this?

Leandros:
Well, I guess I'll give a go then. Could you be so kind to point me to where the input handling is done, and where the configuration lives? Thank you!

collinvandyck:
Search for `KeyCode` or go to `interactive.rs`

raine:
It would be great to be able to configure the `edit` hotkey.

PR #1491, opened 2024-01-03
feat(ui): delete history by Shift+Del

lilydjwg:
This is the same key as Firefox urlbar completion.

This allows the user to get rid of whatever they don't want to appear in the list as they see it.

I changed those magic usize consts to an enum for better semantics.

### DIFF
--- a/atuin/src/command/client/search/interactive.rs
+++ b/atuin/src/command/client/search/interactive.rs
@@ -39,6 +39,7 @@ use ratatui::{
 
 enum InputAction {
     Accept(usize),
+    Delete(usize),
     Copy(usize),
     ReturnOriginal,
     ReturnQuery,
@@ -125,6 +126,7 @@ impl State {
 
         let ctrl = input.modifiers.contains(KeyModifiers::CONTROL);
         let alt = input.modifiers.contains(KeyModifiers::ALT);
+        let shift = input.modifiers.contains(KeyModifiers::SHIFT);
 
         // Use Ctrl-n instead of Alt-n?
         let modfr = if settings.ctrl_n_shortcuts { ctrl } else { alt };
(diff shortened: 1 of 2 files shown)


ellie:
I haven't tested this on Linux just yet, but at least on mac/iterm this isn't triggerable. Putting traces on the key event, key modifiers for "delete" don't come through. Potentially a crossterm problem, more likely a mac setup problem. 

We had this sort of issue before with the alt-n shortcuts

The reason we have not yet had interactive deletion is that it's difficult to do so in a way that offers a nice UX, in a cross platform way. We probably also want confirmation if it's very easy to trigger from the UI.

With every proposed feature at the moment, we need to figure out a new keybinding for it. More #854, #910. Before we end up with emacs-style bindings, I'd like to at the very least make keybindings configurable (#193) 

I started working on an inspector in #1296, which would allow for a menu to be shown for a selected command. Since then, with `enter_accept`, tab is now in use. It could probably be changed to `BackTab` (aka shift+tab). Such a menu would be a great place for deletion.

I like your refactoring here though, thank you! That usize was messy. Would it be possible to get a separate PR with the refactoring changes? I'd love to get that merged

lilydjwg:
> Putting traces on the key event, key modifiers for "delete" don't come through.

Yes, some key combinations have limited terminal support unfortunately.

> We probably also want confirmation if it's very easy to trigger from the UI.

I don't think it's worth it. Firefox doesn't have confirmations at a lot of deletion place. Yes, we'd want it not to be triggered accidentally though. That's why I chose Shift+Del.

Configurable key bindings are definitely the way to go, but it's a lot more work to do. Yesterday someone told me that he chose mcfly over atuin because he wanted to delete from the UI, and so I did it :-)

> Would it be possible to get a separate PR with the refactoring changes?

Will do it later.

c-git:
Just adding my 2cents, I don't think confirmation on delete would be a show stopper for me. I actually prefer the no confirmation in Firefox for both the auto fill and urls because often I want to remove several and the confirmation would be annoying. They are fairly ephemeral so it's not the same but I do like that UX.

ellie:
I've now merged the inspector I mentioned above, which includes interactive deletion

I think it's better suited there! Sorry, but I'll be closing this

PR #1493, opened 2024-01-03
fix(server): typo with default config

ellie:
Without TLS config, the server fails to load defaults. Until this is released, add this to your server config

```
[tls]
enable = false
cert_path = ""
pkey_path = ""
```

### DIFF
--- a/atuin-server/src/settings.rs
+++ b/atuin-server/src/settings.rs
@@ -70,7 +70,7 @@ impl<DbSettings: DeserializeOwned> Settings<DbSettings> {
             .set_default("metrics.port", 9001)?
             .set_default("tls.enable", false)?
             .set_default("tls.cert_path", "")?
-            .set_default("tls.key_path", "")?
+            .set_default("tls.pkey_path", "")?
             .add_source(
                 Environment::with_prefix("atuin")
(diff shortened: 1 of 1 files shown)


ellie:
I've put the new config in the release notes - happy to cut a release if it's _really_ required

If you're automating this, you may be interested in the chart here: https://github.com/atuinsh/helm-charts

buroa:
> I've put the new config in the release notes - happy to cut a release if it's _really_ required
> 
> If you're automating this, you may be interested in the chart here: https://github.com/atuinsh/helm-charts

It's definitely not required, but it would be valuable for those that use a standard `app-template`: https://github.com/bjw-s/helm-charts. Almost all of my standalone apps use this template, which looks like the following: https://github.com/buroa/k8s-gitops/blob/master/kubernetes/apps/home/atuin/app/helmrelease.yaml. Thank you :-)

onedr0p:
Commenting here just to mention the env var fix would be setting `ATUIN_TLS__ENABLE` to `false`.

ellie:
I'll cut a patch in a bit

ellie:
Will be out shortly: https://github.com/atuinsh/atuin/pull/1495

PR #1494, opened 2024-01-03
docs(readme): add repology badge

tranzystorekk:
Hidden by default to avoid messing up the readme look, also configured for 3 columns to limit the horizontal space taken


ellie (approved):
Excellent! Thank you

PR #1496, opened 2024-01-03
chore: remove the teapot response

ellie:
It was fun, but it wasn't as informative as it needs to be

I'm leaving the function name though :)


binary-signal (approved):
Sensible change, being explicit about errors.

binary-signal (approved):

PR #1499, opened 2024-01-04
refactor: use enum instead of magic numbers

lilydjwg:
extracted from #1491.


ellie (approved):
much neater - thank you!

PR #1506, opened 2024-01-05
Prevent input to be interpreted as options for zsh autosuggestions

snaggen:
If you entered -a on the commandline it would be interpreted as an option to atuin. Adding -- will handle this correctly

PR #1511, opened 2024-01-06
fix(bash): prevent input to be interpreted as options for blesh auto-complete

akinomyoga:
This applies the change for zsh-autosuggestions in https://github.com/atuinsh/atuin/pull/1506 to the code for blesh auto-complete.


ellie (approved):
🙏

PR #1512, opened 2024-01-06
refactor: String -> HistoryId

ellie:


atuin-client/src/history/store.rs
```diff
@@ -15,8 +15,8 @@ pub struct HistoryStore {
 
 #[derive(Debug, Eq, PartialEq, Clone)]
 pub enum HistoryRecord {
-    Create(History), // Create a history record
-    Delete(String),  // Delete a history record, identified by ID
+    Create(History),   // Create a history record
+    Delete(HistoryId), // Delete a history record, identified by ID
```
ellie, line 19: This whole PR is basically just to make this interface nice and clear

PR #1513, opened 2024-01-06
refer to image with multi-arch support

matthewberryman:
The docker compose / k8s files all refer to an image without ARM support:
```
➜ docker pull ghcr.io/atuinsh/atuin:main
main: Pulling from atuinsh/atuin
no matching manifest for linux/arm64/v8 in the manifest list entries
```

This change refers to the image with ARM support, to enable usage on Linux arm64 and to improve performance on Apple Silicon machines.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,7 +2,7 @@ version: '3.5'
 services:
   atuin:
     restart: always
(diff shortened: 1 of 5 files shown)


ellie (approved):
Thank you!

PR #1515, opened 2024-01-07
fix(bash): fix and improve the keybinding to `up`

akinomyoga:
This PR includes three different fixes related to the keybinding to <kbd>up</kbd> in Bash. The descriptions are found in commit messages and in code comments.

### DIFF
--- a/atuin/src/command/init.rs
+++ b/atuin/src/command/init.rs
@@ -58,9 +58,18 @@ bindkey -M vicmd 'k' _atuin_up_search_widget";
         println!("{base}");
 
         if std::env::var("ATUIN_NOBIND").is_err() {
-            const BIND_CTRL_R: &str = r#"bind -x '"\C-r": __atuin_history'"#;
-            const BIND_UP_ARROW: &str = r#"bind -x '"\e[A": __atuin_history --shell-up-key-binding'
-bind -x '"\eOA": __atuin_history --shell-up-key-binding'"#;
+            // Note: We do not overwrite [C-r] in the vi-command keymap for
+            // Bash because we do not want to overwrite "redo", which is
+            // already bound to [C-r] in the vi_nmap keymap in ble.sh.
+            const BIND_CTRL_R: &str = r#"bind -m emacs -x '"\C-r": __atuin_history'
+bind -m vi-insert -x '"\C-r": __atuin_history'"#;
+            const BIND_UP_ARROW: &str = r#"bind -m emacs -x '"\e[A": __atuin_history --shell-up-key-binding'
+bind -m emacs -x '"\eOA": __atuin_history --shell-up-key-binding'
+bind -m vi-insert -x '"\e[A": __atuin_history --shell-up-key-binding'
+bind -m vi-insert -x '"\eOA": __atuin_history --shell-up-key-binding'
+bind -m vi-command -x '"\e[A": __atuin_history --shell-up-key-binding'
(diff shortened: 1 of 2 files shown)


akinomyoga:
Now, the inlined script is so long that we might be better to consider outputting just the switches for <kbd>C-r</kbd> and <kbd>up</kbd> and move the actual binding code to `atuin.bash` with the test for the switch. For example, we simplify `fn init_bash(&self)` as

```rs
let base = include_str!("../shell/atuin.bash");
let mut bind_ctrl_r = false;
let mut bind_up_arrow = false;
if std::env::var("ATUIN_NOBIND").is_err() {
    bind_ctrl_r = !self.disable_ctrl_r;
    bind_up_arrow = !self.disable_up_arrow;
}

println!("__atuin_bind_ctrl_r={bind_ctrl_r}");
println!("__atuin_bind_up_arrow={bind_up_arrow}");
println!("{base}");
```

and do branching in `atuin.bash` as

```bash
...

if [[ $__atuin_bind_ctrl_r == true ]]; then
    bind -m emacs -x '"\C-r": __atuin_history'
    ...
fi
if [[ $__atuin_bind_up_arrow == true ]]; then
    ...
fi
...
```

What do you think?

ellie:
I'm happy with this, thank you! Once the conflicts are resolved I'll get it merged

akinomyoga:
Thanks! I have rebased it!

ellie (dismissed):
Thank you! Really appreciate all the improvements you've been doing in this area

akinomyoga:
Hmm, `clippy` requires me to rewrite

```bash
let mut bind_ctrl_r = false;
let mut bind_up_arrow = false;
if std::env::var("ATUIN_NOBIND").is_err() {
    bind_ctrl_r = !self.disable_ctrl_r;
    bind_up_arrow = !self.disable_up_arrow;
}
```

as

```bash
let mut bind_ctrl_r = false;
let bind_up_arrow = if std::env::var("ATUIN_NOBIND").is_err() {
    bind_ctrl_r = !self.disable_ctrl_r;
    !self.disable_up_arrow
} else {
    false
};
```

but this breaks the symmetry of `bind_ctrl_r` and `bind_up_arrow`.

----

OK, this time I can write it in this way:

```bash
let (bind_ctrl_r, bind_up_arrow) = if std::env::var("ATUIN_NOBIND").is_err() {
    (!self.disable_ctrl_r, !self.disable_up_arrow)
} else {
    (false, false)
};
```

though this wouldn't be the general solution in similar cases.

ellie (approved):

akinomyoga:
Probably it reads better when the condition is inversed. I've amended it.

atuin/src/command/init.rs
```diff
@@ -55,19 +55,15 @@ bindkey -M vicmd 'k' _atuin_up_search_widget";
 
     fn init_bash(&self) {
         let base = include_str!("../shell/atuin.bash");
+        let (bind_ctrl_r, bind_up_arrow) = if !std::env::var("ATUIN_NOBIND").is_err() {
```
ellie, line 58:
```suggestion
        let (bind_ctrl_r, bind_up_arrow) = if std::env::var("ATUIN_NOBIND").is_ok() {
```

akinomyoga: Great!

ellie (approved):

akinomyoga:
Thank you for all your help!

PR #1517, opened 2024-01-07
Add timezone configuration option & CLI overrides

cyqsimon:
When reviewing history (usually for the purpose of filling out a work log), I often find it helpful if the timestamps are displayed in the local timezone. This PR adds this feature.

Currently, the `--tz` option either accepts an offset from UTC in the form of `<+|-><hour>[:<minute>[:<second>]]`, or the special value `l`/`local` to use the system's current timezones.

I also wanted to add support for named timezones using `chrono_tz`, but that seems to require pulling in `chrono` as well, which feels very redundant considering that we already depend on `time`. So I left that bit out for now.

## Unresolved questions

- [ ] <s>Are you (maintainers) willing to add `chrono` as a dependency for named timezone support? Or are there alternative, independent crates to `chrono_tz`?</s>
- [x] Do you think we need tests for this? I feel like it's borderline "too simple to warrant testing", but I would be happy to add them too if you so desire.

### DIFF
--- a/atuin-client/Cargo.toml
+++ b/atuin-client/Cargo.toml
@@ -22,7 +22,7 @@ atuin-common = { path = "../atuin-common", version = "17.2.1" }
 
 log = { workspace = true }
 base64 = { workspace = true }
-time = { workspace = true }
+time = { workspace = true, features = ["macros", "formatting"] }
 clap = { workspace = true }
 eyre = { workspace = true }
 directories = { workspace = true }
@@ -53,6 +53,7 @@ thiserror = { workspace = true }
 futures = "0.3"
 crypto_secretbox = "0.1.1"
 generic-array = { version = "0.14", features = ["serde"] }
+serde_with = "3.5.1"
 
 # encryption
 rusty_paseto = { version = "0.6.0", default-features = false }
--- a/atuin-client/src/settings.rs
+++ b/atuin-client/src/settings.rs
@@ -529,7 +529,7 @@ impl Settings {
             .set_default("key_path", key_path.to_str())?
             .set_default("session_path", session_path.to_str())?
             .set_default("dialect", "us")?
-            .set_default("timezone", "+0")?
+            .set_default("timezone", "local")?
             .set_default("auto_sync", true)?
             .set_default("update_check", cfg!(feature = "check-update"))?
             .set_default("sync_address", "https://api.atuin.sh")?
--- a/atuin-client/src/settings.rs
+++ b/atuin-client/src/settings.rs
@@ -1,6 +1,7 @@
 use std::{
     collections::HashMap,
     convert::TryFrom,
+    fmt,
     io::prelude::*,
     path::{Path, PathBuf},
     str::FromStr,
@@ -11,13 +12,18 @@ use clap::ValueEnum;
 use config::{
     builder::DefaultState, Config, ConfigBuilder, Environment, File as ConfigFile, FileFormat,
 };
-use eyre::{eyre, Context, Result};
+use eyre::{bail, eyre, Context, Error, Result};
 use fs_err::{create_dir_all, File};
 use parse_duration::parse;
 use regex::RegexSet;
 use semver::Version;
 use serde::Deserialize;
-use time::{format_description::well_known::Rfc3339, OffsetDateTime};
+use serde_with::DeserializeFromStr;
+use time::{
+    format_description::{well_known::Rfc3339, FormatItem},
+    macros::format_description,
+    OffsetDateTime, UtcOffset,
+};
 use uuid::Uuid;
 
 pub const HISTORY_PAGE_SIZE: i64 = 100;
@@ -120,6 +126,46 @@ impl From<Dialect> for interim::Dialect {
             Dialect::Uk => interim::Dialect::Uk,
             Dialect::Us => interim::Dialect::Us,
         }
+    }
+}
+
+/// Type wrapper around `time::UtcOffset` to support a wider variety of timezone formats.
+///
+/// Note that the parsing of this struct needs to be done before starting any
+/// multithreaded runtime, otherwise it will fail on most Unix systems.
+///
+/// See: https://github.com/atuinsh/atuin/pull/1517#discussion_r1447516426
+#[derive(Clone, Copy, Debug, Eq, PartialEq, DeserializeFromStr)]
+pub struct Timezone(pub UtcOffset);
+impl fmt::Display for Timezone {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        self.0.fmt(f)
+    }
+}
+/// format: <+|-><hour>[:<minute>[:<second>]]
+static OFFSET_FMT: &[FormatItem<'_>] =
+    format_description!("[offset_hour sign:mandatory padding:none][optional [:[offset_minute padding:none][optional [:[offset_second padding:none]]]]]");
+impl FromStr for Timezone {
+    type Err = Error;
+
+    fn from_str(s: &str) -> Result<Self> {
+        // local timezone
+        if matches!(s.to_lowercase().as_str(), "l" | "local") {
+            let offset = UtcOffset::current_local_offset()?;
+            return Ok(Self(offset));
+        }
+
+        // offset from UTC
+        if let Ok(offset) = UtcOffset::parse(s, OFFSET_FMT) {
+            return Ok(Self(offset));
+        }
+
+        // IDEA: Currently named timezones are not supported, because the well-known crate
+        // for this is `chrono_tz`, which is not really interoperable with the datetime crate
+        // that we currently use - `time`. If ever we migrate to using `chrono`, this would
+        // be a good feature to add.
+
+        bail!(r#""{s}" is not a valid timezone spec"#)
     }
 }
 
@@ -251,6 +297,7 @@ pub struct Sync {
 #[derive(Clone, Debug, Deserialize)]
 pub struct Settings {
     pub dialect: Dialect,
+    pub timezone: Timezone,
     pub style: Style,
     pub auto_sync: bool,
     pub update_check: bool,
@@ -304,11 +351,6 @@ pub struct Settings {
     // config! Keep secrets and settings apart.
     #[serde(skip)]
     pub session_token: String,
-
-    // This is determined at startup and cached.
-    // This is due to non-threadsafe get-env limitations.
-    #[serde(skip)]
-    pub local_tz: Option<time::UtcOffset>,
 }
 
 impl Settings {
@@ -487,6 +529,7 @@ impl Settings {
             .set_default("key_path", key_path.to_str())?
             .set_default("session_path", session_path.to_str())?
             .set_default("dialect", "us")?
+            .set_default("timezone", "+0")?
             .set_default("auto_sync", true)?
             .set_default("update_check", cfg!(feature = "check-update"))?
             .set_default("sync_address", "https://api.atuin.sh")?
@@ -591,8 +634,6 @@ impl Settings {
             settings.session_token = String::from("not logged in");
         }
 
-        settings.local_tz = time::UtcOffset::current_local_offset().ok();
-
         Ok(settings)
     }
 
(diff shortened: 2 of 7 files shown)


ellie:
Generally looks good, thank you!

>  Are you (maintainers) willing to add chrono as a dependency for named timezone support? Or are there alternative, independent crates to chrono_tz?

If there are any other options, then I'd rather not. I'm not aware of any other crates, though `time` is not named particularly helpfully for googling things. If `chrono` is the only possible choice then that's ok though

>  Do you think we need tests for this? I feel like it's borderline "too simple to warrant testing", but I would be happy to add them too if you so desire.

Tests would be good, even if very bare. It would be nice to ensure this continues to work with any possible future changes too

Otherwise, would you be able to update the docs too please?

cyqsimon:
So I tried to implement named timezones using `chrono_tz` but as we all know, trying is the first step to failure.

It turns out (unsurprisingly so) that `time` and `chrono` have complete different ways and syntices to handle timezones, and are barely interoperable. With enough effort I think I can get it to work, but the maintainability will surely be bad and frankly I think it's just not worth it.

I'll add tests and call it a day I guess. If ever in the future we decide to migrate to using chrono, I will be happy to add this retroactively.

atuin/src/command/client/history.rs
```diff
@@ -201,32 +269,38 @@ impl FormatKey for FmtHistory<'_> {
...
             "duration" => {
-                let dur = Duration::from_nanos(std::cmp::max(self.0.duration, 0) as u64);
+                let dur = Duration::from_nanos(std::cmp::max(self.history.duration, 0) as u64);
                 format_duration_into(dur, f)?;
             }
             "time" => {
-                self.0
+                self.history
                     .timestamp
+                    .to_offset(self.tz.0)
```
cyqsimon, line 282: I mean, there's really not much to test in terms of applying the offset. It's literally this single function call that should be covered by `time`'s tests.

cyqsimon (commented):
I wrote some basic unit tests. Maybe they'll help ensure that in the future a different implementation won't change the behaviour, but apart from that they really aren't too helpful.

atuin/src/command/client/history.rs
```diff
@@ -475,3 +553,52 @@ impl Cmd {
...
+    use super::Timezone;
+
+    #[test]
+    fn can_parse_local_timezone_spec() -> Result<()> {
+        let local_tz = Timezone(UtcOffset::current_local_offset()?);
+
+        assert_eq!(Timezone::from_str("l")?, local_tz);
+        assert_eq!(Timezone::from_str("local")?, local_tz);
+        Ok(())
+    }
```
cyqsimon, line 573: Also, this test is failing on Linux. I looked at it with a debugger, and noticed a contradiction between `time`'s code and documentation. I'll submit an issue there to ask about it.

cyqsimon: I did some digging, and it appears that there's no mistake in `time`; rather it's my own misunderstanding. Still though, I have to find a way to fix this test. Has to do with some really annoying thread-safety issues so it might take me a while to get it right.

cyqsimon: So I am going down this [***massive*** rabbit hole](https://github.com/time-rs/time/issues/293) that basically boils down to "there's no thread-safe way to get a system's timezone on POSIX systems". To be honest I'm completely flustered at this point.

To deal with this, `time` has implemented [some rather stringent checks](https://github.com/time-rs/time/blob/v0.3.31/time/src/sys/local_offset_at/unix.rs#L129-L154) to ensure safety. This means this test either has to run in a single-threaded environment, or we have to accept unsoundness (which admittedly can only ever happen if libc's `setenv` function is called directly without locking, but still). I'll see what I can do tomorrow.

ellie:
Hey! Just wondering if there's anything I can do to help this one move forwards? 🙏

cyqsimon:
> Hey! Just wondering if there's anything I can do to help this one move forwards? 🙏

Thanks for offering. Here's where I'm at:

As mentioned previously, the safety checks implemented by `time` while justified, is making acquiring the local timezone failure-prone on Unix systems, both in release and in test.

Currently, the specific problem is that `cargo test` runs the tests multithreaded by default, which makes `time` refuse to report the local timezone. I've been trying to find a way around it (without mandating the user run tests with `-j1`) but for now have come up empty-handed.

There's also the long-term concern that this is essentially a landmine if in the future we ever want to make `atuin-history` multithreaded for whatever reason. It's just annoying all-round.

cyqsimon:
I've also explored the very hacky method of obtaining the local timezone using a shell command, something like this:

```rust
#[cfg(target_os = "linux")]
let offset = {
    let tz_raw = process::Command::new("date").arg("+%:::z").output()?.stdout;
    UtcOffset::parse(std::str::from_utf8(&tz_raw)?.trim(), OFFSET_FMT)?
};
#[cfg(not(target_os = "linux"))]
let offset = UtcOffset::current_local_offset()?;
```

Unfortunately, not even this is viable because unlike GNU date, [POSIX date](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/date.html) only supports printing named timezones.

I'm honestly dumbfounded that such a simple problem can cause so much trouble.

ellie:
I've also been unable to find anything that would be a reasonable workaround, and keep local timezone support in this way.

What might work is

1. Add timezone config to Settings
2. Read the local timezone _right at the very beginning_, before we potentially start any multithreaded runtimes.
3. If the config file doesn't specify a timezone, override with the value we read at program start
4. For tests, specify an offset - they have a different entry point, so will never try to read anything from the env

Doesn't quite achieve everything we'd like to, but is 80% of the way there I think.

ellie:
also, thank you for looking into this so much, ik it's been more than expected

cyqsimon:
> 2. Read the local timezone _right at the very beginning_, before we potentially start any multithreaded runtimes.

This is already the case, which is why the feature works but the test fails. I can add some comments in relevant places to remind future maintainers of this requirement.

> 1. Add timezone config to Settings
> 3. If the config file doesn't specify a timezone, override with the value we read at program start

From my POV I don't really see a reason to make a global config for this. But maybe that's just because I'm tunnel-visioned on this single feature while you have a broader view of the whole project. Can you please clarify what other components of `atuin` plan to make use of this config (if indeed this is the case)?

> 4. For tests, specify an offset - they have a different entry point, so will never try to read anything from the env

This test already exists as `can_parse_offset_timezone_spec`. So is it okay if I just get rid of `can_parse_local_timezone_spec`?

ellie:
> This is already the case, which is why the feature works but the test fails. I can add some comments in relevant places to remind future maintainers of this requirement.

I think if we keep the timezone-reading separate to everything else, we don't need to call it in the tests - if it's going to be this difficult, anyway.

> Can you please clarify what other components of atuin plan to make use of this config (if indeed this is the case)?

I'm being asked about timezones a lot right now, essentially for any area that displays time. In the TUI, in the stats, etc. A global config means we can resolve that everywhere.

cyqsimon:
I'm working on adding timezone as an option in settings, but saw that there is already the field `local_tz`. Apparently this was added about the same time as this PR (#1477).

Currently it's marked with `#[serde(skip)]`. I'm thinking, maybe I should just promote this to a full configuration option and use it. What do you think?

cyqsimon:
Rebased.

ellie:
> I'm working on adding timezone as an option in settings, but saw that there is already the field local_tz. Apparently this was added about the same time as this PR (https://github.com/atuinsh/atuin/pull/1477).

Ahhh I'm sorry I totally blanked on that 😭 

> Currently it's marked with #[serde(skip)]. I'm thinking, maybe I should just promote this (with a rename to timezone perhaps) to a full configuration option and use it. What do you think?

Seems reasonable

> By uniting them under one configuration option (which admittedly, is more consistent, and probably is what we want to do anyway), one of them has to break. I would like to know your opinion on this.

I think for general guidance on timezones, we should follow

1. Always always _store_ UTC
2. Display either in the locally fetched timezone, or the configured timezone

atuin-client/Cargo.toml
```diff
@@ -52,6 +52,7 @@ thiserror = { workspace = true }
 futures = "0.3"
 crypto_secretbox = "0.1.1"
 generic-array = { version = "0.14", features = ["serde"] }
+serde_with = "3.5.1"
```
ellie, line 55: I wasn't aware of this crate. Nice!

atuin-client/src/settings.rs
```diff
@@ -123,6 +129,46 @@ impl From<Dialect> for interim::Dialect {
...
+        }
+
+        // offset from UTC
+        if let Ok(offset) = UtcOffset::parse(s, OFFSET_FMT) {
+            return Ok(Self(offset));
+        }
+
+        // IDEA: Currently named timezones are not supported, because the well-known crate
+        // for this is `chrono_tz`, which is not really interoperable with the datetime crate
+        // that we currently use - `time`. If ever we migrate to using `chrono`, this would
```
ellie, line 165: I wouldn't be against it, just not right now. We changed from chrono -> time not _that_ long ago, as chrono wasn't being maintained

It looks like it is being maintained once more! If that remains the case for a while, then sure

atuin/src/command/client/history.rs
```diff
@@ -68,6 +68,14 @@ pub enum Cmd {
         #[arg(action = clap::ArgAction::Set)]
         reverse: bool,
 
+        /// Display the command time in another timezone other than the configured default.
+        ///
+        /// This option takes one of the following kinds of values:
+        /// - the special value "local" (or "l") which refers to the system time zone
+        /// - an offset from UTC (e.g. "+9", "-2:30")
+        #[arg(long)]
+        tz: Option<Timezone>,
```
ellie, line 77: Could we make this `timezone` in order to match the config setting?

atuin-client/src/settings.rs
```diff
@@ -477,6 +519,7 @@ impl Settings {
             .set_default("key_path", key_path.to_str())?
             .set_default("session_path", session_path.to_str())?
             .set_default("dialect", "us")?
+            .set_default("timezone", "+0")?
```
ellie, line 522: I know this would make it more of a breaking change, but thinking forwards; I think it would be better to default to `local`

ellie:
couple of minor comments, but looks great - thank you!

atuin/src/command/client/history.rs
```diff
@@ -68,6 +68,14 @@ pub enum Cmd {
         #[arg(action = clap::ArgAction::Set)]
         reverse: bool,
 
+        /// Display the command time in another timezone other than the configured default.
+        ///
+        /// This option takes one of the following kinds of values:
+        /// - the special value "local" (or "l") which refers to the system time zone
+        /// - an offset from UTC (e.g. "+9", "-2:30")
+        #[arg(long, visible_alias = "tz")]
+        timezone: Option<Timezone>,
```
cyqsimon, line 77: I kept `--tz` as a visible alias. Hope you find this okay =)

atuin/src/command/client/history.rs
```diff
@@ -204,35 +230,41 @@ impl FormatKey for FmtHistory<'_> {
...
+                self.history
                     .hostname
                     .split_once(':')
-                    .map_or(&self.0.hostname, |(host, _)| host),
+                    .map_or("", |(_, user)| user),
             )?,
-            "user" => f.write_str(self.0.hostname.split_once(':').map_or("", |(_, user)| user))?,
             _ => return Err(FormatKeyError::UnknownKey),
         }
         Ok(())
```
cyqsimon, line 270: Lol git is confused

atuin/src/command/client/history.rs
ellie: all good!

atuin/src/command/client/history.rs
ellie: lmao thanks git

ellie:
Only a few conflicts left and I'll get it merged!

cyqsimon:
Rebased.

ellie:
I was convinced I'd merged this, sorry!

ellie (approved):

ellie:
Thank you! Really happy to get this in 🙏

huyz:
How do I enable this?  I can't find docs.  My history still shows times in UTC by default.

cyqsimon:
> How do I enable this? I can't find docs. My history still shows times in UTC by default.

Huh. You're right I did forget to update the docs. TBH I don't know where the source of [this page](https://github.com/atuinsh/atuin/blob/main/crates/atuin-client/config.toml) is.

Meanwhile you can take a look at [the example config file](https://github.com/atuinsh/atuin/blob/main/crates/atuin-client/config.toml).

huyz:
@cyqsimon Strange, no matter what I do in the config.toml, I still end up with `Command Stats` displaying in UTC:
```
Time             2025-08-06 6:59:09.938233 +00:00:00
```

cyqsimon:
@huyz Possible sources of the problem that come to mind are:

1. outdated atuin version
2. misconfigured system timezone
3. containerised or sandboxed environment
4. failure to get local timezone

So I think it's best you create a new issue with comprehensive information on your environment and exact steps to reproduce.

huyz:
@cyqsimon Someone filed the same issue I had: https://github.com/atuinsh/atuin/issues/2567

markovejnovic:
Hate to necro the thread, but worth pointing out (since I was on a wild goose chase) that [`0.3.37`](https://github.com/time-rs/time/blob/main/CHANGELOG.md#0337-2024-12-03) of the time crate supports calling `UtcOffset::current_local_offset` in MT contexts.

PR #1528, opened 2024-01-09
style(bash): use consistent coding style

akinomyoga:
This is just a cosmetic one. I originally planned to submit this after settling all the planned PRs, but some would overlap. I still have two planned PRs, but I'd like to submit this before those two.

Different coding styles and indentations are mixed in the current `atuin.bash`. In this PR, I try to normalize them. I separate the commits for each topic. These are more or less preferences, so if you would like the normalization the other way around, please specify it. I can adjust them (after the confirmation about the consequences if needed).

As usual, I left descriptions in commit messages.

### DIFF
--- a/atuin/src/shell/atuin.bash
+++ b/atuin/src/shell/atuin.bash
@@ -20,11 +20,11 @@ __atuin_preexec() {
 __atuin_precmd() {
     local EXIT=$?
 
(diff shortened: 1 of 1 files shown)


ellie (approved):
LGTM, thank you!

akinomyoga:
Thanks!

PR #1531, opened 2024-01-09
refactor: remove trailing whitespaces

azzamsa:


ellie:
Hey! Thank you for making this PR, but I'm not that keen to merge whitespace changes

1. Any changes to SQL files will change the checksum of the migration, and break installs
2. Stylistic changes should only be made when they can be enforced by tooling, imo

PR #1535, opened 2024-01-09
refactor(bash): refactor and localize `HISTORY => __atuin_output`

akinomyoga:
This is relatively trivial. I do not see the global variable `HISTORY` set by `__atuin_history` being documented. Also, the variable names are different for different shells (e.g. `output` in zsh, and `ATUIN_H` in fish), so this doesn't seem to be a part of the public interface of Atuin's integration. I'd suggest localizing those variables. Note that `ATUIN_H` in fish seems to be already local. I'm not sure how this is processed for the Nu shell integration, but I guess it would be processed without creating a variable.

I also localize `output` set by the Zsh integration. Even if we expose it on purpose, the variable name choice doesn't seem nice to avoid conflicts with the users' variables.

### DIFF
--- a/atuin/src/shell/atuin.bash
+++ b/atuin/src/shell/atuin.bash
@@ -135,25 +135,26 @@ __atuin_history() {
         fi
     fi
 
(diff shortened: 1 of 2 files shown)


ellie (approved):
🚀

PR #1541, opened 2024-01-10
docs(docker): docker-compose -> docker compose

llimllib:
the docker-compose command has been replaced with the docker compose
subcommand. Docker docs:
https://docs.docker.com/engine/reference/commandline/compose/


ellie:
Hey! Sorry to be a pain, but I'm in the process of moving the docs. Their new home is here: https://github.com/atuinsh/docs

It would be fantastic if you could open this PR there instead please? 🙏

Issue #391, opened 2022-05-06
Support vi-style up-down navigation with `j` and `k`


wtfbbqhax:
I love it, I just wish I could navigate up and down the list with `j` and `k`. 

How would this work?
-I have no clue

Okay, I lie. To feel authentic vi-style, it would probably need an "insert" mode to show the search bar and a "visual" mode to hide the bar and allow navigating with alternate up-down keys.

--

going further, a visual mode could allow paging up and down through queries.

GaetanLepage:
Yes, this would be great !

conradludgate:
Ctrl-j/k does work. But as you pointed out we don't have an insert/normal mode distinction

GaetanLepage:
Also, when using the vi-mode in zsh, pressing k in normal mode does not invoke atuin.
In insert mode, pressing the up arrow works fine.

patricksjackson:
1. Ctrl-h/j/k/l are already bound to vi equivalent navigation commands, so this issue probably should be closed unless Atuin wants to consider having a vi-mode? Having a vi-mode seems like a lot of work, since we'd have to maintain a separate vi-compatability layer, and it would have to be behind a flag since you don't want a non-vi user to accidentally get stuck in it via hitting `escape`.
2. Ctrl-j is broken for me somehow, it's interpreted the same as `enter`, even though Ctrl-n (which atuin maps to the same thing) works fine for me. I've made a note to track that down since it's probably something weird in my config.

> when using the vi-mode in zsh, pressing k in normal mode does not invoke atuin

3. This is an unrelated issue, and is because the zsh init script doesn't set the keybindings for all of the modes. That probably needs to be touched by somebody more familiar with zsh than me. I _think_ you can get it to work with `bindkey -M vicmd 'k' _atuin_search_widget`

GaetanLepage:
> 1. Ctrl-h/j/k/l are already bound to vi equivalent navigation commands, so this issue probably should be closed unless Atuin wants to consider having a vi-mode? Having a vi-mode seems like a lot of work, since we'd have to maintain a separate vi-compatability layer, and it would have to be behind a flag since you don't want a non-vi user to accidentally get stuck in it via hitting `escape`.

I agree. A vi-mode might not really be in the scope of this project.


>     2. Ctrl-j is broken for me somehow, it's interpreted the same as `enter`, even though Ctrl-n (which atuin maps to the same thing) works fine for me. I've made a note to track that down since it's probably something weird in my config.

[...]

PR #1553, opened 2024-01-12
feat(ui): vim mode

YummyOreo:
This adds a vim mode: j and k to go up and down the list and i to enter insert mode. has to be enabled in config by adding vim = true

Possible other stuff
- possibly add a color indication?
- things to manipulate the text?

### DIFF
--- a/atuin/src/command/client/search/interactive.rs
+++ b/atuin/src/command/client/search/interactive.rs
@@ -306,7 +306,9 @@ impl State {
             KeyCode::Char('i') if settings.vim && self.vim_mode == VimMode::Normal => {
                 self.vim_mode = VimMode::Insert;
             }
-            KeyCode::Char(c) if !settings.vim || self.vim_mode == VimMode::Insert => self.search.input.insert(c),
+            KeyCode::Char(c) if !settings.vim || self.vim_mode == VimMode::Insert => {
+                self.search.input.insert(c)
+            }
             KeyCode::PageDown if !settings.invert => {
                 let scroll_len = self.results_state.max_entries() - settings.scroll_context_lines;
                 self.scroll_down(scroll_len);
@@ -588,6 +590,28 @@ struct Stdout {
 }
 
 impl Stdout {
+    #[cfg(target_os = "windows")]
+    pub fn new(inline_mode: bool) -> std::io::Result<Self> {
+        terminal::enable_raw_mode()?;
+        let mut stdout = stdout();
+
+        if !inline_mode {
+            execute!(stdout, terminal::EnterAlternateScreen)?;
+        }
+
+        execute!(
+            stdout,
+            event::EnableMouseCapture,
+            event::EnableBracketedPaste,
+        )?;
+
+        Ok(Self {
+            stdout,
+            inline_mode,
+        })
+    }
+
+    #[cfg(not(target_os = "windows"))]
     pub fn new(inline_mode: bool) -> std::io::Result<Self> {
         terminal::enable_raw_mode()?;
         let mut stdout = stdout();
@@ -614,6 +638,21 @@ impl Stdout {
 }
 
 impl Drop for Stdout {
+    #[cfg(target_os = "windows")]
+    fn drop(&mut self) {
+        if !self.inline_mode {
+            execute!(self.stdout, terminal::LeaveAlternateScreen).unwrap();
+        }
+        execute!(
+            self.stdout,
+            event::DisableMouseCapture,
+            event::DisableBracketedPaste,
+        )
+        .unwrap();
+        terminal::disable_raw_mode().unwrap();
+    }
+
+    #[cfg(not(target_os = "windows"))]
     fn drop(&mut self) {
         if !self.inline_mode {
             execute!(self.stdout, terminal::LeaveAlternateScreen).unwrap();
@@ -648,7 +687,7 @@ pub async fn history(
     settings: &Settings,
     mut db: impl Database,
 ) -> Result<String> {
-    let stdout = Stdout::new(settings.inline_height > 0)?;
+    let stdout = Stdout::new(settings.inline_height > 0).unwrap();
     let backend = CrosstermBackend::new(stdout);
     let mut terminal = Terminal::with_options(
         backend,
@@ -659,7 +698,8 @@ pub async fn history(
                 Viewport::Fullscreen
             },
         },
-    ).unwrap();
+    )
+    .unwrap();
 
     let mut input = Cursor::from(query.join(" "));
     // Put the cursor at the end of the query by default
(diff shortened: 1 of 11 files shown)


ellie (dismissed):
LGTM! Thanks!

I'd love to flesh this out some more, but let's not do it here. That match is already _huge_, it would be great to handle keys in a more flexible way

ellie:
Just gonna need a format

akinomyoga:
I'm not sure if we should modify it in this PR, but as far as I understand, the setting of this PR defines the *common* behavior of all the three keymaps in shells (emacs, vi-insert, and vi-command, or similar ones) at the same time. However, it is natural to behave like emacs when Atuin is called from the emacs keymap, vi-insert when from the vi-insert keymap, and vi-normal when from the vi-normal keymap. Wouldn't it be possible to switch the behavior through a command-line option depending on the keymap that the keybinding is defined?

docs/docs/config/config.md
```diff
@@ -320,6 +320,14 @@ This technically defaults to true for new users, but false for existing. We
 have set `enter_accept = true` in the default config file. This is likely to
 change to be the default for everyone in a later release.
 
+## vim
+Default: true
```
akinomyoga, line 324: The documentation says that the vim mode is enabled by default, but the actual default setting seems to be `false`.

https://github.com/atuinsh/atuin/pull/1553/files#diff-40a913e7519b21e144292a549971f0c8c3cc3ed149336446de8a4fbed16ce07fR440

Maybe I miss something?

ellie:
> I'm not sure if we should modify it in this PR

100%, I'm happy to leave this as-is for now, as it's a super-minimal vim mode

> Wouldn't it be possible to switch the behavior through a command-line option depending on the keymap where the keybinding is defined?

Possibly. I'm not sure how I feel about the idea

1. People may not expect vim mode to be automatically enabled, so this could be a surprise. I know there are many people who use Vim, but do not use vim mode in the shell - many who don't even know this exists. Assuming someone would want the same mode everywhere may not be correct. Either way, we'd want to be able to override it
2. Our Vim integration may not be full, or may not match. It could be more jarring to switch from one vim implementation to another automatically, when they differ in details
3. Should Atuin open in insert or normal mode? Should this depend on the mode of the shell? I know you ask this question, but given that the Atuin TUI's primary purpose is _not_ editing text, I think it's not that clear cut

I'd love to explore this in the future when we have more vim parity, but for now, keeping the switch explicit is the best approach imo.

docs/docs/config/config.md
ellie: You're not, this needs updating. It will also need moving to https://github.com/atuin/docs, as this is all about to be deleted

akinomyoga:
Thank you for your thoughts!

Actually, when I adjusted the keybindings of Bash, I also had the same feeling as https://github.com/atuinsh/atuin/issues/391#issuecomment-1884888913 about the current keybinding of <kbd>k</kbd> in Bash's `vi-command` keymap. My request is simply *to make the behavior of history selection consistent with the shell's behavior*.

In my understanding of the mental model, the <kbd>up</kbd> and <kbd>k</kbd> keybindings try to somehow mimic the shell's behavior of recalling the command history. Inside the shell's vi-command mode, <kbd>k</kbd> means to move to the previous entry in the command history. However, this differs from the current behavior of `atuin search`. This PR enables to change the behavior, but the suggested config also affects the behavior in the shell's vi-insert mode. What I would like (and I believe the other would like) is to make it consistent with the shell's keybinding, i.e., <kbd>k</kbd> in vi-normal goes to the previous entry, and <kbd>k</kbd> in vi-insert inserts the letter.

Note that when I use the vi mode in Bash, I switch between vi-insert and vi-command. The current PR could be consistent for people who only use the vi-command mode and never enter the vi-inert mode of the shell, but I cannot imagine how to use the shell in that way.

Can I open an independent PR after this is merged?

> > Wouldn't it be possible to switch the behavior through a command-line option depending on the keymap where the keybinding is defined?
> 
> Possibly. I'm not sure how I feel about the idea
> 
> 1. People may not expect vim mode to be automatically enabled, so this could be a surprise.

Makes sense. Can we have a global option `vim` in addition to a command-line option to `atuin search` that tells the shell's keymap? When the global option `vim` is enabled, we can switch the behavior depending on the shell's keymap.

> I know there are many people who use Vim, but do not use vim mode in the shell - many who don't even know this exists.

I think those who don't use the vim mode in the shell wouldn't be affected when the behavior is determined by the shell's keymap because the shell's keymap doesn't become the vim mode for those people.

> 2. Our Vim integration may not be full, or may not match. It could be more jarring to switch from one vim implementation to another automatically, when they differ in details

I agree that we don't have to try to implement the real Vim mode. But I feel it's a matter independent of the current discussion.

> 3. Should Atuin open in insert or normal mode? Should this depend on the mode of the shell? I know you ask this question, but given that the Atuin TUI's primary purpose is _not_ editing text, I think it's not that clear cut

I wouldn't request to make it an editor. As far as we bind Atuin's search to <kbd>k</kbd>, pressing multiple <kbd>k</kbd> should behave similarly to the shell's binding because we typically press <kbd>k</kbd> multiple times to move around in the history if we are in the vi-normal mode of the shell. However, it shouldn't cause the history movement inside the vi-insert mode.

Another option might be to remove Atuin's keybinding to <kbd>k</kbd> from the default ones. That confuses people. People who would like it can define the keybindings by themselves. Even in that case, at least I would like to have a way to switch the behavior in the shell's vi-normal and vi-insert.

> I'd love to explore this in the future when we have more vim parity, but for now, keeping the switch explicit is the best approach imo.

ellie:
> My request is simply to make the behavior of history selection consistent with the shell's behavior.

You're welcome to PR, I'm happy with those suggestions - I wasn't disagreeing with you at all fwiw, my only concern is automatically enabling vi mode.

I'm happy for us to open in normal mode when invoked from normal, and vice-versa for insert. Just I'd rather avoid doing so unless `vim = true` is in our config.

ellie (dismissed):

atuin/src/command/client/search/interactive.rs
```diff
@@ -5,12 +5,13 @@ use std::{
...
 use atuin_common::utils;
 use crossterm::{
+    cursor::SetCursorStyle,
     event::{
         self, DisableMouseCapture, EnableMouseCapture, Event, KeyCode, KeyEvent, KeyModifiers,
         KeyboardEnhancementFlags, MouseEvent, PopKeyboardEnhancementFlags,
         PushKeyboardEnhancementFlags,
     },
-    execute, terminal,
+    execute, terminal, ExecutableCommand,
```
ellie, line 14:
```suggestion
    execute, terminal,
```

ellie (approved):

Nemo157:
> I'm happy for us to open in normal mode when invoked from normal, and vice-versa for insert.

This doesn't match my intuition for how the modes should start, I use zsh vim keybindings and I tried using this along with

```zsh
bindkey -a / _atuin_search_widget
bindkey -a k _atuin_up_search_widget
```

instead of the default bindings. Both of those are triggered from normal mode, but I would expect `k` to open atuin in normal mode while `/` opens atuin in insert mode, to match how those work in zsh's implementation.

akinomyoga:
Thanks, I forgot to mention it there.

The new PR doesn't switch the behavior based on the key, but you can suggest a detailed behavior there. I can update the PR if we come up with a reasonable interface.

PR #1557, opened 2024-01-13
fix(windows): disables unix specific stuff for windows

YummyOreo:
Windows automatically doesn't allow other users to access other's files (unless they are a admin). So a replacement for windows isn't needed.


ellie (approved):
🙏

PR #1558, opened 2024-01-14
chore: make clipboard dep optional as a feature

lilydjwg:
The cli-clipboard crate doesn't compile on systems other than Windows, Linux and MacOS, e.g. on Android it fails to compile.


ellie (approved):
thanks!

PR #1559, opened 2024-01-14
refactor(search): refactor vim mode

akinomyoga:
Before submitting the change to the configuration interface of the newly introduced vim mode, I'd like to first refactor/fix the behaviors of corner cases. I left details in the commit messages.

@YummyOreo @ellie Could you check these?

### DIFF
--- a/atuin/src/command/client/search/interactive.rs
+++ b/atuin/src/command/client/search/interactive.rs
@@ -703,7 +703,6 @@ struct Stdout {
 }
 
 impl Stdout {
-    #[cfg(target_os = "windows")]
     pub fn new(inline_mode: bool) -> std::io::Result<Self> {
         terminal::enable_raw_mode()?;
         let mut stdout = stdout();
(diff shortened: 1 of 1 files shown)


atuin/src/command/client/search/interactive.rs
```diff
@@ -718,31 +727,16 @@ impl Stdout {
...
-        if !inline_mode {
-            execute!(stdout, terminal::EnterAlternateScreen)?;
+        if cfg!(not(target_os = "windows")) {
+            execute!(
+                stdout,
+                PushKeyboardEnhancementFlags(
+                    KeyboardEnhancementFlags::DISAMBIGUATE_ESCAPE_CODES
+                        | KeyboardEnhancementFlags::REPORT_ALL_KEYS_AS_ESCAPE_CODES
+                ),
+            )?;
```
YummyOreo, line 737: Could you use `#[cfg(not(windows))]` instead `cfg!`? Not too important but removes the code on windows instead of during runtime

akinomyoga: Ah, maybe that fails to compile in Windows?

I expected it to be optimized out in Windows even if `if cfg!(...)` is used, so the code will be removed in Windows anyway. The essential difference is that the semantic analysis such as type checking is skipped with `#[cfg(not(windows))]`, so `#[cfg(not(windows))]` can be used when the code inside doesn't compile in Windows.

Anyway, I don't have a strong preference, so I can change it.

akinomyoga: amended & rebased.

YummyOreo (approved):

ellie (dismissed):

akinomyoga:
fixed format.

akinomyoga:
Sorry, I fixed the wrong part. I again fixed it. Now the checks passed.

ellie (approved):

ellie:
Thanks!

PR #1560, opened 2024-01-14
fix(search): fix invisible tab title in light background

akinomyoga:
In the current `main` branch, I cannot read the selected tab title because the background and foreground colors are the same in terminals with the light background. Here is an example (where the red part is drawn by me):

atuin-pr16-2

This patch specifies the foreground color explicitly. After the fix, it looks as follows:

atuin-pr16-1


ellie (approved):
nice catch! thanks

PR #1561, opened 2024-01-14
fix: shorten text, use ctrl-o for inspector

ellie:
1. Some shells/terminals seem to directly bind ctrl-i to be tab :( Use ctrl-o for the inspector instead
2. Shorten the help text so it doesn't get squished

### DIFF
--- a/atuin/src/command/client/search/interactive.rs
+++ b/atuin/src/command/client/search/interactive.rs
@@ -162,7 +162,7 @@ impl State {
             KeyCode::Tab => {
                 return InputAction::Accept(self.results_state.selected());
             }
-            KeyCode::Char('i') if ctrl => {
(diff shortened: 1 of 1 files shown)


atuin/src/command/client/search/interactive.rs
```diff
@@ -571,20 +571,17 @@ impl State {
                 Span::raw(": edit"),
                 Span::raw(", "),
                 Span::styled("<enter>", Style::default().add_modifier(Modifier::BOLD)),
-                Span::raw(": execute"),
+                Span::raw(": run"),
                 Span::raw(", "),
-                Span::styled("<ctrl-r>", Style::default().add_modifier(Modifier::BOLD)),
```
ellie, line 576: useful, but not crucial to go here imo. I'd like to replace this with a help popup anwyay

Issue #1489, opened 2024-01-02
fish init error


cooperspencer:
I added `atuin init fish | source` to my ~/.config/fish/config.fish, like it is described in https://atuin.sh/docs/advanced-install#fish

I get the following error:
```
- (line 21): $(...) is not supported. In fish, please use '(ATUIN_SHELL_FIS…)'.
    set -l ATUIN_H "$(ATUIN_SHELL_FISH=t ATUIN_LOG=error atuin search $argv -i -- (commandline -b) 3>&1 1>&2 2>&3)"
                    ^
from sourcing file -
	called on line 3 of file ~/.config/fish/config.fish
from sourcing file ~/.config/fish/config.fish
	called during startup
source: Error while reading file “<stdin>”
```

i use atuin version 17.1.0 and fish version 3.3.1

cooperspencer:
just added the official fish repo and updated it to the latest version. now it works.

arcuru:
For others showing up here, the minimum supported fish version is [v3.4.0](https://github.com/fish-shell/fish-shell/releases/tag/3.4.0)

PR #1563, opened 2024-01-14
Fix fish integration on older fishes

### LINKED ISSUE #1562: fish shell init failing on v3.3.1
On fish 3.3.1:

```shell
$ atuin init fish --disable-up-arrow | source
- (line 21): $(...) is not supported. In fish, please use '(ATUIN_SHELL_FIS…)'.
    set -l ATUIN_H "$(ATUIN_SHELL_FISH=t ATUIN_LOG=error atuin search $argv -i -- (commandline -b) 3>&1 1>&2 2>&3)"
                    ^
from sourcing file -
source: Error while reading file '<stdin>'
$ fish --version
fish, version 3.3.1
$ atuin --version
atuin 17.2.1
```

(same setup is working on fish 3.6.1)

mattgodbolt:
- On fish 3.3 the bash-style `$(...)` doesn't work, we should use `(...)`
- Also quoting `"(moo)"` on older fishes gives a literal `moo`
- The result of a `(subcommand)` is a single token, so no need to quote it, anyway

Tested by making the change, executing `cargo run -- init fish > atuin.fish` and then copying and source that script on a system with fish 3.3, as well as 3.6 and observing both still work.

Fixes #1562.

### DIFF
--- a/atuin/src/shell/atuin.fish
+++ b/atuin/src/shell/atuin.fish
@@ -19,11 +19,14 @@ function _atuin_postexec --on-event fish_postexec
 end
 
 function _atuin_search
-    set -l ATUIN_H (ATUIN_SHELL_FISH=t ATUIN_LOG=error atuin search $argv -i -- (commandline -b) 3>&1 1>&2 2>&3)
+    # In fish 3.4 and above we can use `"$(some command)"` to keep multiple lines separate;
+    # but to support fish 3.3 we need to use `(some command | string collect)`.
+    # https://fishshell.com/docs/current/relnotes.html#id24 (fish 3.4 "Notable improvements and fixes")
+    set -l ATUIN_H (ATUIN_SHELL_FISH=t ATUIN_LOG=error atuin search $argv -i -- (commandline -b) 3>&1 1>&2 2>&3 | string collect)
 
     if test -n "$ATUIN_H"
         if string match --quiet '__atuin_accept__:*' "$ATUIN_H"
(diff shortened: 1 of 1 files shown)


mattgodbolt:
I couldn't think of a decent way of testing this, else I'd've tried to add a test to lock in the good behaviour.

ellie:
Ahhh I see

So this was introduced intentionally in this PR, to fix multiline commands being squished onto one line: https://github.com/atuinsh/atuin/pull/1418

If there's no other way to workaround, we may have to switch on Fish version and drop multiline support for 3.3

mattgodbolt:
Oh darn it :) I'll see what I can do on my side. Upgrading fish is not the easiest; in the server I'm using we're on the 22.04LTS version.

mattgodbolt:
This does seem challening: https://fishshell.com/docs/current/language.html#expand-command-substitution
I'll see if I can make this work in all cases, else I'll update this to check the fish version, if that's indeed feasible too.

mattgodbolt:
Now rephrased in a way that works in both fish 3.3.x and 3.6.x. The release notes for fish say that the new syntax `"$(command)"` is identical to `(command | string collect)` and so that's what I now do.

Tested again locally on both 3.3 and 3.6.

Please take another looks

ellie (approved):

ellie:
Lovely! Thank you so much

arcuru (approved):
A similar report came in #1489, I hadn't gotten around to making this PR yet. Thanks :)

PR #1570, opened 2024-01-15
feat(search): introduce keymap-dependent vim-mode

akinomyoga:
This is a possible refinement to the vim mode as suggested in https://github.com/atuinsh/atuin/pull/1553#issuecomment-1890426150.

In this PR, we introduce three keymap modes, `emacs`, `vim-insert`, and `vim-normal` (which was extended from the current `VimMode`). When the config `vim` is enabled, the keys are processed with special rules for the current keymap mode. This is actually the same as the current `main`. The real behavioral change is related to the following config `keymap_mode`:

The keymap mode with which the Atuin search is started is controlled by the config `keymap_mode`. When `keymap_mode` is not specified, the keymap mode on the startup is determined based on the shell's keymap that started the Atuin search. When the config `keymap_mode` is specified, the keymap mode on the startup is forced to be the specified one.

@YummyOreo @ellie What do you think of this interface? If you have another idea, I can adjust the PR.

### DIFF
--- a/atuin-client/config.toml
+++ b/atuin-client/config.toml
@@ -126,15 +126,13 @@
 enter_accept = true
 
 
-## Defaults to false. If enabled, in the vim-normal keymap, you may use 'j' and
-## 'k' to navigate the history list and 'i' to enter Insert mode.
-# vim = false
-
-## The keymap on the startup of `atuin search`.  If this is not set, the keymap
-## on the startup is automatically selected based on the shell's keymap where
-## the keybinding is defined.  Note: To enable the special behavior of the vim
-## keymaps, the setting `vim` needs to be set to `true`.
-# keymap_mode = "vim-normal"
+## Defaults to "emacs".  This specifies the keymap on the startup of `atuin
+## search`.  If this is set to "auto", the startup keymap mode in the Atuin
+## search is automatically selected based on the shell's keymap where the
+## keybinding is defined.  If this is set to "emacs", "vim-insert", or
+## "vim-normal", the startup keymap mode in the Atuin search is forced to be
+## the specified one.
+# keymap_mode = "auto"
 
 #[stats]
 # Set commands where we should consider the subcommand for statistics. Eg, kubectl get vs just kubectl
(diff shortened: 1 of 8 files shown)


Nemo157:
Nice, that works about how I'd expect. I wonder if the generated key-bindings should be changed based on vim-mode too. Currently the `^r` binding overrides the "redo" binding from zsh's normal-mode, and `k`/`/` seem much more familiar bindings to use. I also assume these widgets are internal API so I shouldn't really be binding to them directly.

akinomyoga:
I thought about it again, and I think another option for the configuration interface would be to provide a single configuration `keymap_mode = { "emacs" (default) | "vim-insert" | "vim-normal" | "auto" }` instead of `vim` (i.e. remove `vim`). By making `emacs` default, it doesn't change the behavior for the existing users. When the user wants the automatic determination of the startup keymap mode, the user can specify `keymap_mode = "auto"`. When the user wants to always start the Atuin search with the vim-normal mode, the user can specify `keymap_mode = "vim-normal"`.

akinomyoga:
> I thought about it again, and I think another option for the configuration interface would be

I implemented in the latest commit 96eeca0. I feel this is cleaner, though we have discussed the global config `vim` plus another config in #1553. The default value is `emacs` to keep the behavior for the existing users unchanged. The users who would like the Atuin search to always start with the vim-normal mode can specify `keymap_mode = "vim-normal"` instead of the current way `vim = true`. The users who would like the automatic determination of the startup keymap of the Atuin search can specify `keymap_mode = "auto"`.

ellie (approved):
Looks good to me, thank you!

Would you also be able to update the [docs](https://github.com/atuinsh/docs) please?

ellie:
ah and the shellcheck

akinomyoga:
Thank you! I've fixed SC2048, rebased on top of the latest `main`, and squashed trivial commits.

akinomyoga:
I opened https://github.com/atuinsh/docs/pull/8 in atuin/docs.

PR #1574, opened 2024-01-15
feat(ui): When in vim-normal mode apply an alternative highlighting to the selected line

Nemo157:
This makes it much more obvious whether you're in normal or insert mode.

image

### DIFF
--- a/atuin/src/command/client/search/history_list.rs
+++ b/atuin/src/command/client/search/history_list.rs
@@ -16,6 +16,8 @@ pub struct HistoryList<'a> {
     history: &'a [History],
     block: Option<Block<'a>>,
     inverted: bool,
+    /// Apply an alternative highlighting to the selected row
+    alternate_highlight: bool,
 }
 
 #[derive(Default)]
@@ -65,6 +67,7 @@ impl<'a> StatefulWidget for HistoryList<'a> {
             y: 0,
             state,
             inverted: self.inverted,
+            alternate_highlight: self.alternate_highlight,
         };
 
         for item in self.history.iter().skip(state.offset).take(end - start) {
@@ -81,11 +84,12 @@ impl<'a> StatefulWidget for HistoryList<'a> {
 }
 
 impl<'a> HistoryList<'a> {
-    pub fn new(history: &'a [History], inverted: bool) -> Self {
+    pub fn new(history: &'a [History], inverted: bool, alternate_highlight: bool) -> Self {
         Self {
             history,
             block: None,
             inverted,
+            alternate_highlight,
         }
     }
 
@@ -116,11 +120,14 @@ struct DrawState<'a> {
     y: u16,
     state: &'a ListState,
     inverted: bool,
+    alternate_highlight: bool,
 }
 
 // longest line prefix I could come up with
 #[allow(clippy::cast_possible_truncation)] // we know that this is <65536 length
 pub const PREFIX_LENGTH: u16 = " > 123ms 59s ago".len() as u16;
+static SPACES: &str = "                ";
+static _ASSERT: () = assert!(SPACES.len() == PREFIX_LENGTH as usize);
 
 impl DrawState<'_> {
     fn index(&mut self) {
@@ -157,7 +164,10 @@ impl DrawState<'_> {
         let time = format_duration(since.try_into().unwrap_or_default());
 
         // pad the time a little bit before we write. this aligns things nicely
-        self.x = PREFIX_LENGTH - 4 - time.len() as u16;
+        self.draw(
+            &SPACES[..((PREFIX_LENGTH - self.x) as usize - 4 - time.len())],
+            Style::default(),
+        );
 
         self.draw(&time, style);
         self.draw(" ago", style);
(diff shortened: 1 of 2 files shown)


akinomyoga (commented):
Maybe I'm asking a nonsense question, but how are the cursor movements by such as <kbd>left</kbd>, <kbd>right</kbd>, <kbd>home</kbd>, and <kbd>end</kbd> are presented when we place the cursor in the selected line? Even in the vim-normal mode, we support the cursor movements by those keys in the search text. In addition, in the future, we might support the minimal cursor movements of the vim-normal style such as <kbd>h</kbd>, <kbd>l</kbd>, <kbd>0</kbd>, and <kbd>$</kbd>. In all of these cases, it seems non-trivial how to present the cursor movements if we place the cursor in the selected line.

atuin/src/command/client/search/interactive.rs
```diff
@@ -546,12 +546,20 @@ impl State {
 
         let extra_width = UnicodeWidthStr::width(self.search.input.substring());
 
-        let cursor_offset = if compact { 0 } else { 1 };
-        f.set_cursor(
-            // Put cursor past the end of the input text
-            input_chunk.x + extra_width as u16 + PREFIX_LENGTH + 1 + cursor_offset,
-            input_chunk.y + cursor_offset,
-        );
+        if self.vim_mode == VimMode::Normal {
```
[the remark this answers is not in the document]
Nemo157: Ah whoops yeah, I based this on #1570 at first, which I assumed normalized everything into `keymap_mode` (looking back at it now, it still has all the `settings.vim` checks, it seems like that should be pre-applied by forcing `keymap_mode = KeymapMode::Emacs` when it's disabled).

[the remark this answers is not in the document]
akinomyoga: Ah, yes. I'll modify #1570 in that direction. Thank you for your feedback.

Nemo157:
Mmm, yeah, that's a good point if there's actually going to be normal-mode editing of the input line itself. In that case I'll try and come up with a decent looking marker to add to the input line.

atuin/src/command/client/search/interactive.rs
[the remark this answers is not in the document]
akinomyoga: I updated #1570.

akinomyoga:
Or maybe we can keep the cursor position in the input line, but instead highlight the selected line to make it stand out more than the input line. E.g., by switching the following line to `style = style.reversed()` when the current mode is `vim-normal` (this is just a quick guess. I haven't checked if this is actually the part that should be modified).

https://github.com/atuinsh/atuin/blob/c2439d1ed6fc3e973371b76637c08eea0a8f60c8/atuin/src/command/client/search/history_list.rs#L167-L169

akinomyoga:
> the symbols aren't great but they're the only ascii ones I could find being mentioned related to identifying bash's key mode,

`:` seems confusing with the ex command mode in vi/vim. Fish uses `[I]` vs `[N]`. Bash uses `(ins)` and `(cmd)` by default when enabled by the setting `bind 'set show-mode-in-prompt on'`.

> The reverse highlight is very intense, especially the red.

I prefer the third picture, where the red color is removed. That is actually what I initially thought in my previous reply.

ellie:
Agreed with @akinomyoga - the +/: is confusing, and the third screenshot looks better for sure

Nemo157:
Updated it to just apply the highlighting from the third screenshot, without the symbols.

akinomyoga (approved):

atuin/src/command/client/search/interactive.rs
```diff
@@ -606,8 +606,13 @@ impl State {
         stats
     }
 
-    fn build_results_list(style: StyleState, results: &[History]) -> HistoryList {
-        let results_list = HistoryList::new(results, style.invert);
+    fn build_results_list<'a>(
+        style: StyleState,
+        results: &'a [History],
+        keymap_mode: KeymapMode,
+    ) -> HistoryList<'a> {
```
ellie, line 613:
```suggestion
    fn build_results_list(
        style: StyleState,
        results: &[History],
        keymap_mode: KeymapMode,
    ) -> HistoryList<'_> {
```

The lifetime here can be elided

Nemo157: Fixed it, left in the `<'_>` elided lifetime marker as would be required by the `elided_lifetimes_in_paths` lint as it makes it obvious that there's a lifetime dependency here.

ellie (approved):
Great, lgtm! Thanks for the work here 🙏

atuin/src/command/client/search/history_list.rs
```diff
@@ -157,20 +164,25 @@ impl DrawState<'_> {
         let time = format_duration(since.try_into().unwrap_or_default());
 
         // pad the time a little bit before we write. this aligns things nicely
-        self.x = PREFIX_LENGTH - 4 - time.len() as u16;
+        self.draw(
+            &SPACES[..((PREFIX_LENGTH - self.x) as usize - 4 - time.len())],
```
akinomyoga, line 168: I hit a regression while testing the `main` branch for another fix. This line produces the following error

```
thread 'main' panicked at atuin/src/command/client/search/history_list.rs:168:20:
byte index 18446744073709551615 is out of bounds of `                `
```

when the <kbd>up</kbd> arrow is pressed immediately after running the previous command in Zsh.

@Nemo157 I guess you know what is happening and how to fix it. Could you take a look at it?

Nemo157: Yep, I can add some limits here so it will just not pad if it would overflow.

akinomyoga: Thanks!

PR #1576, opened 2024-01-15
Stop control characters being printed to terminal

tpoliaw:
Detail of change is in the commit message but easiest steps to reproduce is to open a new terminal and run the following

```sh
echo one
echo '^[[31mtwo' # with `^[` entered as literal using ctrl-v + ctrl-[
echo three
atuin history list -s
```

With the previous behaviour, the control sequence would not be visible (and therefore not copyable) and everything after the second 'echo' would be red.

Running `atuin search -i --filter-mode session` afterwards would display the `echo two` command as `echo '[31mtwo'`. Selecting it would print `echo 'two'` without the control sequence and two onwards in red.

### DIFF
--- a/atuin-common/src/utils.rs
+++ b/atuin-common/src/utils.rs
@@ -1,4 +1,6 @@
+use std::borrow::Cow;
 use std::env;
+use std::fmt::Write;
 use std::path::PathBuf;
 
 use rand::RngCore;
@@ -100,6 +102,34 @@ pub fn is_bash() -> bool {
     env::var("ATUIN_SHELL_BASH").is_ok()
 }
 
+/// Extension trait for anything that can behave like a string to make it easy to escape control
+/// characters.
+///
+/// Intended to help prevent control characters being printed and interpreted by the terminal when
+/// printing history as well as to ensure the commands that appear in the interactive search
+/// reflect the actual command run rather than just the printable characters.
+pub trait Escapable: AsRef<str> {
+    fn escape_control(&self) -> Cow<str> {
+        if !self.as_ref().contains(|c: char| c.is_ascii_control()) {
+            self.as_ref().into()
+        } else {
+            let mut remaining = self.as_ref();
+            // Not a perfect way to reserve space but should reduce the allocations
+            let mut buf = String::with_capacity(remaining.as_bytes().len());
+            while let Some(i) = remaining.find(|c: char| c.is_ascii_control()) {
+                // safe to index with `..i`, `i` and `i+1..` as part[i] is a single byte ascii char
+                buf.push_str(&remaining[..i]);
+                let _ = write!(&mut buf, "\\x{:02x}", remaining.as_bytes()[i]);
+                remaining = &remaining[i + 1..];
+            }
+            buf.push_str(remaining);
+            buf.into()
+        }
+    }
+}
+
+impl<T: AsRef<str>> Escapable for T {}
+
 #[cfg(test)]
 mod tests {
     use time::Month;
@@ -182,5 +212,35 @@ mod tests {
         }
 
         assert_eq!(uuids.len(), how_many);
+    }
+
+    #[test]
+    fn escape_control_characters() {
+        use super::Escapable;
+        // CSI colour sequence
+        assert_eq!("\x1b[31mfoo".escape_control(), "\\x1b[31mfoo");
+
+        // Tabs count as control chars
+        assert_eq!("foo\tbar".escape_control(), "foo\\x09bar");
+
+        // space is in control char range but should be excluded
+        assert_eq!("two words".escape_control(), "two words");
+
+        // unicode multi-byte characters
+        let s = "🐢\x1b[32m🦀";
+        assert_eq!(s.escape_control(), s.replace("\x1b", "\\x1b"));
+    }
+
+    #[test]
+    fn escape_no_control_characters() {
+        use super::Escapable as _;
+        assert!(matches!(
+            "no control characters".escape_control(),
+            Cow::Borrowed(_)
+        ));
+        assert!(matches!(
+            "with \x1b[31mcontrol\x1b[0m characters".escape_control(),
+            Cow::Owned(_)
+        ));
     }
 }
--- a/atuin/src/command/client/history.rs
+++ b/atuin/src/command/client/history.rs
@@ -5,7 +5,7 @@ use std::{
     time::Duration,
 };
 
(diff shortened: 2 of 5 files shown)


atuin-common/src/utils.rs
```diff
@@ -100,6 +102,34 @@ pub fn is_bash() -> bool {
...
+        if !self.as_ref().contains(|c: char| c.is_ascii_control()) {
+            self.as_ref().into()
+        } else {
+            let mut part = self.as_ref();
+            // Not a perfect way to reserve space but should reduce the allocations
+            let mut buf = String::with_capacity(part.as_bytes().len());
+            while let Some(i) = part.find(|c: char| c.is_ascii_control()) {
+                // safe to index with `..i`, `i` and `i+1..` as part[i] is a single byte ascii char
+                buf.push_str(&part[..i]);
+                let _ = write!(&mut buf, "\\x{:02x}", part.as_bytes()[i]);
```
akinomyoga, line 122: This might be just a preference, but is there a possibility to use the caret notation instead of the hex notation `\xHH`? I'm more familiar with the caret notation in the command line and thus the command strings (as the shells' line editors show the control characters in the caret notation). (Note: I'm not a maintainer, so I don't have any power. If you don't like this, you can ignore it.)

```suggestion
                buf.push('^');
                buf.push(match part.as_bytes()[i] {
                    0x7F => '?',
                    code => char::from_u32(u32::from(code) + 64).unwrap(),
                });
```

akinomyoga (approved):
This fix is also good for the multiline commands. Previously, lines were connected without any separators in the history list, so the structural information of the multiline commands was lost (or at least, was harder to grasp at a glance). Now the lines are separated by the escaped newlines, and these are visible. This is nice.

atuin-common/src/utils.rs
ellie: +1 here

ellie:
Generally looks good to me! Thank you 🙏 

If we could get a `cargo fmt` for CI please

tpoliaw:
(it looked like other PRs had used the force push approach so I copied them)

atuin-common/src/utils.rs
akinomyoga: > but the main problem with using `^[` etc is that you can't copy/paste the output to re-run the command as they'll be interpreted as two characters instead of one.

Ah, OK. I now understand the third paragraph of the cover of this PR. Maybe, is `echo '[31mtwo'` in the cover a typo of `echo '\x1b[31mtwo'`?

But then, I'm afraid to say that the hex notation still has the same problem for copying and pasting as the caret notation. The raised example of `echo` works with the hex notation because of the specific behavior of a command of a specific shell (i.e., echo in Bash and Fish doesn't work that way). The hex notation is not handled at the shell level, so we cannot expect it to work in general. I think there are even opposite cases, where the caret notation works while the hex notation fails. For example, in Zsh, `bindkey ^R some_widget` works, but `bind \x0D some_widget` doesn't work as expected.

In the first place, the users can also copy and paste the commands in the terminal's scrollback buffer, where the control characters in the commands are presented with the caret notation. I'm not sure if using the hex notation in Atuin would be actually useful in solving the essential issue. If a user wants to include a raw control character in commands, I feel the user finally needs to learn that copying and pasting the displayed command doesn't work in general.

ellie:
> Ellie, how do you prefer to handle updating PRs, push new noisy commits and squash at the end (assuming it's accepted of course), or update the same commit and force push to the branch? (or something else?)

I'll update the contributing guide to clarify, but as we squash-and-merge I'm happy for people to do whatever they wish within the PR itself. It's much easier that way vs policing how people build up their work

ellie (approved):
Excellent! Thank you for working on this 🙏 

atuin-common/src/utils.rs
tpoliaw: I think I'm coming round to this idea more now. It's less convenient for the cases I had that originally come across but I can see the `^[` version is more general and better matches the way the terminal displays control chars. As Ellie initially `+1`ed your original comment I'll change it now. Is it worth inverting the colours for escape characters (similar to how zsh does it) to distinguish it from the literal `^` + `[`?

akinomyoga: > Is it worth inverting the colours for escape characters (similar to how zsh does it) to distinguish it from the literal `^` + `[`?

Yeah, that sounds better. However, we might need adjustments for the ongoing PR #1574, which uses the terminal's reverse video for another meaning. In principle, we can treat *(reverse + reverse)* as *not-reverse*, but the terminal control sequences do not offer such a model. Someone needs to implement it in the application layer, but I'm not sure if that is already implemented in any of the Rust libraries that Atuin currently depends on.

tpoliaw: Ok, I'll leave the reverse for now, it can be figured out whenever 1574 is merged.

akinomyoga: > > It could be possible to use caret notation for the interactive search but the hex version when things are printed to stdout but I'm not sure the inconsistency would be worth it.
> 
> I think we usually switch the output format depending on whether `stdout` (or the destination stream) is connected to a TTY. We can escape the control characters when `stdout` is connected to TTY, or otherwise, output the raw command string.

Do you have a plan to fix this issue? The current `main` branch doesn't work for the multiline command:

image

`\x0a` is not treated as a newline of the shell syntax.

tpoliaw: Then yes, looks like I need to. Would it be better to add the tty check or should the search result always be printed as is? Or a flag could switch between the two (although I'd accept if there's a reluctance to add more flags to search)?

akinomyoga: Hmm, I leave it to you. It's similar to the coloring of the output of the commands. Some commands do not support colors at all. Some commands enable colors with the check for TTY through `--color=auto`. Some commands perform the TTY switching by default. All are understandable.

tpoliaw: #1585 and #1586 for these two changes.

ellie: are they both ok for review? I see they're marked as drafts

tpoliaw: Sure, I just figured I should test these changes more thoroughly than I did for this one and didn't have time when I submitted them. Given main is currently broken though I thought it was worth getting something in.

tpoliaw: Thanks, I'll have a look at what changed. And I will do but it might not be for a while.

PR #1580, opened 2024-01-16
feat: make deleting from the UI work with record store sync

ellie:
Yes I know I'm adding more coupling 😭 

Plans are; 

1. Get this working fine with history
2. When introducing new stores (less legacy, less tangled), figure out a nice design
3. Port history onto new design


ellie:
All of this can be enabled with

```
[sync]
records = true
```

In config. Once enabled, please don't disable again. The old sync may not react well, and old deleted records may "magically" re-appear (they'll only be deleted in the new store, for now)

ellie:
will give a once over and merge in the morning, I'm 😴

PR #1585, opened 2024-01-17
Escape control characters using caret notation

tpoliaw:
Instead of using their equivalent hex codes.


akinomyoga:
I think `#[test]` cases should also be updated.

ellie (approved):

PR #1586, opened 2024-01-17
Print literal control characters to non terminals

tpoliaw:
Previous 'fix' to prevent control sequences being interpreted when they
shouldn't have been also prevented them being used when they should have
been. This checks if the output is to a terminal (where control
sequences shouldn't be interpreted) before escaping control characters.

### DIFF
--- a/atuin/src/command/client/search.rs
+++ b/atuin/src/command/client/search.rs
@@ -1,3 +1,5 @@
+use std::io::{stderr, IsTerminal as _};
+
 use atuin_common::utils::{self, Escapable as _};
 use clap::Parser;
 use eyre::Result;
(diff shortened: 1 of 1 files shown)


tpoliaw:
I noticed while testing this that I'd missed the command preview view. I have a fix for that in its own branch [here](https://github.com/tpoliaw/atuin/tree/preview_controls), would it be better to merge all these related PRs into one 'fix the other commit' PR rather than spreading them out?

atuin/src/command/client/search.rs
```diff
@@ -167,7 +169,11 @@ impl Cmd {
 
         if self.interactive {
             let item = interactive::history(&self.query, settings, db, &history_store).await?;
-            eprintln!("{}", item.escape_control());
+            if stderr().is_terminal() {
+                eprintln!("{}", item.escape_control());
+            } else {
+                eprintln!("{}", item);
```
ellie, line 175:
```suggestion
                eprintln!("{item}");
```

ellie (approved):

PR #1589, opened 2024-01-18
Use existing db querying for history list

tpoliaw:
When printing the history list with either the session or cwd filter
enabled, use to same query method as without either to ensure that the
other options (hide deleted entries etc) are respected.

### DIFF
--- a/atuin/src/command/client/history.rs
+++ b/atuin/src/command/client/history.rs
@@ -1,5 +1,4 @@
 use std::{
-    env,
     fmt::{self, Display},
     io::{self, Write},
     time::Duration,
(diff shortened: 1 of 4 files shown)


atuin/src/command/client/history.rs
```diff
@@ -349,37 +351,16 @@ impl Cmd {
...
-        };
-        let cwd = if cwd {
-            Some(utils::get_current_dir())
-        } else {
-            None
+        let filters = match (session, cwd) {
+            (true, true) => [Session, Directory],
+            (true, false) => [Session, Global],
+            (false, true) => [Global, Directory],
+            (false, false) => [settings.filter_mode, Global],
```
tpoliaw, line 358: I don't really like this. Is there a clever way to allow passing either `&[one_thing]` or `&[two, things]` to a function that accepts `&[T]` without creating a vec?

ellie: Not really - the slice is a reference to something, so the "something" needs to exist. A vec would be fine imo

ellie (approved):

PR #1590, opened 2024-01-19
fix: add acquire timeout to sqlite database connection

### LINKED ISSUE #1503: atuin makes the shell unresponsive under high I/O pressure
Hi!
Under high I/O pressure on my system, atuin makes my shell hang for a long period, then fail with this error:
```
Error: pool timed out while waiting for an open connection

Location:
    atuin/src/command/client.rs:50:22
```
or
```
Error: error returned from database: (code: 5) database is locked

Caused by:
    (code: 5) database is locked

Location:
    atuin/src/command/client/history.rs:199:17
```
I'd love to see a `register_timeout` in milliseconds, similar to `network_timeout`. So when `atuin` would register a new entry in history,  if after `register_timeout` it still hasn't successfully registered the entry, it aborts the transaction and quits.

At least for me, better to have a few entries discarded than having a completely unresponsive shell under high IO/pressure.

What do you think?

ellie:
This should fix #1503

I wasn't able to trigger enough IO pressure for the SQL connection to be a problem.

This adds `local_timeout` to the client config. This is a float, and represents the number of seconds (units in line with the other timeouts, though those are ints). Users may well want to reduce this if they regularly have issues, but by default I think 2s is fine and avoids a non-responsive system in bad situations.

### DIFF
--- a/atuin-client/config.toml
+++ b/atuin-client/config.toml
@@ -134,6 +134,12 @@ enter_accept = true
 ## the specified one.
 # keymap_mode = "auto"
 
+# network_connect_timeout = 5
+# network_timeout = 5
+
+## Timeout (in seconds) for acquiring a local database connection (sqlite)
+# local_timeout = 5
+
 #[stats]
 # Set commands where we should consider the subcommand for statistics. Eg, kubectl get vs just kubectl
 #common_subcommands = [
--- a/atuin-client/src/database.rs
+++ b/atuin-client/src/database.rs
@@ -2,6 +2,7 @@ use std::{
     env,
     path::{Path, PathBuf},
     str::FromStr,
+    time::Duration,
 };
 
 use async_trait::async_trait;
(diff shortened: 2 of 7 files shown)


ellie:
Oh also I am still planning on adding an Atuin daemon, which should totally avoid the issue referenced above. Just

1. That's a MUCH larger change
2. We'd probably want timeouts there too anyway

happysalada:
if you have a moment to make a new release, I'd be happy to update the version on nixos and make another test.
Nixos on rebuilds does a lot of IO, so it will be very easy to check that the problem didn't hide another problem.
Finally, thank you for making this and for all the time you spent on this project!

ppom0:
Thanks a lot @ellie, it seems great! I'll wait for another release, but no stress about this ofc.

My use case is the same as happysalada, I'm using NixOS too, and its *rebuilds* are very IO intensive.

PR #1593, opened 2024-01-19
fix: Only escape control characters when writing to terminal

tpoliaw:
The latest in the control characters saga. I opted for the nested enum option as
it made accessing the history fields easier in the FormatKey implementation.


ellie:
looks good! just the one clippy error and then ready to merge. thanks!

ellie (approved):

ellie:
Thank you!

PR #1595, opened 2024-01-21
feat(search): make cursor style configurable

akinomyoga:
The vim mode introduced in #1553 changes the terminal's cursor style on a mode change, but the current implementation has the following issues.

* The terminal's cursor style set by the Atuin search remains even after Atuin exits.  This causes an inconsistency with the shell's settings for the cursor style.
* Also, the cursor style for each keymap mode is currently hardcoded in the source code, which is not necessarily consistent with the user's cursor-style settings in the shell.
* Since the current implementation does not set the cursor style for the initial keymap mode but only sets the cursor style when the keymap mode is changed, it also causes inconsistency in the cursor style and the actual keymap when the shell's keymap and Atuin's initial keymap mode are different.

This patch solves those issues by introducing an opt-in configuration variable `keymap_cursor`.  By default, the vim mode does not change the cursor style because there is no way to determine the cursor style consistent with the shell settings automatically.  We enable the feature only when the user specifies the preferred cursor style in each mode in their config.  Also, the cursor style is set on the startup of the Atuin search (based on the initial keymap mode) and is reset on the termination of the Atuin search (based on the shell's keymap mode that started the Atuin search).

**edit**: let me ping @YummyOreo, who introduced the feature

### DIFF
--- a/atuin-client/Cargo.toml
+++ b/atuin-client/Cargo.toml
@@ -52,6 +52,7 @@ thiserror = { workspace = true }
 futures = "0.3"
 crypto_secretbox = "0.1.1"
 generic-array = { version = "0.14", features = ["serde"] }
+crossterm = { workspace = true }
 
 # encryption
 rusty_paseto = { version = "0.6.0", default-features = false }
--- a/atuin-client/Cargo.toml
+++ b/atuin-client/Cargo.toml
@@ -52,7 +52,6 @@ thiserror = { workspace = true }
 futures = "0.3"
 crypto_secretbox = "0.1.1"
 generic-array = { version = "0.14", features = ["serde"] }
-crossterm = { workspace = true }
 
 # encryption
 rusty_paseto = { version = "0.6.0", default-features = false }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -214,7 +214,6 @@ dependencies = [
  "base64 0.21.5",
  "clap",
  "config",
- "crossterm",
  "crypto_secretbox",
  "directories",
  "eyre",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -214,6 +214,7 @@ dependencies = [
  "base64 0.21.5",
  "clap",
  "config",
+ "crossterm",
  "crypto_secretbox",
  "directories",
  "eyre",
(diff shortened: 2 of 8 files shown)


atuin-client/Cargo.toml
```diff
@@ -52,6 +52,7 @@ thiserror = { workspace = true }
 futures = "0.3"
 crypto_secretbox = "0.1.1"
 generic-array = { version = "0.14", features = ["serde"] }
+crossterm = { workspace = true }
```
ellie, line 55: I'd really like to avoid adding the crossterm dependency here. Where possible, keeping UI/interaction style dependencies to the `atuin` crate is desireable. Given that it's only needed for the `SetCursorStyle` enum, would it be possible to remove this?

akinomyoga: OK. It's possible by giving up the conversion through the `From<>` trait. I'll update it.

akinomyoga: In commit 15cbded2e0831c6aaa3d479d8ba0a91ce7501e50, I moved the conversion function into `atuin` and removed the dependency from `atuin-client`. Could you check it?

ellie: looks good, thanks!

ellie (approved):

akinomyoga:
Thank you!

ellie:
I was originally planning on doing so, but if you'd like to take a look that would be fantastic!

My plan was

1. Bring out a map of `KeyCode -> Action`. Actions would be `InsertKey(char)`, `MoveTo`, `Delete`... etc.
2. Refactor the UI so that it passes the keycode to (1), then handles the resulting action. Probs break it down to vim/emacs/etc
3. Once this is there, make it configurable like you say. It would be good to be able to express "ctrl-p: moveup(1)" or similar.

But feel free to do this however you think is best. Probably worth opening something to track the work

akinomyoga:
Thank you for your thoughts! I was afraid of some conflicts. I'll submit a PR, but that PR doesn't actually change the current structure. I just sort and factorize some `match` branches.

akinomyoga:
I created a PR at #1606.

PR #1601, opened 2024-01-21
fix: report alternate keys so shift still works

conradludgate:
Kitty terminal was sending `shift-a` as `shift` + `a` and not `A`, and we have no special handling of shift, so atuin was only acknowledging lowercase `a`.

https://sw.kovidgoyal.net/kitty/keyboard-protocol/#report-alternate-keys fixes it for me, although I don't know if this has any other side-effects


ellie (approved):
seems reasonable, thanks!

Issue #68, opened 2021-05-08
integrate with fzf


neerajbadlani:
Unfortunately , I tried tui search , but still want to use fzf history search . 
I am trying to integrate with fzf using history list , it throws following error .
Can u please help ? 
Thanks 

[master:.configfiles] atuin history list
Error: environment variable not found

Location:
    src/command/history.rs:151:56

conradludgate:
Yeah that seems to be a bug. It's looking for the `ATUIN_SESSION` environment variable, even though its not needed in your example

ellie:
Yup the bug is with the bash script! The `ATUIN_SESSION` variable is actually needed for history to be properly recorded though, so I'll get the bash plugin fixed. 

As well as that, you'll probably want a nice way of listing *only* the command for `fzf`, so that'll need sorting too

ellie:
Hey @neerajbadlani!

Once #89 merges, you should be able to do something like this

```
atuin h l --cmd-only | fzf
```

neerajbadlani:
Thanks @ellie , above works and is close to what i want . However I see duplicate entries in history . 

Also , new to rust ecosystem , since brew install --HEAD dint work , so had to do manual installation . 
How do i tell rust to install in specific directory ?
```
[main:atuin] cargo install --root $(brew --prefix)
error: Using `cargo install` to install the binaries for the package in current working directory is no longer supported, use `cargo install --path .` instead. Use `cargo build` if you want to simply build the package.
```

conradludgate:
`cargo install --path . --root $(...)` should do the trick

neerajbadlani:
Thanks , I am able to integrate it with fzf properly now , exactly similar to what fzf provides , but with machine machine and session sync . 
2 things remaining : 1 . Duplicate Entries 2 . Sorted Histories ( Recent , Frequency , .. ) 
Should I close this and open another issue for those ?

thedrow:
Maybe we should ask fzf's maintainers to detect if atuin is in the path and use that instead of history.

Dentrax:
I'm looking forward to this feature! @ellie  Each time I write a command and press the <kbd>UP</kbd> key to auto-complete current cmd, TUI appears and I had a bit distracting DX. It would be great if we are able to use it with `fzf` feature enabled only. 😇 

i.e. something like the following, but `atuin` powered. :)

[...]

PR #1602, opened 2024-01-21
Add a count_id formatting option to search command

alcroito:
The {count_id} formatting token will output a counter incremented by
one for each search result, starting with 1.

This allows creating a formatting style similar to the output of the
bash builtin 'history' command.

### DIFF
--- a/atuin/src/command/client/history.rs
+++ b/atuin/src/command/client/history.rs
@@ -68,7 +68,7 @@ pub enum Cmd {
         #[arg(action = clap::ArgAction::Set)]
         reverse: bool,
 
-        /// Available variables: {command}, {directory}, {duration}, {user}, {host}, {exit} and {time}.
+        /// Available variables: {command}, {directory}, {duration}, {user}, {host}, {exit}, {count_id} and {time}.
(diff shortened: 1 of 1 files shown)


alcroito:
Hi.
I'm open to suggestions on renaming the formatting option to something better than `count_id`

ellie:
Hey!

I'm not too sure if this is necessary. It's pretty easy to add line numbers with standard unix tooling

EG

```
# using "number lines"
atuin history list | nl

# (ab)using grep
atuin history list | grep -n '^'

# there's probably an awk/sed/etc thing too
```

Could you let me know a bit more about your use case please?

alcroito:
> Could you let me know a bit more about your use case please?

Sure. 
I wanted to replicate the exact UI of using fzf's ctrl-R, by feeding it `atuin history` / `atuin search` results.
Basically the setup described here https://github.com/atuinsh/atuin/issues/68#issuecomment-1567410629

By default fzf's ctrl-R history results list the history count id as well, whereas the setup above was missing that.
So i implemented the count_id in rust to replicate that.

I tried with `| nl` and that seems to work fine. Thanks for the advice.

I will close the PR.

PR #1608, opened 2024-01-22
feat: make store init idempotent

ellie:
This ensures that the user can run

```
atuin history init-store
```

a whole bunch of times, without getting duplicates in the store

Note that without this, nothing would break. You'd just end up with a much larger record store than necessary, which would be de-duped when built into a history database. Not ideal.


ellie:
wait no not like that 🤦‍♀️

PR #1610, opened 2024-01-22
fix(history): dedupe was removing history

ellie:
Related: https://forum.atuin.sh/t/search-ignoring-commands/74/5?u=ellie

When a user ran a duplicated command, but in another session, it was removed by filters. This is because the subquery that was once used did not have the same filters applied as the main query.

Instead of messing with subqueries, `group by` instead. This aligns with the search() function

### DIFF
--- a/atuin-client/src/database.rs
+++ b/atuin-client/src/database.rs
@@ -306,10 +306,7 @@ impl Database for Sqlite {
         }
 
         if unique {
-            query.and_where_eq(
-                "timestamp",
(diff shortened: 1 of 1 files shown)


atuin-client/src/database.rs
```diff
@@ -306,10 +306,7 @@ impl Database for Sqlite {
         }
 
         if unique {
```
ellie, line 308: we probably want to reuse search at some point, rather than sorta duplicate it

the main difference is that search automatically de-dupes and has a few other options, while list does not.

PR #1615, opened 2024-01-23
Add change-password command & support on server

### LINKED ISSUE #1374: Change password
**Describe the bug**
There is no clear way to change password

**To Reproduce**
1. Register with shitty password
1. You're screwed

**Expected behavior**
Account command let user change or rotate password. This flow is documented

**Error logs**
Please try and provide error logs where possible

 - OS: [e.g. macOS]
 - Shell [bash/zsh/fish/nu]
 - Version [e.g. 16.0.0]

**Self hosted**
If you are self hosting Atuin server, please detail your setup here. What version are you using, and how is it deployed?

**Additional context**

TymanWasTaken:
Fixes #1374

This PR both adds an API endpoint to the server (`PATCH /account/password`) and adds a corresponding cli command to use it. To change an account password, both a valid session and the current account password are necessary (the current password being verified to ensure it can't be changed without already knowing the password). This simply modifies the hash in the database, returning 200 on success. I tried to keep the code style similar to the surrounding code, so hopefully it is fine :)

By the way, while implementing this, I did notice that the delete account endpoint has no type of verification whatsoever, and doesn't even have a confirmation on the CLI, which kind of makes the required password for changing the password pointless as you can just delete re-register as a workaround. If you think that is something that should be added, I will gladly add it in this PR or another.

This is personally tested, but I recommend testing it yourself before merging, I may have missed something

### DIFF
--- a/atuin-client/src/api_client.rs
+++ b/atuin-client/src/api_client.rs
@@ -10,8 +10,9 @@ use reqwest::{
 
 use atuin_common::{
     api::{
-        AddHistoryRequest, CountResponse, DeleteHistoryRequest, ErrorResponse, LoginRequest,
-        LoginResponse, RegisterResponse, StatusResponse, SyncHistoryResponse, ChangePasswordRequest
+        AddHistoryRequest, ChangePasswordRequest, CountResponse, DeleteHistoryRequest,
+        ErrorResponse, LoginRequest, LoginResponse, RegisterResponse, StatusResponse,
+        SyncHistoryResponse,
     },
     record::RecordStatus,
 };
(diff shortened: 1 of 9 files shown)


ellie:
This is great, thank you! Just two things

1. I know our test coverage right now isn't amazing, but I'd like to improve that. Would it be possible to get a test for this endpoint please?

Something similar to the register test would be awesome 🙌 

https://github.com/atuinsh/atuin/blob/2bd7114cea89a471dcf6f4c279048db4be5d7731/atuin/tests/sync.rs#L102-L127

2. You're right - confirmation would be good. I don't think it should require the password though, as users may wish to delete an account they no longer have the password for (but maintain an active session for). Maybe just a 'Are you sure? [y/N]" type thing

TymanWasTaken:
I added a test that passes on my machine which should accurately test changing the password: it registers a user, changes the password on it, and attempts to login with the new password to confirm it actually worked.

ellie (approved):
Fantastic, thank you! I pushed one commit that was just the output of `cargo fmt`, hope that's ok

PR #1623, opened 2024-01-23
Fix: Check for format errors when printing history

tpoliaw:
Improve the errors reported when a format string is not valid, eg

    atuin search cargo --format '{invalid}'

gives

    history output failed with: The requested key "invalid" is unknown

instead of

    history output failed with: formatter error

Was the runtime-format library forked/written for use in atuin - it looks like
the timelines match? How feasible would it be either vendor the parts atuin
needs or refactor it so that the API exposes errors in a more conventional way?
I notice the repo doesn't have issues enabled or I'd raise it there.

### DIFF
--- a/atuin/src/command/client/history.rs
+++ b/atuin/src/command/client/history.rs
@@ -152,7 +152,12 @@ pub fn print_list(
     for h in iterator {
         let fh = FmtHistory(h, CmdFormat::for_output(&w));
         let args = parsed_fmt.with_args(&fh);
-        check_for_write_errors(write!(w, "{args}{entry_terminator}"));
+        let write = write!(w, "{args}{entry_terminator}");
+        if let Err(err) = args.status() {
+            eprintln!("ERROR: history output failed with: {err}");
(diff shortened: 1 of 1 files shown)


ellie:
lgtm, thanks!

> Was the runtime-format library forked/written for use in atuin - it looks like
the timelines match? How feasible would it be either vendor the parts atuin
needs or refactor it so that the API exposes errors in a more conventional way?
I notice the repo doesn't have issues enabled or I'd raise it there.

I'd say that's a question for @conradludgate - are you planning on maintaining/developing it further?

Happy to vendor if needed

ellie (approved):

conradludgate:
> I notice the repo doesn't have issues enabled or I'd raise it there.

Sorry, wasn't intentional. Enabled now

conradludgate:
> are you planning on maintaining/developing it further?

Yeah, I'm still maintaining it. If it's clear what exactly needs changing to make the error handling more convenient then I'm happy to make and release any changes.

Issue #1626, opened 2024-01-25
Provide JSON schema


artuross:
Hey,

Atuin has few configuration options and even comes with `default-config` that prints default configuration with short description for each setting.

Another popular project, [starship](https://starship.rs/config/#configuration) provides a JSON schema for the settings which integrates nicely with VS Code (probably other editors/IDEs too).

Introducing JSON schema would significantly improve configuration editing (not that there's that much options).

mikavilpas:
A suggestion: an example of a setup where the schema can be generated in a type safe fashion can be found here https://github.com/mfontanini/presenterm/blob/5a1e2d029f4f1775864dc0e16db8b8031bf4d882/src/main.rs?plain=1#L367-L372

It uses the [schemars](https://docs.rs/schemars/latest/schemars/) crate which generates json schemas for serde compatible structs. The implementation can be done with attributes, or also dynamically. Example:
https://github.com/mfontanini/presenterm/blob/5a1e2d029f4f1775864dc0e16db8b8031bf4d882/src/config.rs?plain=1#L17

The users can use the latest version or any other version by including a magic comment in their config. Here are examples from my own config ([yaml example](https://github.com/mikavilpas/dotfiles/blob/990b9fe310ad9209c1e1bd5a37f419a14758eb77/.config/presenterm/config.yaml?plain=1#L1) for presenterm, and a [toml example](https://github.com/mikavilpas/dotfiles/blob/990b9fe310ad9209c1e1bd5a37f419a14758eb77/.config/yazi/yazi.toml?plain=1#L1) for [yazi](https://yazi-rs.github.io/))

PR #1627, opened 2024-01-25
add json schema for config

hsnprsd:
Adds JSON schema completion & validation to `config.toml` file on Visual Studio Code.
The `config-schema.json` file should be accessible at https://atuin.sh/config-schema.json.

Fixes: https://github.com/atuinsh/atuin/issues/1626.

### DIFF
--- a/atuin-client/config-schema.json
+++ b/atuin-client/config-schema.json
@@ -1,0 +1,257 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "title": "Atuin Configuration",
+    "description": "Atuin Configuration",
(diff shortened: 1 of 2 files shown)


hsnprsd:
@ellie Is this fine?

ellie:
Not quite. Realistically, we won't keep a static JSON schema up to date reliably. It should be generated dynamically.

Imo it should be hosted as part of the api, which means we can have an API handler here to generate it. We can then point clients to api.atuin.sh/api/v0/client-config-schema.json.

One other thing to note is that there are actually two config files. One for the client, and one for the server

PR #1629, opened 2024-01-26
feat(shell)!: bind the Atuin search to "/" in vi-normal mode

akinomyoga:
This is a suggestion for the keybinding to <kbd>/</kbd> motivated by @Nemo157's comments https://github.com/atuinsh/atuin/pull/1553#issuecomment-1892022291 and https://github.com/atuinsh/atuin/pull/1570#issuecomment-1892163553.

Also, I suggest removing keybinding to <kbd>C-r</kbd> in the `vicmd` keymap of Zsh since it overwrites the existing widget (redo) which is not a search. Since this involves removal of a keybinding, I think we need to discuss it carefully.

See also the commit message.

### DIFF
--- a/atuin/src/command/init.rs
+++ b/atuin/src/command/init.rs
@@ -33,8 +33,8 @@ impl Cmd {
 
         if std::env::var("ATUIN_NOBIND").is_err() {
             const BIND_CTRL_R: &str = r"bindkey -M emacs '^r' _atuin_search_widget
-bindkey -M vicmd '^r' _atuin_search_vicmd_widget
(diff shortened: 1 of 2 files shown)


ellie:
> Also, I suggest removing keybinding to C-r in the vicmd keymap of Zsh since it overwrites the existing widget (redo) which is not a search. Since this involves removal of a keybinding, I think we need to discuss it carefully.

I think this makes sense. I would expect that vim mapping users would understand this, and it's fairly straightforward to add the mapping back again should they wish.

ellie (approved):

PR #1630, opened 2024-01-26
fix: Skip padding time if it will overflow the allowed prefix length

Nemo157:
Fixes issue from https://github.com/atuinsh/atuin/pull/1574#discussion_r1467552000 (though, I wasn't able to reproduce, the failure implies that `PREFIX_LENGTH` is inadequate for some situations, and the previous implementation would have been overwriting the already printed duration).


Nemo157:
Another option would be to then clamp to `1..` so that there is always at least 1 space of padding.

ellie (approved):
:pray

akinomyoga:
Thank you!

PR #1631, opened 2024-01-26
feat(zsh): update widget names

akinomyoga:
This is another suggestion. See the commit message.

For more contexts, the following is the list of how the other zsh frameworks name their widgets:

- https://github.com/jeffreytse/zsh-vi-mode ... `zvm_*`
- https://github.com/babarot/zsh-vimode-visual ... `vi-visual-*`
- https://github.com/junegunn/fzf/blob/master/shell/key-bindings.zsh ... `fzf-*-widget`
- https://github.com/ohmyzsh/ohmyzsh/
  - `plugins/{fuck,history-substring-search,man,per-directory-history,sudo,shrink-path}` ... `<PLUGIN>-*` (where `<PLUGIN>` is the name of the plugin)
  - `plugins/{npm,percol}` ... `<PLUGIN>_*`
  - `plugins/git-escape-magic` ... `<PLUGIN>.*`
  - `plugins/{globalias,copybuffer,fancy-ctrl-z}` ... `<PLUGIN>` (These define a single widget that has the same name as the plugin)
  - `plugins/zsh-navigation-tools` ... `znt-*-widget`
  - `plugins/zsh-interactive-cd` ... `zic-*`
  - `plugins/jump` ... `_mark_expansion`
  - `plugins/dircycle` ... `insert-cycled*`
  - `plugins/safe-paste` ... `_bracketed_paste_*`
  - `plugins/dirhistory` ... `dirhistory_zle_dirhistory_*`

Also, it should be noted that ZLE also prepares the special widget names `zle-*` for the users to customize the behavior. The most popular namespacing seems to be `<ProjectName>-*`.

### DIFF
--- a/atuin/src/command/init.rs
+++ b/atuin/src/command/init.rs
@@ -32,17 +32,17 @@ impl Cmd {
         println!("{base}");
 
         if std::env::var("ATUIN_NOBIND").is_err() {
-            const BIND_CTRL_R: &str = r"bindkey -M emacs '^r' _atuin_search_widget
-bindkey -M viins '^r' _atuin_search_viins_widget
-bindkey -M vicmd '/' _atuin_search_widget";
+            const BIND_CTRL_R: &str = r"bindkey -M emacs '^r' atuin-search
+bindkey -M viins '^r' atuin-search-viins
+bindkey -M vicmd '/' atuin-search";
 
-            const BIND_UP_ARROW: &str = r"bindkey -M emacs '^[[A' _atuin_up_search_widget
-bindkey -M vicmd '^[[A' _atuin_up_search_vicmd_widget
-bindkey -M viins '^[[A' _atuin_up_search_viins_widget
-bindkey -M emacs '^[OA' _atuin_up_search_widget
-bindkey -M vicmd '^[OA' _atuin_up_search_vicmd_widget
-bindkey -M viins '^[OA' _atuin_up_search_viins_widget
-bindkey -M vicmd 'k' _atuin_up_search_vicmd_widget";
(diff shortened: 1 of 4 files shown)


ellie:
I'm happy with the change, but would it be possible to have aliases to the existing names? 

If we rename them without warning, some setups will break. One release is enough warning imo, we can remove the aliases after that

akinomyoga:
> I'm happy with the change, but would it be possible to have aliases to the existing names?
> 
> If we rename them without warning, some setups will break.

Maybe I missed something, but I already left the old names (as described also in the commit message), so the existing setups shouldn't break. Or does it need to be a real `alias` in the shell's sense?

Or should I also leave the new ones `_atuin{,_up}_search_vi{cmd,ins}_widget`? These are recently added and did not exist in the release version, so I dropped those new names and only left `_atuin{,_up}_search_widget`, which exist in the release version.

> One release is enough warning imo, we can remove the aliases after that

I didn't intend to remove it soon, but if you'd like the warning message, I can add the warning message.

akinomyoga:
I cleaned up slashes in docs that escaped `_` before. Now we removed `_` so don't have to keep the backslashes.

ellie (approved):
Thank you! Would it be possible to get a PR to the main docs too please? 🙏

akinomyoga:
Thanks!

> Thank you! Would it be possible to get a PR to the main docs too please? 🙏

Sure! I'm not able to create it now, but I'll later submit it (probably after a few or several hours).

akinomyoga:
> > Thank you! Would it be possible to get a PR to the main docs too please? 🙏
> 
> Sure! I'm not able to create it now, but I'll later submit it (probably after a few or several hours).

https://github.com/atuinsh/docs/pull/19

Issue #998, opened 2023-05-22
Support customizing data displayed for TUI search


lazywei:
I'm building `atuin` on HEAD,

This is working as intended
```
atuin search --format "{time} {command}"
```

But the following doesn't, it just ignore the specified format
```
atuin search -i --format "{time} {command}"
```

Is this expected?

conradludgate:
Yes, that is expected. I've been considering changing it to properly respect what flags are supported under interactive or not, but I haven't managed to find a decent solution yet

iloveitaly:
My use case here is I don't care about anything other than `{command}` in my results and would love the tui list to be nice and clean.

PR #1638, opened 2024-01-27
feat: make history list format configurable

### LINKED ISSUE #1616: Make format for history configurable
I often forget which directory I ran a command from, and when I then find `./start.sh` in my history I'm only halfway there.

With the help of #619 I can find the directory, but I'd like this config:

`history_format` -- default format for `atuin history list`

dongxuwang:
fix #1616 

## Checks

### DIFF
--- a/atuin-client/config.toml
+++ b/atuin-client/config.toml
@@ -84,7 +84,7 @@
 ## alt-0 .. alt-9
 # ctrl_n_shortcuts = false
 
-## default history list item format fallback when not appending in cmd
+## default history list format - can also be specified with the --format arg
 # history_format = "{time}\t{command}\t{duration}"
 
 ## prevent commands matching any of these regexes from being written to history.
--- a/atuin-client/config.toml
+++ b/atuin-client/config.toml
@@ -84,6 +84,9 @@
 ## alt-0 .. alt-9
 # ctrl_n_shortcuts = false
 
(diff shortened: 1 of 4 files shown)


atuin-client/config.toml
```diff
@@ -84,6 +84,9 @@
 ## alt-0 .. alt-9
 # ctrl_n_shortcuts = false
 
+## default history list item format fallback when not appending in cmd
```
ellie, line 87:
```suggestion
## default history list format - can also be specified with the --format arg
```

would it also be possible to update this on the docs repo please?

https://github.com/atuinsh/docs

ellie (commented):
thanks!

I mentioned in the comment too, but if you could also update the [docs](https://github.com/atuinsh/docs) that would be fantastic 🙏

ellie (dismissed):

ellie (approved):

dongxuwang:
> thanks!
> 
> I mentioned in the comment too, but if you could also update the [docs](https://github.com/atuinsh/docs) that would be fantastic 🙏

I'll try to update, thanks

dlvhdr:
should this also work in interactive mode? It doesn't seem to and I would love it

ben-alkov:
> should this also work in interactive mode? It doesn't seem to and I would love it

#998

PR #1645, opened 2024-01-29
feat: add prefers_reduced_motion flag

conradludgate:
Implementation of ideas discussed in #1290.

Can be enabled by setting either
* `PREFERS_REDUCED_MOTION=1`
* `ATUIN_PREFERS_REDUCED_MOTION=1`
* `prefers_reduced_motion = true` in client config.

The implementation uses a custom `now` function, which is either `OffsetDateTime::now_utc`, or it uses a cached time value.

## Checks

### DIFF
--- a/atuin-client/config.toml
+++ b/atuin-client/config.toml
@@ -148,6 +148,10 @@ enter_accept = true
 ## Timeout (in seconds) for acquiring a local database connection (sqlite)
 # local_timeout = 5
 
+## Set this to true and Atuin will minimize motion in the UI - timers will not update live, etc. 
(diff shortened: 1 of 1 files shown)


ellie:
Looks good! Might be worth also adding NO_MOTION? I know it's technically not "none", but just to align with NO_COLOR

Not a standard or anything like the color one, but in case it does become one

ellie:
Thank you for this!

I wonder if we can end up with NO_MOTION implemented in a bunch of tools, like [NO_COLOR](https://no-color.org/)

PR #1653, opened 2024-01-30
Set ATUIN_ variables for development in devshell

matthiasbeyer:
I accidentially broke my atuin database by executing `atuin` from the PR I was working on without setting these variables and had to manually roll back my local database.

That shouldn't happen, so we set the database and record store path in the devshell to something that does not overwrite our normal databases. We also warn if these files already exist, because when entering the devshell, a user might want to start from a clean slate here.

## Checks


ellie (approved):
thanks!

PR #1655, opened 2024-01-30
Add "filter_mode_list" configuration option

### LINKED ISSUE #366: Allow custom filtering rules
Hi, thanks for this project. 

I would like to see all the search filters together, but in a custom ordered manner. Like I want to see a search UI with three sections - Directory, Host, and Global. That way I don't have to cycle between the filter modes and can just start searching with minimal cognitive effort.

### LINKED ISSUE #1654: Feature request: Allow changing order of filter modes
I would like to change the order in which the filter modes are used and actually be able to drop some of them (for example, I do not care about `directory` filter mode).

This is probably related to #366 - but I am not sure whether we're talking about the very same thing here.

matthiasbeyer:
Closes #1654 
Maybe closes #366 

## Checks

### DIFF
--- a/atuin/src/command/client/search/interactive.rs
+++ b/atuin/src/command/client/search/interactive.rs
@@ -359,23 +359,10 @@ impl State {
             }
             KeyCode::Char('u') if ctrl => self.search.input.clear(),
             KeyCode::Char('r') if ctrl => {
-                let filter_modes = if settings.workspaces && self.search.context.git_root.is_some()
-                {
-                    vec![
-                        FilterMode::Global,
-                        FilterMode::Host,
-                        FilterMode::Session,
-                        FilterMode::Directory,
-                        FilterMode::Workspace,
-                    ]
-                } else {
-                    vec![
-                        FilterMode::Global,
-                        FilterMode::Host,
-                        FilterMode::Session,
-                        FilterMode::Directory,
-                    ]
-                };
+                let mut filter_modes = settings.filter_mode_list.clone();
+                if settings.workspaces && self.search.context.git_root.is_some() {
+                    filter_modes.retain(|fm| *fm != FilterMode::Workspace);
+                }
 
                 let i = self.search.filter_mode as usize;
                 let i = (i + 1) % filter_modes.len();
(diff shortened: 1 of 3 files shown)


matthiasbeyer:
I did not yet test-drive this (also because I ran into #1653 while I started working on this).

I leave the discussion whether such a feature is desired to the issue (#1654 ) :+1:

atuin/src/command/client/search/interactive.rs
```diff
@@ -359,23 +359,10 @@ impl State {
...
-                    vec![
-                        FilterMode::Global,
-                        FilterMode::Host,
-                        FilterMode::Session,
-                        FilterMode::Directory,
-                    ]
-                };
+                let mut filter_modes = settings.filter_mode_list.clone();
+                if settings.workspaces && self.search.context.git_root.is_some() {
+                    filter_modes.retain(|fm| *fm != FilterMode::Workspace);
```
matthiasbeyer, line 364: I just noticed that this might be wrong, the condition should be wrapped in a `!(...)` I guess.

I'll leave it this way until there's a decision whether this feature is desired or not. :smile_cat:

matthiasbeyer:
Also, this feature somewhat contradicts the `filter_mode` setting I guess, because implicitely the first entry in the `filter_modes_list` is the default, right? :thinking: We might want to have some discussion about this as well.

matthiasbeyer:
Any ideas about this?

ellie:
I think it would make sense to add a `[search]` block

So

```toml
[search]
filters = ["session", "directory", "global"]
```

When set, it will override `filter_mode` and any other related settings.

Wrt workspaces, I'd suggest the following

1. If a filter list is set, does not contain workspace, and workspaces = true. We should work with the current behaviour, of switching to workspace mode.
2. If a filter list is set, and _does_ contain workspace, then follow the order set in the list - regardless of whether workspaces = true.

mattbaker:
I'd love to see this feature!

markovejnovic:
I believe this has been implemented in #2430.

Would suggest setting

```toml
search.filters = ["global", "host", ...]
```

PR #1657, opened 2024-01-30
fix: never overwrite the key

ellie:
Now that local history is stored encrypted, new_key should not overwrite an existing one. This may be frustrating, but will remove the risk of Atuin generating a new key and the user losing their old one.

## Checks


ellie:
if we wish to generate a new key, the options are

1. Manually delete it, re run
2. Delete it in code, very explicitly. Rather than overwrite.

PR #1660, opened 2024-01-31
feat: add an exact search mode

lilydjwg:
This pr adds a new search mode called "exact" as skim calls it, in which the query are matched as substrings of the target. In this way, there are much fewer results than the so-called fuzzy search mode, thus provide more relevant results as long as the query is correct.

I usually can't relate fuzzy results to my queries, so I always use this kind of substring match. There is also a full text search mode, which I find less useful as only one substring can be provided.

## Checks

### DIFF
--- a/atuin-client/src/database.rs
+++ b/atuin-client/src/database.rs
@@ -432,7 +432,6 @@ impl Database for Sqlite {
 
         match search_mode {
             SearchMode::Prefix => sql.and_where_like_left("command", query),
-            SearchMode::FullText => sql.and_where_like_any("command", query),
             _ => {
                 // don't recompile the regex on successive calls!
                 lazy_static! {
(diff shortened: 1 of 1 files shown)


ellie:
I'd be happy to change fulltext to be this behaviour to be honest. At this point I'd rather not add any new search modes unless they're sufficiently different.

ellie:
> Do you want to rename "fulltext" to "exact" or "substr" for the change?

I think keep it as fulltext - with this change we're likely to break the installs for anyone currently using fulltext. Or at the very least alias it so that the config still works OK

I have absolutely no idea how many people use it though, so we shall have to see if this change is well accepted or not.

ellie:
> I failed to add an alias due to [mehcode/config-rs#502](https://github.com/mehcode/config-rs/issues/502).

in that case, let's change fulltext to be this behaviour :D

ellie (approved):
thanks!

Issue #1661, opened 2024-01-31
Move to a background daemon


ellie:
See discussion for full details: https://forum.atuin.sh/t/moving-atuin-to-a-daemon/78

This has been on my mind for a long time, and after some discussion I think it should go ahead. 

This would allow for

1. Far less risk of blocking shells (slow filesystems and other such errors)
2. Easy scheduling of periodic maintenance (eg, sqlite vacuum)
3. Real time sync - sync will not need to be prompted by shell activity

Opening this issue to track

dbaynard:
After #2006 (which is working well, for me), what remains to resolve this?

(I came here from #1455 which says it's blocked on this issue — I'm interested in the exit code filtering in the ctrl-r UI.)

markovejnovic:
Closing this since Atuin has been running on the daemon for a while. #2008 tracks further progress.

PR #1667, opened 2024-02-02
set durability for sqlite to recommended settings

nebkor:
On my ZFS system, commands were being blocked by history insertion due to `FULL` durability being enabled by default in SQLx for the WAL, sometimes for over a second, but at least half the time there was at least 500ms before the command was run. According to the [documentation](https://www.sqlite.org/pragma.html#pragma_synchronous), `NORMAL` is recommended.

With this change, my terminal commands are no longer being blocked.

## Checks
- [x ] I am happy for maintainers to push small adjustments to this PR, to speed up the review cycle
- [ x] I have checked that there are no existing pull requests for the same thing

### DIFF
--- a/atuin-client/src/database.rs
+++ b/atuin-client/src/database.rs
@@ -14,7 +14,10 @@ use rand::{distributions::Alphanumeric, Rng};
 use regex::Regex;
 use sql_builder::{esc, quote, SqlBuilder, SqlName};
 use sqlx::{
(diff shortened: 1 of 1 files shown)


ellie:
I'm hesitant due to

> A transaction committed in WAL mode with synchronous=NORMAL might roll back following a power loss or system crash.

but I think that seems reasonable for now. once #1661 is complete we can do always change back to full with no issues.

ellie (approved):
Thank you!

ellie:
ref #952

nebkor:
Thank you so much!

PR #1669, opened 2024-02-03
fix(bash/preexec): erase prompt last line before Bash renders on it

### LINKED ISSUE #1668: `enter_accept` breaks liquidprompt with bash-preexec
"breaks" as in:

after interactively selecting-and-running a command from atuin's TUI the previous `PS1` is shown, and there are leftover characters right of the cursor if that previous `PS1` happened to be shorter than the current one (if that formulation even makes sense?  LP does the seemingly-normal thing of putting logic in `${pre{cmd,exec}_functions[@]}` that does stuff and sets `PS1` to something).

I tried flipping atuin & LP inits around in `.bashrc` to see if that's some sort of an order issue, but that didn't help.  no idea what else to try or if this is even a tractable problem; suggestions welcome!

akinomyoga:
Fixes #1668

The descriptions are found in commit messages.

## Checks

### DIFF
--- a/atuin/src/shell/atuin.bash
+++ b/atuin/src/shell/atuin.bash
@@ -106,10 +106,16 @@ __atuin_accept_line() {
     # Reprint the prompt, accounting for multiple lines
     local __atuin_prompt __atuin_prompt_offset
     __atuin_evaluate_prompt
+    local __atuin_clear_prompt
+    __atuin_clear_prompt=$'\r'$(tput el)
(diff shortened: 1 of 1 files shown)


cmm:
this does seem to work, thanks!

akinomyoga:
@cmm Thank you for checking!

ellie (approved):
thank you 🙏

PR #1675, opened 2024-02-05
chore(ci): run rust build/test/check on 3 platforms

ellie:
I don't see too much point in also running the format/clippy on all platforms, and integration tests are maybe a little too much. Otherwise, this should ensure we don't break builds for mac/windows by accident

Note: This is still not "true" support for windows. I'd like to ensure builds/tests run OK, but I'd like to be able to properly test on that platform before we commit to supporting it.

## Checks

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -11,7 +11,10 @@ env:
 
 jobs:
   build:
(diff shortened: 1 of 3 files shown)


ellie:
cc @YummyOreo, you'll probably be happy to see this :D

PR #1689, opened 2024-02-08
minor formatting updates to the default config.toml

### LINKED ISSUE #1687: Default config file [stats] block commented out
In the [default config file](https://github.com/atuinsh/atuin/blob/e2f080ccff59f3b6b7112092531f6f060f85aaa2/atuin-client/config.toml#L161) the line indictating `[stats]` is commented out.  So when I tried to set `common_prefix` and `common_subcommands` atuin ignored them, because it didn't know they were supposed to be `stats.common_prefix`, etc.

I was trying to find an example, but I _think_ in previous "default" config files I've seen, the sections or blocks are uncommented, and only the specific values are commented out.  I also think the "comments" in the config file are double-commented, so that you can select a block of text in your IDE and uncomment and you don't end up with the parser trying to parse the things that are supposed to be comments.  E.g.,

```toml
[stats]
# # Set commands where we should consider the subcommand for statistics. Eg, kubectl get vs just kubectl
#common_subcommands = [
#  "cargo",
#  "go",
#  "git",
#  "npm",
#  "yarn",
#  "pnpm",
#  "kubectl",
#]

# # Set commands that should be totally stripped and ignored from stats
#common_prefix = ["sudo"]
```

instead of 

```toml
#[stats]
# Set commands where we should consider the subcommand for statistics. Eg, kubectl get vs just kubectl
#common_subcommands = [
#  "cargo",
#  "go",
#  "git",
#  "npm",
#  "yarn",
#  "pnpm",
#  "kubectl",
#]
#
# Set commands that should be totally stripped and ignored from stats
#common_prefix = ["sudo"]
```

Anyways, not a big deal and could arguably be chalked up to "user error", feel free to close if you think it's fine.

drmorr0:
Resolves #1687.  Some minor formatting updates to the default config.

## Checks

### DIFF
--- a/atuin-client/config.toml
+++ b/atuin-client/config.toml
@@ -134,7 +134,6 @@
 # This applies for new installs. Old installs will keep the old behaviour unless configured otherwise.
 enter_accept = true
 
-
 ## Defaults to "emacs".  This specifies the keymap on the startup of `atuin
 ## search`.  If this is set to "auto", the startup keymap mode in the Atuin
 ## search is automatically selected based on the shell's keymap where the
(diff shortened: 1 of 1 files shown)


ellie (approved):
Thanks!

drmorr0:
Ooh, exciting, I just filled out the form!

PR #1697, opened 2024-02-09
feat: Add 'a', 'A', 'h', and 'l' bindings to vim-normal mode

alerque:
The current 'i' binding to switch to insert mode is entirely unintuitive since what I almost always want to do is append to the current query. The fact that the cursor extends past the current input (which vim doesn't do with default settings) adds to the problem. The 'a' key is what I would reach for, but 'A' makes a lot of sense too so I added that.

The 'h' and 'l' bindings for moving the cursor also help makes things a bit more usable.

## Checks


ellie (approved):
Looks good, thank you! 

PR #1716, opened 2024-02-13
add ctrl-k shortcut

### LINKED ISSUE #1715: emacs ctrl-k does not work in search mode
PR incoming.

ctrl-k truncates a line from the cursor. This was a trivial fix with the `truncate(index)` function on rust strings.

I do think that maybe using a "proper" readline implementation could save a lot of duplicated work, for instance: https://github.com/kkawakam/rustyline

rseymour:
fixes  #1715

Adds the control-k (k for unalive) shortcut which is hardwired into my brain.  I do humbly suggest that rustyline or some other readline equivalent be used for cursor handling to prevent reimplementing every shortcut from scratch tho.

## Checks

### DIFF
--- a/atuin/src/command/client/search/cursor.rs
+++ b/atuin/src/command/client/search/cursor.rs
@@ -191,6 +191,10 @@ impl Cursor {
         }
     }
 
+    pub fn truncate(&mut self) {
+        self.source.truncate(self.index);
+    }
+
     pub fn clear(&mut self) {
         self.source.clear();
         self.index = 0;
--- a/atuin/src/command/client/search/interactive.rs
+++ b/atuin/src/command/client/search/interactive.rs
@@ -380,6 +380,7 @@ impl State {
                 }
             }
             KeyCode::Char('u') if ctrl => self.search.input.clear(),
+            KeyCode::Char('k') if ctrl => self.search.input.truncate(),
             KeyCode::Char('r') if ctrl => {
                 let filter_modes = if settings.workspaces && self.search.context.git_root.is_some()
                 {


ellie:
Hey!

I'm afraid that readline parity is a non-goal. I'm aware this may cause frustration with some keybindings, but the issue is that 

1. we have very limited key combinations available as-is
2. readline uses... pretty much all of them. many of which are several different ways of doing the same thing

If we were to go for parity, there would be no keys leftover for our own uses.

see: https://forum.atuin.sh/t/default-key-binding-changes/41

ellie:
I'm going to close this for now, but if you have thoughts/suggestions happy to discuss.

rseymour:
I agree with your sentiment. I'd love a readline compatible mode, but I think sticking with tradition isn't always the right way for new software. For now I'm just disabling up and ctrl-r and leaving it like that.

iwahbe:
I'd like to second the suggestion to add a `ctrl-k` binding that works as emacs users expect, scoped exclusively to the `KeymapMode::Emacs` setting.

It's pretty common for me (and I suspect other Emacs users) to move the cursor and then `ctrl-k` to remove the latter half of a line, instead of deleting deleting from the end. I do this at least 10 times a day in atuin (where it doesn't work). It makes sense that readline parity is not a goal.


I would suggest scoping it exclusively to the `emacs` key bindings, where users already have muscle memory for up: [`ctrl-p`](https://github.com/atuinsh/atuin/blob/3d6b16354697e6036a57819d79f2f6084392938a/atuin/src/command/client/search/interactive.rs#L444-L446) (which atuin already supports).

This wouldn't cost atuin a keybinding, since `ctrl-k` already sends the `return self.handle_search_up(settings, false);` signal (duplicated with `ctrl-p`).


> I'm going to close this for now, but if you have thoughts/suggestions happy to discuss.

P.S. If this is out of line please tell me and I'll stop commenting.

Issue #1719, opened 2024-02-14
Ctrl+m as Enter stopped working


timka:
My muscle memory loves using ctrl+m as Enter. I believe this is [some old VT100 sequence](https://github.com/helix-editor/helix/issues/6551#issuecomment-1931335427) similar to ctrl+[ functioning as Esc which doesn't work too BTW. However I don't use the later so can't say for sure if this is a regression.

I suspect this may be a result of https://github.com/atuinsh/atuin/pull/1170.

My old version was 11.0.0. Now running 18.0.1.

The old version also wasn't affected by https://github.com/atuinsh/atuin/issues/492

I tried to just revert the binary to 11.0.0 (oh, silly me!) but it didn't work due to DB migrations. Fortunately I had a backup.

akinomyoga:
> I suspect this may be a result of https://github.com/atuinsh/atuin/pull/1170.

#1170 is about the switching between CSI and SS3 in the key codes of <kbd>up</kbd>/<kbd>down</kbd>/<kbd>right</kbd>/<kbd>left</kbd>.

The problem of <kbd>C-m</kbd> should be related to #1505. In terminals, <kbd>C-m</kbd> and <kbd>RET</kbd> are usually transmitted by the same code `\x0D`. The `crossterm` translates them to `KeyCode::Enter` [1] and Atuin performs an action for `KeyCode::Enter` [2]. After #1505, if the terminal supports the kitty keyboard protocol, <kbd>C-m</kbd> is transmitted by a distinct key sequence and translated to `KeyCode::Char('m')` plus `KeyModifiers::CONTROL`, which isn't processed by Atuin.

- [1] [`src/event/sys/unix/parse.rs:547` @ `crossterm-rs/crossterm`](https://github.com/crossterm-rs/crossterm/blob/99fe255f33f774f04c09755572503349c915112b/src/event/sys/unix/parse.rs#L547)
- [2] [`atuin/src/command/client/search/interactive.rs:308`](https://github.com/atuinsh/atuin/blob/1a432b61ee4f52f3ff0da1c991aa5d605f700869/atuin/src/command/client/search/interactive.rs#L308)

PR #1720, opened 2024-02-14
feat(search): process Ctrl+m for kitty keyboard protocol

akinomyoga:
Fixes https://github.com/atuinsh/atuin/issues/1719

The background is explained in the commit mesasge.

I know there is tension between the readline bindings and Atuin's specific bindings for the available key-sequence space, but I think <kbd>C-m</kbd> cannot be used for Atuin's feature anyway because it can only be used with the terminals with an extended keyboard protocol.

## Checks


ellie (approved):
Totally makes sense, thank you!

PR #1721, opened 2024-02-14
feat: support syncing aliases

ellie:
```
atuin config alias set k kubectl
```

to set an alias. in this example, alias k -> kubectl

```
atuin config alias delete k
```

to delete an alias, in this example deleting the one we just set.

Aliases will be copied across shells, and across machines. All that's needed is a shell reload (in the future maybe let's remove that too)

## Checks

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -185,6 +185,7 @@ dependencies = [
  "async-trait",
  "atuin-client",
  "atuin-common",
+ "atuin-config",
  "atuin-server",
  "atuin-server-postgres",
  "base64 0.21.7",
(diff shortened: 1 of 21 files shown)


ellie:
Some timing - building the entire alias list from scratch took 14ms on my machine, with a few aliases set. Optimisation will still be a good thing (we're doing too much work, really), but not as critical as I thought

ellie:
This doesn't support Nu right now, but Nu cannot directly eval output. Currently Atuin writes init to a file and nu sources that.

I'd like to make _all_ the shells do that automatically, and update the file when required. Or at least something like it. Leave out Nu support for now

PR #1727, opened 2024-02-17
Revert "fix(bash): avoid unexpected `atuin history start` for keybindings"

ellie:
Reverts atuinsh/atuin#1509

cc @akinomyoga 

Looks like this is causing some of the "bash not recording commands" issues, from what I've managed to replicate. I'd rather take inaccurate timings vs missed commands :/

### DIFF
--- a/atuin/src/shell/atuin.bash
+++ b/atuin/src/shell/atuin.bash
@@ -17,14 +17,6 @@ export ATUIN_SESSION
 ATUIN_HISTORY_ID=""
 
 __atuin_preexec() {
-    if [[ ! ${BLE_ATTACHED-} ]]; then
-        # With bash-preexec, preexec may be called even for the command run by
-        # keybindings.  There is no general and robust way to detect the
-        # command for keybindings, but at least we want to exclude Atuin's
-        # keybindings.
-        [[ $BASH_COMMAND == '__atuin_history'* && $BASH_COMMAND != "$1" ]] && return 0
-    fi
-
     local id
     id=$(atuin history start -- "$1")
     export ATUIN_HISTORY_ID=$id


akinomyoga:
I submitted a proper fix to https://github.com/rcaloras/bash-preexec/pull/152.

ellie:
Great!

Does that fix mean we don't need this patch at all? I'm probably going to need to release v18.0.2 shortly

akinomyoga:
> Does that fix mean we don't need this patch at all? 

Yes, after the PR is merged, we shouldn't need this revert patch.

As explained in https://github.com/rcaloras/bash-preexec/pull/152, the problem is that preexec is called for the keybinding and passes the previous command. This behavior of bash-preexec caused overwriting the entry of the previous command and produced 0s duration. #1509 originally tried to fix the problem at Atuin's side (as at that time, I haven't thought about the possibility that it is possible to fix it at bash-preexec's side). Now that we reverted #1509, I guess we start to have the 0s-duration issue again.

Let me look at the behavior more closely. Maybe I'll submit a rework of #1509.

ellie:
> Yes, after the PR is merged, we shouldn't need this revert patch.

I'm afraid we will still need it, as we have no way of ensuring existing users upgrade. 

> I guess we start to have the 0s-duration issue again.

In my testing so far, without this revert, we don't really record history at all. I'd rather have the 0-s duration than not record history

akinomyoga:
> In my testing so far, without this revert, we don't really record history at all.

Let me check the behavior. As far as I look at the change in #1509, I don't quite get why removing this part would fix the problem.

akinomyoga:
As far as I test the latest `main` after this revert, the problem doesn't seem to be solved. Hmm.

In the first place, since bash-preexec doesn't pass the command to Atuin at all, there doesn't seem to be a way for Atuin to get the command. If the command is added to Atuin's history with Atuin of the current `main` with the current bash-preexec, it's a mystery to me where Atuin got the command.

akinomyoga:
Ah, OK. I think you started and canceled the Atuin history search at every line in the testing. In that case, the command is added to Atuin's history before the Atuin history search is started. In that use pattern, the problem doesn't arise indeed.

However, in the use case where the Atuin history search is called now and then, the commands are still randomly added.

ellie:
Hmmm I'm sorry, I think you're right. Perhaps this revert was a little hasty 🙈 

The behaviour that does seem consistent, and not random, is that 

1. A command after cancelling an Atuin search is _not_ recorded
2. A command after accepting an atuin search _is_ recorded

Try running a bunch in a series between accepting/cancelling a search and you'll see that the first isn't recorded depending on whether you've accepted or cancelled

akinomyoga:
I'm now preparing a rework of #1509. I'll submit it soon. Could you take a look at the reworked version?

ellie:
> I'm now preparing a rework of #1509. I'll submit it soon. Could you take a look at the reworked version after I submit it?

Of course! I've undone the revert as it was a red-herring. 

I've also just checked v17.2.1, and the behaviour I've described in my previous comments occurs there too

PR #1729, opened 2024-02-17
fix(bash): rework #1509 to recover from the preexec failure

akinomyoga:
Explanations are in the commit message. I've quickly tested it locally and confirmed that the user commands are at least not missing, but I think @ellie can also test it locally to carefully see the behavior.

Refs. #1003, #1509, #1727, #1728, [Forum 175](https://forum.atuin.sh/t/atuin-bash-and-ble-sh/175)

## Checks


ellie (commented):
This works great! I tried a few variations that did not work previously and this seems to have resolved the issue. Thank you 🙏

ellie (approved):

PR #1735, opened 2024-02-19
refactor(nushell): update `commandline` syntax, closes #1733

### LINKED ISSUE #1733: [nushell] breaking change coming
Heads-up friendly folks at atuin! At nushell, we're working on a PR that will probably be a breaking change for people who use atuin with nushell integration.

We're breaking up the commandline built-in command into several subcommands. Please see the PR for more details. It will be deprecated for a while. You'll get a warning message, but it should still work in version 0.91.0 but after that we'll probably remove the deprecation.
https://github.com/nushell/nushell/pull/11877

stevenxxiu:
This should be merged when the next version of *Nushell* releases, where it'll change the `commandline` syntax.

## Checks

### DIFF
--- a/atuin/src/shell/atuin.nu
+++ b/atuin/src/shell/atuin.nu
@@ -31,7 +31,7 @@ def _atuin_search_cmd [...flags: string] {
     [
         $ATUIN_KEYBINDING_TOKEN,
         ([
-            `commandline (ATUIN_LOG=error run-external --redirect-stderr atuin search`,
+            `commandline edit (ATUIN_LOG=error run-external --redirect-stderr atuin search`,
             ($flags | append [--interactive, --] | each {|e| $'"($e)"'}),
             `(commandline) | complete | $in.stderr | str substring ..-1)`,
         ] | flatten | str join ' '),
--- a/atuin/src/shell/atuin.nu
+++ b/atuin/src/shell/atuin.nu
@@ -28,10 +28,12 @@ let _atuin_pre_prompt = {||
 }
 
 def _atuin_search_cmd [...flags: string] {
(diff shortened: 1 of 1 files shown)


arcuru:
Is it possible to check the Nushell version and support both versions?

stevenxxiu:
I could do that, but this adds the additional burden of removing the code later down the track.

I feel like *Nushell* is still unstable, so people should update regularly anyway.

fdncred:
Nushell does populate `$env.NU_VERSION` for an easy way to check the nushell version running. One can also do `nu --version` and get the same version string.

arcuru:
Nushell itself is apparently going to support the old way for 1 release before deprecating. Some of Atuin's users may be on platforms where they aren't running the [absolute latest Nushell](https://repology.org/project/nushell/versions).

This change as written will cause Atuin to deal with the Nushell breakage, as anyone on older versions of Nushell will not get an appropriate warning message and will think Atuin has the bug. That will be far more burden than removing the code down the line.

Nushell having breaking changes is fine, but all programs trying to work with Nushell, like Atuin, will be trying to support older versions where possible. Having a point in time where a user has to update both Nushell and Atuin at exactly the same time is unreasonable.

stevenxxiu:
True. I suppose *Atuin* and *Nushell* won't always update at similar times.

I've added backwards compatibility for *Nushell* <= v0.90.

stevenxxiu:
With the code being backwards compatible, this could probably be reviewed and merged before the *Nushell* release. If there's not more breaking changes on this front. @ellie @fdncred

ellie (approved):
lgtm, thank you!

PR #1736, opened 2024-02-20
Update atuin.nu

WindSoilder:
Reflect to nushell pr: https://github.com/nushell/nushell/pull/11877
Since commandline with positional argument will be deprecated on nushell 0.91.0 version, and it uses `commandline edit -r` instead.

## Checks

I don't think we need to merge it until nushell 0.91.0 is coming out.  Just create a pr for reminding


ellie:
Dupe of https://github.com/atuinsh/atuin/pull/1735

Thanks though :)

PR #1740, opened 2024-02-20
fix(import/zsh-histdb): missing or wrong fields

Aloxaf:
This PR do two things:
1. Imports `exit_status` and `session` columns from zsh-histdb
2. Converts `duration` to nanoseconds for importing

## Checks

### DIFF
--- a/atuin-client/src/import/zsh_histdb.rs
+++ b/atuin-client/src/import/zsh_histdb.rs
@@ -32,9 +32,12 @@
 //                       duration int);
 //
 
+use std::collections::HashMap;
+use std::env;
 use std::path::{Path, PathBuf};
 
 use async_trait::async_trait;
+use atuin_common::utils::uuid_v7;
 use directories::UserDirs;
 use eyre::{eyre, Result};
 use sqlx::{sqlite::SqlitePool, Pool};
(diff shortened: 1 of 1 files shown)


atuin-client/src/import/zsh_histdb.rs
```diff
@@ -76,12 +82,9 @@ impl From<HistDbEntry> for History {
...
             )
             .duration(histdb_item.duration)
-            .hostname(
-                String::from_utf8(histdb_item.host)
-                    .unwrap_or_else(|_e| String::from(""))
-                    .trim_end()
-                    .to_string(),
-            );
+            .exit(histdb_item.exit_status)
+            .session(format!("{}:{}", hostname, histdb_item.session))
```
ellie, line 86: This isn't enforced (yet) by types, but by convention; Atuin uses UUIDs for the session ID. Specifically a uuidv7, but really any would work

Is it possible to ensure we do the same here?

The importer can group what it's importing by session, and then map that to a new session ID

Aloxaf:
> Just a couple of comments

Thanks for your suggestions. I have updated my code.

> Are there any other fields we are missing?

No more fields.
But I notice that `duration` should be measured in nanoseconds. Considering this is a minor bug, I also fixed it in this PR.

ellie (approved):
looks good! thank you

Issue #1742, opened 2024-02-20
[FR] Command to retro-actively apply history_filter on history


xvello:
As discussed offline, i have a lot of `kubectl delete` lines in my history after operations on the dev cluster. I really want to avoid executing these on a prod cluster, so I'd like to make sure they are never shown in the search.

I set `history_filter` to avoid capturing more of these, but I'd really like to remove them from the DB. I am aware that one can remove commands with ad-hoc calls to `atuin search --delete`, but I feel like a command to apply the filters would give me the incentive to keep adding to `history_filter` and avoid a repeat.

I'm happy to implement this once we agree on the design for it. My suggestion:

- a new `atuin history prune` subcommand will scan the whole history, match `history_filter` on it, and issue the appropriate deletions
- the subcommand can be extended later to pass arbitrary patterns, but by default will use the values from the config
- as a destructive operation, we'll want to allow to preview the changes before they are executed (to make sure a buggy patterns matches everything). We could either:
  - add a unix-style `-n` / `--dry-run` argument and default to delete (easier)
  - list the commands and ask for confirmation before deleting (and add a `-f` / `--force` argument to skip the confirmation)
- the most naive implementation would reuse `History::should_save`, so that all filtering config is applied without increasing maintenance cost
  - as a once-a-quarter operation, I think it's fine to have it be CPU-intensive and slow. We could try delegating the filtering to the `Database` but it would add more complexity, and possibly bugs if the regexp behaviours don't align perfectly 
  - we need to inspect every single history record. Calling `Database::list` with `unique = true` would give us this, at the cost of unbounded memory usage. It might be good enough for a first iteration
  - to keep the memory usage bounded, we'd need a new access method from `Database` that would allow to iterate or paginate. Alternatively, we can use `Database::range` on a user-provided timestamp range (default on last 7 days maybe?)

What are your thoughts @ellie?

ellie:
Hey! Thanks for writing this one up 😊 

> a new atuin history prune subcommand will scan the whole history, match history_filter on it, and issue the appropriate deletions

I think that sounds great to me, both in description + name

> the subcommand can be extended later to pass arbitrary patterns, but by default will use the values from the config

I'm not sure that would be the best idea, I'd rather keep deleting stuff elsewhere. Prune makes sense for enforcing config, and it would be great to keep the "edit your config" incentive

> as a destructive operation, we'll want to allow to preview the changes before they are executed...

[...]

PR #1743, opened 2024-02-20
feat: add history prune subcommand

xvello:
Closes https://github.com/atuinsh/atuin/issues/1742 by implementing the new `atuin history prune`  subcommand discussed in this issue. This command defaults to removing matching lines, which can be previewed via the `--dry-run` command flag.

TODO:

- [ ] tests: tested manually without sync. I am reusing the `search --delete` code path, so remote deletes should work? I'd like to add a black-box test in `atuin/tests/` to add a few history entries then prune part of them, but is it desirable?
- [ ] confirmation prompt: we could add a confirmation prompt before we delete the entries. Are there similar prompts I could reuse? Feel free to directly push the code for it.
- [ ] documentation: in the https://github.com/atuinsh/docs repo?

## Checks

### DIFF
--- a/atuin/src/command/client/history.rs
+++ b/atuin/src/command/client/history.rs
@@ -109,6 +109,13 @@ pub enum Cmd {
     },
 
     InitStore,
+
+    /// Delete history entries matching the configured exclusion filters
+    Prune {
+        /// List matching history lines without performing the actual deletion.
+        #[arg(short = 'n', long)]
+        dry_run: bool,
+    },
 }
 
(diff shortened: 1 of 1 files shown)


ellie (approved):
Excellent! Thank you!

If we could also get an update to the docs, that would be greatly appreciated 🙏 

https://github.com/atuinsh/docs

xvello:
> If we could also get an update to the docs, that would be greatly appreciated 🙏

Will do

PR #1744, opened 2024-02-21
feat(client): add config option keys.scroll_exits

### LINKED ISSUE #922: Feature request: Make exit-on-down-arrow behavior optional
If I scroll to the bottom of the results list and press down, Atuin closes (as implemented in #659). This would make perfect sense if I was using the up-arrow binding to seed the search with what I already typed at the prompt, but I disabled Atuin's up-arrow binding in favor of the one built into my shell.

In my case, I've pressed CTRL+R and typed a search, then scrolled up and down in the list and accidentally exited Atuin and lost my search.

Could this behavior be made optional, or maybe not trigger based on what method was used to activate Atuin?

tessus:
If the config option is set the `false`, using the up/down key won't exit the TUI when scrolled past the first/last entry.

Example:

```
[keys]
scroll_exits = false
```

The default is `true`, which is the current behavior.

/ref https://forum.atuin.sh/t/add-option-to-stop-down-arrow-exiting-search/94

fixes #922

## Checks

### DIFF
--- a/atuin/src/command/client/search/interactive.rs
+++ b/atuin/src/command/client/search/interactive.rs
@@ -225,7 +225,7 @@ impl State {
         is_down: bool,
     ) -> InputAction {
         if is_down {
-            if enable_exit && self.results_state.selected() == 0 && settings.keys.scroll_exits {
+            if settings.keys.scroll_exits && enable_exit && self.results_state.selected() == 0 {
                 return Self::handle_key_exit(settings);
             }
             self.scroll_down(1);
--- a/atuin/src/command/client/search/interactive.rs
+++ b/atuin/src/command/client/search/interactive.rs
@@ -225,7 +225,7 @@ impl State {
         is_down: bool,
     ) -> InputAction {
         if is_down {
-            if enable_exit && self.results_state.selected() == 0 {
+            if enable_exit && self.results_state.selected() == 0 && settings.keys.scroll_exits {
                 return Self::handle_key_exit(settings);
             }
             self.scroll_down(1);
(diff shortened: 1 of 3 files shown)


atuin/src/command/client/search/interactive.rs
```diff
@@ -225,7 +225,7 @@ impl State {
         is_down: bool,
     ) -> InputAction {
         if is_down {
-            if enable_exit && self.results_state.selected() == 0 {
+            if enable_exit && self.results_state.selected() == 0 && settings.keys.scroll_exits {
```
akinomyoga, line 228:
```suggestion
            if settings.keys.scroll_exits && enable_exit && self.results_state.selected() == 0 {
```
It's logically equivalent, but we can check the setting before entering the implementation for `scoll_exits`. The check `enable_exit && self.results_state.selected()` is a part of the implementation of `scroll_exits`.

akinomyoga (commented):
The config `scroll_exists` should also be mentioned in [`atuin-client/config.toml`](https://github.com/atuinsh/atuin/blob/main/atuin-client/config.toml).

atuin/src/command/client/search/interactive.rs
tessus: yes, this makes sense.

ellie (approved):
LGTM, thank you!

Could we also get an update to the docs please? https://github.com/atuinsh/docs

tessus:
> Could we also get an update to the docs please? https://github.com/atuinsh/docs

Sure, will do in a sec.

> Seeing as this is your first time contributing, if you would like a holographic contributors-only Atuin sticker, then please fill out this

I'd be delighted. Thank you.

PR #1745, opened 2024-02-21
feat: support regex with r/.../ syntax

lilydjwg:
## Checks

As discussed in https://forum.atuin.sh/t/a-regex-search-mode/169/16.

I hope there are no missed edge cases.


ellie:
Looks good overall, but would it also be possible to get some tests (given the complexity of this) please? 

You should be able to start with copying one of the tests at the bottom of the same file

Thanks!

PR #1748, opened 2024-02-21
Clarify operation result for working with aliases

Adda0:
This PR tries to clarify the info messages when working with aliases. No functional changes were made.

The delete message mentions which alias with which value was deleted so that the user can confirm that this is indeed the alias they wanted to remove. If not, they can reset it again. In case of an error, the error message format of `rm` is followed: `problem: reason` (`rm: cannot remove 'file': No such file or directory`).

The set alias message clearly separates the beginning and end of alias `name=value` from the normal text. Otherwise, one needs to parse the `name=value with name=value2` where `with` blends in with the end of the `value`.
```
$ atuin config alias set m "make run"
# Change this:
Overwriting alias m=make with m=make run
# to this:
Overwriting alias 'm=make' with 'm="make run"'.
```

If the changes are not to your liking, feel free to modify or close the PR.

## Checks

### DIFF
--- a/atuin/src/command/client/config/alias.rs
+++ b/atuin/src/command/client/config/alias.rs
@@ -19,10 +19,10 @@ impl Cmd {
         let found: Vec<Alias> = aliases.into_iter().filter(|a| a.name == name).collect();
 
         if found.is_empty() {
-            println!("Aliasing {name}={value}");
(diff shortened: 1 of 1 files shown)


ellie (approved):
works for me! thanks

PR #1751, opened 2024-02-22
remove shell history related passage

fakeid30:
## Checks

Since atuin removed the ability to generate github like activity graph, the passage related to whether any server is in use is redundant. Thus, I removed it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,10 +53,6 @@ As well as the search UI, it can do things like this:
 atuin search --exit 0 --after "yesterday 3pm" make
 ```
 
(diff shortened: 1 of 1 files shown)


ellie:
How is it redundant? 

The activity graph was a fun extra that I literally just need to rewrite. It's never been the main point of the server

ellie:
> Since atuin removed the ability to generate github like activity graph, the passage related to whether any server is in use is redundant. Thus, I removed it.

I'm not sure what you mean by this? It remains true, regardless of activity graphs.

Fwiw, I am planning on re-introducing the activity graphs, just not in their old form.

Issue #1754, opened 2024-02-22
Ability to remove headers?


digital-cuttlefish:
Hello folks!
I just discovered Atuin and I quite like it!
I'm trying to make it very compact, and I was wondering if we could remove some of the headers
<img width="1496" alt="Untitled" src="https://github.com/atuinsh/atuin/assets/141751473/562ad552-f745-4e89-85bf-e4f09cf54f91">

Mainly the one in the red rectangle (because it doesn't provide much utility), but if we could also remove the blue one that'd be super.

ellie:
You can do this with `show_help = false` in your config file!

Hope that helps

digital-cuttlefish:
Thanks for the quick reply!
Could we also get something to hide the tabs please :)

https://github.com/atuinsh/atuin/blob/cf2cbd23f03286e8e7d12f47fec1023dd119106c/atuin/src/command/client/search/interactive.rs#L502-L527

PR #1755, opened 2024-02-22
Add config setting for showing tabs

digital-cuttlefish:
From https://github.com/atuinsh/atuin/issues/1754

Yet to do local testing, but this is what I had in mind.

## Checks

### DIFF
--- a/atuin-client/config.toml
+++ b/atuin-client/config.toml
@@ -122,6 +122,9 @@
 ## amount of commands in your history.
 # show_help = true
 
+## Configure whether or not to show tabs for search and inspect
+# show_tabs = true
+
 ## Defaults to true. This matches history against a set of default regex, and will not save it if we get a match. Defaults include
 ## 1. AWS key id
 ## 2. Github pat (old and new)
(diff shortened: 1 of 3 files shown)


tessus:
I like the idea, but I think the name of the config option is a bit off. It sounds that atuin won't show any tabs anymore. If you don't read the explanation that is.

I'd opt for a name like `show_tab_info` or something along those lines. But that's just me.

digital-cuttlefish:
Oh, I actually wanted the tabs themselves to be hidden.
Where should I look for that?

tessus:
No, your code is perfectly fine afaik by looking at it. I haven't tested it, but it looks fine.
I only mentioned that the option name you used `show_tabs` is a bit ambiguous - at least for me. But maybe I am wrong. Let's see what the devs have to say.

ellie (approved):
I get @tessus' point about show_tabs being a little ambiguous, but I'm happy with this as-is due to how it matches nicely with show_help.

If you could also make a PR to the docs, that would be great! https://github.com/atuinsh/docs

digital-cuttlefish:
https://github.com/atuinsh/docs/pull/47

PR #1760, opened 2024-02-22
feat: Add '/', '?', and 'I' bindings to vim-normal mode

mwpastore:
Add 'I' binding to vim-normal mode (a la 'A' introduced in #1697) to jump into vim-insert mode at the beginning of the search input.

Also add '/' and '?' bindings to vim-normal mode to clear the search input and jump into vim-insert mode. This mimics the UX in e.g. `set -o vi` (bash) or `bindkey -v` (zsh) mode when you are using 'k' and 'j' to browse history lines and can type '/' or '?' to start a new search. (In a perfect world it would target the search in the forward or backward range starting at your current position in the history, but this is a reasonable first step.)

## Checks

### DIFF
--- a/atuin/src/command/client/search/interactive.rs
+++ b/atuin/src/command/client/search/interactive.rs
@@ -265,6 +265,18 @@ impl State {
         // First handle keymap specific keybindings.
         match self.keymap_mode {
             KeymapMode::VimNormal => match input.code {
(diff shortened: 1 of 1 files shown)


ellie (approved):
Looks good to me! Thank you

PR #1768, opened 2024-02-23
fix(client): no panic on empty inspector

### LINKED ISSUE #1713: atuin crashes on "Inspect" with empty history
This edge case could probably be handled more elegantly so that `atuin` doesn't crash with a cryptic error message.

# Steps to reproduce

1. `TMPDIR=$(mktemp -dp.) && cd $TMPDIR`
2. `^R` to start `atuin`
3. Toggle "Directory" mode
   - NOTE: Because we're in a new directory, "Directory" mode will show no entries.
5. `^O` to enter the "Inspect" panel

`atuin` should crash with the following error:

```
thread 'main' panicked at atuin/src/command/client/search/interactive.rs:970:35:
index out of bounds: the len is 0 but the index is 0
```

# Environment

macOS 14.3.1 (23D60)

```
❯ atuin --version
atuin 18.0.1
❯ uname -a
Darwin M-AI813782 23.3.0 Darwin Kernel Version 23.3.0: Wed Dec 20 21:30:44 PST 2023; root:xnu-10002.81.5~7/RELEASE_ARM64_T6000 arm64
```

tessus:
atuin no longer panics, if the inspector is invoked from an empty search window.

image

fixes #1713

## Checks

### DIFF
--- a/atuin/src/command/client/search/interactive.rs
+++ b/atuin/src/command/client/search/interactive.rs
@@ -581,7 +581,7 @@ impl State {
 
             1 => {
                 if results.is_empty() {
-                    let message = Paragraph::new("Nothing to inspect. :-(")
+                    let message = Paragraph::new("Nothing to inspect")
                         .block(
                             Block::new()
                                 .title(
--- a/atuin/src/command/client/search/interactive.rs
+++ b/atuin/src/command/client/search/interactive.rs
@@ -39,7 +39,7 @@ use ratatui::{
     prelude::*,
     style::{Color, Modifier, Style},
     text::{Line, Span, Text},
-    widgets::{Block, BorderType, Borders, Paragraph, Tabs},
+    widgets::{block::Title, Block, BorderType, Borders, Padding, Paragraph, Tabs},
     Frame, Terminal, TerminalOptions, Viewport,
 };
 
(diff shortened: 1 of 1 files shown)


atuin/src/command/client/search/interactive.rs
```diff
@@ -580,12 +580,26 @@ impl State {
             }
 
             1 => {
-                super::inspector::draw(
-                    f,
-                    results_list_chunk,
-                    &results[self.results_state.selected()],
-                    &stats.expect("Drawing inspector, but no stats"),
-                );
+                if results.is_empty() {
```
ellie, line 583: Would it be possible to shift this logic into the inspectors draw call please? That is, within `super::inspector::draw`

atuin/src/command/client/search/interactive.rs
```diff
@@ -580,12 +580,26 @@ impl State {
...
+                    let message = Paragraph::new("Nothing to inspect. :-(")
```
ellie, line 584:
```suggestion
                    let message = Paragraph::new("Nothing to inspect")
```

ellie (approved):
Thank you!

atuin/src/command/client/search/interactive.rs
tessus: Oops, sorry. orhun uses `:(` in some of the messages in rustypaste. I kinda liked it and it must have stuck...

tessus:
Thanks for making the changes. I was in bed fast asleep.

PR #1770, opened 2024-02-25
fix(nushell): readd up-arrow keybinding, now with menu handling

### LINKED ISSUE #1025: nushell: atuin should not always trigger on ⬆️

remmycat:
Fixes #1025

This seems to work, see the nushell docs for `until`, and how it can account for the menu handling: https://www.nushell.sh/book/line_editor.html#until-type

## Checks

### DIFF
--- a/atuin/src/command/client/init.rs
+++ b/atuin/src/command/client/init.rs
@@ -56,20 +56,23 @@ impl Cmd {
     )
 )";
             const BIND_UP_ARROW: &str = r"
-# The up arrow keybinding has surprising behavior in Nu, and is disabled by default.
-# See https://github.com/atuinsh/atuin/issues/1025 for details
-# $env.config = (
-#     $env.config | upsert keybindings (
-#         $env.config.keybindings
-#         | append {
-#             name: atuin
-#             modifier: none
-#             keycode: up
-#             mode: [emacs, vi_normal, vi_insert]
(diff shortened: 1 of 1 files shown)


remmycat:
Since it's easy to test out, we should probably wait a little for feedback confirming that this fixes the issue 🙂

arcuru:
Awesome! I've confirmed that this snippet works as well.

This probably needs a small matching PR in https://github.com/atuinsh/docs to remove https://github.com/atuinsh/docs/blob/a814efba030b985942c09c215122f655f36272c0/src/content/docs/configuration/key-binding.mdx#L52

I'll also ping https://github.com/nushell/nushell/issues/10414 so there is a link to the solution :)

remmycat:
I see - added the docs PR :)

ellie (approved):
Looks good! Thank you!

Thanks for testing @arcuru 🙏 

Issue #1153, opened 2023-08-06
colors should be configurable


leonard-slass:
Hello,

The blue and red are too dark on my terminal. I have a basic xterm with black background. My `TERM` variable is set to `xterm-256color`.

![image](https://github.com/atuinsh/atuin/assets/68689230/402ed850-8077-422b-afc0-a413feb7f75a)

I have patched the colors in the code:

```
@@ -136,7 +136,7 @@ impl DrawState<'_> {
         let status = Style::default().fg(if h.success() {
             Color::Green
         } else {
-            Color::Red
+            Color::Yellow
         });
         let duration = Duration::from_nanos(u64::try_from(h.duration).unwrap_or(0));
         self.draw(&format_duration(duration), status);
@@ -144,7 +144,7 @@ impl DrawState<'_> {
 
     #[allow(clippy::cast_possible_truncation)] // we know that time.len() will be <6
     fn time(&mut self, h: &History) {
-        let style = Style::default().fg(Color::Blue);
+        let style = Style::default().fg(Color::Cyan);
 
         // Account for the chance that h.timestamp is "in the future"
         // This would mean that "since" is negative, and the unwrap here
@@ -164,7 +164,7 @@ impl DrawState<'_> {
     fn command(&mut self, h: &History) {
         let mut style = Style::default();
         if self.y as usize + self.state.offset == self.state.selected {
-            style = style.fg(Color::Red).add_modifier(Modifier::BOLD);
+            style = style.fg(Color::Yellow).add_modifier(Modifier::BOLD);
         }
 
         for section in h.command.split_ascii_whitespace() {
```

This gives me better color, but ideally, it should be possible to do so in the configuration file.

![image](https://github.com/atuinsh/atuin/assets/68689230/3609b5b8-6d82-4668-939e-5a5992e3221d)

Have a good day.

tpoliaw:
This sounds like a terminal configuration issue as much as an atuin one - do you have similar problems from other applications? The red/yellow/blue used in the code are the 16 terminal base colours so you should be able to customise them as you want using `.Xresources`. One guide for this [for xterm is here](https://www.futurile.net/2016/06/15/xterm-256color-themes-molokai-terminal-theme/).

leonard-slass:
[...]

PR #1782, opened 2024-02-26
Custom Styling

kolektiv:
As mentioned in #1153 - I've had a quick little play at a simple way for people to set styles. I've tried to make it fairly minimal rather than introduce a whole theme system or something similar. Some things to note:
* I've just used Ratatui colors as a basic component.
    * That means that atuin-client now needs a dependency on Ratatui (but I don't think that's an issue; it's a dependency of crates which already have it).
    * Ratatui color deserialization was very basic indeed (only named colors) until 0.26, after which it supports hex (RGB) and indexed colors, so I've moved Ratatui to a workspace dependency and bumped the version (haven't tested for side effects!)
* Styles can be set under a new `styles` table in the config. I've currently implemented support for two styles, just as a proof - the command in the history list, and the selected command in the history list. They can be set something like the following (obviously, compact table notation is also fine):
```toml
[styles]
command = "#555555"

[styles.command_selected]
bold = true
foreground = "#90fa00"
```
* Style values can be of two kinds (as seen above):
    * A simple color value, in which case it will be treated as the foreground color.
    * A table of style properties which make up a style, and which consists of...
        * foreground (color)
        * background (color)
        * underline (color)
        * bold (bool)
        * crossed_out (bool)
        * italic (bool)
        * underlined (bool)
    * Properties which are not set will not be set on the style (there is currently no inheritance, etc.)

Here's how that particular configuration looks in my terminal (using the compact style).

Screenshot 2024-02-26 at 16 46 51

Opinions/objections are welcome. Obviously, a couple of slight structural tweaks are needed, but I've tried to avoid anything too invasive.

## Checks

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -248,6 +248,7 @@ dependencies = [
  "parse_duration",
  "pretty_assertions",
  "rand",
+ "ratatui",
  "regex",
  "reqwest",
  "rmp",
@@ -563,6 +564,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "df8670b8c7b9dae1793364eafadf7239c40d669904660c5960d74cfd80b46a53"
 
 [[package]]
+name = "castaway"
+version = "0.2.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8a17ed5635fc8536268e5d4de1e22e81ac34419e5f052d4d51f4e01dcc263fcc"
+dependencies = [
+ "rustversion",
+]
+
+[[package]]
 name = "cc"
 version = "1.0.83"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


kolektiv:
Hold off on reviewing this, changes have been lost on the way...

kolektiv:
OK this is pretty representative of the proposed approach now I think, seems fairly straightforward.

tessus:
I think this needs a rebase.

kolektiv:
> I think this needs a rebase.

Yes, probably so @tessus - I'm very out of practice with git and forks these days, if you let me know what to run I'll do it!

tessus:
in the main branch you have to run `git pull upstream main` (replace `upstream` with the name of the remote for the atuin repo, not your fork).
Then switch to your branch and do a `git rebase main`
After that a `git push -f` is required.

That should do it.

tessus:
Hmm, I just tried myself and something seems off. You removed code that is currently in main. A rebase won't fix this.

P.S.: Ah, you also didn't create your own branch. So that could explain it. I am afraid the PR can't be merged as is. You might have to interactively rebase and remove the commit where you merged main into your main. Let me try to see, if I can fix this.

kolektiv:
Ah hmm - I let GitHub do some kind of sync - I'll look into it in the morning, and see if I can work out how to fix it up.

tessus:
I tried several things including cherry-picking your commits. I think one would have to go through all your commits and undo the removal of the code you did.

tessus:
342384593dd662ec757ebca1bb5a7ee25c5c2f0c is the commit that removes code that shouldn't be removed

kolektiv:
I'll close this pull request and create a non-broken one for simplicity!

PR #1784, opened 2024-02-26
feat(stats): add linux sysadmin commands to common_subcommands

tessus:
I've been using atuin on my Linux box for some time now and I have noticed that a few commands that are used rather often are not in the list of common_subcommands.
This change adds these commands to the list.

see also https://github.com/atuinsh/docs/pull/38

## Checks


ellie (approved):
thanks!

PR #1785, opened 2024-02-27
Custom Styles

kolektiv:
(Recreated PR!)

As mentioned in #1153 - I've had a quick little play at a simple way for people to set styles. I've tried to make it fairly minimal rather than introduce a whole theme system or something similar. Some things to note:
* I've just used Ratatui colors as a basic component.
    * That means that atuin-client now needs a dependency on Ratatui (but I don't think that's an issue; it's a dependency of crates which already have it).
    * Ratatui color deserialization was very basic indeed (only named colors) until 0.26, after which it supports hex (RGB) and indexed colors, so I've moved Ratatui to a workspace dependency and bumped the version (haven't tested for side effects!)
* Styles can be set under a new `styles` table in the config. I've currently implemented support for two styles, just as a proof - the command in the history list, and the selected command in the history list. They can be set something like the following (obviously, compact table notation is also fine):
```toml
[styles]
command = "#555555"

[styles.command_selected]
bold = true
foreground = "#90fa00"
```
* Style values can be of two kinds (as seen above):
    * A simple color value, in which case it will be treated as the foreground color.
    * A table of style properties which make up a style, and which consists of...
        * foreground (color)
        * background (color)
        * underline (color)
        * bold (bool)
        * crossed_out (bool)
        * italic (bool)
        * underlined (bool)
    * Properties which are not set will not be set on the style (there is currently no inheritance, etc.)

Here's how that particular configuration looks in my terminal (using the compact style).

Screenshot 2024-02-26 at 16 46 51

Opinions/objections are welcome. Obviously, a couple of slight structural tweaks are needed, but I've tried to avoid anything too invasive.

## Checks

### DIFF
--- a/atuin/src/command/client/search/interactive.rs
+++ b/atuin/src/command/client/search/interactive.rs
@@ -189,11 +189,15 @@ impl State {
         }
 
         let ctrl = input.modifiers.contains(KeyModifiers::CONTROL);
+        let esc_allow_exit = !(self.tab_index == 0 && self.keymap_mode == KeymapMode::VimInsert);
 
         // core input handling, common for all tabs
         match input.code {
             KeyCode::Char('c' | 'g') if ctrl => return InputAction::ReturnOriginal,
-            KeyCode::Esc if !(self.tab_index == 0 && self.keymap_mode == KeymapMode::VimInsert) => {
+            KeyCode::Esc if esc_allow_exit => {
+                return Self::handle_key_exit(settings);
+            }
+            KeyCode::Char('[') if ctrl && esc_allow_exit => {
                 return Self::handle_key_exit(settings);
             }
             KeyCode::Tab => {
@@ -320,7 +324,7 @@ impl State {
                 _ => {}
             },
             KeymapMode::VimInsert => {
-                if input.code == KeyCode::Esc {
+                if input.code == KeyCode::Esc || (ctrl && input.code == KeyCode::Char('[')) {
                     self.set_keymap_cursor(settings, "vim_normal");
                     self.keymap_mode = KeymapMode::VimNormal;
                     return InputAction::Continue;


kolektiv:
Just as an extra note if this is agreed as an approach, next steps would presumably be
* Implement the rest of the style-able features (extend the Styles type, etc. and any drawing functions which currently create styles)
* Create some basic docs?

atuin-client/Cargo.toml
```diff
@@ -54,6 +54,7 @@ futures = "0.3"
 crypto_secretbox = "0.1.1"
 generic-array = { version = "0.14", features = ["serde"] }
 serde_with = "3.5.1"
+ratatui = { workspace = true }
```
ellie, line 57: I'm not a _huge_ fan of this overall, but I think it's ok for now. Totally out of scope for this PR, but I'd like to reorg our crates a bit anyway.

atuin-client/src/settings.rs
```diff
@@ -303,6 +304,107 @@ impl Default for Stats {
     }
 }
 
+#[derive(Clone, Debug, Default, Deserialize)]
+pub struct Styles {
```
ellie, line 308: I know this isn't the only thing here, but would you be open to the following

1. Split settings into a module
2. Break the style stuff into it's own file

Don't worry about the rest for now, but `settings.rs` is getting fairly unwieldy now imo!

ellie:
I like this approach, thank you! Just a few comments, nothing major though

ellie:
Oh one other thought

It would be neat if we could also style the UI beyond colours this way. Like the differences between compact/full

EG

```toml
[styles.search_box]
border = none
etc
```

that way, full/compact could just be presets. Not at all required for this PR though

kolektiv:
> It would be neat if we could also style the UI beyond colours this way. Like the differences between compact/full

Yeah it would. It would be a bit of an expansion - at the moment the contents of styles maps basically 1-1 to the Ratatui type, and if you started to include a more general concept of style it would be more of a theme than a set of styles, but I agree that would be a good direction. I'll have a think about that one - it may be that something like `theme = layout styles + text styles` is a good rule of thumb, and would be a good evolution from here.

atuin-client/src/settings.rs
ellie: > When you say another module, are you thinking about a settings.rs as a root, containing only top-level construction-type code and then a settings/ dir containing things like styles.rs for example?

Exactly that!

And yes, the re-org I was considering is more-or-less what you say. However, I'd also like to break all "history" specific stuff out into an `atuin-history` crate. Then, we can make the main `atuin` binary just connect a few things together. Kinda like `atuin-config`.

Still a bit of a fuzzy idea though, and I'd like to see how other things shake out first. I'd also like to consider moving to KDL for config.

(gone way beyond the scope of this PR now, but just context setting)

kolektiv: That makes sense, I'll have a quick play around and see how dramatic it would be to include that in this set of changes. Just taking the struct members out shouldn't mean any changes to consumers, if we wanted to group some of the current settings into new types (e.g. paths) then code consumers would need updating and that might get quite extensive (shouldn't have a problem with the config file itself if new structs are serde flattened). Will see how it looks!

kolektiv: Ok, I've started playing with a bit of a settings refactor, which can be seen in the latest commit. What I've done for now...
* Pulled groups of settings out into submodules of settings with vaguely sensible names (I hope) and any specialized types where relevant
* Re-exported those types from the original settings module, to avoid needing to change other code

What I've also done for one of those sets of properties (the `display` submodule) is to pull the properties into a new settings type at that submodule level, as well as a function for setting defaults for that submodule on the `ConfigBuilder` instance. I've then replaced the raw properties on settings with the new settings type, but with a `serde` attribute to flatten the type, so the structure of the actual config file is unchanged. That obviously does need some changes elsewhere, so those changes are reflected wherever there is now a "`.display`" included in a settings path.

The new submodule settings type is a potential option, just to see how well that fits in. I don't have any strong opinions on that one, see what you think and whether it's worth the slight extra code.

atuin-client/src/settings/display.rs
```diff
@@ -0,0 +1,161 @@
...
+
+use config::{builder::DefaultState, ConfigBuilder, Value, ValueKind};
+use eyre::Result;
+use ratatui::style::{Color, Style, Stylize};
+use serde::{Deserialize, Deserializer};
+
+// Settings
+
+#[derive(Clone, Debug, Deserialize)]
+pub struct Settings {
```
kolektiv, line 11: This is the new settings type mentioned...

atuin-client/src/settings/display.rs
```diff
@@ -0,0 +1,161 @@
...
+    pub show_help: bool,
+    #[serde(alias = "display")]
+    pub style: Display,
+    #[serde(default)]
+    pub styles: Styles,
+}
+
+// Defaults
+
+pub(crate) fn defaults(
```
kolektiv, line 26: ...along with the a simple function to set defaults.

atuin-client/src/settings.rs
```diff
@@ -1,374 +1,110 @@
...
 use semver::Version;
 use serde::Deserialize;
-use serde_with::DeserializeFromStr;
-use time::{
-    format_description::{well_known::Rfc3339, FormatItem},
-    macros::format_description,
-    OffsetDateTime, UtcOffset,
-};
+use time_lib::{format_description::well_known::Rfc3339, Duration, OffsetDateTime};
 use uuid::Uuid;
```
kolektiv, line 29: Exporting all of the moved types saves making changes elsewhere for now...

atuin-client/src/settings.rs
```diff
@@ -1,374 +1,110 @@
...
-    pub show_help: bool,
-    pub exit_mode: ExitMode,
-    pub keymap_mode: KeymapMode,
-    pub keymap_mode_shell: KeymapMode,
-    pub keymap_cursor: HashMap<String, CursorStyle>,
-    pub word_jump_mode: WordJumpMode,
-    pub word_chars: String,
-    pub scroll_context_lines: usize,
-    pub history_format: String,
-    pub prefers_reduced_motion: bool,
```
kolektiv, line 54: This is where the new display settings type actual gets used, flattened, rather than having a set of properties here

atuin/src/command/client/search.rs
```diff
@@ -158,7 +158,7 @@ impl Cmd {
             settings.filter_mode = self.filter_mode.unwrap();
         }
         if self.inline_height.is_some() {
-            settings.inline_height = self.inline_height.unwrap();
+            settings.display.inline_height = self.inline_height.unwrap();
```
kolektiv, line 161: Any display settings are now a level deeper because of the new settings type, even though it's not visible externally in config.

atuin-client/src/settings.rs
ellie: pretty reasonable imo!

ellie:
um ok what github?

Could you reopen this please, but use a branch that is not main? It behaves very strangely in terms of permissions.

The rebased code is here: https://github.com/atuinsh/atuin/tree/kolektiv/styles

I could open the PR, but would rather it be credited to you of course.

---

tldr is that GitHub allows me to push to your branch, if you open a PR. I rebased this locally, pushed, and it just overwrote your main with the current main 🙃 ffs. I also now don't have permission to sort this out, as the PR was auto-closed and cannot be reopened.

Sorry about 😭

ellie:
Otherwise - thanks for sorting this out. I'm pretty happy with the approach here!

I'm a little unsure on the submodule `Settings` type, but I think it's ok for now and tidier than what we have.

The only other thing I'd like, is to ensure that a variety of already existing config files still load OK. 

This doesn't necessarily need to be done here, as we have broken this a few times in the past. So perhaps a wider issue of config testing

kolektiv:
Hah, I'm glad it's not only me that regularly gets tripped up by GitHub when it decides to be "helpful". No problem, what I'll do is grab that branch in my fork and work on there, and see where we end up. I'll go through and add in a "complete" (hopefully) set of styles.

I'm also not 100% sure on the submodules `Settings` type, but I think it's marginally neater, and it can always be done away with if/when a new approach is taken to config, so I'll go ahead and do it for each submodule. There "shouldn't" (TM) be any changes to config file reading like this, but I completely agree that actually checking that is probably sensible!

I had a look at KDL incidentally, as you mentioned it somewhere above - I'd not come across it before. Interesting. My only personal pain from that is that all my config is generated by nix, and I don't think it can generate KDL at the moment, but that's clearly a pretty selfish reason not to do it if it makes sense!

joaomendoncaa:
Sorry for the ping everyone but any advancements on this initiative @kolektiv? Excited about it!

kolektiv:
Yes, I'm going to try and pick it back up this weekend! I ended up travelling at short notice, so I'll put soldering together from here 🙂

xhalo32:
Hi, I added some more styling options and merged with the main branch in [my fork](https://github.com/xhalo32/atuin/tree/xhalo32/styles).

As this feature still requires some work and is now entangled with the settings refactoring, should we split the settings refactoring part into its own PR?

joaomendoncaa:
> Hi, I added some more styling options and merged with the main branch in [my fork](https://github.com/xhalo32/atuin/tree/xhalo32/styles).
> 
> As this feature still requires some work and is now entangled with the settings refactoring, should we split the settings refactoring part into its own PR?

If they are not mutually dependent on each other I'd love if the custom styles could be merged even if settings refactoring is not complete. Thanks for all the work.

PR #1787, opened 2024-02-27
feat: add `CTRL+[` binding as `<Esc>` alias

liljaylj:
in Vim/Neovim there is alias for `<Esc>` mapping -- `<C-[>`. it is useful if your `<Esc>` key is hard to hit on your keyboard.

for reference:

- https://vimhelp.org/insert.txt.html#i_CTRL-%5B
- https://neovim.io/doc/user/insert.html#i_CTRL-%5B

## Checks

### DIFF
--- a/atuin/src/command/client/search/interactive.rs
+++ b/atuin/src/command/client/search/interactive.rs
@@ -189,11 +189,15 @@ impl State {
         }
 
         let ctrl = input.modifiers.contains(KeyModifiers::CONTROL);
+        let esc_allow_exit = !(self.tab_index == 0 && self.keymap_mode == KeymapMode::VimInsert);
 
         // core input handling, common for all tabs
         match input.code {
             KeyCode::Char('c' | 'g') if ctrl => return InputAction::ReturnOriginal,
-            KeyCode::Esc if !(self.tab_index == 0 && self.keymap_mode == KeymapMode::VimInsert) => {
+            KeyCode::Esc if esc_allow_exit => {
+                return Self::handle_key_exit(settings);
+            }
+            KeyCode::Char('[') if ctrl && esc_allow_exit => {
                 return Self::handle_key_exit(settings);
             }
             KeyCode::Tab => {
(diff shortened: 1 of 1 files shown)


tessus:
2 questions:

1. on which keyboard is `Esc` hard to hit - just asking out of curiosity because I have only seen the `Esc` key in the top left corner: us-en, german, korean, spanish, ...

2. does this shortcut work in terminal multiplexers?

akinomyoga:
> 1. on which keyboard is `Esc` hard to hit - just asking out of curiosity because I have only seen the `Esc` key in the top left corner: us-en, german, korean, spanish, ...

I actually always use <kbd>C-[</kbd> to send <kbd>ESC</kbd> because I have difficulty in pressing <kbd>ESC</kbd>. Only when I use the keyboard of a laptop machine, without breaking the home position, I can reach <kbd>ESC</kbd> with my little finger. On all the keyboards of the full size, it's impossible for me to reach the <kbd>ESC</kbd> key.

> 2. does this shortcut work in terminal multiplexers?

This PR is only relevant when an advanced keyboard protocol is enabled. If the advanced keyboard protocol is enabled, it *should work better* in terminal multiplexers because the protocol is free from the heuristic and obscure distinction of Meta modifier vs <kbd>ESC</kbd>, which the terminal multiplexer needs to process. (But, I'm not sure `crossterm` supports enabling the advanced keyboard protocol of the outside terminal when terminal multiplexers are involved.)

akinomyoga (approved):
For the exactly same reasons as #1720, I agree to have this binding.

tessus:
Isn't there already `ctrl-c` and `ctrl-g` an alias to `Esc`? At least in my atuin, `ctrl-c` and `ctrl-g` exits atuin the same way as `Esc` does.

akinomyoga:
<kbd>ESC</kbd> and <kbd>C-[</kbd> are equivalent in the traditional terminal protocol, and we should assign the same functionality to both even in the advanced keyboard protocol unless we would drop the support for the traditional terminal protocols.

Given <kbd>C-c</kbd> and <kbd>C-g</kbd> already exit the Atuin session, another option would be to remove the keybinding to <kbd>ESC</kbd>, but that doesn't seem useful to me.

ellie (approved):
thank you!

Issue #798, opened 2023-03-22
Replace native up arrow more seamlessly


endigma:
Is it possible to replace the native up arrow in the shell not with the full screen TUI but with normal up arrow behavior, albeit with the atuin db?

endigma:
No, disabling *disables* the usage of the atuin history, I just want traditional up arrow behavior but **using the atuin history db**.

conradludgate:
We have a WIP to add an 'inline view' #648 but it's currently blocked on a few things. 

This UI is not final, but it would work a little like this

https://cdn.discordapp.com/attachments/954121165826326642/1054084181531054200/image.png

Might flip the list vertically and have the search at the top, closer to the prompt

conradludgate:
Ah I see what you mean now. I'll see if that can easily be achieved already

RafaelKr:
Interesting, I'm using zsh (with oh-my-zsh) with this config:
```
# Bind ctrl-r but not up arrow
eval "$(atuin init zsh --disable-up-arrow)"
```

My `.zsh_history` file is still used synchronously (history.db and .zsh_history are both filled) and the up arrow behaviour is working the same as before installing Atuin.

@endigma Which shell are you using?

endigma:
Up arrow does still work normally, it uses fish_history, not the atuin db.

patricksjackson:
I am fairly sure that you can now add this to the scripts using the `--offset` flag added in #825.

endigma:
Yep, that looks like about all that would be required to get this done, I'd still like to see it built in.

patricksjackson:
I whipped up a version of this for fish, since I've wanted this myself and that's the shell I use: https://github.com/ellie/atuin/compare/main...patricksjackson:atuin:replace-bind-up. Here's a link to the full file, that you can use in place of `atuin init fish` (note I added the bindings to the script, even though those are normally appended by `atuin`) https://github.com/patricksjackson/atuin/blob/d4c1c2b4137f39dcbfe57c9cf4c71d7614ea564b/src/shell/atuin.fish

It's not totally trivial but it does seem to work. Though there may be some rough edges compared to using the inbuilt `up-or-search` though. Pretty much the only benefit of doing it with atuin is that a user could [set the filter-mode](https://github.com/patricksjackson/atuin/blob/8db035bfcca07a70a78d42d3ed64887993b04816/src/shell/atuin.fish) if they wanted something other than `session` for the up key binding.

[...]

PR #1789, opened 2024-02-27
style: Reorganize config file

arcuru:
This is intended to be more of an example/discussion than a PR that should be merged. This PR format just may be a bit easier to make comments on, but we can move discussion somewhere else.

The config file is getting pretty unwieldy. Ellie has been pushing towards adding some more structure to it, but I think we need to re-organize some of the existing options.

There are a lot of contributions lately that are adding new options, I think this is a good time to make sure there is a cohesive plan to the options layout.

We have several logical groups of options that can pretty cleanly be grouped together. The most important is I think the keybindings. We can move the `--disable-up-arrow` option into the config file, more cleanly customize the filter/search/workspace modes, and have a switch for #798 if that gets implemented.

@ellie, feel free to close if you already had your own goals here :)

## Checks

### DIFF
--- a/atuin-client/config.toml
+++ b/atuin-client/config.toml
@@ -1,3 +1,5 @@
+version = 2
+
 ## where to store your database, default is your system data directory
 ## linux/mac: ~/.local/share/atuin/history.db
 ## windows: %USERPROFILE%/.local/share/atuin/history.db
@@ -22,20 +24,16 @@
 ## for example: "+9", "-05", "+03:30", "-01:23:45", etc.
 # timezone = "local"
 
-## enable or disable automatic sync
-# auto_sync = true
-
 ## enable or disable automatic update checks
 # update_check = true
 
-## address of the sync server
-# sync_address = "https://api.atuin.sh"
+## Timeout (in seconds) for acquiring a local database connection (sqlite)
+# local_timeout = 5
 
-## how often to sync history. note that this is only triggered when a command
-## is ran, so sync intervals may well be longer
-## set it to 0 to sync after every command
-# sync_frequency = "10m"
+## default history list format - can also be specified with the --format arg
+# history_format = "{time}\t{command}\t{duration}"
 
+# FIXME: These 3 options are a group, and might be worth adding to their own table??
 ## which search mode to use
 ## possible values: prefix, fulltext, fuzzy, skim
 # search_mode = "fuzzy"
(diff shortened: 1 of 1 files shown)


tessus:
I think adding more structure is a great idea. I can speak from experience that doing it without being a breaking change _can_ be challenging. Additionally I want to mention that allowing "old" config files to be read - for a longer period of time (e.g. several releases) - is quite the maintenance burden. 

I am not saying that one should just go cold turkey and introduce a new structure without an (automated) migration path.
What I am saying is that the migration path should be part of the discussion.

atuin-client/config.toml
```diff
@@ -22,20 +24,16 @@
...
-## enable or disable automatic sync
-# auto_sync = true
-
 ## enable or disable automatic update checks
 # update_check = true
 
-## address of the sync server
-# sync_address = "https://api.atuin.sh"
+## Timeout (in seconds) for acquiring a local database connection (sqlite)
+# local_timeout = 5
```
ellie, line 31: tbh I'd like to eliminate all top-level/root config items

atuin-client/config.toml
```diff
@@ -48,57 +46,68 @@
 ## in any directory within a git repository tree (default: false)
 # workspaces = false
 
-## which filter mode to use when atuin is invoked from a shell up-key binding
-## the accepted values are identical to those of "filter_mode"
-## leave unspecified to use same mode set in "filter_mode"
-# filter_mode_shell_up_key_binding = "global"
+[sync]
```
ellie, line 49: may also want 

```
enabled = true/false
```

some users seem to want confirmation that Atuin is offline.

atuin-client/config.toml
```diff
@@ -48,57 +46,68 @@
...
-## characters that count as a part of a word
-# word_chars = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
+# FIXME: Customize the keybinding used for the entrypoint from the shell
+# UNIMPLEMENTED, reserved for future use of custom shell keybindings
+# key = "up" 
 
-## number of context lines to show when scrolling by pages
-# scroll_context_lines = 1
+# Customize any option in the ui table here
+# [keybindings.ctrl_r.ui]
```
ellie, line 82: I like this way of describing things! Much neater.

ellie (commented):
Generally for this!

More thoughts

## 1. Nest everything under a history table. 

I've mentioned in another comment somewhere that I'd like to start moving history-specific implementations to a new crate (`atuin-history`) It would make sense for it to have its own history struct, which could go under its own table. 

Bearing in mind that at present there's also alias + kv syncing, which aren't yet widely used but this would allow for future flexibility.

## 2. Switch to KDL

Not necessarily a prerequisite, but I've been thinking about it more lately. IMO KDL seems much nicer for config than TOML is, and is much more flexible.

Spec: https://github.com/kdl-org/kdl
Rust implementation: https://github.com/kdl-org/kdl-rs

It's already used by Zellij, if you like an example: https://zellij.dev/documentation/configuration

TOML was something I never really thought about in the beginning, but in hindsight it's not the best choice. Nesting is non-obvious (order matters, rather than indentation). I also find large TOML config files difficult to read.

Another option could be something executable. EG, see how NeoVim can be configured with Lua. I can still see benefits there, but think that our config should be very quickly/easily parsed and not require executing. We can always have plugins in the future that can manipulate config.

## 3. Migration

Realistically, having v2 as a replacement rather than a toggle at the top may be better. 

The migration path is fairly smooth then, as we can install the v2 config with defaults automatically for all new users, and use it to override old config for existing users. It could potentially be auto-generated.

The requirement here would be that it has to live under a different path.

`config.kdl` could work if we change format, otherwise `settings.kdl` may also be nice.

atuin-client/config.toml
tessus: I also think that all config settings should belong to a `[section]`. If one is not sure, it's still possible to create a `Misc` or `Other` section. On the other hand, if it is not clear, maybe the config option is not useful in the first place (with exceptions of course).

YummyOreo:
What if we separated config for the tui and config for atuin itself too?

atuin-client/config.toml
```diff
@@ -108,10 +117,25 @@
 ## start with ^ or end with $, they'll match anywhere in CWD.
 ## For details on the supported regular expression syntax, see
 ## https://docs.rs/regex/latest/regex/#syntax
-# cwd_filter = [
+# cwd = [
```
tessus, line 120: `cmd` and `cwd` look very similar. Also, one doesn't have to write these more than once, thus I believe that the following are more descriptive (for non-experts):

`command` or `commands`
and
`directory` or `directories`

ellie:
Closing this one because it's been a draft for a long time! Please lmk if you'd like to reopen/continue discussion/etc! Thank you for the work, either way <3

arcuru:
Oops, yea I definitely forgot about this one sorry. I don't have plans to finish this.

PR #1791, opened 2024-02-28
feat(nushell): add nushell completion generation

remmycat:
As mentioned on Discord, I noticed that clap offers completion generation for nushell in a separate crate. I went ahead and implemented these for the existing gen-completions command.

Some extensive implementation notes:
- When installing the new dependency, I also updated the clap and clap_complete versions in the manifests. I thought it couldn't hurt to have the whole group up to date when installing a new part. Looking at the lock file changes, I believe it had these versions already installed anyway.
- I moved the gen-completions code into a new module, similar to the other commands. With the extra impl it took up quite a bit of space in `mod.rs` otherwise.
- After trying out a few variants of composing the two crates into one command, I landed on defining an enum ourselves and adding some repetitive glue code. It was the most ergonomic version I could come up with, but I'm happy for alternative suggestions
- **Question:** I'm not sure if the nushell generations are wanted in the release tarball? I just noticed it when searching for mentions of completions.
- Another bit of code for the completions that I wasn't sure about was in `atuin.nix`. But the function used there, `installShellCompletion`, seems to only support bash fish and zsh in this way, so I didn't touch it (also I have next to no experience with nix, so I wouldn't want to break anything).
- I tested the generated completions using the latest nushell version, `0.90.1`, and everything worked well for me. 
- I'm gonna have a look at preparing a corresponding docs update PR tomorrow, or latest on the weekend, at least to update the list of supported shells for completion generation. 

## Checks

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -154,7 +154,7 @@ jobs:
         esac;
 
         # Shell completions
-        for sh in 'bash' 'fish' 'zsh'; do
+        for sh in 'bash' 'fish' 'zsh' 'nushell'; do
           $QEMU_PREFIX "${{ steps.strip.outputs.BIN_PATH }}" gen-completions -s $sh -o "${ARCHIVE_DIR}/completions"
         done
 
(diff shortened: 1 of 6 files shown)


ellie (approved):
Looks good! thank you!

PR #1793, opened 2024-02-29
Custom Styles (inc. Settings refactor)

kolektiv:
WIP

Continuing previous pull request...

The Settings refactor is probably a little better than the previous growing flat structure, but it could do with a proper overhaul at some point I think. Still, it should do for now. This should be enough to try and test various configs - in theory it shouldn't have changed the format required, but that's not tested...

- [x] Settings Refactor
- [ ] Styles completed

## Checks

### DIFF
--- a/atuin-client/src/api_client.rs
+++ b/atuin-client/src/api_client.rs
@@ -25,7 +25,7 @@ use semver::Version;
 use time::format_description::well_known::Rfc3339;
 use time::OffsetDateTime;
 
-use crate::{history::History, sync::hash_str};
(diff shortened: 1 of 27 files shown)


ellie:
Looks good! Thanks. 

Did you intend to make this PR against main, or is there more to come?

kolektiv:
Definitely WIP - more to come on this one and then we can see about main 🙂

joaomendoncaa:
This is looking fantastic @kolektiv! 🙌

markovejnovic:
Closing this as it appears to be stale. I suspect that the theming support added in #2236 should be sufficient.

PR #1794, opened 2024-02-29
Guard against ancient versions of bash (3.2) where Atuin does not work.

fragmede:
Hi.

## Checks
- [✔️ ] I am happy for maintainers to push small adjustments to this PR, to speed up the review cycle
- [✔️ ] I have checked that there are no existing pull requests for the same thing
  * I searched issues for "bash 3.2" and got nothing relevant.

I'm on OS X, version `Sonoma 14.2.1 (23C71)`, which ships an ancient version of bash (3.2).

Atuin doesn't work under bash 3.2. it runs when I hit "up", but then when I hit enter, the command doesn't get put into the line to be run if I hit enter a second time.

Of course, the first thing to do with that is to upgrade to a recent version of bash using homebrew.

Unfortunately, that doesn't always take fully, resulting in bash 3.2 getting run... sometimes.

This adds a guard to install.sh, erroring out, saying that Atuin doesn't work in bash 3.2.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,5 +1,10 @@
 #! /usr/bin/env bash
 
+if [[ "${BASH_VERSION%%.*}" -eq 3 ]]; then
+    echo "Atuin has limited support for Bash 3.2. The Atuin config enter_accept cannot be turned off." >&2
+    echo "To turn off enter_accept, please upgrade your version of bash (possibly via homebrew or ports)" >&2
+fi
+
 set -euo pipefail
 
 cat << EOF


akinomyoga:
Bash 3.2 is supported by Atuin. As far as I test it, Atuin works with both Bash 3.2/bash-preexec and Bash 3.2/ble.sh.

> the command doesn't get put into the line to be run if I hit enter a second time.

I guess you set `enter_accept = false` because otherwise there is no second <kbd>enter</kbd>. If that is the case, you missed "PLEASE NOTE" section on [README](https://github.com/atuinsh/atuin?tab=readme-ov-file#bash):

> *(Quoted from PLEASE NOTE in Bash section in README)*
>
> To use Atuin in `bash < 4` with bash-preexec, the option `enter_accept` needs to be turned on (which is so by default).

install.sh
```diff
@@ -1,5 +1,10 @@
 #! /usr/bin/env bash
 
+if [[ "${BASH_VERSION%%.*}" -eq 3 ]]; then
+    echo "Atuin doesn't work on bash 3.2. Please upgrade your version of bash (possibly via homebrew or ports) and try again"
+	exit
```
akinomyoga, line 5: It is still good to notify users that Atuin doesn't offer full support for Bash 3.2.

```suggestion
    echo "Atuin has limited support for Bash 3.2. The config enter_accept cannot be turned off." >&2
    echo "To turn off enter_accept, please upgrade your version of bash (possibly via homebrew or ports)" >&2
```

fragmede:
> If that is the case, you missed "PLEASE NOTE" section on [README](https://github.com/atuinsh/atuin?tab=readme-ov-file#bash):

I did miss that, but it doesn't seem to be the default.

If I run `/bin/bash --noprofile`it doesn't work (mac osx sonoma 14.2.1, nor does it work if I run `set enter_accept=true` in bash before setting up atuin.

I'm also [not finding much online](https://www.google.com/search?q=%22enter_accept%22+-atuin+bash) about this option.

akinomyoga:
`enter_accept` is Atuin's option. It's not Bash's option. It's turned on in Atuin's default configuration automatically created at `~/.config/atuin/config.toml`. But, if the file creation is blocked by some permissions, etc, the config may not be turned on.

> nor does it work if I run `set enter_accept=true` in bash before setting up atuin.

Also, `set x=y` is not the way to set a config. `set arg` sets `$1` to be `arg`. You are just assigning the string `"enter_accept=true"` to `$1`, which doesn't have any effect.

fragmede:
(oops yeah I meant export not set)

Clearing out `~/.config/atuin/config.toml` gets it to work for me under bash 3.2. I installed atuin a while ago, before the config file set that option to true by default.

Updated the text slightly.

akinomyoga:
> (oops yeah I meant export not set)

Ah, I see. Then, that makes more sense.

> Clearing out `~/.config/atuin/config.toml` gets it to work for me under bash 3.2. I installed atuin a while ago, before the config file set that option to true by default.

I see. The option was recently added.

> Updated the text slightly.

Have you reverted it again? After force-pushing, nothing seems to be changed from the very initial version.

ellie:
Is this PR still required? Atuin does work under bash 3.2. Perhaps not as well as with later versions, but it works nonetheless. Looks like this was over some config issues?

install.sh
akinomyoga, line 5: The previous suggestion was just discarded after the application. I'm not sure if that is intentional, but I think at least the current PR needs updating. @fragmede Could you take a look at it again?

```suggestion
    echo "Atuin has limited support for Bash 3.2. The config enter_accept cannot be turned off." >&2
    echo "To turn off enter_accept, please upgrade your version of bash (possibly via homebrew or ports)" >&2
```

akinomyoga (changes requested):

install.sh
fragmede: Hows 3a3ce3eee333 look?

akinomyoga: LGTM

akinomyoga (approved):

ellie:
Thank you both!

ellie (approved):

PR #1802, opened 2024-03-01
ci: automatically update CONTRIBUTORS once a month

tessus:
The commit message says it all.

This action updates the CONTRIBUTORS file on the first day of the month. You can also run the workflow manually.

The list format and sort order is the same as the current one. Bots are excluded.

I didn't open an issue on the forum, because it's rather an infra/ci issue than an atuin feature request.

## Checks

### DIFF
--- a/.github/workflows/update-contributors.yml
+++ b/.github/workflows/update-contributors.yml
@@ -1,0 +1,11 @@
+name: Update CONTRIBUTORS
+on:
+  schedule:
+    - cron: "0 0 1 * *"
+  workflow_dispatch:
+jobs:
+  update_contributors:
+    runs-on: ubuntu-latest
+    steps:
+      - name: update
+        uses: tessus/update-contributors-list@v1


ellie:
We'll either need to do a big update to the mailmap, or the action will need to support deduping for this

Right now there are people who have contributed with slightly differing names or emails, meaning that with this they will be listed a few times

tessus:
I didn't add any deduplication, because the current CONTRIBUTORS file doesn't have any deduplication either and because the process is not without problems.

Automatic deduplication is unfortunately not error proof. While it is rather unlikely, there could be people with the same name. Apart from that, matching similar names is equally problematic and should not be be done in an automated fashion.

What I can see from the current file is that there are some people with the same name but different email addresses. One email is from github (when they did something on github) and another email (when they used git on the command line).
In such a case it is highly likely that they are the same person.
It is then possible to just git rid of one entry, e.g. the github one, in an automated fashion. However, what do you do when there are more than 2 and/or the email addresses are both/all not these no-reply github addresses? (e.g. `ContrA <a@super.com>, ContrA <a.z@evenbetter.com>, ContrA <a@example.com>`). Shall the process remove one email randomly? Or 2 emails randomly?

As for an email map, it is something that would have to be created/updated by the contributors. An additional file would be needed, maybe `.github/dedup.txt` that has the following format:

```
ContributorA Name: email-2-keep@whatever.com, 2ndemail-2-drop@gh.com, maybe-a-3rd-also-2-drop@example.com
```
Or better yet, only list the emails to remove:
```
ContributorA Name: 2ndemail-2-drop@gh.com, maybe-a-3rd-also-2-drop@example.com
```

But this also requires this text file to be up-to-date and without any format errors. e.g. if someone uses a `;` instead of the `:` by mistake, the parsing won't work.
But maybe it is a file that is not updated by contributors, but by someone who would remove those duplicate entries manually, if it wasn't for an automated process. 

It is a text file on purpose so that it is easily parseable via standard unix commands (within a shell script.)

As you can see, I thought about that a lot, but didn't do anything, because it's IMO not worth the maintenance overhead, and the current CONTRIBUTORS file has sometimes multiple entries for the same name as well.

Please let me know what you want to do. I am happy to give such a dedup file a go. Writing a bash script that parses such a file and filters out the specified entries is easy. (I can do that in probably 5 minutes.) The hard part is to add it to a gh action. (This could take several hours with all the testing.)

ellie:
> because the current CONTRIBUTORS file doesn't have any deduplication either and because the process is not without problems.

It actually does, I normally give it a scan through when updating to make sure obvious dupes are remove. With this change, I'd be listed three times, Conrad would be listed twice, etc.

> As for an email map, it is something that would have to be created/updated by the contributors. An additional file would be needed, maybe .github/dedup.txt that has the following format:

Not at all. It would make sense for this PR to also include an updated mailmap with the obvious cases. Anything ambiguous, we can just leave as-is. 

Perfect is the enemy of good, so it would be best to at least sort the cases that are clear and obvious duplicates.

tessus:
mailmap is usually used for another purpose. Parsing this file is also a bit more complex, but I can make it work.

I want to remove mails based on a file. Even though it is possible to do it the other way around: only add the email that should be kept for a person.
e.g. in your case you would add your name (exactly the one that shows up in CONTRIBUTORS) and the only email that should be kept, all others will then be removed.
This makes the process certainly more complex, but it is possible.

You were a bit vague on how to proceed. How shall I use this particular file, when it already has some meaning and is used somewhere else?
Due to the fact that a file already exists and that I do not know what for, how shall I process the file?

- as a list of emails to keep (a bit more complex)
- as a list of emails to remove (easy)

tessus:
Actually I found a way to make this work with emails to keep. It is also less maintenance for the mailmap file. 
I think it's a good solution 
I'll work on it and update the PR.

ellie:
Can we use the mailmap file please?

It's pretty much intended for this use case: https://git-scm.com/docs/gitmailmap

It will work with a whole bunch of tooling and ensure we don't get dupes anywhere.

tessus:
> Can we use the mailmap file please?

Of course. In this case I'll wait until the file has been updated.

You know which email you want to use and since you have de-duplicated the file manually before you (most likely) also know it for the other duplicates.

tessus:
Please update the mailmap file accordingly, so that the correct email addresses are used.

I've tested it and there are no more duplicates. I am still trying to update the gh action, but I ran into a small issue and hope to get this sorted soon. (For some reason the command, which I run locally, does not work in the gh action, but this makes no sense, since it's not really so much different than the previous command and it is in fact a valid and correct command.)

tessus:
I was able to fix the gh action. The only part missing is the updated mailmap.

You can run the following command locally to see what the action will produce:

```
git log |git shortlog -se |tr '\t' ':' |cut -d: -f2- |sort -uf |sed '/\[bot\]/d' >CONTRIBUTORS
```

ellie:
> Please update the mailmap file accordingly, so that the correct email addresses are used.

I'll try and get to this soon, but it's not something that's first on the list for me right now.

If you'd like this to be merged sooner, the current CONTRIBUTORS list has the email addresses that are preferred! So it can just be based on that. I generally prefer that we use people's personal emails rather than work, as they're more likely to remain stable.

Otherwise - would it be possible to remove the third party action and include the implementation directly into our own workflows please? I'm a little cautious about using actions that are small and not widely used

Thanks!

tessus:
> the current CONTRIBUTORS list has the email addresses that are preferred!

I can have a look. It's just that I remember there to be several duplicates. But with the additional hint of private emails being the better choice, I think I can sort it out. At least in most cases.

> would it be possible to remove the third party action and include the implementation directly into our own workflows please?

It is certainly possible, but not a good idea. I created this action, because I liked the fact that this project used a simple plain text file and all the other actions were geared towards markdown files or a section in the README.md. It's also something I can use in my repos.
I can certainly copy the action verbatim over to this project, but it's duplicating code which comes with all the maintenance overhead it entails. But if this is what you want I kindly ask that I can add a comment that if people add features and/or make changes they also notify the upstream (action) repo. Would that be ok?

> I'm a little cautious about using dependencies that are small and not widely used

On one side I can understand, on the other side with this logic no action will ever be widely used. ;-)
The current version (v2) is tagged and thus frozen, so nothing will change when using the action. I am not sure why duplicting the code is better (or safer).

I already have a few ideas for v3: your project uses an alphabetically sorted list, but other projects might want to sort it by the number of commits. (optional input for the action)

akinomyoga:
Or, we could host `update-contributors-list` under the `atuin` Organization.

tessus:
As I mentioned earlier, I want and will use it for my repos.

ellie:
> It's just that I remember there to be several duplicates.

I'd also prefer that there be a possible dupe than we mix up two separate people who happen to have the same name. 

> It is certainly possible, but not a good idea.

Taken from your viewpoint I can see why you may think this is the case. However, every external dependency is an attack vector. _Especially_ when integrated into a CI pipeline. GitHub does allow you to pin actions by SHA, which is perhaps the only thing I would consider in this case.

Focusing purely on Atuin, I would say that it is a good idea.

> but it's duplicating code which comes with all the maintenance overhead it entails

I don't want to be too reductive, but given that almost the entirety of the action is

```
git log |git shortlog -se |tr '\t' ':' |cut -d: -f2- |sort -uf |sed '/\[bot\]/d' >CONTRIBUTORS
```

and (at least in our case) `mailmap`, there won't be too much maintenance involved.

> On one side I can understand, on the other side with this logic no action will ever be widely used. ;-)

Without being taken as too rude, your action being widely used is not my concern. The security and long term maintenance of Atuin is.

I'd appreciate it if you do not try to use Atuin to promote your own project.

> The current version (v2) is tagged and thus frozen, so nothing will change when using the action

Tags can be force-pushed

Some other concerns I'd have are

1. I would like for it to happen on a PR, and not to run on the main branch. It should be clear that this is the case in your documentation
2. I'd like the option to run this on releases, rather than on a schedule

I'm afraid that my appetite to include this in Atuin is significantly reduced. I'd be happy to revisit in the future, but as it stands right now I'm unsure of the motives here and feel that giving you feedback leads to more friction.

tessus:
> I'd appreciate it if you do not try to use Atuin to promote your own project.

atuin's contributors file gave me the idea to write that action, because I wanted to use such a simple plain text file in my projects as well. Then I thought why copy a workflow when I can create a composite action (and I guess this is the reason why github made the action ecosystem available).
Since atuin's file was outdated I created a PR to update that file and automate it.

This is not promoting my project, which is hardly a project but a simple action. I asked for a comment in case of the copy and paste route, so that any changes (for the better) can be added to my action as well. This is not promoting my project, but trying to live the opensource spirit. 

I don't care whether this action is used or not, but I am not ok with being accused of trying to promote my action when I only wanted to improve this project.

PR #1803, opened 2024-03-01
fix: missing characters in preview

### LINKED ISSUE #1792: preview is missing a few characters
While implementing my [idea](https://forum.atuin.sh/t/show-preview-only-if-command-longer-than-terminal-width/212?u=tessus) and testing my code change, I found out that there is actually a display issue with the preview text.

This has nothing to do with my code change and I have reproduced it on git main (commit 593dc410eb3d5d7d409b780b9c7bd22ee4aa8654).

I used the following fairly long command (you can use whatever is longer than your terminal width):

```
echo "this is a very long command. I have to come to a width longer than 200 characters. I hate typing all that much. But what can you do when you have to test stuff and see if everything works as expected. kc tessarek."
```

As you can see a few characters around the borders are missing:

image

I haven't had a chance to debug the issue any further. I might be able to look into it on the weekend, but maybe a ratatui expert can figure this out in a few seconds.

tessus:
fixes #1792

Padding did not help. The only way was to decrease the width by 2 in non-compact mode.

## Checks

### DIFF
--- a/atuin/src/command/client/search/interactive.rs
+++ b/atuin/src/command/client/search/interactive.rs
@@ -639,6 +639,11 @@ impl State {
         let input = self.build_input(style);
         f.render_widget(input, input_chunk);
 
+        let preview_width = if compact {
+            preview_width
+        } else {
+            preview_width - 2
+        };
         let preview =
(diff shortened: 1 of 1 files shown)


ellie (approved):
thanks!

PR #1804, opened 2024-03-01
feat: show preview auto

tessus:
This change introduces a new config option `show_preview_auto = true|false` with `true` being the default.

I wanted to originally change the current boolean option `show_preview` with an option that takes also the value `auto`. But this would be a breaking change and should be done when the config/settings are updated to the new format. I am more than happy to create another PR then.

/ref [Show preview only if command longer than terminal width](https://forum.atuin.sh/t/show-preview-only-if-command-longer-than-terminal-width/212?u=tessus)

## Checks

### DIFF
--- a/atuin/src/command/client/search/interactive.rs
+++ b/atuin/src/command/client/search/interactive.rs
@@ -500,27 +500,41 @@ impl State {
         let invert = settings.invert;
         let border_size = if compact { 0 } else { 1 };
         let preview_width = f.size().width - 2;
-        let preview_height = if settings.show_preview && self.tab_index == 0 {
-            let longest_command = results
-                .iter()
-                .max_by(|h1, h2| h1.command.len().cmp(&h2.command.len()));
-            longest_command.map_or(0, |v| {
-                std::cmp::min(
-                    settings.max_preview_height,
-                    v.command
-                        .split('\n')
-                        .map(|line| {
-                            (line.len() as u16 + preview_width - 1 - border_size)
-                                / (preview_width - border_size)
-                        })
-                        .sum(),
-                )
-            }) + border_size * 2
-        } else if compact || self.tab_index == 1 {
-            0
-        } else {
-            1
-        };
+        let preview_height =
+            if settings.show_preview_auto && self.tab_index == 0 && !results.is_empty() {
+                let length_current_cmd =
+                    results[self.results_state.selected()].command.len() as u16;
+                // The '- 19' takes the characters before the command (duration and time) into account
+                if length_current_cmd > preview_width - 19 {
+                    std::cmp::min(
+                        settings.max_preview_height,
+                        (length_current_cmd + preview_width - 1 - border_size)
+                            / (preview_width - border_size),
+                    ) + border_size * 2
+                } else {
+                    1
+                }
+            } else if settings.show_preview && self.tab_index == 0 {
+                let longest_command = results
+                    .iter()
+                    .max_by(|h1, h2| h1.command.len().cmp(&h2.command.len()));
+                longest_command.map_or(0, |v| {
+                    std::cmp::min(
+                        settings.max_preview_height,
+                        v.command
+                            .split('\n')
+                            .map(|line| {
+                                (line.len() as u16 + preview_width - 1 - border_size)
+                                    / (preview_width - border_size)
+                            })
+                            .sum(),
+                    )
+                }) + border_size * 2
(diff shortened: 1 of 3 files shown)


atuin/src/command/client/search/interactive.rs
```diff
@@ -500,27 +500,41 @@ impl State {
...
-                        })
-                        .sum(),
-                )
-            }) + border_size * 2
-        } else if compact || self.tab_index == 1 {
-            0
-        } else {
-            1
-        };
+        let preview_height =
```
ellie, line 503: I appreciate the reasoning for this change, but I think we need to refactor this. The code below has gotten beyond unwieldy. 

Would it be possible to break this down a bit and make it easier to read/understand?

I'm a little concerned that with all the PRs to this section lately we're going to end up with a ball of string.

tessus: I didn't change the overall structure, but only added the branch `settings.show_preview_auto && self.tab_index == 0 && !results.is_empty()` (lines 504-516). The rest was changed by `cargo format`.

I am not sure how to refactor this, because there's a reason for all the `if`s and `else`s. Using a `mut` won't change much because the `else if`s are still necessary. A match would make it even longer.

I do not know how to refactor it to make it any clearer. The structure is very logical. What it could need are comments for the calculations. Yet again, I did not come up with those, except the `- 19` one, which I explained in a comment.

If you have an idea how to refactor this, please let me know.

ellie: I think we should start breaking things out into functions that render each component, and contain the logic for them separately to the overall rendering. 

A nice start there could be a function for calculting the preview height, and a test to ensure that it is behaving correctly.

I appreciate that you're not responsible for everything here. I have to consider that each change is an improvement in terms of featureset, and is also not making the overall structure more difficult to maintain.

To be blunt, please

1. Break out this calculation into its own function, ideally with pure input/output
2. Write a test for it to both demonstrate its functionality and ensure it is behaving as intended

ellie: then we can just use

```rust
let preview_height = calc_preview_height(stuff...);
```

tessus: This is a great idea, although this also means that all the unwieldy code is most likely just moved to another funtion (with maybe a few improvements). I love TDD, but it's unfortunately mostly ignored these days.
However, the refactoring is more than I bargained for when I created the PR.

I really want to refactor it as you suggested and add tests, but I can't say how long it will take. I am currently a bit busy with other stuff.

I also understand, if you rather close the PR. Or maybe you can add a `help wanted` tag to the PR. Who knows, maybe somebody wants to give it a go.

tessus: I was looking into this and I am not entirely sure how to tackle it.
I started with the following function:

```rust
fn calc_preview_height(
        &mut self,
        compact: bool,
        border_size: u16,
        preview_width: u16,
        results: &[History],
        settings: &Settings,
    ) -> u16 {
```

The problem is that `compact`, `border_size`, and `preview_width` are needed later, so I can't move and hide them in the new function.

I also need `self.tab_index` and `self.results_state.selected()` in the new function, but this makes the tests rather complex. How to easily mock the State structure? I could also add them as arguments (and remove self), in which case the function `calc_preview_height` would have 7 arguments.

What's your take on this?

ellie: > How to easily mock the State structure

I'd avoid passing in self, and especially a mutable reference to it. You only need two attributes of it, so could just pass in those. Alternatively, it may make more sense to as much of the preview window logic as possible into build_preview.

tessus: alright. I've refactored it, added tests, and rebased to main.

P.S.: 
> I'd avoid passing in self, and especially a mutable reference to it. 

I only did that at first, because pretty much any other function did that.

tessus: I want to point out that when the new config format is available, the options `show_preview_auto` and `show_preview` should be merged like so:

```
show_preview = auto|true|false
```

Should I miss the new format, please create an issue and assign it to me. I'll create a PR then.

tessus: @ellie Just in case you missed this ^^^

ellie: ah nice, thanks for the heads up

PR #1807, opened 2024-03-01
Nushell: check if command history is disabled

IanManske:
Somewhat recently in Nushell, we added a [CLI option](https://github.com/nushell/nushell/pull/11550) to disable command history. This is very similar to fish's private mode. Like in #577, this PR skips `atuin history start` if history is disabled.

## Checks

### DIFF
--- a/atuin/src/shell/atuin.nu
+++ b/atuin/src/shell/atuin.nu
@@ -6,6 +6,9 @@ hide-env -i ATUIN_HISTORY_ID
 let ATUIN_KEYBINDING_TOKEN = $"# (random uuid)"
 
 let _atuin_pre_execution = {||
+    if ($nu | get -i history-enabled) == false {
+        return
+    }
     let cmd = (commandline)
     if ($cmd | is-empty) {
         return


remmycat:
I'm wondering if it makes sense to take nushell's setting here?

I think the way things are today, the two history implementations have no connection apart from the option to import nushell's history into atuin.

Personally I would like to see a kind of "native" integration of atuin into nushell, but until then I would consider disabling nushell's history in my config _because_ atuin is handling it for me already and I don't need to have it recorded twice 🤔 

(But I can see the argument how it can become a privacy issue if a user believes they did disable the history everywhere from nushell's config)

IanManske:
Good point, if it was a config option, then it would probably make sense to disregard it. But yeah, it's a CLI option like you mentioned, so I think it makes sense to handle it like fish's private mode given their similarity.

ellie (approved):
Thanks for this!

Makes sense to have this as a "private mode". I do agree with the discussion though - if Nu introduces a global history off switch, ignoring it would be the best approach

PR #1823, opened 2024-03-04
fix(nix): set meta.mainProgram in the package

arcuru:
Small change, as nixpkgs is starting to require meta.mainProgram to be set for the package. I also updated the link to the nixpkgs package after some refactoring in that repo broke the link. They don't intend to move it again.

## Checks


ellie (approved):
Thanks! Appreciate you keeping this up to date

PR #1825, opened 2024-03-05
make atuin compile on non-win/mac/linux platforms

sunshowers:
Hi!

I've been trying to get atuin set up on the illumos machine I built for work @oxidecomputer, and I ran into a few issues which are fixed here:

1. The `clipboard` feature was only supported on Windows, Mac and Linux. I've added a platform gate for that.
2. The `atomic-write-file` crate needed an update to the version of `nix` -- that is included.
3. As part of this, I found a [security bug](https://rustsec.org/advisories/RUSTSEC-2024-0020.html) in the whoami crate. The bug has been fixed upstream and I've included it (it's required to prevent a crash that happens on illumos).

   whoami 1.5.0 deprecates the `hostname` function, which produced some fresh warnings. While fixing the warnings I also took the liberty of doing some code rearrangement, adding a few functions that wrap some common operations. I didn't really know where to put those functions, so I created a new `utils` module for it. If you have a better place to put them, I'm happy to change the PR.

With these issues fixed, I'm happy to say that atuin works fantastically on illumos with zsh!

Feel free to make any changes to this PR if you like before landing it, or to ask for review.

As a followup I'm also happy to set up a cross-compile build for atuin on illumos. It's a bit harder to run tests in CI for illumos at the moment, but I'm trying to get a project started up to make that happen in the future as well.

## Checks
- [x] I am happy for maintainers to push small adjustments to this PR, to speed up the review cycle
- [x] I have checked that there are no existing pull requests for the same thing

PR #1830, opened 2024-03-06
chore(ci): add cross-compile job for illumos

sunshowers:
Followup to #1825.

I also tried getting FreeBSD to work, but didn't get too far sadly. I don't have the time to debug this so I just put in a comment for now. (With my changes, Atuin works great on FreeBSD as well.)

## Checks

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -44,6 +44,43 @@ jobs:
     - name: Run cargo build main
       run: cargo build --all --locked --release 
 
+  cross-compile:
+    strategy:
+      matrix:
+        # There was an attempt to make cross-compiles also work on FreeBSD, but that failed with:
+        #
+        # warning: libelf.so.2, needed by <...>/libkvm.so, not found (try using -rpath or -rpath-link)
+        target: [x86_64-unknown-illumos]
+    runs-on: ubuntu-latest
+
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/rust.yml
```diff
@@ -44,6 +44,43 @@ jobs:
...
+      run: cross build -p atuin-common --locked --target ${{ matrix.target }}
+
+    - name: Run cross build client
+      run: cross build -p atuin-client --locked --target ${{ matrix.target }}
+
+    - name: Run cross build server
+      run: cross build -p atuin-server --locked --target ${{ matrix.target }}
+
+    - name: Run cross build main
+      run: cross build --all --locked --target ${{ matrix.target }}
```
sunshowers, line 82: I noticed that the indentation for this is slightly off from what YAML (or at least Prettier) recommends:

```yaml
steps:
  - uses: actions/checkout@v3
  - name: Install cross
    # ...
```

I changed it to match the rest of the file, but it may be worth fixing this at some point. (Again, happy to send a PR for this.)

ellie: > Again, happy to send a PR for this.

If you have the time, that would be very much appreciated! 🙏

ellie (approved):
Nice one, thank you!

I actually expected it to be a bit slower, pretty impressed that it only took ~5 mins to build

PR #1857, opened 2024-03-11
feat(search): process [C-h] and [C-?] as representations of backspace

### LINKED ISSUE #1753: In Atuin UI interprets backspace as literal ^h
Opening the Atuin UI and pressing the backspace key writes an h to the input. To delete the previous character I need to press ctrl+backspace.

I assume it has to do with backspace, `^H`, being interpreted as the literal `^H` in some way, hence why ctrl "cancels out" the `^`.

I can reproduce in:
 - [cool-retro-terminal](https://github.com/Swordfish90/cool-retro-term)
 - xterm
 
OS: Arch linux
Atuin version: 18.0.1

akinomyoga:
Fixes #1753

As discussed in #1753, I think this should ideally be fixed in `crossterm`, but there are general limitations by the terminal protocols and the support by the terminals in the market. I here instead suggest solving this at Atuin's side. The description is found in the commit message and also in a code comment.

## Checks

### DIFF
--- a/atuin/src/command/client/search/interactive.rs
+++ b/atuin/src/command/client/search/interactive.rs
@@ -379,6 +379,21 @@ impl State {
             KeyCode::Backspace => {
                 self.search.input.back();
             }
+            KeyCode::Char('h' | '?') if ctrl => {
+                // Depending on the terminal, [Backspace] can be transmitted as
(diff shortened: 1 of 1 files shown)


ellie (approved):
This seems reasonable to me. Thank you for the fix!

Issue #1864, opened 2024-03-12
Add a systemd service file


svenstaro:
Hey, I think it would make sense to have a systemd service file upstream. It should source environment variables from `/etc/default/atuin`.

I think something like this could work:
```ini
[Unit]
Description=Start the atuin sync service

[Service]
ExecStart=atuin server start
Restart=on-failure
User=atuin
EnvironmentFile=/etc/default/atuin
PrivateTmp=yes
PrivateUsers=yes
PrivateDevices=yes
NoNewPrivileges=true
ProtectSystem=strict
ProtectHome=yes
ProtectClock=yes
ProtectControlGroups=yes
ProtectKernelLogs=yes
ProtectKernelModules=yes
ProtectKernelTunables=yes
ProtectProc=invisible

[Install]
WantedBy=multi-user.target
```

If you like this, I could make an MR for this along with a sysusers file to create the `atuin` user.

ellie:
Sounds good to me! I'd also really appreciate it if you could update the [docs](https://github.com/atuinsh/docs) with information on how to use this 🙏

inglor:
I tried to do that for Arch Linux package and had issues with atuin server searching for configuration in hardcoded $HOME variable and wanted write access to it. Also it would be nice for DB_URI to support (passwordless) socket in linux - couldnt' figure out how to set this up in sqlx.

For reference my work in progress files:

`atuin.env`
```
ATUIN_CONFIG_DIR=/var/lib/atuin
ATUIN_HOST="127.0.0.1"
ATUIN_PORT=8888
ATUIN_OPEN_REGISTRATION=true
ATUIN_DB_URI="postgres://username:password@localhost/atuin" # socket how?
```
`atuin.service`
```
[Unit]
Description=Atuin server
After=network-online.target
Wants=network-online.target systemd-networkd-wait-online.service

[Service]
User=atuin
Group=atuin
ExecStart=/usr/bin/atuin server start
ReadOnlyPaths=/etc/atuin

EnvironmentFile=/etc/atuin/atuin.env
ReadWritePaths=/var/lib/atuin

# Hardening options
CapabilityBoundingSet=
AmbientCapabilities=
NoNewPrivileges=true
ProtectHome=true
ProtectSystem=strict
ProtectKernelTunables=true
ProtectKernelModules=true
ProtectControlGroups=true
PrivateTmp=true
PrivateDevices=true
LockPersonality=true

[Install]
WantedBy=multi-user.target
```
[...]

PR #1879, opened 2024-03-14
Add Systemd config for self-hosted server

svenstaro:
Fixes https://github.com/atuinsh/atuin/issues/1864.

## Checks

### DIFF
--- a/systemd/atuin-server.service
+++ b/systemd/atuin-server.service
@@ -1,0 +1,29 @@
+[Unit]
+Description=Start the Atuin server syncing service
+After=network-online.target
+Wants=network-online.target systemd-networkd-wait-online.service
+
+[Service]
+ExecStart=atuin server start
+Restart=on-failure
+User=atuin
+Group=atuin
+
+Environment=ATUIN_CONFIG_DIR=/etc/atuin
+ReadWritePaths=/etc/atuin
+
+# Hardening options
+CapabilityBoundingSet=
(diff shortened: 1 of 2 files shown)


ellie:
The env makes sense, but it might be better to just use the config file. We autocreate with the defaults which _hopefully_ leads to less people asking for help with things

Setting ATUIN_CONFIG_DIR=/etc/atuin and ensuring that directory is created would make that work

svenstaro:
I reworked this following the suggestions in #1864. I think this is probably good to go.

systemd/atuin.service
```diff
@@ -0,0 +1,29 @@
+[Unit]
```
ellie, line 1: Could we also use the name `atuin-server` please? I think it's fairly important to make the distinction, especially considering the Atuin Daemon that will soon also have a systemd unit file

ellie:
Just to clarify here too, please make it very clear in the unit file that this is for running _atuin server_, and not anything client related

systemd/atuin.service
svenstaro: Alright, renamed this and the sysusers file.

svenstaro:
@ellie Friendly ping (for this and https://github.com/atuinsh/docs/pull/45) :) Are you cool with this now?

ellie:
Both look good now, thank you!

ellie (approved):
Thank you!

PR #1898, opened 2024-03-22
feat(search): show filter mode shortcut

douglascamata:
My usage of `atuin` got much better after I learned about this shortcut and I wondered why not let everyone know about it. 

## Checks

### DIFF
--- a/atuin/src/command/client/search/interactive.rs
+++ b/atuin/src/command/client/search/interactive.rs
@@ -728,6 +728,9 @@ impl State {
                 Span::raw(", "),
                 Span::styled("<ctrl-o>", Style::default().add_modifier(Modifier::BOLD)),
                 Span::raw(": inspect"),
+                Span::raw(", "),
+                Span::styled("<ctrl-r>", Style::default().add_modifier(Modifier::BOLD)),
+                Span::raw(": filter mode"),
             ]))),
 
             1 => Paragraph::new(Text::from(Line::from(vec![


tessus:
`<ctrl-s>` has never been listed in the help either. It should then be listed as well for completeness. Although I can't say for sure. Let's wait for what the devs/maintainers think.

(I have deactivated the help, so I don't have a preference.)

douglascamata:
@tessus whaaaat, I didn't know `<ctrl-s>`! Super helpful, thanks. 

I would say it should be listed there too. There's little discoverability for shortcuts in Atuin, yet they are so powerful. 

I think the project even deserves some sort of "shortcut list" popover under a shortcut like `<ctrl-?>`, then the real state of the main screen can be preserved. But this is way beyond my league to implement.

tessus:
> I think the project even deserves some sort of "shortcut list" popover under a shortcut like `<ctrl-?>`, then the real state of the main screen can be preserved. But this is way beyond my league to implement.

I think this is not a bad idea. I can help implement this, similar as to what is shown in tmux when invoking the help screen. But I think this has to wait a bit. Afaik there is the idea of allowing to make shortcuts configurable. I'll keep this idea in the back of my mind.

ellie:
I'd really rather avoid adding things to the help text at this point. The more we add there, the denser it gets. That causes two problems

1. More help text will just be clipped off the end on smaller terminals
2. People tend to ignore dense help text. I am still regularly asked how to edit rather than execute, even though it's at the top of the terminal, README, docs, etc. If anything, I'd prefer to streamline the help text

I think the only thing I'm OK with adding there is a shortcut showing a popup with the full default keybindings displayed.

I'm pretty sure that ctrl-? won't work correctly in most terminals, so it will probably need to be <kbd>ctrl-a ?</kbd>, see #1875 for prefix bindings

For more information on basic usage (including filter mode switching), see: https://docs.atuin.sh/guide/basic-usage/

Filter modes are also mentioned at the top of the README, and in the keybinding list here: https://docs.atuin.sh/configuration/key-binding/#atuin-ui-shortcuts

PR #1899, opened 2024-03-22
fix: install script echo

ivan-toriya:
## Checks

Fix for the incorrect distro name detection during install.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3624,9 +3624,9 @@ checksum = "2047c6ded9c721764247e62cd3b03c09ffc529b2ba5b10ec482ae507a4a70160"
 
 [[package]]
 name = "sysinfo"
-version = "0.30.6"
+version = "0.30.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


tessus (commented):
LGTM.

The only issue could be that `OS` holds the name in all lowercase. I don't mind, but it changes the current output.

There's an easy way to uppercase the first character in bash >=4, but this script supports older bash versions and thus it would break that premise.

Unless of course you only use it for bash versions >=4. I don't know how important capitalizing the first character is, some people really care about the output.

You could do something like this at the beginning of the function:

```
[[ "${BASH_VERSION%%.*}" -ge 4 ]] && OS="${OS^}"
```

Let's see what the maintainers/devs say.

ellie:
> The only issue could be that OS holds the name in all lowercase. I don't mind, but it changes the current output.

I'm ok with this, would rather avoid complication trying to change formatting

@ivan-toriya happy to merge this, could we get a rebase please?

ellie (approved):
Thanks!

PR #1900, opened 2024-03-24
Set the synchronous option for the records DB connection to "normal".

nebkor:
Just like #1667, when connecting to a DB using WAL mode for journaling, the recommended setting for synchronizing the data is "NORMAL", while the SQLx default is "FULL", which causes hangs on ZFS systems.

Without this change, the hangs are not as long or frequent as before #1667, but are still present. With this change, there are no hanging commands.

## Checks


nebkor:
Nevermind; I'm still getting occasional tiny hangs, and an error that the connection to the records DB timed out.

PR #1913, opened 2024-04-01
Update atuin.nu

WindSoilder:
Nushell will deprecate `--redirect-stderr` flag on upcoming release 0.92.0 due to this pr: https://github.com/nushell/nushell/pull/11934

This pr is going to make atuin compatible with new version

## Checks


ellie (approved):
thanks!

hustcer:
@WindSoilder After init `atuin` with the latest init script with `nushell` v0.92 if I input `git st` in the command line and press `Ctrl + r` the history search UI appears, however `git st` I inputed before didn't show in the command editor and I have to input it again, do you have the same issue?

WindSoilder:
> do you have the same issue?

Sorry for lately reply, I never thought I can use atuin with this way.  But yeah, I can reproduce your issue.

Issue #1905, opened 2024-03-27
[Bug]: error when parsing non json


papertigers:
### What did you expect to happen?

`atuin sync` runs successfully.

### What happened?

```
Error: error decoding response body: expected value at line 1 column 1

Caused by:
    expected value at line 1 column 1

Location:
    /Users/link/src/atuin/atuin-client/src/api_client.rs:164:21
```

It would be nice if atuin at least included the status code for client errors.  I ended up adding some additional logging and found that nginx in front of my atuin server was not allowing for a sufficient body size.  I ended up seeing the following:
```
     Running `target/debug/atuin sync`
[atuin-client/src/api_client.rs:162:9] json = Ok(
    "<html>\r\n<head><title>413 Request Entity Too Large</title></head>\r\n<body>\r\n<center><h1>413 Request Entity Too Large</h1></center>\r\n<hr><center>nginx/1.25.3</center>\r\n</body>\r\n</html>\r\n",
)
```

Thoughts on doing something like attempting to parse the error and if that fails passing up an error with `bail!()` that at least gives you the status code back?

### Atuin doctor output

```yaml
❯ atuin doctor
Atuin Doctor
Checking for diagnostics


Please include the output below with any bug reports or issues

atuin:
  version: 18.1.0
  sync:
    cloud: false
    records: false
    auto_sync: true
    last_sync: 2024-03-27 19:44:10.339676 +00:00:00
shell:
  name: zsh
  plugins:
  - atuin
system:
  os: Darwin
  arch: arm64
  version: '14.3'
  disks:
  - name: Macintosh HD
    filesystem: apfs
  - name: Macintosh HD
    filesystem: apfs
```
```


### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

**Comment by ellie** [comment]
I've just merged a fix, thank you!

Fwiw, you'll likely have a better experience on sync v2

Put

```
[sync]
records = true
```

In your client config. No changes needed with the server

It will soon become the default, and has been well tested by a bunch of people at this point. Sync v2 is generally faster and more efficient

PR #1915, opened 2024-04-02
fix: report non-decodable errors correctly

ellie:
ref #1905

## Checks

### DIFF
--- a/atuin-client/src/api_client.rs
+++ b/atuin-client/src/api_client.rs
@@ -157,13 +157,17 @@ async fn handle_resp_error(resp: Response) -> Result<Response> {
         );
     }
 
-    if status.is_client_error() {
-        let error = resp.json::<ErrorResponse>().await?.reason;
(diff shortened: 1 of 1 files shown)


atuin-client/src/api_client.rs
```diff
@@ -157,13 +157,17 @@ async fn handle_resp_error(resp: Response) -> Result<Response> {
...
     }
 
-    if status.is_client_error() {
-        let error = resp.json::<ErrorResponse>().await?.reason;
-        bail!("Could not fetch history, client error: {error}.")
-    } else if status.is_server_error() {
-        let error = resp.json::<ErrorResponse>().await?.reason;
-        bail!("There was an error with the atuin sync service: {error}.\nIf the problem persists, contact the host")
-    } else if !status.is_success() {
+    if !status.is_success() {
```
ellie, line 160: more nested than I'd like, but fine for now 🙃

PR #1917, opened 2024-04-03
feat: add 'ctrl-a a' to jump to beginning of line

tessus:
The `ctrl-a` shortcut (to jump to the beginning of the line) was replaced by the prefix.

tmux handles it in a way so that if `ctrl-a` is the prefix, `ctrl-a a` will jump to the beginning of the line.
This change implements the same machanism in atuin.

## Checks

P.S.: I didn't open an issue, because it is a follow-up to #1875


akinomyoga (approved):

ellie (approved):
thanks! 🚀

PR #1928, opened 2024-04-08
fix(bash): do not use "return" to cancel initialization

akinomyoga:
In #1533, I used `return 0` for the "include guards" of Atuin's initialization. However, I noticed that it would unexpectedly cancel the entire processing of `~/.bashrc` because it is evaluated by `eval`. This is my fault. In this PR, I suggest enclosing the entire content of the integration code within a `if`-statement.

### Estimation on the impact

I expect the impact of `return 0` would not be so big since the problem only happens when (i) `~/.bashrc` is sourced twice in an interactive session, (ii) `eval "$(atuin init bash)"` is performed outside the interactive sessions, or (iii) it is performed in a very old version of Bash for which Atuin's doesn't offer the support. For case (ii), with a typical `.bashrc` starting with `case $- in *i*) ;; *) return ;; esac` (or equivalent), `eval "$(atuin init bash)"` is usually performed only in interactive sessions. For case (iii), Atuin's initialization anyway produces an error message, so the setup is technically wrong. If the user use mixed versions of Bash, the user is supposed to check the Bash version at the caller side in `~/.bashrc` before evaluating `eval "$(atuin init bash)"`.

## Checks

### DIFF
--- a/atuin/src/shell/atuin.bash
+++ b/atuin/src/shell/atuin.bash
@@ -1,15 +1,16 @@
 # Include guard
-[[ ${__atuin_initialized-} == true ]] && return 0
-__atuin_initialized=true
-
-# Enable only in interactive shells
-[[ $- == *i* ]] || return 0
-
-# Require bash >= 3.1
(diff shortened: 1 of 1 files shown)


ellie (approved):
ahh I see, nice fix. thanks!

PR #1933, opened 2024-04-09
feat: prevents stderr from going to the screen when syncing nushell

YummyOreo:
I have no idea why this works, but it does! The problem was that when atuin was syncing after a command, it would print the progress bar, this fixes it!

## Checks


ellie (approved):
Makes sense! (https://www.nushell.sh/commands/docs/complete.html#complete-for-system for anyone else)

Though I think `complete` is a bit of a weird name

Thanks for fixing :)

PR #1935, opened 2024-04-09
feat(gui): add base structure

ellie:
I have a few motivations here, and absolutely understand that not all _current_ users will use or value this

1. I'd like to make it easier to explore past history, commands, and stats. TUIs can be very powerful, but there are limits to what we can render
2. I'd like to make it easier to discover and explore features
3. I'd like to make Atuin more accessible. Especially as we start to add new features like the daemon, running and installing Atuin can be intimidating to new/inexperienced terminal users
4. While the search UI works well for its purpose, things like dotfiles/kv/etc will be smoother to use this way

Current WIP:
Screenshot 2024-04-09 at 08 57 17

It does use a webview, but it absolutely is not electron. It uses [Tauri](https://tauri.app/), which is a Rust backend and system webview frontend. No chrome, no node. We can reuse the Atuin crates, and also have a much smaller footprint.

## Checks


arcuru:
I haven't run this yet, but this is a great idea.

ellie:
just gonna remove the tauri backend from the workspace. doesn't make sense to run all the GUI testing along with the rest, especially when it isn't published

I'll restructure after merging this PR. I'll merge as-is - it's fairly functional, but still needs a bunch of work. I don't want this PR to diverge much further

PR #1936, opened 2024-04-09
fix: use spawn_blocking for file access during async context

conradludgate:
Context:
update checking runs asynchronously to allow DB and UI rendering to occur concurrently with the otherwise superficial update check system.

The file accesses that take place during this check run synchronously on the main thread and can block the runtime from running the render routine. I wouldn't expect the file accesses to take long but filesystems can be weird I guess?

(extra context, [tokio::fs::File also does spawn_blocking](https://docs.rs/tokio/latest/src/tokio/fs/file.rs.html#567-570), so I would stick with fs-err here for the better error messages and not use tokio::fs)

## Checks

### DIFF
--- a/atuin-client/src/settings.rs
+++ b/atuin-client/src/settings.rs
@@ -527,7 +527,13 @@ impl Settings {
         if !self.needs_update_check()? {
             // Worst case, we don't want Atuin to fail to start because something funky is going on with
             // version checking.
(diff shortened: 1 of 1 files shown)


ellie:
is this ready for review? (still a draft)

conradludgate:
I was planning to shuffle around when the queries happen too, but I can do that separately

ellie (approved):
ty 🚀

Issue #1921, opened 2024-04-04
[Bug]: Atuin doesn't store new commands for user 'root'


ibahnasy:
### What did you expect to happen?

Atuin stores new executed commands fine for any normal user and this should happen to the `root` user too.

### What happened?

This doesn't happen unless manually ran: `atuin import bash`.

### Atuin doctor output

```yaml
Atuin Doctor
Checking for diagnostics

[Shell] If you are using Bash, Atuin requires that either bash-preexec or ble.sh be installed. We cannot currently detect ble, so if you have it setup then ignore this! Read more here: https://docs.atuin.sh/guide/installation/#bash

Please include the output below with any bug reports or issues

atuin:
  version: 18.1.0
  sync: null
shell:
  name: bash
  plugins:
  - atuin
system:
  os: Ubuntu
  arch: x86_64
  version: '23.10'
  disks:
  - name: /dev/mapper/ubuntu--vg-ubuntu--lv
    filesystem: ext4
  - name: /dev/nvme0n1p2
    filesystem: ext4
  - name: /dev/mapper/ubuntu--vg-ubuntu--lv
    filesystem: ext4
  - name: /dev/nvme0n1p1
    filesystem: vfat
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

ibahnasy:
I fixed that by defining the ble.sh line before atuin in .bashrc:
```
source ~/.local/share/blesh/ble.sh
eval "$(atuin init bash)"
```

ibahnasy:
There is an issue with ble.sh I reported it here: https://github.com/akinomyoga/ble.sh/issues/430

ellie:
Sounds like this is resolved and was due to misconfiguration?

PR #1937, opened 2024-04-09
docs(bash-preexec): describe the limitation of missing commands

akinomyoga:
I noticed that the problem of missing commands with `bash-preexec` is not mentioned in README. This PR adds a mention.

Also, the quick instruction at the beginning of README doesn't mention the possible minor problems with Bash integration, so I added a link to the later Bash section, which describes the problems in more detail.

As another change, I made those "notes" more outstanding using [the GFM blockquote](https://github.com/orgs/community/discussions/16925) so that users do not miss them. There are still reports by users who missed those notes (like #1921).

## Checks

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,7 +122,13 @@ atuin sync
 ```
 
 Then restart your shell!
-  
+
+> [!NOTE]
+>
+> **For Bash users**: The above sets up `bash-preexec` for necessary hooks, but
+> `bash-preexec` has limitations.  For details, please see the [Bash](#bash)
+> section below.
+
 ## Offline only (no sync)
   
 ```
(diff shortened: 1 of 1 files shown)


ellie (approved):
lgtm

may also be worth adding something to `atuin doctor` when it detects someone using bash-preexec

akinomyoga:
Thanks!

> may also be worth adding something to `atuin doctor` when it detects someone using bash-preexec

Yeah. Maybe a troubleshooting section can be added in README.

ellie:
> Yeah. Maybe a troubleshooting section can be added in README.

I'd rather add it to the docs and the doctor. My experience so far has been that people tend to ignore dense information - the docs has search, and if someone opens an issue without doing their own search, the doctor should catch that too

akinomyoga:
Ah, OK. I misread your previous comment!

akinomyoga:
BTW, I have a commit https://github.com/akinomyoga/atuin/commit/0598e4a0376623c486377110a6532abbdf66bfc2 for `atuin doctor`. It tries to check if Atuin's integration with bash-preexec or ble.sh is properly set up. Even if bash-preexec/ble.sh is loaded in the session, there is a possibility that the integration is not properly set up e.g. when the Atuin is initialized before ble.sh is sourced as in #1921. What do you think of this kind of detection?

ellie:
> What do you think of this kind of detection?

Seems reasonable to me!

akinomyoga:
Thank you! Then, I'll later submit a PR after some refinements.

PR #1963, opened 2024-04-20
nushell: use version.(major|minor|patch) if available

poliorcetics:
I added `(version).major` and others like it in https://github.com/nushell/nushell/issues/12561 to make atuin's life easier regarding version parsing.

At the moment we obviously have to keep the support for older versions but in a year (or more?) we will be able to reasonably just use the new members I think and make that part of the script much cleaner 🤞 

## Checks


ellie (approved):
Nice! Thank you!

Issue #1953, opened 2024-04-16
[Bug]: fish: `atuin dotfiles alias import` imports non-alias lines


rigrig:
### What did you expect to happen?

I ran `atuin dotfiles alias import` under fish, and expected it to import (only) my fish aliases.

### What happened?

Atuin imported my aliases, *and* the additional output that my fish generates on startup.

I have a fish script to show a formatted fortune cookie when starting:
```fish
> cat ~/.config/fish/conf.d/fortune.fish 
if status is-interactive
        fortune -a | cowsay -f (cowsay -l | tail -n+2 | xargs shuf -en1) | lolcat -t
end
```
Giving something like this:
![image](https://github.com/atuinsh/atuin/assets/2941436/ccaf8f0e-7008-4ad6-a222-bacb5c3ec6ea)

Atuin seems to have captured those lines as alias:
` atuin dotfiles alias list` gives
```
                ||     ||=
                ||----w |=
           (___)\       )\/\=
         \ (ooo)\_______=
        \  ^___^=
 -----------------------------------------=
 _________________________________________=
/ If you permit yourself to read meanings \=
\ 416                                     /=
alias addr 'ip addr show dev enp2s0 | fgrep "inet "'=
... (snip: more aliases)
alias xl 'exa --icons --git --classify --long --header  --group-directories-first  --time-style=long-iso'
|                                         |=
| -- Michael Keith, "The Bar-Code Beast", |=
| The Skeptical Enquirer Vol 12 No 4 p    |=
| conclusion you like.                    |=
| into (rather than drawing meanings out  |=
| of) the evidence, you can draw any      |=
```

Opening a new terminal now always greets me with
```
- (line 77): Expected a command, but found '||'
alias                 ||     ||=''
                             ^^
from sourcing file -
        called on line 2 of file ~/.config/fish/conf.d/atuin.fish
from sourcing file ~/.config/fish/conf.d/atuin.fish
        called on line 248 of file /usr/share/fish/config.fish
from sourcing file /usr/share/fish/config.fish
        called during startup
source: Error while reading file '<stdin>'
```

### Atuin doctor output

```yaml
atuin:
  version: 18.2.0
  sync:
    cloud: false
    records: true
    auto_sync: true
    last_sync: 2024-04-16 10:11:50.497574363 +00:00:00
shell:
  name: fish
  default: unknown
  plugins: []
system:
  os: Linux Mint
  arch: x86_64
  version: '21.3'
  disks:
[...]

PR #1970, opened 2024-04-22
fix(dotfiles): alias import with init output

ellie:
Ref #1953

Added a test case and fix

cc @rigrig, does this resolve the issue for you?

## Checks

### DIFF
--- a/crates/atuin-dotfiles/src/shell.rs
+++ b/crates/atuin-dotfiles/src/shell.rs
@@ -16,16 +16,22 @@ pub struct Alias {
     pub value: String,
 }
 
-pub fn parse_alias(line: &str) -> Alias {
-    let mut parts = line.split('=');
+pub fn parse_alias(line: &str) -> Option<Alias> {
+    let parts: Vec<&str> = line.split('=').collect();
+
+    if parts.len() <= 1 {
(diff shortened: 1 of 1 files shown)


rigrig:
Actually...
I didn't pay a lot of attention when reporting that: I just deleted *all* aliases from my db to fix the problem, and the fortune was the obvious breakage.
But looking at the code now, it looks broken in another way: Fish `alias` output doesn't include a `=` sign, so it would just be e.g.:
```
alias emacs 'TERM=xterm-24bits emacs -nw --foo=bar'
alias k kubectl
```
(Fish allows setting them both ways though: both `alias k=kubectl` and `alias k kubectl` work)

ellie:
Ah I see! Well that needed fixing anyway, glad it's done.

#1972 should sort that

rigrig:
Urgh, it's still *sometimes* broken now: some ASCII art contains `=`...
I guess maybe it could redirect the `alias` output to a file and read that? But that seems like a lot of effort 😞

I don't really mind having to turn off the script for the initial import, and I (only just) figured out I can just `atuin dotfiles alias delete` those stray "aliases", instead of messing with the database.

PR #1973, opened 2024-04-22
docs: fix "From source" `cd` command

rigrig:
`atuin` was moved into the `crates/` directory in 95cc4720

## Checks


ellie (approved):
ah yes! thanks

PR #1976, opened 2024-04-23
fix(dotfiles): unquote aliases before quoting

ellie:
We want to quote all aliases when setting config. Aliases in the store may be quoted, but may also be unquoted.

First normalise to unquoted, then quote.

## Checks

### DIFF
--- a/crates/atuin-common/src/utils.rs
+++ b/crates/atuin-common/src/utils.rs
@@ -2,6 +2,8 @@ use std::borrow::Cow;
 use std::env;
 use std::path::PathBuf;
 
+use eyre::{eyre, Result};
+
 use rand::RngCore;
 use uuid::Uuid;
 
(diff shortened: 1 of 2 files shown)


tessus:
May I ask what happens with aliases like these:

```
alias wd='echo "`whoami`@`hostname -s`:`pwd`"'
alias cdate='if [ -n "${TZ+x}" ]; then   export SAVETZ=$TZ;   export TZ=${CLIENT_TZ};   date ;   export TZ=$SAVETZ;   unset SAVETZ; else   export TZ=${CLIENT_TZ};   date;   unset TZ ; fi'
alias dt='date +"%Y-%m-%d %H:%M:%S %z"'
```

ellie:
They all work without issue!

ellie:
Yep all good for imports too

tessus:
perfect. thanks for the info. looking at the code alone, I wasn't sure how quotes in the middle of the alias are handled.

P.S.: I had the chance to lookup string manipulation in Rust and I get the code now.

PR #1985, opened 2024-04-29
feat: add env syncing support for nushell

YummyOreo:
Adds support for env syncing for nushell, it auto updates because it just parses `autin dotfiles var list` at runtime.

## Checks

### DIFF
--- a/crates/atuin-dotfiles/src/shell/nu.rs
+++ b/crates/atuin-dotfiles/src/shell/nu.rs
@@ -1,0 +1,7 @@
+pub async fn var_config() -> String {
+    // Because nushell won't autoupdate, we just parse the output of `atuin dotfiles var list` in
+    // nushell and load the env vars that way
+
+    // we should only do this if the dotfiles are enabled
+    String::from(r#"atuin dotfiles var list | lines | parse "export {name}={value}" | reduce -f {} {|it, acc| $acc | upsert $it.name $it.value} | load-env"#)
+}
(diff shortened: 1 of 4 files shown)


crates/atuin-dotfiles/src/shell/nu.rs
```diff
@@ -0,0 +1,8 @@
+pub async fn var_config() -> String {
+    // Because nushell won't autoupdate, we just parse the output of `atuin dotfiles var list` in
+    // nushell and load the env vars that way
+
+    String::from(
+        r#"atuin dotfiles var list | lines | parse "export {name}={value}" | reduce -f {} {|it, acc| $acc | upsert $it.name $it.value} | load-env"#,
```
ellie, line 6: Does nushell have the same shellvar/env var difference?

`atuin dotfiles var list` will output 

```
foo=bar
```

for a shellvar, and 

```
export foo=var
```

for an env var

We'd probably want to handle both cases

YummyOreo: The only way I've found to do local variables would be to update the plugin file after it runs and waits till the user opens a new shell. That would fix the updating problem and alias problems.

ellie: I think that makes the most sense, in lieu of being able to load anything dynamically!

markovejnovic:
Thank you for the PR, but dotfiles have been removed in #4193 and we will no longer be supporting dotfiles.

PR #1991, opened 2024-04-30
refactor: preview_auto to use enum and different option

tessus:
This is a follow-up PR to #1804. It implements an enum and adds a new table `[preview]` (should be used for `max_preview_height` and `show_preview` when the new settings are implemented, although the keys should be renamed then) and the `show_preview_auto` option has been renamed as mentioned in https://github.com/atuinsh/atuin/pull/991#issuecomment-2081664065

```yaml
[preview]
strategy = auto (default) | static (max_of_all_cmds)   # was show_preview_auto
```

However, I am contemplating whether to change the naming:

strategy -> mode
auto -> dynamic

I believe 
- `mode` is used in other settings and thus it might be more consistent.
- `dynamic` as it determines the height dynamically. `auto` also fits, so I really have no clue which one is better.

@ellie what do you think?

## Checks

### DIFF
--- a/crates/atuin-client/config.toml
+++ b/crates/atuin-client/config.toml
@@ -71,12 +71,7 @@
 
 ## enable or disable showing a preview of the selected command
 ## useful when the command is longer than the terminal width and is cut off
-# show_preview = false
-
-## enable or disable automatic preview. It shows a preview, if the command is
-## longer than the width of the terminal. It respects max_preview_height.
-## (default: true)
-# show_preview_auto = true
+# show_preview = true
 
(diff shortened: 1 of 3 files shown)


ellie (approved):
thank you!

ellie:
I think mode vs strategy is a bit ambiguous as while we use mode elsewhere, it's not quite the same thing (search mode, filter mode, etc are all for searching history and aren't "smart" decision makers)

Auto seems fine to me

tessus:
Great. I'll update the docs accordingly later tonight.

PR #1995, opened 2024-05-02
fix(dotfiles): allow clearing aliases, disable import

ellie:
At the moment there are far too many edge cases to handle importing aliases.

1. We need an interactive shell to print aliases. Without it, most shells won't report much.
2. Many people have their shells print things on startup (graphics, fortunes, etc). This could be detected as an attempt to set an alias.

 Rather than spend the next year finding import edge cases, I'm
 disabling it for now. There's probably a better way we can do this?

## Checks

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -207,6 +207,7 @@ dependencies = [
  "itertools",
  "log",
  "ratatui",
+ "regex",
  "rpassword",
  "runtime-format",
  "rustix",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,7 +9,7 @@ exclude = ["ui/backend"]
 [workspace.package]
 version = "18.2.0"
 authors = ["Ellie Huxtable <ellie@elliehuxtable.com>"]
-rust-version = "1.67"
+rust-version = "1.70"
 license = "MIT"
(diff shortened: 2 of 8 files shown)


tessus:
Doesn't this change now clear aliases no matter what? e.g. also manually set aliases?

ellie:
No matter what? The intent is for it to allow deleting all aliases to remove Atuin set config 

I'm not sure on a use case for only deleting imported aliases - this was prompted by the linked issue but I think it's a good thing to have regardless

tessus:
No, maybe I expressed myself poorly. I meant the following: in the function `run`, the alias store is cleared. Isn't `run` being run as an init function when atuin starts? This means all aliases are removed every time atuin is started.

P.S.: Since gh only shows the diff, I might have misunderstood the `run` function.

ellie:
Ah I see! No, it doesn't. Run is part of the `atuin dotfiles alias` command, which contains a pattern match. The pattern match switches based on which subcommand is required, so whether that's `alias set`, `alias clear`, etc.

tessus:
Thanks. Next time I'll click on `view file`...
Sorry for the noise.

PR #2000, opened 2024-05-04
Added "accept-to-edit" for (Back)Space / Home / Left

### LINKED ISSUE #1906: Feature Suggestion: Left Arrow/Backspace to start edit
I often mistype commands, and my muscle memory has me hit up arrow, left arrow to start editing it. With atuin this of course doesn't do anything, I need to hit up arrow, tab, left arrow. 

Hitting up arrow shows me commands, which is perfect. Hitting left right after, with no search, does nothing. My suggestion is, if there is no search and the user presses left, start editing the currently selected command. Basically, left is like tab+left if the search box is empty. I think the same suggestion would make sense for backspace with no search.

I think this would help with people's long entrenched muscle memory, and also be a natural way of working when scrolling back through commands with up arrow. What do you think?

ps, Apologies if this has been suggested. I couldn't find in google or in github search

rwlr:
The above keys edit the selected command (as with hitting Tab), but like if using them in the shell. They only trigger when the search input is currently empty, and must be enabled individually with the respective boolean config options:

* exit_with_backspace: right-trim command and remove last character

* exit_with_space: right-trim and append a space character

* exit_with_home: start editing at position 0

* exit_with_cursor_left: right-trim and start editing at last character

* exit_positions_cursor: enable cursor positioning for Home and Left keys (bash-only)

Updated bash init to support cursor positioning. For other shells, Home and Left keys will behave like Tab.

Resolves #1906

## Checks

### DIFF
--- a/crates/atuin/src/command/client/search/interactive.rs
+++ b/crates/atuin/src/command/client/search/interactive.rs
@@ -49,13 +49,19 @@ use ratatui::{
 const TAB_TITLES: [&str; 2] = ["Search", "Inspect"];
 
 pub enum InputAction {
-    Accept(usize),
+    Accept(usize, InputAcceptKind),
     Copy(usize),
     Delete(usize),
     ReturnOriginal,
     ReturnQuery,
     Continue,
     Redraw,
+}
+pub enum InputAcceptKind {
+    Default,
+    Backspace,
+    Space,
+    Offset(i64),
 }
 
 #[allow(clippy::struct_field_names)]
@@ -214,7 +220,10 @@ impl State {
             KeyCode::Char('c' | 'g') if ctrl => Some(InputAction::ReturnOriginal),
             KeyCode::Esc if esc_allow_exit => Some(Self::handle_key_exit(settings)),
             KeyCode::Char('[') if ctrl && esc_allow_exit => Some(Self::handle_key_exit(settings)),
-            KeyCode::Tab => Some(InputAction::Accept(self.results_state.selected())),
+            KeyCode::Tab => Some(InputAction::Accept(
+                self.results_state.selected(),
+                InputAcceptKind::Default,
+            )),
             KeyCode::Char('o') if ctrl => {
                 self.tab_index = (self.tab_index + 1) % TAB_TITLES.len();
 
@@ -273,7 +282,7 @@ impl State {
         if settings.enter_accept {
             self.accept = true;
         }
-        InputAction::Accept(self.results_state.selected())
+        InputAction::Accept(self.results_state.selected(), InputAcceptKind::Default)
     }
 
     #[allow(clippy::too_many_lines)]
@@ -374,7 +383,10 @@ impl State {
             }
             KeyCode::Char(c @ '1'..='9') if modfr => {
                 return c.to_digit(10).map_or(InputAction::Continue, |c| {
-                    InputAction::Accept(self.results_state.selected() + c as usize)
+                    InputAction::Accept(
+                        self.results_state.selected() + c as usize,
+                        InputAcceptKind::Default,
+                    )
                 })
             }
             KeyCode::Left if ctrl => self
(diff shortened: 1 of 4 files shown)


crates/atuin/src/command/client/search/interactive.rs
```diff
@@ -1134,12 +1175,53 @@ pub async fn history(
...
+                                // counting from the start
+                                length.min(offset)
+                            } else {
+                                // start editing at specified position,
+                                // counting from the (trimmed) end
+                                0.max(length - offset.abs())
+                            };
+                            // Bash's READLINE_POINT is required or positioning the cursor; we still allow
+                            // to go back to the shell even when not using bash (if user's configure the
+                            // respective options), but the actual positioning is only possible with bash
```
akinomyoga, line 1217: I think it's possible with Zsh. The text before the cursor position can be stored in `LBUFFER`, and the text after the cursor position can be stored in `RBUFFER`

crates/atuin/src/shell/atuin.bash
```diff
@@ -263,6 +264,17 @@ __atuin_history() {
...
+        # Edit selected command at specified cursor position
+        __atuin_output=${__atuin_output#__atuin_edit_at__:}
+        local __atuin_edit_at_rx='^([0-9]+):(.*)$'
+        if [[ $__atuin_output =~ $__atuin_edit_at_rx ]]; then
+            READLINE_LINE="${BASH_REMATCH[2]}"
+            READLINE_POINT=${BASH_REMATCH[1]}
+        else
+            READLINE_LINE=$__atuin_output
+            READLINE_POINT=${#READLINE_LINE}
+        fi
```
akinomyoga, line 277: It could probably simplified as

```suggestion
    elif [[ $__atuin_output == $__atuin_edit_at_rx ]]; then
        # Edit selected command at specified cursor position
        READLINE_LINE="${BASH_REMATCH[2]}"
        READLINE_POINT=${BASH_REMATCH[1]}
```

with

```bash
local __atuin_edit_at_rx='^__atuin_edit_at:([0-9]+):(.*)$'
```

defined earlier.

akinomyoga (commented):
For the finer settings of the keybindings, we might want to add/find a general framework in the future (#193). A naive question is whether we would add detailed switches, which would be deprecated if #193 is implemented.

crates/atuin/src/command/client/search/interactive.rs
rwlr: Thanks for the pointer. I'll have a look at L/RBUFFER and zsh.

crates/atuin/src/command/client/search/interactive.rs
```diff
@@ -1134,12 +1175,53 @@ pub async fn history(
...
+                            let position = if offset >= 0 {
+                                // start editing at specified position,
+                                // counting from the start
+                                length.min(offset)
+                            } else {
+                                // start editing at specified position,
+                                // counting from the (trimmed) end
+                                0.max(length - offset.abs())
+                            };
+                            // Bash's READLINE_POINT is required or positioning the cursor; we still allow
```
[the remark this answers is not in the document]
rwlr: Yes, typo.

crates/atuin/src/shell/atuin.bash
rwlr: Indeed, just with =~.

rwlr:
> For the finer settings of the keybindings, we might want to add/find a general framework in the future (#193). A naive question is whether we would add detailed switches, which would be deprecated if #193 is implemented.

I agree. Just looking at my changes, you'd probably also want to add Ctrl-W, Ctrl-Left, and who knows what else...being able to achieve the same via some kind of configuration would be much nicer than having to change code.

Looking through #193 it seems #1763 also had the same kind of feature in mind ("accept-to-edit" keys).

barries:
This would reduce friction in a very common use case for me, king of typos.

I very often want to go back to a recent command and add or fix an option, it's very natural for me to Up-Arrow to the command and want to hit Left-Arrow or Backspace to edit it. So, `exit_with_cursor_left` (or the slightly more grammatical `exit_on_cursor_left`) and `exit_with_backspace` (`exit_on_backspace`?) would rock.

romain-ai:
any update on this ? there may be a workaround / key remap fix that is possible ?

ellie:
Hey! Thanks for the work here, but I've decided to go ahead with a more minimal and simpler implementation in #3090 

The new default has up-left or up-backspace default to the same as up-tab

As said previously, something like #193 would be great longer term.

Issue #2020, opened 2024-05-13
Do not print password in error message on server startup


yggdr:
### What did you expect to happen?

Some filtering on the configured `db_uri` to redact the configured password.

### What happened?

When reinstalling atuin, due to some misconfiguration on my part, it printed this error message on startup:
```
Error: failed to connect to db: PostgresSettings { db_uri: "postgres://atuin:PASSWORD_HERE@10.0.9.105/atuin" }

Caused by:
[…]
```
which included the password.

### Atuin doctor output

```yaml
Atuin Doctor
Checking for diagnostics

[Filesystem] ZFS is known to have some issues with SQLite. Atuin uses SQLite heavily. If you are having poor performance, there are some workarounds here: https://github.com/atuinsh/atuin/issues/952

Please include the output below with any bug reports or issues

atuin:
  version: 18.2.0
  sync:
    cloud: false
    records: true
    auto_sync: true
    last_sync: 2024-05-13 12:28:42.386454986 +00:00:00
shell:
  name: 'python: xonsh'
  default: unknown
  plugins:
  - atuin
system:
  os: FreeBSD
  arch: amd64
  version: '14.0'
  disks:
[…]
```
```


### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

ellie:
Resolved by #2032

PR #2032, opened 2024-05-20
fix: redact password in database URI when logging

jeremycline:
Hi,

I think this is a bit of a niche fix since I imagine not tons of people self-host and perhaps it's something those who are running servers _do_ want, so I won't be the least bit bothered if this isn't merged.

Previously, in the event that there was a configuration issue and the atuin server failed to connect to PostgreSQL, it would log the database URI including the password.

For example, if the password authentication failed the following log message would be printed:

Error: failed to connect to db: PostgresSettings { db_uri: "postgres://atuin:definitelymypassword@db.example.com/atuin" }

This change sets the password to "****" when printing it via Debug:

Error: failed to connect to db: PostgresSettings { db_uri: "postgres://atuin:****@db.example.com/atuin" }

Hopefully few people use **** as the actual password.

## Checks

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -442,6 +442,7 @@ dependencies = [
  "sqlx",
  "time",
  "tracing",
+ "url",
  "uuid",
 ]
 
--- a/crates/atuin-server-postgres/Cargo.toml
+++ b/crates/atuin-server-postgres/Cargo.toml
@@ -21,3 +21,4 @@ sqlx = { workspace = true }
 async-trait = { workspace = true }
 uuid = { workspace = true }
 futures-util = "0.3"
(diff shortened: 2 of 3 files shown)


tessus:
LGTM.

I have also often seen that the length of the password is kept (just the characters are replaced by `*`), but the current solution is more secure as it doesn't allow to infer the password length. 👍 

Some people also deem the username, hostname, and db name as privileged information, but unless you work for one of the 3 or 4 letter gov orgs, there is no need to entertain that. ;-)

Edit: Can you maybe add the following to your comment above: fixes #2020

ellie (approved):
Looks good, thank you! 🥳 

Issue #2031, opened 2024-05-19
[Bug]: Unable to install on ARM64 - failed to select a version for `env_logger`.


victorhooi:
### What did you expect to happen?

Atuin should be able to compile successfully on ARM64.

### What happened?

I attempted to install Atuin using the commands under Quickstart in the Atuin README.

However, it failed to compile an ARM64 package - I'm not a Rust expert, but it seems like some kind of version mismatch on env_logger?

```
bash (curl --proto '=https' --tlsv1.2 -sSf https://setup.atuin.sh | psub)

 _______  _______  __   __  ___   __    _
|   _   ||       ||  | |  ||   | |  |  | |
|  |_|  ||_     _||  | |  ||   | |   |_| |
|       |  |   |  |  |_|  ||   | |       |
|       |  |   |  |       ||   | |  _    |
|   _   |  |   |  |       ||   | | | |   |
|__| |__|  |___|  |_______||___| |_|  |__|

Magical shell history

Atuin setup
https://github.com/atuinsh/atuin

Please file an issue if you encounter any problems!

===============================================================================

Detected Linux!
Checking distro...
debian detected, but not amd64
Unknown or unsupported OS or architecture
Please check the README at https://github.com/atuinsh/atuin for manual install instructions
If you have any problems, please open an issue!
Do you wish to attempt an install with 'cargo'? [Y/N] y
Attempting install with cargo
    Updating crates.io index
  Downloaded atuin v18.2.0
  Downloaded 1 crate (92.8 KB) in 0.12s
  Installing atuin v18.2.0
    Updating crates.io index
error: failed to compile `atuin v18.2.0`, intermediate artifacts can be found at `/tmp/cargo-installaQ0Kou`

Caused by:
  failed to select a version for `env_logger`.
      ... required by package `atuin v18.2.0`
  versions that meet the requirements `^0.11.2` are: 0.11.3, 0.11.2

  the package `atuin` depends on `env_logger`, with features: `anstream` but `env_logger` does not have these features.
   It has an optional dependency with that name, but that dependency uses the "dep:" syntax in the features table, so it does not have an implicit feature with that name.


  failed to select a version for `env_logger` which could resolve this conflict
```

### Atuin doctor output

```yaml
Not available - as Atuin was unable to be installed.
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

ellie:
Hey! Thanks for the report

I'm struggling to replicate this one 🤔 

If you run 

[...]

PR #2068, opened 2024-05-31
install.sh: log cargo and rustc version

xvello:
Issue https://github.com/atuinsh/atuin/issues/2031 was caused by an outdated rustc version. 

The build should ideally fail with a better error message (is the workspace's `rust-version` lower than this dep's?), but in the meantime let's log both cargo and rustc versions to help with support.

## Checks

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -156,6 +156,10 @@ __atuin_install_cargo(){
 
 	fi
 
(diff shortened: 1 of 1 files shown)


ellie (approved):
thank you! 😊

PR #2075, opened 2024-06-02
feat(doctor): report sqlite version

xvello:
Report the sqlite version to detect unsupported versions possibly brought in by third-party patching.

We could also grab the `libsqlite3_sys::SQLITE_VERSION` constant, but it requires listing this transitive dependencies in the deps, as sqlx_sqlite does not re-export this constant. Opening a sqlite connection also allows querying the build flags, which we might want to log too.

## Checks


ellie (approved):
looks goooood, thank you! 🚀

Issue #2030, opened 2024-05-18
[Bug]: could not login due to version mismatch


newro:
### What did you expect to happen?

succeed login and sync ok

### What happened?

[2024-05-18T13:17:19Z DEBUG reqwest::connect] starting new connection: https://api.atuin.sh/
[2024-05-18T13:17:19Z DEBUG hyper::client::connect::dns] resolving host="api.atuin.sh"
[2024-05-18T13:17:19Z DEBUG hyper::client::connect::http] connecting to 172.67.70.19:443
[2024-05-18T13:17:19Z DEBUG hyper::client::connect::http] connected to 172.67.70.19:443
[2024-05-18T13:17:19Z DEBUG rustls::client::hs] No cached session for DnsName("api.atuin.sh")
[2024-05-18T13:17:19Z DEBUG rustls::client::hs] Not resuming any session
[2024-05-18T13:17:19Z DEBUG rustls::client::hs] Using ciphersuite TLS13_AES_256_GCM_SHA384
[2024-05-18T13:17:19Z DEBUG rustls::client::tls13] Not resuming
[2024-05-18T13:17:19Z DEBUG rustls::client::tls13] TLS1.3 encrypted extensions: [ServerNameAck, Protocols([ProtocolName(6832)])]
[2024-05-18T13:17:19Z DEBUG rustls::client::hs] ALPN protocol is Some(b"h2")
[2024-05-18T13:17:19Z DEBUG h2::client] binding client connection
[2024-05-18T13:17:19Z DEBUG h2::client] client connection bound
[2024-05-18T13:17:19Z DEBUG h2::codec::framed_write] send frame=Settings { flags: (0x0), enable_push: 0, initial_window_size: 2097152, max_frame_size: 16384 }
[2024-05-18T13:17:19Z DEBUG h2::proto::connection] Connection; peer=Client
[2024-05-18T13:17:19Z DEBUG h2::codec::framed_write] send frame=WindowUpdate { stream_id: StreamId(0), size_increment: 5177345 }
[2024-05-18T13:17:19Z DEBUG hyper::client::pool] pooling idle connection for ("https", api.atuin.sh)
[2024-05-18T13:17:19Z DEBUG h2::codec::framed_write] send frame=Headers { stream_id: StreamId(1), flags: (0x4: END_HEADERS) }
[...]

PR #2076, opened 2024-06-02
fix(client): better error reporting on login/registration

xvello:
Closes https://github.com/atuinsh/atuin/issues/2030 - unless you downgraded `api.atuin.sh` by mistake, this is misleading error reporting on non-200 responses:

- Use the `handle_resp_error` helper on the registration and login code paths for consistency. Tune its strings to be a bit more generic, feel free to reword
- Only check the server version if the request is a 200 (if we get a 503 from the CDN, the version check would run and wrongly report)
- Remove the version fallback from `ensure_version` if the header is not preset. Since we bumped the major version, the fallback is not needed + the header has been present for 6 months

Tangentially related, should `handle_resp_error` directly call `ensure_version` now, to ensure a consistent behaviour?

## Checks

### DIFF
--- a/crates/atuin-client/src/api_client.rs
+++ b/crates/atuin-client/src/api_client.rs
@@ -61,14 +61,10 @@ pub async fn register(
         .json(&map)
         .send()
         .await?;
+    let resp = handle_resp_error(resp).await?;
 
     if !ensure_version(&resp)? {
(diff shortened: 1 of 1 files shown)


xvello:
Yes, my weekend has been very rainy, thank you very much :umbrella:

Issue #2049, opened 2024-05-27
[Bug]: no disable


iconoclasthero:
### What did you expect to happen?

i want this it stop. i tried atuin disable and atuin off
either one of those should work to stop something I installed by clicking a url that installed something automatically via curl.
no answer here:

$ atuin help
atuin 18.2.0
Ellie Huxtable <e@elm.sh>
Magical shell history

Usage:
  atuin <COMMAND>

Commands:
  history          Manipulate shell history
  import           Import shell history from file
  stats            Calculate statistics for your history
  search           Interactive history search
  sync             Sync with the configured server
  login            Login to the configured server
  logout           Log out
  register         Register with the configured server
  key              Print the encryption key for transfer to another machine
  status           Display the sync status
  account          Manage your sync account
  kv               Get or set small key-value pairs
  store            Manage the atuin data store
  dotfiles         Manage your dotfiles with Atuin
  init             Print Atuin's shell init script
  info             Information about dotfiles locations and ENV vars
  doctor           Run the doctor to check for common issues
  default-config   Print example configuration
  server           Start an atuin server
  uuid             Generate a UUID
  contributors     
  gen-completions  Generate shell completions
  help             Print this message or the help of the given subcommand(s)

Options:
  -h, --help     Print help
  -V, --version  Print version

### What happened?

i installed it.

https://docs.atuin.sh/uninstall/
doesn't tell me how to remove it in linux.

### Atuin doctor output

```yaml
no.
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

iconoclasthero:
also, this is a problem in my ~/.bashrc:
```
. "$HOME/.googlerc"

#. "$HOME/.cargo/env"

#THIS MUST BE AT THE END OF THE FILE FOR SDKMAN TO WORK!!!
export SDKMAN_DIR="$HOME/.sdkman"
[[ -s "$HOME/.sdkman/bin/sdkman-init.sh" ]] && source "$HOME/.sdkman/bin/sdkman-init.sh"




[[ -f ~/.bash-preexec.sh ]] && source ~/.bash-preexec.sh
eval "$(atuin init bash)"
```

iconoclasthero:
WRT to your code of conduct, i consider it a violation of the code of conduct for a developer to have a curl line like that.  Also, not to have a way to stop something like atuin which dramatically changes how i interact with my shell after install.  Also, not to have a clear way to uninstall something like this.  

[...]

Issue #1546, opened 2024-01-11
How to completely uninstall?


sunglim:
![image](https://github.com/atuinsh/atuin/assets/756988/368438ba-dd5a-4002-8b80-a6e9c35b2bbc)

I'm using mac and bash

yesterday I installed atuin, but now the screen is completely broken. I removed atuin via homebrew. but uparrow still blame atuin is not existing.

I searched 'uninstall atuin' on google. and read the https://docs.atuin.sh/.  but couldn't find any document about uninstallation.

ellie:
Dupe of #1529, #649, #111, #372, #671 

I'll add a section to the docs, however:

You need to edit your zshrc/bashrc/fish config to remove the Atuin integration. Uninstalling the binary depends on your system.

12932:
> Mind if I ask why Atuin didn't work out for you?

If you look at their screenshot the output is sort of messed up. I personally wasn't aware that atuin provided this feature automatically when pressing the up arrow to go through shell history

ellie:
> If you look at their screenshot the output is sort of messed up

Hmm I see. Probably an altscreen issue from tmux/screen/their terminal/whatever. Fixable with a small config change with most. Annoyingly there's seemingly infinite combinations of setups, so making it work for everyone right out of the box is a game of whack-a-mole. 

Looks like I can see some vim/some editor output somewhere in there too

> I personally wasn't aware that atuin provided this feature automatically when pressing the up arrow to go through shell history

Yeah we hear this a lot - mind if I ask which parts of the docs you read for the install? I tried to make sure it's mentioned almost all over, but it's still missed a lot so I'd like to improve that. People feel strongly about the up arrow (both for and against), though there is no consensus as to whether it should be enabled by default or not.

foice:
> Mind if I ask why Atuin didn't work out for you?

Not being able to edit the line I just recovered from this otherwise very smart way to navigating the history. Going back to fuzzy finder fzf ...

ellie:
You can press tab to select the command without executing it, and then edit it there

Or 

enter_accept = false

In your config to make this the default

foice:
This helps a lot to get closer to the desired behavior!

Issue #1529, opened 2024-01-09
Uninstall option


furicle:
While no one of sound mind would want to stop using atuin after they've figured out out, sometimes weird things happen, and you need to move on.

It would be nice if atuin offered an uninstall option, if run on a system where atuin is already present.

ellie:
Hey!

> While no one of sound mind would want to stop using atuin

<3

The issue with doing the uninstall via Atuin, is that we have no idea how the user installed it. If installed via the system package manager, that is the route it will need to be uninstalled by. Our install script is literally just a switch that will use the system package manager wherever possible. 

It could maybe come with a switch to do the reverse, and uninstall instead, but the user will still need to edit their shell config by hand

furicle:
Yes, just a switch to do the reverse I _think_ would be helpful.

Could something as simple as a grep though the shell config files for 'atuin' at least suggest to the user what they'd need to remove from their shell config as well?

I'm not familiar enough with all the possible config files to know just how effective that would be.

ellie:
It really depends on how you installed it, Atuin is available in many system package managers. If you let me know

- how you installed it
- which system you use

Then I'd be able to suggest that for you.

More here: https://docs.atuin.sh/uninstall/

> cause of unpredictable behavior and lot of terminal weird behaviour only on systems where i installed and used it.

Are you able to elaborate?

Issue #1485, opened 2024-01-02
how to unistall 


AugustinMauroy:
i had install attuin to test it but I din't need it. So how to uninstall it.

arcuru:
Well, you should uninstall it using the same method you installed it, and then remove the atuin specific lines from your shell init script.

If you used the install script here, it may have installed from your package manager or by using `cargo install`. You should uninstall it using the method it was installed with. The install script will have printed output that told you how it was installing it.

Since you haven't specified your OS, package manager, or shell, I can't help you any further than that.

AugustinMauroy:
I use MacOS, and I use simple shell command provided by the website

binary-signal:
which command did you use specifically  ?

ellie:
If you used the install script, and you're using macOS, you _probably_ want to run

```
brew uninstall atuin
```

and then remove the Atuin line from your `.zshrc`

We may want to add an install script that does the reverse, or maybe just make it an arg on the script.

AugustinMauroy:
@binary-signal 

```
 bash <(curl --proto '=https' --tlsv1.2 -sSf https://setup.atuin.sh)
 ```

**Comment by binary-signal** [comment]
**Assuming you run ubuntu and used the default install options**

To uninstall atuin 

First run
```shell
sudo  dpkg --purge atuin # purge will remove the configuration too let's hope.
```

Then delete the following folders, just in case --purge command has left leftovers behind.
```shell
rm -rf ~/.config/atuin
rm -rf ~/.local/share/atuin
```

Finally, you need to remove from .zshrc or .bashrc or (insert your shell config file) the line added by atuin, 
example for zsh config file
```shell
# file: .zshrc
bla bla bla
bla bla bla 
....
atuin init zsh #  delete this line, save changes, relaunch shell
```

That's all.

PR #2085, opened 2024-06-04
chore: switch to cargo dist for releases

ellie:
I'd like to see how CI behaves here before merging anything.

From https://axo.dev

cargo-dist handles building releases far better than we can, and do so
for several large projects now.

We will need to change our install script to use the cargo-dist
installer. To start off, I think it would make sense to use this as a fallback instead of using cargo install

Historically, we have used the system package manager wherever possible.
Once 100% switched to the new installer, this will no longer be the case. If
the user wishes to use their package manager, and Atuin is maintained
there, then they can choose to do so.

This way, we can ensure that users are running a known build, can easily
uninstall (just delete the atuin dir), easily update, etc. Builds will
use our lockfile, and can have their checksum verified. Later, I'd like
to introduce build signing.

As Axo are focused on release engineering, they will likely have
resolved many more issues than we have - libc versions, etc.

I'm not particularly happy with our response of "just use your package
manager", as many users seem to have difficulty there. It's unclear what
our installer has done, as this behaviour varies massively across
systems. It's also unclear how some package maintainers may have patched
things

I'm hoping that some better release tooling will lead to more confidence
in the process, and therefore more frequent releases.

Uninstall clarity: #111, #372, #640, #1485, #1546, #2049, #1529


ellie:
I'd like to figure out how to update nix, but min rust versions can be dealt with elsewhere

I'm going to merge this, and use it to ship a pre-release. We already have weekly releases, so this would be a pretty nice replacement for that

Wrt the install script, after some thought I'd rather not totally replace it immediately. To begin with, we can use the cargo-dist installer as a fallback instead of cargo-install

PR #2091, opened 2024-06-05
chore: build illumos releases

ellie:
## Checks


ellie:
ok this didn't work and I'm not surprised, but I wonder how far we can get

ellie:
gonna wait until cross compiling is better supported

PR #2122, opened 2024-06-12
chore(build): compile protobufs with protox

### LINKED ISSUE #2119: Remove protoc build dependency
Hi, I noticed you're using `protoc` in your build, which makes many aspects of the build more complicated (nix, Dockerfile, cross, github workflows...) But more importantly, users cannot easily `cargo install` from source anymore. There is a pure-rust protobuf compiler implementation called [protox](https://github.com/andrewhickman/protox), which can be used as a library in the build script of `atuin-daemon`. I have a working version of this locally.

My question is, are you aware of `protox` and did you already decide to not use it for some reason?

Otherwise, if you have no objections, I'll prepare a PR with the simplified build.

senekor:
protox is a pure-rust implementation of the protobuf compiler. Therefore, it can be managed by cargo.

This removes the implicit dependency on protoc being available in the environment for the build.

## Checks

closes #2119

### DIFF
--- a/crates/atuin-daemon/build.rs
+++ b/crates/atuin-daemon/build.rs
@@ -1,4 +1,17 @@
-fn main() -> Result<(), Box<dyn std::error::Error>> {
-    tonic_build::compile_protos("./proto/history.proto")?;
-    Ok(())
+use std::{env, fs, path::PathBuf};
+
+use protox::prost::Message;
+
+fn main() -> std::io::Result<()> {
+    let file_descriptors = protox::compile(["history.proto"], ["./proto/"]).unwrap();
+
+    let file_descriptor_path = PathBuf::from(env::var_os("OUT_DIR").expect("OUT_DIR not set"))
+        .join("file_descriptor_set.bin");
+    fs::write(&file_descriptor_path, file_descriptors.encode_to_vec()).unwrap();
+
+    tonic_build::configure()
+        .build_server(true)
+        .file_descriptor_set_path(&file_descriptor_path)
+        .skip_protoc_run()
+        .compile(&["root.proto"], &["."])
 }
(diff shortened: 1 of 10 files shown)


crates/atuin-daemon/build.rs
```diff
@@ -1,4 +1,17 @@
...
+
+    let file_descriptor_path = PathBuf::from(env::var_os("OUT_DIR").expect("OUT_DIR not set"))
+        .join("file_descriptor_set.bin");
+    fs::write(&file_descriptor_path, file_descriptors.encode_to_vec()).unwrap();
+
+    tonic_build::configure()
+        .build_server(true)
+        .file_descriptor_set_path(&file_descriptor_path)
+        .skip_protoc_run()
+        .compile(&["root.proto"], &["."])
```
ellie, line 16:
```suggestion
        .compile(&["history.proto"], &["."])
```

senekor: Huh. copypasta. I was wondering why this even worked with the typo, so I dug a bit in the source of prost-build. Turns out when `skip_protoc_run` is set, this string is used for almost nothing except a `cargo:rerun-if-changed` directive. So the build works the same, but the copypasta would lead to the build being cached incorrectly if the protobuf definitions are changed.

By the way, do you prefer incremental commits on a PR or clean history with amend / rebase & force-push?

ellie: I was also surprised it built ok! That makes sense, thanks for digging into it.

Whichever works best for you. If it's the sort of larger/complex PR where incremental commits tell a better story and are easier to review then it's appreciated, but it's all squashed anyway so not a big deal.

ellie (approved):
Excellent, thank you!

senekor:
Cool, thanks! :slightly_smiling_face:

ellie:
Oh! You're also contributor number 200 🥳

PR #2123, opened 2024-06-12
De-couple atuin nix build from nixpkgs rustc version

### LINKED ISSUE #2089: Rust 1.78 in nix
As part of #2085, I've been trying to ensure that we properly stick to a MSRV, as stated in CONTRIBUTING.md

> We commit to supporting the latest stable version of Rust - nothing more, nothing less, no nightly.

Currently, this is 1.78. However updating the toolchain file leads to nix complaining that only 1.77 is available. 

I'm not familiar enough with nix to update this, and would love to have a way of updating MSRV to the latest stable that doesn't require some nix magic

1.79 is out in a few days, so this will need changing again

Sciencentistguy:
Closes #2089 

This PR moves the nix flake to using a rust toolchain provided by fenix (https://github.com/nix-community/fenix), which downloads the toolchain declared in `./rust-toolchain.toml`, removing the requirement on nixpkgs to update rustc for us to bump MSRV.

When the value in `./rust-toolchain.toml` is changed, the nix build must be run locally once (with `nix build -L`). This will print the new sha256 hash for the new rust toolchain, that must then be inserted into `flake.nix` before the build will succeed. This may be automatable via a github action - that is not included in this PR.

## Checks

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,5 +1,26 @@
 {
   "nodes": {
+    "fenix": {
+      "inputs": {
+        "nixpkgs": [
+          "nixpkgs"
+        ],
+        "rust-analyzer-src": "rust-analyzer-src"
+      },
+      "locked": {
+        "lastModified": 1717827974,
+        "narHash": "sha256-ixopuTeTouxqTxfMuzs6IaRttbT8JqRW5C9Q/57WxQw=",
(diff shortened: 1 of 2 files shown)


ellie (approved):
nice! thanks for sorting this out :D

PR #2139, opened 2024-06-13
feat(gui): automatically install and setup the cli/shell

ellie:
Big WIP right now, still need to handle the shell plugin install, clicky clicky bits, etc

I'd like to eventually polish and move some of the auto setup stuff to `atuin setup`

## Checks


ellie:
if there's not atuin in the PATH, run the installer. If there's no shell plugin, install it

PR #2140, opened 2024-06-14
perf: create idx cache table

ellie:
Let's stop that P99 from creeping up eh?

Screenshot 2024-07-01 at 13 48 15

Postgres doesn't like group by aggregates very much (it's OLTP afterall). I'll follow up with the actual caching logic

## Checks

### DIFF
--- a/crates/atuin-server-postgres/migrations/20240614104159_idx-cache.sql
+++ b/crates/atuin-server-postgres/migrations/20240614104159_idx-cache.sql
@@ -1,6 +1,3 @@
--- using group by and max is slow. as the store grows, latency is creeping up.
--- get a handle on it!
-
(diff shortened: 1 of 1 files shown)


crates/atuin-server-postgres/migrations/20240614104159_idx-cache.sql
```diff
@@ -0,0 +1,43 @@
+-- using group by and max is slow. as the store grows, latency is creeping up.
+-- get a handle on it!
+-- Really this sort of workload (lots of aggregation) is better suited to OLAP,
+-- but let's just stick with postgres until we can't any more
```
ellie, line 4:
```suggestion

```

PR #2162, opened 2024-06-18
fix(gui): add support for checking if the cli is installed on windows

YummyOreo:
On windows, this makes the default shell `powershell` and adds support for checking if atuin is installed on windows in the gui
## Checks

### DIFF
--- a/crates/atuin-common/src/shell.rs
+++ b/crates/atuin-common/src/shell.rs
@@ -95,7 +95,7 @@ impl Shell {
                 "dscl localhost -read \"/Local/Default/Users/$USER\" shell | awk '{print $2}'",
             ])?
         } else if cfg!(windows) {
-            return Ok(Shell::Powershell)
+            return Ok(Shell::Powershell);
         } else {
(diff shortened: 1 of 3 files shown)


ellie (approved):
Excellent - thank you so much!

I'm still pretty surprised that's all we needed for Windows. I'm doing a bunch more with ptys fairly soon which I imagine _might_ break things more though 😢

CarterLi:
Just find this in the changelog of the latest release. Does it mean that we now support Windows?

YummyOreo:
> Just find this in the changelog of the latest release. Does it mean that we now support Windows?

Tentatively. It builds, but there isnt a powershell plugin yet (but there is a nushell plugin).

ellie:
It will usually build and run on Windows thanks to @YummyOreo's work, but we don't really have the bandwidth (time, hardware, etc) to test and support it properly.

CarterLi:
I'll have a try

CarterLi:
image

It works!

CarterLi:
This is a binary if someone don't want to built it himself. Don't know why it's so large though

[atuin.zip](https://github.com/user-attachments/files/15973911/atuin.zip)

Issue #816, opened 2023-03-26
Add history export (and re-import) option, machine-readable


shoffmeister:
It could be a useful addition to atuin to offer history export, reimport/merge in a machine-readable, open, stable format.

I did find the SQLite database in ~/.local/share/atuin, but that's "locked away" in a sense, IMHO, requiring SQLite tooling.

Using JSON as data exchange format, it would be possible to
* backup transparently the history - e.g. `atuin history list --json`, sitting next to the `--format` option
* replace existing history, e.g. `atuin history replace --json < my-history.json`
* merge / append existing history, e.g. `atuin history append --json < my-history.json`
* run filtering, e.g. `atuin history list --json | jq ' <fancy-filtering-here> ' | atuin history append --json`

This would be particularly useful in scenarios where there is no Atuin server around, only sneakernet.

Disclaimer: I only just installed `atuin`, with one major attraction being the global shell history across bash sessions (something that `fzf` does not / cannot offer).

conradludgate:
> * backup transparently the history - e.g. atuin history list --json, sitting next to the --format option
> * merge / append existing history, e.g. atuin history append --json < my-history.json

You might know we already have a sync service. This uses JSON already, but the payloads are encrypted.

In theory we could have a cli version too. But it needs a proper design. We have some ideas on the roadmap. I'll see if we can fit this in with our new ideas too.

Thanks for the suggestion

tennox:
"Workaround" script (nushell, but you get the point): 
```nushell
atuin search -c $session_dir --format '"command": "{command}", "directory": "{directory}", "duration": "{duration}", "user": "{user}", "host": "{host}", "time": "{time}", "exit": "{exit}", "relativetime": "{relativetime}"' 
        | trim | each {|s| '{' + $s + '}'}
```
(note: can't put `{` in format string, not even `\{` :sweat_smile: )

tombh:
Here's a little script I'm using to keep my Atuin history synced with my Nushell history. I run it to both import and to sync when I delete items from history.

The only reason I sync Atuin and Nushell is just because I like those ZSH-style autocompletions as you type, which as far as I know Atuin only currently natively supports for ZSH itself.



```sh
# Remember to change the DB paths to your own.
echo "\
  ATTACH DATABASE '/home/tombh/.local/share/atuin/history.db' AS atuin;\
  DELETE FROM history;\
  INSERT INTO history(command_line, start_timestamp, hostname, cwd, duration_ms, exit_status) \
[...]

PR #2170, opened 2024-06-19
Feature atuin import from file

mijoharas:
Hi, I wanted a one-off offline sync for my atuin history.db. I've left it in a rough and ready state to see if there's any interest in it, and to get feedback on if the feedback is wanted.

(would address https://github.com/atuinsh/atuin/issues/816 ).

Reasons I'd personally like this feature:
* People aren't always in a situation where they have access to the internet, and a one-off import to copy history across can be useful.
* The syncing server seems non-trivial to self-host, and there is a lot of code to review if you want to be sure you can trust the external syncing server.

Either way, figured it was easy enough to write some code to start a discussion. Let me know your thoughts.

## Checks

### DIFF
--- a/crates/atuin-client/src/import/atuin_histdb.rs
+++ b/crates/atuin-client/src/import/atuin_histdb.rs
@@ -1,0 +1,73 @@
+// import old shell history from atuin-histdb file!
+// automatically hoover up all that we can find
+
+use std::path::{Path, PathBuf};
+
+use async_trait::async_trait;
+use eyre::{eyre, Result};
+
+use crate::database::{Sqlite, Database, current_context};
+use crate::settings::FilterMode;
+
+use super::Importer;
+use crate::history::History;
+use crate::import::Loader;
+
+pub struct AtuinExternalDb {
+    histdb: Vec<History>,
+}
+impl AtuinExternalDb {
+    pub fn histpath() -> Result<PathBuf> {
+        let path = Path::new("/tmp/history.db").to_path_buf();
+        if path.exists() {
+            Ok(path)
+        } else {
+            Err(eyre!("Could not find history file hardcodded at /tmp/history.db"))
+        }
+    }
+}
+
+/// Read db at given file, return vector of entries.
+async fn hist_from_db(dbpath: PathBuf) -> Result<Vec<History>> {
+    let db_to_migrate = Sqlite::new(dbpath, 0.1).await?;
+    let external_history = db_to_migrate.list(
+        &[FilterMode::Global],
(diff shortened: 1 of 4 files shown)


ellie:
Hey! Thanks for bringing this up

I think it's worth addressing what the full set of requirements are here. Are users OK with duplicate data if they run it twice? Do they expect sync to work after this has been done?

I'd rather not use the database as the transport mechanism here either - really something like `atuin history dump` should dump a format like jsonlines, and then the importer can read and import that

mijoharas:
Hey, schema changes and versioning is definitely one of the reasons I wanted to raise this Pr early for feedback. And thanks for all the good points:

>  Are users OK with duplicate data if they run it twice?
I should have mentioned (but I only tested it after I had raised the PR)., while we don't do anything in the code to remove duplicates, running it twice doesn't actually cause duplicates. I assume it's because we include the id of the record so the duplicate insert fails (I didn't dig into this.)

> Do they expect sync to work after this has been done?
It's a good question, and probably something that I'd approach via documentation (sync is obviously the go-to ongoing sync feature, this new thing could be stressed as an option for if you have no internet, and we could note that it won't setup sync, and that setting up sync is probably what most users would want). But I've obviously got a lot less insight into your users than you do, so let me know your thoughts!

> I'd rather not use the database as the transport mechanism here either - really something like atuin history dump should dump a format like jsonlines, and then the importer can read and import that.
This seems like a reasonable approach to me, and should also be fairly straightforward. The final question is:

> I'd be concerned that changes to the db schema could break imports in the future
This feels like it would be an issue whether we go for jsonlines or the history dump anyways, so something we should consider. Easiest way to solve is to ensure that both atuin's have the exact same schema version or something. I'm not super familiar with the data format, is there a schema number or something? if so, (and if we wanted to proceed with a json-lines version of this) how should it be enforced? would the json-lines file have the first line be something like `{"atuin_history_dump_schema_version": "1"}` ?

Let me know your thoughts, and thanks for the response!

crablab:
Drive-by +1 for this. 

I take backups of my home directory and don't use `atuin sync`. When reinstalling Atuin on a new machine, I simply copy the backed up `~/.local/share/atuin` across. As Ellie mentions, this isn't great if there's a local database migration (I think you'd get around that by downgrading the binary, copying and then upgrading with the migration steps)

markovejnovic:
Hi folks, thank you for this PR! Since the code has diverged so much from this point, and the PR appears to have gone stale, I will be closing it. If you feel like that's unfair, please ping me and I'll re-open the PR and provide some help/feedback on it.

PR #2172, opened 2024-06-20
chore: Add systemd user units for atuin daemon

Flowdalic:
Now that https://github.com/atuinsh/atuin/pull/2171 is merged, we can add sensible systemd user units for atuin daemon.

## Checks

### DIFF
--- a/systemd/atuin-daemon.service
+++ b/systemd/atuin-daemon.service
@@ -1,0 +1,10 @@
+[Unit]
+Description=atuin shell history daemon
+Requires=atuin-daemon.socket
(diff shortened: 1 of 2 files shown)


hvisage:
could you also please consider a central atuin server for multiple users?

I'd like to use it in systems where systemd is explicitly not available.

ellie:
I'm afraid the daemon is designed to be single-user - you can run it without systemd

Finkregh:
I did not check, but would it be an alternative to add these to the docs?

PR #2190, opened 2024-06-24
feat: add user account verification

ellie:
Add `atuin account verify` for verifying the user account email address

This is not required, and so far has no real benefit. It can be used in a follow up for resetting passwords. 

We've used Postmark for the forum for a long while now, and I'm super happy with it. Verified emails is only really beneficial for Atuin Cloud, so I'm not currently considering a generic SMTP server approach. Happy to do so if people want it, but I'd rather not maintain something nobody is using.

## Checks

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -300,8 +300,10 @@ dependencies = [
 name = "atuin-common"
 version = "18.3.0"
 dependencies = [
+ "base64 0.22.1",
  "directories",
  "eyre",
+ "getrandom",
  "lazy_static",
  "pretty_assertions",
  "rand",
@@ -401,6 +403,7 @@ dependencies = [
  "fs-err",
  "metrics",
  "metrics-exporter-prometheus",
+ "postmark",
  "rand",
  "reqwest",
  "rustls",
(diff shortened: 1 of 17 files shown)


tessus:
I suspect that people who self-host the server would also want the option to reset passwords. And the ability for the servwr to send emails is a pre-req for that, isn't it?

But maybe I am wrong.

ellie:
If they're hosting it themselves, it's possible without email infrastructure - which is frustrating to maintain, or costs money. From the last time I asked, most people host instances either just for themselves, or for a very small number of people.

I've had to reset maybe 20 accounts, out of about 10,000. It's unlikely most people self hosting will run into the same issues, and I've put this off for as long as possible. We're just now reaching numbers where it's required.

Again though, like my PR said; should anyone self hosting actually want this, then it's possible. I'm just not at the point where I'm willing to write or maintain features that nobody actually requires or uses.

tessus:
Personally I don't need it. But I have seen that people use applications in an interesting fashion. e.g. look at the vaultwarden project. It was supposed to be a bw compatible solution for "personal" use. However, people use it in companies for hundreds and thousands of users.

I can imagine that the people who use vw in that fashion would also use atuin the same way.

On the other side, who knows, maybe they just open a PR against atuin to add the option to specify an SMTP server.

My comment was rather meant to emphasize how popular atuin has become.

nc7s:
Hi, I'm the main maintainer of the [Debian atuin package](https://tracker.debian.org/pkg/atuin). It's been a nice experience packaging and using it in Debian.

Following the release of 18.4.0, however, we see the postmark crate introduced as dependency. This PR did give a reason, but not a strong one from Debian's perspective as a distro. Postmark is a nice provider, I kinda like it, but we don't quite like introducing a package supporting only a specific service into Debian. Plus, as of writing it's only used by atuin-server (as seen on crates.io), which is an indicator for the likelihood of low maintenance. Maybe this is a point in the direction of general sending aka SMTP, if the demand ever arises. There's [lettre](https://lib.rs/crates/lettre) for that.

I'll have to [patch](https://salsa.debian.org/debian/atuin/-/blob/debian/sid/debian/patches/disable-email-verify.patch) the functionality out for now, due to both the reasons above, and the absence of the postmark crate in Debian which means additional screening period (known as the NEW queue).

ellie:
Nice to hear it's been easy to package!

Totally makes sense. I don't see this being needed or useful for any deployment beyond api.atuin.sh for now, and will happily consider general SMTP if the demand is there.

I'm separately considering moving away from Postmark for what we use, so if that happens we'll likely support SMTP in a future release.

PR #2201, opened 2024-06-25
chore: do not use package managers in installer

ellie:
For a long time I wanted to try and prioritise using package managers, so we would have the smallest impact on the users machine possible. Unfortunately with this change, the install script will no longer be doing this.

Firstly, we're limited by which package managers we are in. The fallback to build from source _really_ sucks on systems like a raspberry pi (running a flavour of debian). For that specific case, it's worth noting that Debian's repositories are very stable - even if we were packaged there, Atuin would rapidly become so out of date that it would no longer sync. Hence the proliferation of ppas.

Increasing the breadth of package managers we are included in becomes tricky in terms of support. Some have much stricter requirements than others, including dynamic linking as much as possible or high levels of patching - both make supporting users and replicating their issues harder, and is not something we can support. We've already seen a couple of issues that will only occur when using the system sqlite, vs the pinned, tested, and bundled version we build.

Secondly (the primary issue), it's diffcult to tell users _precisely_ what the script has done. The "it depends on your system and package manager" has been met with frustrated responses, and I really want to streamline things there. 

The cargo-dist installer provides a consistent install location, consistent binary, consistent update path, and is much easier for us to support.

I am very aware that some users may have issues with ~/.atuin being used.  Really, the automatic installer _is not for them_. The idea with this installer is to streamline installs as much as possible for users that just wish to have things working quickly, and to make it easier for us to support them. There's more discussion about this on the [forum post
](https://forum.atuin.sh/t/the-new-atuin-directory/372?u=ellie)

If you're someone who prefers to have your setup just how you like it, we provide [a whole bunch](https://docs.atuin.sh/guide/installation/#manual-installation) of alternate install methods that will work well for you! You can probably just go ahead and use your package manager (and then setup the shell plugin!)  

## Checks

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,5 +1,6 @@
 #! /usr/bin/env bash
 
+
 if [[ "${BASH_VERSION%%.*}" -eq 3 ]]; then
     echo "Atuin has limited support for Bash 3.2. The Atuin config enter_accept cannot be turned off." >&2
     echo "To turn off enter_accept, please upgrade your version of bash (possibly via homebrew or ports)" >&2
@@ -20,12 +21,17 @@ Magical shell history
 
 Atuin setup
 https://github.com/atuinsh/atuin
+https://forum.atuin.sh
 
-Please file an issue if you encounter any problems!
+Please file an issue or reach out on the forum if you encounter any problems!
 
 ===============================================================================
 
 EOF
+
+__atuin_install_binary(){
+  curl --proto '=https' --tlsv1.2 -LsSf https://github.com/atuinsh/atuin/releases/latest/download/atuin-installer.sh | sh
+}
 
 if ! command -v curl &> /dev/null; then
     echo "curl not installed. Please install curl."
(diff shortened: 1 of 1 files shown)


PauloMigAlmeida:
[For future visitors]

Let me guess, you got here because you did the following:

- You do not like installing things using `curl xxx | sh` so you inspected `install.sh` for due diligence
- You realised that `install.sh` actually runs `curl xxx/atuin-installer.sh | sh` but you couldn't find `atuin-installer.sh` file in the repo
- You also noticed that there are other ways of [manually installing it](https://docs.atuin.sh/cli/guide/installation/#manual-installation) without `curl xxx | sh` but at this point you thought that you really needed to know where `atuin-installer.sh` comes from, right?

Look no further mate!

1. This file is generated as part of the Github release action that runs in this [workflow](https://github.com/atuinsh/atuin/blob/main/.github/workflows/release.yml)
2. The actual generator of that `<x>-installer.sh` file is [cargo-dist](https://github.com/astral-sh/cargo-dist)
3. The template used to generate all the commands that will run during the installation can be found [here](https://github.com/astral-sh/cargo-dist/blob/622170f09a1521bc0782332317076e4805737cef/cargo-dist/templates/installer/installer.sh.j2)

Issue #2214, opened 2024-07-01
Feature Request: Set terminal title when interactive mode is launched via shell hooks


fidgetingbits:
If atuin is triggered in a shell by using the up/ctrl-r key, the terminal title will not be updated to reflect that the atuin TUI is open. This is the case even you use a plugin to set the titles for executed commands (ie: [zsh-term-title](https://github.com/pawel-slowik/zsh-term-title) or similar). This seems to because (at least on zsh) atin tui is being started from within a hook itself, before other hooks end up running. If you manually run `atuin search --interactive` and use a plugin similar to above, the title will be updated to indicate atuin is running.

I use a voice coding system (talon) that relies on analyzing the window title to know what voice grammars to load for a given context, and want to be able to detect when atuin is running so I can load atuin-specific grammars accordingly.

I'll send a PR for addressing this on zsh, but I think it may be worth adding similar functionality on all of the shells that atuin supports (assuming the devs agree the feature is wanted):

- [ ] zsh (PR# )
- [ ] bash (PR #)
- [ ] fish (PR #)
- [ ] nu (PR #)
- [ ] xsh (PR #)

I'm also not entirely sure but I suspect this maybe should be behind a new setting, since some users likely set an explicit title in their shells that they never want changed.

PR #2215, opened 2024-07-01
feat: Set terminal title to indicate atuin tui is active

fidgetingbits:
Attempts to address #2214 for zsh. Currently this just naively changes the title to "atuin" and back to "zsh". The default zsh settings will auto update the title (unless using zsh setting `DISABLE_AUTO_TITLE="true"`), so possibly setting it back to "zsh" afterwards isn't even needed.

afaict it's not possible to query the current title, so we can't switch it back to what was originally there. So I think this does run the risk of clobbering someones title, if they have set `DISABLE_AUTO_TITLE="true"`, in which case it may be best to have an atuin option that's default to off? I figured I'd send this PR to show the idea anyway, since it's working for my use case, and can start a discussion about how best to do it.

### DIFF
--- a/crates/atuin/src/shell/atuin.zsh
+++ b/crates/atuin/src/shell/atuin.zsh
@@ -27,6 +27,12 @@ fi
 export ATUIN_SESSION=$(atuin uuid)
 ATUIN_HISTORY_ID=""
 
+_atuin_update_title() {
+    if [[ "$TERM" == xterm* ]]; then
+        print -n "\e]0;$1\a"
+    fi
+}
+
 _atuin_preexec() {
     local id
     id=$(atuin history start -- "$1")
@@ -51,7 +57,7 @@ _atuin_precmd() {
 _atuin_search() {
     emulate -L zsh
     zle -I
-
+    _atuin_update_title atuin
     # swap stderr and stdout, so that the tui stuff works
     # TODO: not this
     local output
@@ -64,12 +70,12 @@ _atuin_search() {
         RBUFFER=""
         LBUFFER=$output
 
-        if [[ $LBUFFER == __atuin_accept__:* ]]
-        then
+        if [[ $LBUFFER == __atuin_accept__:* ]]; then
             LBUFFER=${LBUFFER#__atuin_accept__:}
             zle accept-line
         fi
     fi
+    _atuin_update_title zsh
 }
 _atuin_search_vicmd() {
     _atuin_search --keymap-mode=vim-normal


crates/atuin/src/shell/atuin.zsh
```diff
@@ -27,6 +27,12 @@ fi
 export ATUIN_SESSION=$(atuin uuid)
 ATUIN_HISTORY_ID=""
 
+_atuin_update_title() {
+    if [[ "$TERM" =~ "xterm*" ]]; then
```
akinomyoga, line 31: Shouldn't this be 

```suggestion
    if [[ $TERM == xterm* ]]; then
```

? The regex match (`=~`) against `xterm*` matches `xter`, `xterm`, `xtermm`, `xtermmmm`, etc. Also, the regex matching is not a matching to the entire word but can match a substring of the supplied string.

crates/atuin/src/shell/atuin.zsh
```diff
@@ -27,6 +27,12 @@ fi
 export ATUIN_SESSION=$(atuin uuid)
 ATUIN_HISTORY_ID=""
 
+_atuin_update_title() {
+    if [[ "$TERM" == "xterm*" ]]; then
```
akinomyoga, line 31: `"xterm*"` doesn't work. If you quote `*` inside the double quotes `"..."`, it only matches the literal character `'*'`. Here, we expect a wildcard `*` of the glob pattern, which should not be quoted.

```suggestion
    if [[ "$TERM" == xterm* ]]; then
```

fidgetingbits: Ah right, and you did recommend that the first time, sorry about that. Need more coffee...

akinomyoga (commented):
I naively think that maybe we can output the OSC sequences from `atuin search`. Then, we don't have to write the settings for each shell.

Also, it would be better if we could configure the title during `atuin search` and that after `atuin search`.

fidgetingbits:
> I naively think that maybe we can output the OSC sequences from `atuin search`. Then, we don't have to write the settings for each shell.
> 
> Also, it would be better if we could configure the title during `atuin search` and that after `atuin search`.

I like that idea and had considered it. The main thing I wasn't so sure about is, is what to set it back to, since each shell has some default (and different) title behavior I think. So just opted to do it per shell to start. Being able to just enable it with a setting would be great since most users likely don't care about it, and would be much easier by adding it to `atuin search`.

So do you envision a setting that toggles the behavior on/off, a setting for what the title is when the TUI is open (default to 'atuin' maybe), and another one for what to reset it to? Probably the title to reset to could default to nothing (so not change it), as many shells (like zsh) would auto-reset it on it's own anyway?

akinomyoga:
First, I think we should make this feature opt-in, i.e. it would set the terminal title only when it is turned on by the configuration. The users can have their own settings for the terminal title. This feature overwrites the users' settings if it is enabled by default. The users who want Atuin to set the terminal title can enable this feature explicitly. Then, the user can set the reset sequence to the desired one.

> I like that idea and had considered it. The main thing I wasn't so sure about is, is what to set it back to, since each shell has some default (and different) title behavior I think.

- I'm not a user of Zsh, but Zsh doesn't seem to change the title by default. This PR changes back the title to `zsh`, but is this really Zsh's behavior? For example, the `zsh-term-title` plugin you mentioned in #2214 seems to set it to a somewhat different string: the command name plus the directory name. Is `zsh` your personal preference or needed by "talon" you mentioned in #2214? Meanwhile, Fish seems to set the terminal title by default to the directory name by default, while Bash doesn't set the terminal title by default. It seems to me that even if the shell is known, it is in general difficult to reset the title to the one that the user expects.
- If the terminal title can be just reset to the shell name (as in this PR), `atuin search` may detect the parent shell and switch the behavior, though I'm not sure if that is useful.
- As another possibility, one can use the push/pop terminal title sequences, [`XTWINOPS(22)`](https://terminalguide.namepad.de/seq/csi_st-22/) and [`XTWINOPS(23)`](https://terminalguide.namepad.de/seq/csi_st-23/), but this is not implemented by many terminals and might cause error messages in the terminal not supporting them.

> Being able to just enable it with a setting would be great since most users likely don't care about it, and would be much easier by adding it to `atuin search`.

Yeah, that would be what I suggest.

> So do you envision a setting that toggles the behavior on/off, a setting for what the title is when the TUI is open (default to 'atuin' maybe), and another one for what to reset it to?

Yeah, or the toggling option can be omitted by allowing the title to be empty; the empty title means that the feature is turned off.

> Probably the title to reset to could default to nothing (so not change it), as many shells (like zsh) would auto-reset it on it's own anyway?

I'm not sure, but I think the terminal title is usually configured to be updated after a user command is executed. I'm not sure if we can expect that the shell settings will reset the title after each keybinding in general.

fidgetingbits:
> First, I think we should make this feature opt-in, i.e. it would set the terminal title only when it is turned on by the configuration. The users can have their own settings for the terminal title. This feature overwrites the users' settings if it is enabled by default. The users who want Atuin to set the terminal title can enable this feature explicitly. Then, the user can set the reset sequence to the desired one.

Okay, I agree with this. I suspect the majority of users won't use it as it's mostly an accessibility thing, so opt-in totally makes sense.

>     * I'm not a user of Zsh, but Zsh doesn't seem to change the title by default. This PR changes back the title to `zsh`, but is this really Zsh's behavior? For example, the `zsh-term-title` plugin you mentioned in [Feature Request: Set terminal title when interactive mode is launched via shell hooks #2214](https://github.com/atuinsh/atuin/issues/2214) seems to set it to a somewhat different string: the command name plus the directory name. Is `zsh` your personal preference or needed by "talon" you mentioned in [Feature Request: Set terminal title when interactive mode is launched via shell hooks #2214](https://github.com/atuinsh/atuin/issues/2214)? 

I do match on "zsh" using talon, but can match on anything unique as long as something is set. I just double checked a vanilla terminal without my voice coding setup loaded and you're right, it's not set to zsh by default, it's set to the cwd.

> Meanwhile, Fish seems to set the terminal title by default to the directory name by default, while Bash doesn't set the terminal title by default. It seems to me that even if the shell is known, it is in general difficult to reset the title to the one that the user expects.

Agreed.

>     * As another possibility, one can use the push/pop terminal title sequences, [`XTWINOPS(22)`](https://terminalguide.namepad.de/seq/csi_st-22/) and [`XTWINOPS(23)`](https://terminalguide.namepad.de/seq/csi_st-23/), but this is not implemented by many terminals and might cause error messages in the terminal not supporting them.

Cool! Never heard of that. But ya, I'd probably just opt to not bother so special casing various terminals isn't required.

> > Being able to just enable it with a setting would be great since most users likely don't care about it, and would be much easier by adding it to `atuin search`.
> 
> Yeah, that would be what I suggest.
> 
> > So do you envision a setting that toggles the behavior on/off, a setting for what the title is when the TUI is open (default to 'atuin' maybe), and another one for what to reset it to?
> 
> Yeah, or the toggling option can be omitted by allowing the title to be empty; the empty title means that the feature is turned off.

Ya that's simpler.

> > Probably the title to reset to could default to nothing (so not change it), as many shells (like zsh) would auto-reset it on it's own anyway?
> 
> I'm not sure, but I think the terminal title is usually configured to be updated after a user command is executed. I'm not sure if we can expect that the shell settings will reset the title after each keybinding in general.

Okay, thanks for all the feedback. I'll work on adding it to `atuin search` and update the PR/issue accordingly.

PR #2234, opened 2024-07-05
feat(bash/blesh): hook into BLE_ONLOAD to resolve loading order issue

akinomyoga:
Currently, to set up Atuin with ble.sh in Bash, we require `ble.sh` to be loaded before `eval "$(atuin init bash)"`. This PR removes this requirement by using the shell variable `BLE_ONLOAD`, which is a hook called by `ble.sh` in its loading phase.

The hook has existed for a long time, but I was forgetting it. I remembered it and used it to run Atuin's integration on ble.sh's loading. I also update the corresponding message in `atuin doctor`.

## Checks


ellie (approved):
nice! thanks for the improvement

akinomyoga:
Thanks!

Issue #320, opened 2022-04-21
GNU screen requires altscreen enabled


bradrf:
[As noted in the tui-rs issues](https://github.com/fdehau/tui-rs/issues/344), when running atuin under GNU screen, repainting will fail and overlay the atuin windowing over whatever text is already on the terminal's output (see video attached below).

There's a simple workaround to enable screen's [`altscreen`](https://www.gnu.org/software/screen/manual/screen.html#Redisplay) (just add it to your `.screenrc` file). If others run into this, atuin might consider producing some kind of warning (if you run into this, please :+1: this post to let atuin folks know it's a broader problem than just one person).

https://user-images.githubusercontent.com/295715/164484861-88ea359c-ad98-4943-920f-3d95311e9468.mp4

jbellis:
+1 for improving this, at a minimum can we detect screen/tmux/weird terminals and display a warning?

jbellis:
Additionally: setting inline_height=5 fixes the problem for me under screen.  (This also forces style=compact, but style=compact alone does not fix the problem.)

welbow:
I ran into this problem for a different reason - I purposefully disable the alternate screen in my terminal emulator so everything is in the scrollback buffer. I found a way to wrap a function around atuin's bindings in my bash shell to print enough blank lines to give the same effect as switching to the alternate screen.

I am not expecting anything like this to be included in atuin, but I am posting in case it helps someone else.

```
_original_atuin_history=$(declare -f __atuin_history)

__atuin_history() {
  local lines
  lines=$(tput lines)
  printf '\n%.0s' $(seq 1 "$lines")
  eval "${_original_atuin_history/__atuin_history/__atuin_history_orig}"
  __atuin_history_orig "$@"
  printf '\n'
}
```

ellie:
you can also use `inline_height = <a number larger than 0>` in your config to avoid using alt screen

welbow:
> you can also use `inline_height = <a number larger than 0>` in your config to avoid using alt screen

I tried that several times, logged out and back in to ensure the profile is reloaded and didn't notice any change there. I even set style = compact.

ellie:
hmm I see, bit of a weird one because we definitely do not use the altscreen in that mode. perhaps there's something else going on with your terminal in that config

I'm going to close this one because there's nothing we are going to action here! Feel free to open something about your issue though

PR #2249, opened 2024-07-10
chore: enable inline height and compact by default

ellie:
I've been running Atuin this way for a really long time, and have heard from quite a few users now that they weren't a fan of Atuin using the whole screen.

This is how it looks:

inline-height

However I have also learned that you only hear from people who _don't_ like the way things currently are, right up until you change them - and then hear from everyone that was happy and quiet.

I'll try and circulate this PR as much as I can. It's the same as adding

```
style = "compact"
inline_height = 40
```

to the config file.

If this change is merged, it will change how Atuin looks for people who don't have anything set in their config. If they don't like it, they can restore how things were with

```
style = "auto"
inline_height = 0
```

We could change the default config file so this only happens for new users, but I'd like to be able to change defaults at some point and would like to see the general opinion first.

I can't think of a better way of getting feedback, short of opt-in telemetry. That's a different discussion :)

**If you like this, please react with 👍**

**If you don't, please react with 👎** 

## Checks

### DIFF
--- a/crates/atuin-client/src/settings.rs
+++ b/crates/atuin-client/src/settings.rs
@@ -686,8 +686,8 @@ impl Settings {
             .set_default("sync_frequency", "10m")?
             .set_default("search_mode", "fuzzy")?
             .set_default("filter_mode", "global")?
-            .set_default("style", "auto")?
-            .set_default("inline_height", 0)?
+            .set_default("style", "compact")?
+            .set_default("inline_height", 40)?
             .set_default("show_preview", true)?
             .set_default("preview.strategy", "auto")?
             .set_default("max_preview_height", 4)?


djc:
I've been running with compact/40 anyway, so this seems great!

tessus:
I am not a fan of most atuin's default settings. When it comes to these two, I am all for setting `inline_height`, although I'd go with a lower number. If you set it to 40 it might use all of the terminal, so not a big change between that and not setting it. (e.g. I use a Retina 2880 × 1800, my terminal is 55 lines, and I set my inline_height to 25.) 
I don't like the compact mode, so that's a no for me.

Personally I don't mind either way, because I almost never use default values (in any SW) for anything that is important.

ellie:
Size isn't actually the only reason for this change. Some of the other issues I've heard people have with full screen:

1. Regardless of size, the transition flash between inline/full screen is jarring for some people
2. The lack of context makes it feel a bit disconnected - even with 40 lines, there's still usually some context visible on most terminals
3. This doesn't depend on altscreen, so will work with screen/etc without changing config (see #320)

With a lower value than 40, the inspector won't work very well. It might be something users want to change, but can't work as a default. The inspector does need some work to function better on smaller terminals, but it is a balance between "I want it to work on smaller terminals" and "I want to see/explore more information". Maybe the precise number will change in the future.

> I think that having the full-screen mode by default is a smart thing to do, as it gives the new users some guidance about where to focus their attention

This was the initial reasoning, however I hear very often that the full screen is either something people change immediately, or that it's a reason they stop using Atuin entirely. I've made enough mistakes trying to guess what people might want that I'd rather try to use numbers.

> Personally I don't mind either way, because I almost never use default values (in any SW) for anything that is important.

I totally expect that users will be changing these values, but I have absolutely no insight as to whether or not they do so. I'm trying to optimize for new users, and minimize the amount of config that they need to change to feel comfortable. This allows me to gauge if most people deviate from defaults, or if most people are happy with things how they are.

plttn:
At least for me, inline when not bounded by terminal height overlaps with the last line of my shell (Fish+Tide):

```
~ ❯                                                                                                            docker-desktop
```

becomes
```
~Atuin v18.3.0n ❯                   <esc>: exit, <tab>: edit, <enter>: edit, <ctrl-o>: inspect               history count: 12219p
```
If this didn't occur, I'd lean this change, but I prefer fullscreen at the moment. I'm sure this is a separate bug that should be filed 😅

lilydjwg:
I'm using compact mode with all lines below or 40% of lines, whichever is larger.

(I tried to invert the UI when cursor was near the bottom briefly, but it didn't feel good.)

dcarosone:
I'm not particularly concerned with the default, as for personal use I will configure options anyway.

But I'm curious about the altscreen issue..  I use screen, with altscreen enabled, as a workaround for mosh not supporting altscreen.  What's the behaviour in compact mode, with respect to screen contents? Do they just get overwritten, but it's supposed to be ok because it's not the whole screen?  My preference is to preserve the screen in full.

tessus:
The screen is preserved. When not using fullscreen atuin (altscreen), the text is not overwritten, but the TUI is created below the prompt. See my [post](https://forum.atuin.sh/t/tui-opens-below-cursor-thus-shifts-the-prompt-up/405) on the forum. I've also provided a video what's happening.

philtweir:
Possibly not a helpful datapoint, but as someone who types Ctrl+r by default for commands over 2 chars, my use-case is "minimalist like traditional reverse search" -- I am not sure how common that is, but I nearly did a 5m try-and-move-on with Atuin as I had not spotted `compact` as an option immediately. As it happened, I was sufficiently fed up with traditional reverse search that I went back and searched through the docs/PRs again and then realised it was a setting (essentially, I hadn't realised what to search for first time). So strong propose for this suggestion, personally, but no idea if I'm a common case.

> Regardless of size, the transition flash between inline/full screen is jarring for some people

For my type of use-case, absolutely, especially if hitting Ctrl+r at least once for nearly every command entered.

> The lack of context makes it feel a bit disconnected - even with 40 lines, there's still usually some context visible on most terminals

Again, I use it as a helper to writing a command, rather than a full-featured search tool, so context is key for my type of use.

That said, I normally use Atuin in "ultracompact" 3-8 lines though, and have a customization that autohides non-critical UI lines when inline_height is under 10, so arguably my use-case should really be out-of-scope :D 

Screenshot from 2024-07-15 13-53-13

(OT: @ellie with the theming landing, I have rebased this & the inspector work onto today's main, so it could be timely for me to PR, if this "ultracompact" use could be relevant for others and the compact UI is about to be highlighted more? No pressure, only if generally desirable obvs!)

ellie:
> (I tried to invert the UI when cursor was near the bottom briefly, but it didn't feel good.)

Same, I could never get the inverted UI to feel quite right, no matter what I tried. Some people seem to like it though 🤷‍♀️ I think it's because, depending on scroll position, the cursor either jumps up in inverted mode, or jumps down in not-inverted mode.

> What's the behaviour in compact mode, with respect to screen contents?

For extra clarity, empty space is created below the prompt and the terminal essentially just scrolls a bit to make room for the TUI. 

> Again, I use it as a helper to writing a command

That is a pretty good datapoint! I've heard it from a few people now, and wonder how common it actually is.

Otherwise, I'd definitely be interested in a PR with your ultra-compact mode please! 🙏

PR #2284, opened 2024-07-15
Fix logic for store_failed=false

### LINKED ISSUE #2227: [Bug]: History stopped recording

Show history correctly when I press `UP-Arrow` or `atuin history list`

### What happened?

History stopped at some point in time and won't update. I reinstalled the atuin binary and also removed the `~/.config/atuin` and `~/.local/share/atuin` directories. Same issues afterwards. I've run `atuin login` and `atuin sync` and they have worked correctly.

### Atuin doctor output

```yaml
{
  "atuin": {
    "version": "18.3.0",
    "sync": {
      "cloud": true,
      "records": true,
      "auto_sync": true,
      "last_sync": "2024-07-02 10:49:25.55878 +00:00:00"
    },
    "sqlite_version": "3.44.0"
  },
  "shell": {
    "name": "zsh",
    "default": "zsh",
    "plugins": [
      "atuin"
    ],
    "preexec": "built-in"
  },
  "system": {
    "os": "Darwin",
    "arch": "arm64",
    "version": "14.1.1",
    "disks": [
      {
        "name": "Macintosh HD",
        "filesystem": "apfs"
      },
      {
        "name": "Macintosh HD",
        "filesystem": "apfs"
      },
      {
        "name": "projects",
        "filesystem": "apfs"
      }
    ]
  }
}
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

JRGould:
fixes  #2227

Just updated the conditional here to check the `exit` param instead of `h.exit` since it seems like the latter will always be `-1` here as that's the default value. Also changed the operator to `>` instead of `!=` just to reflect the fact that exit codes are expected to only be positive integers 0-255

I also added `.envrc` to the `.gitignore` file as using `direnv` for development was recommended in the CONTRIBUTING doc 

## Checks

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 .vscode/
 result
 publish.sh
+.envrc
 
 ui/backend/target
 ui/backend/gen
--- a/crates/atuin/src/command/client/history.rs
+++ b/crates/atuin/src/command/client/history.rs
@@ -384,7 +384,7 @@ impl Cmd {
             return Ok(());
         }
 
-        if !settings.store_failed && h.exit != 0 {
+        if !settings.store_failed && exit > 0 {
             debug!("history has non-zero exit code, and store_failed is false");
 
             // the history has already been inserted half complete. remove it


ellie (approved):
Nice fix, thanks! I realise I only ever tested this with the daemon enabled, which does not store intermediate history with a -1 exit code 🤦‍♀️ 

We do also have a [Discord](https://discord.gg/Fq8bJSKPHh) and a [forum](https://forum.atuin.sh) if you'd like to ask any questions, or just fancy hanging out!

PR #2286, opened 2024-07-16
Fixed preview height

tobiasge:
Uses max_preview_height as a fixed value to prevent the list from "jumping".

## Checks

### DIFF
--- a/crates/atuin-client/config.toml
+++ b/crates/atuin-client/config.toml
@@ -209,6 +209,7 @@ records = true
 ## possible values: auto, static
 ## auto: length of the selected command.
 ## static: length of the longest command stored in the history.
+## fixed: use max_preview_height as fixed height.
 # strategy = "auto"
 
 [daemon]
--- a/crates/atuin-client/src/settings.rs
+++ b/crates/atuin-client/src/settings.rs
@@ -410,6 +410,10 @@ pub enum PreviewStrategy {
     // Preview height is calculated for the length of the longest command stored in the history.
     #[serde(rename = "static")]
     Static,
+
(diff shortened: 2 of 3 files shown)


tessus:
The current "static" is unfortunately not so static. When I implemented the `dynamic` part, I had the false impression that the height of the original preview window sizing was determined by the longest command in the entire history. This was/is not the case, but it is determined by the longest command of the history that is currently shown (basically the commands that fit into the height of the TUI minus a few rows).
The "jumping" would also be alleviated with my other PR #2064 (see the example videos), if you set the preview window position to the opposite of the input line. Then there is no jumping. This is also a reason why I use `invert=true` and the preview is at the bottom. No jumping.

Back then I didn't want to change the original behavior, because I had the false impression I mentioned earlier.

Maybe it is even ok to replace the current "static" strategy with your code, but this depends on @ellie of course.

P.S.: It might not be a bad idea to rename `static` to something more appropriate, e.g. `result`, `resultset`, `window`, `visible`, `shown`, ...

ellie:
> This was/is not the case, but it is determined by the longest command of the history that is currently shown (basically the commands that fit into the height of the TUI minus a few rows).

It's actually just based on the results of the current query. Not the longest in the entire database, and not the longest that's currently on the screen. This is shown here:

https://github.com/atuinsh/atuin/blob/39824db32a1e6a1b794f3d0b8a1cce385dece19b/crates/atuin/src/command/client/search/interactive.rs#L571-L584

ellie (approved):
Thank you! Pretty neat solution to the problem 🙏

PR #2294, opened 2024-07-20
Restore default theme

philtweir:
Following the [discussion](https://forum.atuin.sh/t/release-v18-4-0-beta-2/417/9) about the 18-4-0.beta-2 release, I made the suggested styling change. This PR now:

* makes default named "default" not ("")
* uses `ContentStyle` internally, rather than `Color`, allowing for unset colours
* tidies up `theme.rs` for internal consistency
* avoids explicit use of colours _except_ in the stats, where it is colouring the indicator bars, and will fallback to the original set if the colours are unset
* allows any Crossterm-deserializable colour to be given, via an `@` prefix: `@ansi_(255)` or `@dark_grey`, providing a means to reference ANSI codes [**needs documented**]

As a result, changes will be needed in the docs.

## Checks

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1001,6 +1001,7 @@ dependencies = [
  "libc",
  "mio",
  "parking_lot",
+ "serde",
  "signal-hook",
  "signal-hook-mio",
  "winapi",
--- a/crates/atuin-client/Cargo.toml
+++ b/crates/atuin-client/Cargo.toml
@@ -71,7 +71,7 @@ indicatif = "0.17.7"
 tiny-bip39 = "1"
 
 # theme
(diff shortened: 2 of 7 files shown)


philtweir:
@ellie I decided to go ahead with the refactor, and just redo it - I am happy to split out some of the changes here to make them separately mergeable (or not, if preferred) but keen to get your feedbacck.

ellie (approved):
Couple of minor comments, but this is great - thanks for the changes! I've tested and ensured that the colours are the correct defaults on a few terminals

philtweir:
Thanks @ellie - sorry for the delay! Will include those comment updates in the ultracompact PR!

PR #2339, opened 2024-08-03
fix(theme): Restore original colours

philtweir:
## What does this PR do?

Restores the colours from 18.3.0, which had changed due to a naming mismatch between ratatui and crossterm. This can be seen in the [ratatui compatibility code](https://github.com/ratatui-org/ratatui/blob/b344f95b7cfc3b90221c84cfef8afcf6944bafc1/src/backend/crossterm.rs#L275).

Screenshot from 2024-08-03 05-14-34

To make sure, here are three versions - the most recent 18.4.1 beta (bright), the current fix and 18.3.0 running bash in default xterm config without RCs.
 
## Checks

### DIFF
--- a/crates/atuin-client/src/theme.rs
+++ b/crates/atuin-client/src/theme.rs
@@ -258,7 +258,10 @@ lazy_static! {
             "default".to_string(),
             None,
             HashMap::from([
-                (Meaning::AlertError, StyleFactory::from_fg_color(Color::DarkRed)),
+                (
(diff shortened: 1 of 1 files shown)


ellie (approved):
Lovely lovely, thanks for following through on this one, really appreciate your work 🙏

philtweir:
No worries - thanks for the review, and sorry for not catching that earlier!

PR #2344, opened 2024-08-05
chore: update to rust 1.80

ellie:
## Checks

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,7 +7,7 @@ exclude = ["ui/backend"]
 [workspace.package]
 version = "18.4.0-beta.3"
 authors = ["Ellie Huxtable <ellie@atuin.sh>"]
-rust-version = "1.79"
+rust-version = "1.80"
 license = "MIT"
 homepage = "https://atuin.sh"
 repository = "https://github.com/atuinsh/atuin"
--- a/crates/atuin/src/command/client/search/interactive.rs
+++ b/crates/atuin/src/command/client/search/interactive.rs
@@ -764,7 +764,7 @@ impl State {
         );
     }
 
-    fn build_title(&mut self, theme: &Theme) -> Paragraph {
+    fn build_title(&self, theme: &Theme) -> Paragraph {
         let title = if self.update_needed.is_some() {
             let error_style: Style = theme.get_error().into();
             Paragraph::new(Text::from(Span::styled(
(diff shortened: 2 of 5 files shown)


ellie:
I ran `nix flake update`, but it looks like things are more broken than an outdated lockfile 😞 

Do either of you know what's up? cc @arcuru @Sciencentistguy 

Here's the error:

```
<super super long stack trace that's impossible to read>

… while evaluating the argument passed to builtins.fromTOML

       … while calling the 'readFile' builtin
         at /nix/store/f5ykynf933aab7nd1arqz6dm79hdh91q-source/lib/trivial.nix:722:24:
          721|   importTOML = path:
          722|     builtins.fromTOML (builtins.readFile path);
             |                        ^
          723|

       … while realising the context of path '/nix/store/99hlq6kn4ww9y617wc093pnng24kbidc-channel-rust-1.80.toml'

       error: hash mismatch in fixed-output derivation '/nix/store/iv1pn1rf4c487gz0vv2yxwgmila27dbm-channel-rust-1.80.toml.drv':
         specified: sha256-Ngiz76YP4HTY75GGdH2P+APE/DEIx2R/Dn+BwwOyzZU=
            got:    sha256-6eN/GKzjVSjEhGO9FhWObkRFaE1Jf+uqMSdQnb8lcB4=
```

arcuru:
It's just a hash mismatch, it looks like he never explained how to update.

The method that you're using for the Rust toolchain requires pinning the hash of the toolchain file in flake.nix. So you need to manually update that as well every time you update the Rust Toolchain.

```
error: hash mismatch in fixed-output derivation '/nix/store/iv1pn1rf4c487gz0vv2yxwgmila27dbm-channel-rust-1.80.toml.drv':
         specified: sha256-Ngiz76YP4HTY75GGdH2P+APE/DEIx2R/Dn+BwwOyzZU=
            got:    sha256-6eN/GKzjVSjEhGO9FhWObkRFaE1Jf+uqMSdQnb8lcB4=
```

Just change the line in the flake.nix file from "sha256-Ngiz..." to the new "sha256-6eN..." and it should work fine. I tested locally.

ellie:
Thanks you!

That's weird 🤔  It doesn't update when I run `nix flake update`?

Sciencentistguy:
Yeah the fenix library requires you to specify the hash of your toolchain file. In theory this isn't needed, as it is a local file and versioned by git, but the library does not have the ability to detect this. (the hash is needed for non-local files to ensure reproducability)

In general with Nix issues if there's a hash mismatch the solution is to update the mismatched hash with the new one, fwiw.

ellie:
Thank you both! I managed this with the last update too 🤦‍♀️ 

sorry for the needless ping!

PR #2350, opened 2024-08-06
Fix zsh_histdb import issue

tru2dagame:
## Fix Import Error by Modifying Order Clause

This PR addresses an import zsh-hist-db error by changing the order clause in the query. 
The current implementation uses `ORDER BY history.start_time`, which is causing issues. The fix involves using `ORDER BY history.id` instead. The order by `id` and `start_time` produces the same sequence, and since `id` is the primary key, this change may improve the query performance.

###### Error Screenshot:
image

###### Histdb Query Testing:
image

image

## Checks


markovejnovic:
Hey @tru2dagame! Thank you for your PR! Do you know why those issues were arising? I grepped for `database disk is malformed` and couldn't find it in atuin's source. Would love to know more before merging the fix

markovejnovic:
The PR appears to have gone stale, so I'll be closing it. Please ping me if you want me to reopen it.

PR #2355, opened 2024-08-08
Remove debug statement when changing account password

TymanWasTaken:
This was introduced in my PR #1615 without being noticed somehow, I just happened to look at the diff of that PR and saw the macro I must have accidentally left in it. I just checked and it does indeed print the request when you attempt to change your password in the CLI, whoops!

## Checks

### DIFF
--- a/crates/atuin-client/src/api_client.rs
+++ b/crates/atuin-client/src/api_client.rs
@@ -392,8 +392,6 @@ impl<'a> Client<'a> {
             .send()
             .await?;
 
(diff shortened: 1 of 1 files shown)


ellie (approved):
Thanks!

Not really a big deal leaving this in though tbh

1. It's the response, not the request
2. Printing the details when _debug_ logging is on is fairly reasonable 

I don't feel strongly either way though so happy to merge this

PR #2369, opened 2024-08-22
Update Dockerfile to use Bookworm for both runtime and build

cultpony:
The runtime needs to use the same debian release as the build environment, otherwise this just crashes on startup

## Checks


ellie (approved):
Thank you! 🙏

PR #2370, opened 2024-08-24
Fix reverse import order for zsh and nu

### LINKED ISSUE #178: `atuin import zsh` imports history in reverse order?
I've just installed `atuin` in my Ubuntu 20.04, ZSH 5.8 pc using the deb package in Releases.

Running `atuin import zsh` (or `auto`) successfully imports my command history, but the imported commands then show up in reverse order when searching it through `atuin`. That makes no sense... Commands saved _after_ the import and correctly appended to the history file appear in the right order.

Is it a timestamping issue? My history file had no timestamp or extra info, just the plain commands. I noted that `atuin` adds timestamps to ZSH history file.

For now, I've just reverted the history file with `tac` before importing and got it done. 👍🏼

ajesipow:
Should fix https://github.com/atuinsh/atuin/issues/178

Contains the following changes:

1. Added `DoubleEndedIterator` implementation for `UnixByteLines`.

2. Reversed import order for `zsh`
I had to make some additional changes to handle commands spanning multiple lines: the previous command is prepended to the current one if it contains a line break (i.e. we need to keep the original import order for commands on multiple lines).

3. Reversed import order for `nu`
Simply reversed the order for `nu`. Commands spanning multiple lines do not need to be handled differently since they are stored in a single line in the history.

The import order for `bash`, `fish` and `replxx` seem to be correct already.
I also did not touch `resh` as I could not test locally since it tries importing `".resh_history.json"` whereas `resh` [seems to be storing its history](https://github.com/curusarn/resh/blob/master/troubleshooting.md#recorded-history) in `history.reshjson`. So that appears to be a separate issue unless I missed something.

## Checks

### DIFF
--- a/crates/atuin-client/src/import/mod.rs
+++ b/crates/atuin-client/src/import/mod.rs
@@ -38,7 +38,7 @@ fn unix_byte_lines(input: &[u8]) -> impl DoubleEndedIterator<Item = &[u8]> {
         bytes: input,
         i: 0,
         // Set to the last element
-        i_rev: input.len().checked_sub(1).unwrap_or(0),
+        i_rev: input.len().saturating_sub(1),
     }
 }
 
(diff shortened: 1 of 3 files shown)


taylordotfish:
This issue has now been fixed by #3597 and #3598.

PR #2371, opened 2024-08-26
Allow init --skip-env and --only-env

calebhearth:
In (some?) shells, it's sometimes preferable to set up env vars in a
different place than other configuration like most of what `atuin init`
sets up. For example, ZSH reads ~/.zshenv for all shells while ~/.zshrc
is only read for interactive shells.

This allows splitting the output of init such that in these two
different config files we could load environment variables and load all
other config seperately:

~/.zshenv:

    eval "$(atuin init zsh --env-only)"

~/.zshrc:

    eval "$(atuin init zsh --skip-env --disable-up-arrow)"

Which means that when launching noninteractive shells (I do this
frequently for running tests from Vim, for example, but it's more
generally useful) the specified env values will be sourced and we won't
redefine them in login shells.

## Checks
- [x] I am happy for maintainers to push small adjustments to this PR,
      to speed up the review cycle

Note: I know I didn't create an issue or post anywhere else first. If
this is totally not a desired feature my feelings won't be hurt if you
say so and close this.


calebhearth:
I took a look at the clippy errors. I was able to fix a docs suggestion but extracting an enum and sharing across files proved more complicated. I've not done much rust before and while I had no problem (and my initial instinct had been to) extracting a 3 value enum, actually sharing that across files exceeded my skill level.

calebhearth:
@Ellie I know you're busy with Desktop, but wondering if you have a bit to look at this. I'm happy to try and fix clippy if you want to point me in the right direction, or if you know how to do the fix easily feel free.

cosgroveb:
<3 this is a worthwhile enhancement

cosgroveb:
CONTIRUBUTING.md doesn't say it but maybe @ellie would be more responsive with a corresponding issue?

CHANGELOG.md
```diff
@@ -84,6 +84,7 @@ All notable changes to this project will be documented in this file.
 - Right Arrow to modify selected command ([#2453](https://github.com/atuinsh/atuin/issues/2453))
 - Provide additional clarity around key management ([#2467](https://github.com/atuinsh/atuin/issues/2467))
 - Add `atuin wrapped` ([#2493](https://github.com/atuinsh/atuin/issues/2493))
+- *(init)* Allow splitting environment variable loading from remainder of setup
```
ellie, line 87: This is autogenerated! No need to revert changes, but they will be overwritten

ellie (changes requested):
This looks great! Thank you. Sorry it took me so long to get to it. Could you just do a `cargo fmt` please?

calebhearth:
Sure can, I'll take a look tonight or tomorrow.

calebhearth:
@Ellie I realized that the previous implementation didn't actually solve the problem. This one does, though it's a bit more verbose.

I think an ideal solution looks more like the original flags (note that they've changed from --skip-env and --only-env here), but I was having trouble figuring out how to avoid the clippy complaint about the struct having too many bools. Maybe someone with more clap-fu would be able to convert arbitrary flags directly into values in an enum!

Here's the `atuin init zsh --help` output:

```
Print Atuin's shell init script

Usage: atuin init [OPTIONS] <SHELL>

Arguments:
  <SHELL>
          Possible values:
          - zsh:   Zsh setup
          - bash:  Bash setup
          - fish:  Fish setup
          - nu:    Nu setup
          - xonsh: Xonsh setup

Options:
      --disable-ctrl-r
          Disable the binding of CTRL-R to atuin

      --disable-up-arrow
          Disable the binding of the Up Arrow key to atuin

      --emit <EMIT>
          Limit initialization to a subset of atuin's init, useful for splitting when in a shell's lifecycle various configuration is performed
          
          [default: all]

          Possible values:
          - all:          Emit atuin shell init and managed dotfiles (environment variables and aliases)
          - necessary:    Only emit atuin shell init script without any Dotfiles-related setup (no env vars, no aliases)
          - env-only:     Only emit environment variables (useful for `~/.zsh_env` or `~/.bash_env`)
          - aliases-only: Only emit aliases

  -h, --help
          Print help (see a summary with '-h')
```

calebhearth:
@ellie I've just rebased this, in case you're still interested in the change.

markovejnovic:
Thank you for the PR, but dotfiles have been removed in #4193 and we will no longer be supporting dotfiles.

PR #2382, opened 2024-09-08
Update rustls and axum-server

tobiasge:
Updated rustls and axum-server to the current versions. Sync was tested with my self-hosted server.
## Checks

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -437,7 +437,7 @@ dependencies = [
  "postmark",
  "rand",
  "reqwest 0.11.27",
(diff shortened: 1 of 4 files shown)


tessus:
Unfortunately the pipeline for illumos fails.

I have run into the same issue with #2062 but I don't know how to fix it. I don't even know why there is a pipeline for illumos (maybe it is a stack that is used extensively, but I just don't know about it). Before this project I have never heard of illumos.

ellie (approved):
nice one - thank you!

Issue #1555, opened 2024-01-13
multiline commands in Zsh history get corrupted after import


oconnor663:
I had this command in my Zsh history:

```
for i in *.xml ; do ;
    echo "=== $i ==="
    if ! ./parse.py "$i" ; then
        break
    fi
done
```

I then ran `atuin import zsh` and put `eval "$(atuin init zsh)"` in my `.zshrc`. Now when I ctrl-R and pull up that command, it looks like this:

```
for i in *.xml ; do ;\
    echo "=== $i ==="\
    if ! ./parse.py "$i" ; then\
        break\
    fi\
done
zsh: parse error near `then'
```

It looks like some backslashes are getting inserted into the command that shouldn't be there. Some sort of escaping bug?

conradludgate:
(first, thanks for your work on BLAKE3)

It's possible that zsh inserted the backslashes into the history to make it clear that they are a single multiline command and not many single line commands. I need to double check the code/zsh histfile.

In general, bash/zsh/fish histfiles are full of ambiguous/weird cases. Why couldn't they just use jsonlines...

lilydjwg:
I don't have this issue because zsh rewrites my multiline commands into one line in the history but I do remember this kind of issue happened to me somewhere (not atuin) in the past.

> Why couldn't they just use jsonlines...

Because JSON was born too late :-)

danielfleischer:
Solution in #98 helped, but this needs to be addressed in the zsh import code.

ellie:
we actually came across this before + tried to fix with some issues, need to give it a read through to remember the context though

see #100 #98

PR #2390, opened 2024-09-17
zsh import multi-line fix

danielfleischer:
When importing zsh history, sometimes multi-line commands end with `\\` instead of `\`. These are not identified and ingested as-is and inserted at use time with `\\`. PR is about cleaning these cases. Test added.

Issue is mentioned in #98, #1555.

## Checks


markovejnovic:
Turns out that https://github.com/atuinsh/atuin/pull/2799 already fixed this. Thank you for your PR regardless!

PR #2394, opened 2024-09-21
fix(daemon): Add context to error when unable to connect

jeremycline:
Recently, it seems, the socket location for the daemon moved and this caused me to scratch my head briefly since I saw errors from the client connecting to the daemon, but the daemon was clearly running and the socket seemed to exist.

This patch includes more context when the client fails to connect to the daemon. The path is included to help the user understand where the client was looking, and `wrap_err_with()` is used to show the user the cause of the error. This changes the error message from:
```
Error: failed to connect to local atuin daemon. Is it running?
```
to:
```
Error: failed to connect to local atuin daemon at /run/user/1001/atuin.sock. Is it running?

Caused by:
   0: transport error
   1: No such file or directory (os error 2)
   2: No such file or directory (os error 2)
```

## Checks

### DIFF
--- a/crates/atuin-daemon/src/client.rs
+++ b/crates/atuin-daemon/src/client.rs
@@ -1,4 +1,4 @@
-use eyre::{eyre, Result};
+use eyre::{Context, Result};
 #[cfg(windows)]
 use tokio::net::TcpStream;
 use tonic::transport::{Channel, Endpoint, Uri};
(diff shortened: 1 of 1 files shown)


ellie (approved):
> Recently, it seems, the socket location for the daemon moved

Huh that's weird. It should not have moved 🤔 

But yes otherwise this totally makes sense, thank you

jeremycline:
> > Recently, it seems, the socket location for the daemon moved
> 
> 
> 
> Huh that's weird. It should not have moved 🤔 
> 

I admit I didn't spend enough time looking at exactly why I hit this error; it's likely my systemd unit does something silly. Anyway, thanks!

PR #2408, opened 2024-10-04
Add a parameter to the sync to specify the download/upload page

m42e:
This adds a parameter to change the upload/download page size, this could help if one gets a '413' in a selfhosted scenario.

## Checks


crates/atuin/src/command/client/store/pull.rs
```diff
@@ -19,6 +19,11 @@ pub struct Pull {
     /// This will first wipe the local store, and then download all records from the remote
     #[arg(long, default_value = "false")]
     pub force: bool,
+
+    /// Page Size
+    /// How many records to upload at once. Defaults to 100
```
ellie, line 24:
```suggestion
    /// How many records to download at once. Defaults to 100
```

ellie:
I'd be pretty concerned if users would want a page_size to _reduce_ the value from 100 tbh, it should be plenty small enough that a 413 isn't a problem. Wanting to increase it for a potentially faster sync might make sense, though

Is this an issue you've been hitting?

ellie:
I'm ok to merge this, but please be careful storing huge records in the db - it's really not meant for this. We're discussing adding support for much larger records in the future

PR #2451, opened 2024-11-19
chore: docker healtcheck

mortee:
## Checks


mateuscomh:
@mortee suggestion: great idea implement heathcheck. But to optimize the image size and reduce unnecessary dependencies, consider replacing curl and jq with netcat (nc) in the health check.  This change will minimize the number of packages installed in the final runtime image and improve efficiency.

markovejnovic:
Hey, @mortee, apologies for this not landing prior. What is the motivation for this PR, as in what is the motivation behind adding a wrapper script to invoke curl/nc, rather than just invoking `curl` directly on `/healthz`?

mortee:
The motivation should be clear: for the container to have healthcheck by default, and saving the user the trouble of thinking one up, and having to tamper with the container in order to have the tools for it available.

When it comes to the wrapper script: I think it's cleaner, and pretty common practice as opposed to putting the convoluted command directly in the Dockerfile.

markovejnovic:
talked with @ellie and we decided that perhaps the best path forwards is to just install curl so the code is more readable! thank you for your PR!

PR #2485, opened 2024-12-15
docs: slightly nicer default config.toml order

ShalokShalom:
This groups the relevant settings together, and it also clearifies the wording

Less hustle looking for the right setting now

## Checks

P.S: Do I understand it right, that there is on setting for cltr+r yet?


markovejnovic:
thank you!

PR #2494, opened 2024-12-26
feat(client): add prune filters for age and length

imsakg:
- Introduce options to prune history by minimum age and maximum length.
- Default `dry_run` to true and allow optional specification.
- Improve command filtering for efficient pruning.

## Checks

### DIFF
--- a/crates/atuin/src/command/client/history.rs
+++ b/crates/atuin/src/command/client/history.rs
@@ -114,8 +114,17 @@ pub enum Cmd {
     /// Delete history entries matching the configured exclusion filters
     Prune {
         /// List matching history lines without performing the actual deletion.
-        #[arg(short = 'n', long)]
+        #[arg(short = 'n', long, default_value = "true")]
+        // accept no value
+        #[arg(num_args(0..=1), default_missing_value("true"))]
+        // accept a value
         dry_run: bool,
+        /// Delete all history lines which are older than the configured minimum age (seconds).
+        #[arg(long)]
+        min_age: Option<u64>,
+        /// Delete all history lines which are longer than the configured maximum length.
+        #[arg(long)]
+        max_length: Option<usize>,
     },
 }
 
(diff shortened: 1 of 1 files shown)


imsakg:
Example usages:

- `atuin history prune --max-length 200`
- `atuin history prune --min-age 86400 --dry-run false`

tessus:
I don't really understand the use-case for this. `history prune` only deletes history that matches the `history_filter`.

The use-case for pruning was to remove entries from the history after one changes the `history_filter`. So why would you want to not record command `x`, but keep all commands `x` not older than N seconds?

Don't get me wrong, I am not against your PR (and I don't make any decisions for this project anyway), I am just saying that I do not understand your use-case.

imsakg:
Hey @tessus,

Certainly, I did not get it wrong. Let me clarify my use case and the purpose of this pull request.

Consider a scenario where you are executing a command and frequently changing flags until you identify the correct one. Once you have determined the appropriate flag, you may not wish to retain the other executed commands in your Atuin database, as they contribute to unnecessary bloat.

For example:
1. `my_command -x my_file`
2. `my_command -x -y my_file`
3. `my_command -x -y -z my_file`
4. `my_command -a my_file` <-- This is the correct command.

Additionally, there is another use case to consider. Suppose you are executing commands that you intend to use only once. Such commands will also be stored in your Atuin database, and it is likely that you do not wish to keep them.

For instance:

`DB_CONNECTION_STRING=aws-some-db-service-some-weird-url-geolocating-tag-bla-bla-bla my_program_that_connecting_to_db query "Hello World"`

This pull request allows you to remove those unnecessary commands based on their length or age.

tessus:
Interesting, I understand. Mostly, I guess.

I would have pruned all and then used the "correct" version of the command once (which is either in my clipboard or I saved it in a temp file)
However, if `my_command` is in the history_filter, it wouldn't be recorded either way, so still a bit confused there. This workflow makes more sense for `atuin search --delete` operations than pruning, but maybe I'm still missing something.

markovejnovic:
Hi folks, thank you for this PR! Since the code has diverged so much from this point, and the PR appears to have gone stale, I will be closing it. If you need me to reopen it, ping me and I'll re-open the PR and provide some help/feedback on it.

PR #2174, opened 2024-06-20
filter out various environment variables containing potential secrets

drmorr0:
Add secret key filters for `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, `AZURE_STORAGE_ACCOUNT_KEY`, and `GOOGLE_SERVICE_ACCOUNT_KEY`.

I don't know how common these environment variables are. I know there are a lot of libraries that look in AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY env vars for credentials, but I've done a bunch of searching and it seems like that pattern is less common for the other cloud providers?

## Checks
- [x] I am happy for maintainers to push small adjustments to this PR, to speed up the review cycle
- [x] I have checked that there are no existing pull requests for the same thing

PR #2501, opened 2024-12-28
feat(client): update AWS secrets env var handling checks

ardrigh:
Update to use AWS env vars, replace duplicated entries.

`AWS_SECRET_ACCESS_KEY` the secret paired to Access Key IDs

`AWS_SESSION_TOKEN` the secret session token for STS sessions

AWS Access Keys should be caught by the test above for `"AKIA[0-9A-Z]{16}"` though this might also be updated to use a matching format with these two entries (including the env var name).

Relates to https://github.com/atuinsh/atuin/pull/2174

AWS documentation: https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html

## Checks


ardrigh:
With the v18.4 these secrets appear in the history

image

ellie (approved):
nice fix! thank you

PR #2508, opened 2024-12-29
fix(wrapped): fix crash when history is empty

alexandregv:
## Checks

## Details

I noticed a crash when there is no history entries for the specified year, 2023 in my case, so I added a simple check and message.  
I'm not sure of the behavior if someone has no history for 2023 but has some for 2024 and does a `atuin import`. If it's not recommened, please let me know and I will update the error message. Feel free to change it too.

Before:
image

After:
image


ellie (approved):
Ohhh yes, nice catch. Thank you for this! 🥳

PR #2519, opened 2025-01-02
style: Avoid calling `unwrap()` when we don't have to

andrewaylett:
Use `if let` rather than `is_some()` followed by `unwrap()`, and coerce errors instead of calling `unwrap()` when available.

## Checks


ellie (approved):
much better! thank you

PR #2521, opened 2025-01-03
test: Use testcontainers for integration tests requiring Postgres

andrewaylett:
Rather than requiring the environment to supply a suitable database, we may use TestContainers to provision a fresh database for each test. This allows contributors to run the entire test suite without project-specific setup, although it does require a working Docker environment.

## Checks

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -155,16 +155,6 @@ jobs:
   integration-test:
     runs-on: ubuntu-latest
 
-    services:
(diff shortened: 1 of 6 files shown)


ellie:
Thank you for opening this, however I really don't want to add a dependency on Docker for our tests to complete. It should be up to the developer how they wish to run the database

Personally, I run a non-containerized postgres locally for all of my development. Docker has a non-negilgeable impact on battery usage for laptops

PR #2523, opened 2025-01-04
chore: Remove unneeded dependencies

andrewaylett:
These dependencies are unused in actual code, and the test I've removed is a remnant from a move to use an external library -- it was useful to show that the mechanical transformation was correct, but it's only testing that library nowadays.

## Checks


ellie (approved):
thank you! really appreciate you cleaning up!

PR #2527, opened 2025-01-06
feat: added star history chart

github-star-history:
Star History chart added, hope you like it! :)

## Checks

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,5 +130,9 @@ If you find any security issues, we'd appreciate it if you could alert ellie@atu
 
 Made with [contrib.rocks](https://contrib.rocks).
 
+# Star History
+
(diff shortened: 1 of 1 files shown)


ellie:
Looks like the API is down - I'd be open to this in theory, but it seems like this might be automated and unreliable

github-star-history:
The API has maintained 100% uptime since December 2023. For repositories with a large number of stars, fetching all the data can take some time. To optimize performance, we cache images for 24 hours. When the cache expires, the first request triggers a cache refresh, which takes approximately 40 seconds for this repo. As a result, the first request after the 24-hour period may fail with a timeout error. However, this triggers the update process, ensuring that all subsequent requests within the next 24 hours are served in just 2-3 milliseconds.

PR #2543, opened 2025-01-12
Add PowerShell module

### LINKED ISSUE #84: Powershell Support
I absolutely love the project and would love to be able to use it within powershell

ltrzesniewski:
This adds PowerShell support 🎉 

I built this script around @lzybkr's [prototype](https://github.com/atuinsh/atuin/issues/84#issuecomment-1689168533), so I added him as co-author (I hope that's ok). I wouldn't know where to start without his contribution.

I'm not a PowerShell expert, so this was a nice opportunity to learn some stuff. I think it's ok, but I would appreciate if someone more knowledgeable in the matter could review this though.

It would be nice if other PowerShell users could test this with their configs and report any issues. I wouldn't be surprised if there are some remaining bugs or missing features.

Fixes #84

## Installation

```powershell
atuin init powershell | Out-String | Invoke-Expression
```

This requires `atuin` to be in the path and the [PSReadLine](https://github.com/PowerShell/PSReadLine) module to be installed, which is the case by default.

## Tests

I tested this on the following:
- PowerShell 7.4.6 / PSReadLine 2.3.5 / Windows (the latest one)
- PowerShell 5.1.22621.4391 / PSReadLine 2.0.0 / Windows (the one shipped with Windows)
- PowerShell 7.4.6 / PSReadLine 2.3.5 / Ubuntu WSL (strangely, it didn't behave exactly like the Windows version)

I also tested this with and without my custom [Oh My Posh](https://ohmyposh.dev/) prompt. It works fine in both cases, except that since my OMP config contains `"newline": true`, my prompt is multiline and shifts downwards by a single line on each `atuin search -i` invocation. I suppose fixing that would require an environment variable which specifies how many lines the prompt takes, but I don't think it's a big issue and I can add such a fix later, maybe after receiving some feedback from other users.

## Other notes

I added the following before the first draw call, since I got Atuin drawing over my PS prompt without entirely erasing it:

```rust
if settings.inline_height > 0 {
    terminal.clear()?;
}
```

## Checks


Master-Hash:
I've tested it and everything works well. Thank you for your work!

justunsix:
Thank you @ltrzesniewski and lzybkr, appreciate the work ❤️

## Successful Test on

- PowerShell 7.4.6 / PSReadLine 2.3.5 / Windows 11 10.0.26100 N/A Build 26100
- PowerShell 5.1.26100.2161 / PSReadLine 2.3.6 / Windows 11 10.0.26100 N/A Build 26100

### About Tests

Details on test steps and commands run at https://github.com/justunsix/rust-tests/blob/main/atuin-tests/pr2543/README.md 

- How to build from PR branch
- Commands executed to test
- Test platform versions

- Also tested successfully with nushell 0.101.0 in same environment

### Error not related to PR

Sharing an error found during testing in case others find it during their testing, though it is a local issue and not related to the PR itself.

The error also occurs when testing with atuin/main in nushell in the same environment, so error
is not caused by changes in the PR.

- During testing, there was an error running `atuin search -i`:

```sh
Error: migration 20230531212437 was previously applied but has been modified
>>
>> Location:
>>     C:\Users\user1\usr\reference\atuin\crates\atuin-client\src\record\sqlite_store.rs:61:9
Error:: The term 'Error:' is not recognized as a name of a cmdlet, function, script file, or executable program.
Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
```

- Detailed trace and local fix: https://github.com/justunsix/rust-tests/blob/main/atuin-tests/pr2543/README.md#error-migration-20230531212437-and-fix-to-delete-old-files
- `C:\Users\user1\usr\reference\atuin\` is where the PR was built

- There was a problem with the local `~/.local/share/atuin/records.db`. Fix was to delete `records.db` and have a new one generated.
- The system could run powershell and nushell fine with atuin and they successfully shared `~/.local/share/atuin/`

ltrzesniewski:
@justunsix thanks for testing and the detailed report! Sorry, I should have mentioned the database issue in the OP, I'll add a warning right now. 😬

ltrzesniewski:
I found an issue with encoding and pushed a fix: `PSConsoleReadLine` prompt editing functions are now called in UTF-8 mode in order to handle Atuin search results correctly when those contain codepoints invalid in the current console encoding.

I took the opportunity to add support for an `ATUIN_POWERSHELL_PROMPT_OFFSET` environment variable (please tell me if you'd prefer a better name) that adds an offset to the Y position of the prompt when returning from `atuin search -i`, which is useful when the prompt is multi-line.

The compiler was also complaining about an elided lifetime, but not every time for some reason, so I made it explicit.

justunsix:
Re-tested after the push and [confirming successful tests](https://github.com/justunsix/rust-tests/blob/main/atuin-tests/pr2543/README.md) on:

- PowerShell 7.4.6 / PSReadline 2.3.5 / Ubuntu 24.04.1 LTS
- PowerShell 7.4.6 / PSReadLine 2.3.5 / Windows 11 10.0.26100 N/A Build 26100
- PowerShell 5.1.26100.2161 / PSReadLine 2.3.6 / Windows 11 10.0.26100 N/A Build 26100

Additions from the earlier testing:

- Confirmed using a multiple line prompt and setting `$env:ATUIN_POWERSHELL_PROMPT_OFFSET=-3` as an example with a 4 line prompt prevents  extra lines being added in the terminal when using `atuin search -i` and making a selection. For tests, used a [Starship prompt](https://starship.rs/).

- Tested more atuin commands across different sessions
- Added testing on Ubuntu

Comments:

- Regarding the offset, was thinking later the docs will need to be updated and inform the user of the setting. 
- So, I drafted what could be in the [installation, Installing the shell plugin](https://github.com/atuinsh/docs/blob/main/src/content/docs/guide/installation.mdx) section summarizing what the PR can do; can be discussed later.

```mdx
<TabItem label="powershell">
    Add

    ```shell
    atuin init powershell | Out-String | Invoke-Expression
    ```

    to your $PROFILE file.
    
    If you use a prompt that has more than one line, also add 
    
    ```shell
    $env:ATUIN_POWERSHELL_PROMPT_OFFSET=-X
    ```
    
    to your $PROFILE file. 
    
    Replace X with the number of lines in your prompt, but take away 1. 
    For example, if your prompt has 2 lines, the environment variable would be -1.
</TabItem>
```

stark1tty:
Amazing, cannot wait! Thanks for the hard work.

ltrzesniewski:
⚠️ This stopped working after I updated PowerShell to v7.5.0 😠

Atuin is launched but nothing appears on screen. I don't see anything relevant in the changelog. I guess I'll have to take a look.

ellie:
Awh man :( What part breaks?

ltrzesniewski:
I can see that Atuin is launched and even works:

image

Pressing key up N times and key down N+1 times will close it, pressing enter will execute a command, tab will insert it, etc.

The problem is that the UI doesn't show up on screen. You just see the prompt as usual. 😕

ltrzesniewski:
FWIW, it still works on Linux with pwsh 7.5.0

ltrzesniewski:
This is caused by https://github.com/PowerShell/PowerShell/pull/20853 - looks like I'll have to redirect Atuin's standard streams from the PowerShell script. Ugh.

I'll need to think a bit more about it, and it doesn't look like a one-line change, so I'm converting this PR to draft for now. 😞

ltrzesniewski:
I worked around the issue by adding a hidden command line option which tells Atuin where to write the result instead of going through stderr. I hope that's ok for you.

ltrzesniewski:
I was a bit bored so I added dotfiles support since it was missing 😅 

A few notes:

- I refactored a bit how the `build()` functions work in `AliasStore` and `VarStore`, since they were unnecessarily querying the same data multiple times, and I didn't want to add another pair of queries.
- Built-in PowerShell aliases (`Set-Alias`) are quite useless since they don't let you add implicit arguments (you can alias `git`, but not `git status` for instance), so I used functions instead.  
  See https://github.com/PowerShell/PowerShell/issues/12962
- Using functions means you can't make syntax errors in your aliases. The output of `atuin init powershell` won't parse if you do.
- Environment variables should be fine as long as their names are valid.
- The *"Error: migration 20230531212437 was previously applied but has been modified"* message was caused by checking-out files from git in CRLF mode, I submitted a separate PR for this: #2601

ellie:
Thank you for all that! Appreciate you keeping this up to date. I'm happy with the hidden command flag + have no strong preference for stderr.

A couple of thoughts

1. After the breakage from upstream changes earlier, do you think this is likely to happen again? I'd like to avoid merging, announcing support, and then having upstream break things for us. Additionally, do you think this requires more testing from others, or is it good to go? I'm happy to share this PR around a bit if needs be.
2. If this is merged, how engaged do you intend on remaining with the project? Not asking for any commitment of course, but mostly just concerned that if I merge this and something breaks, I'm not really in a place to support/fix powershell issues at the moment

Thanks again for all of your work here! <3

ltrzesniewski:
> I'm happy with the hidden command flag + have no strong preference for stderr.

I'd then keep using this since it works on 7.5.0 and seems more robust overall.

> After the breakage from upstream changes earlier, do you think this is likely to happen again? I'd like to avoid merging, announcing support, and then having upstream break things for us.

I understand the concern. Well, quite honestly, it looks like a genuine bug in PowerShell (it produces a file with a *"The handle is invalid."* error message and a PR is currently open to fix this).

I hope such issues won't happen often, but I'm unable to guarantee that, unfortunately. I just did my best to make the same script work on 5.1 (the old PowerShell version shipped with Windows) and 7.4 (the latest version at the time this PR was opened).

> Additionally, do you think this requires more testing from others, or is it good to go? I'm happy to share this PR around a bit if needs be.

I use this, and it works well for me (except for that 7.5.0 thing 😓). I'm happy a few other people were also able to test it, so I'd say it's good to go. The dotfiles feature is less robust as I indicated earlier, but I guess it could also break on other shells if you give it invalid input.

If you'd prefer to be more cautious, you may release it as a beta first.

> If this is merged, how engaged do you intend on remaining with the project? Not asking for any commitment of course, but mostly just concerned that if I merge this and something breaks, I'm not really in a place to support/fix powershell issues at the moment

I'll be happy to help! I usually keep supporting the new features I add to other projects as I wouldn't want to deliver broken stuff. Feel free to tag me if there's an issue with PowerShell, and I'll do my best to help.

> Thanks again for all of your work here! <3

My pleasure! 🙂

ltrzesniewski:
> What gets written to this file/location and would it leak history/commands to disk?

In the PowerShell script, Atuin writes the command line you select to a temporary file, which gets read by the script then deleted. The file should always get deleted as it is done in a `try { } finally { }` block [here](https://github.com/ltrzesniewski/atuin/blob/5cccb331e408b67b41375d1a25a33b64875269c8/crates/atuin/src/shell/atuin.ps1#L104).

the-mentor:
Hi @ltrzesniewski did you test this in powershell on MacOS and Linux or only on windows?

ltrzesniewski:
> Hi @ltrzesniewski did you test this in powershell on MacOS and Linux or only on windows?

I tested it on Windows and WSL Linux, like specified in the initial post. I don't have MacOS.

mickey1700:
@ltrzesniewski I've tested the branch, and everything functions well except for a ~1-second delay when exiting Atuin - either by selecting an item or pressing `Ctrl+G`. After exiting, there's a noticeable lag before returning to the shell prompt. Could this be related to my environment, or are you observing the same behavior?

God-damnit-all:
@mickey1700 Run Get-Module and post what version PSReadLine is.

ltrzesniewski:
@mickey1700 some delay is expected, unfortunately, as [PSReadLine polls for commands every 300ms](https://github.com/PowerShell/PSReadLine/blob/9b6a47d7980109b652d33e7c26c5763ccfbf8862/PSReadLine/ReadLine.cs#L193). It shouldn't be as long as 1 second though, but it's noticeable.

the-mentor:
If I understand correctly atuin is starting a new powershell process and maybe if your profile is large it takes longer?

Just a guess I could be totally wrong

ltrzesniewski:
@the-mentor nope, powershell runs atuin, and work continues in the same powershell process when atuin exits.

I'm pretty sure this is due to how PSReadLine polls for commands, but I didn't spend a lot of time on this. I thought maybe an issue could be opened about this in PSReadLine about shortening this delay.

God-damnit-all:
> @the-mentor nope, powershell runs atuin, and work continues in the same powershell process when atuin exits.
> 
> I'm pretty sure this is due to how PSReadLine polls for commands, but I didn't spend a lot of time on this. I thought maybe an issue could be opened about this in PSReadLine about shortening this delay.

Yeah, I was going to bring that up, but an older version of the module might have other delays.

the-mentor:
I have lots of powershell experience so when I get a chance I'll try to see if I can look the code and test it out.

mickey1700:
> @mickey1700 Run Get-Module and post what version PSReadLine is.

I'm on the latest stable version: 2.3.6.  Are you using a beta version and not seeing the issue on your end?

> @mickey1700 some delay is expected, unfortunately, as [PSReadLine polls for commands every 300ms](https://github.com/PowerShell/PSReadLine/blob/9b6a47d7980109b652d33e7c26c5763ccfbf8862/PSReadLine/ReadLine.cs#L193). It shouldn't be as long as 1 second though, but it's noticeable.

Thanks - I definitely expected some delay, similar to what can be observed with `PSFzf`. However, in my case, the delay with `atuin` feels noticeably longer.

> If I understand correctly atuin is starting a new powershell process and maybe if your profile is large it takes longer?
> 
> Just a guess I could be totally wrong

Good thought! I tested with the `-noprofile` flag to rule that out, but unfortunately, I saw the same results.

@ltrzesniewski Let me know if there’s anything else I can try to help pinpoint the cause.

God-damnit-all:
> I'm on the latest stable version: 2.3.6. Are you using a beta version and not seeing the issue on your end?

I have a delay, I just asked because "1 second" seemed a little long to me.

mickey1700:
> > I'm on the latest stable version: 2.3.6. Are you using a beta version and not seeing the issue on your end?
> 
> I have a delay, I just asked because "1 second" seemed a little long to me.

Fair enough. I measured the delay during a random invocation — it took around 800 ms to return to the shell. Sometimes it's faster, other times a bit slower.

ltrzesniewski:
> I definitely expected some delay, similar to what can be observed with `PSFzf`. However, in my case, the delay with `atuin` feels noticeably longer.

This PR [runs up to 4 PSReadLine commands](https://github.com/ltrzesniewski/atuin/blob/84379cd70fa1e708b762f81bfbcc757fc7d8c852/crates/atuin/src/shell/atuin.ps1#L81-L99) after Atuin exits: `InvokePrompt`, `RevertLine`, `Insert`, `AcceptLine` (along with `GetBufferState` before running Atuin). I suppose that 300ms delay could be awaited between several of those commands.

Vacant0mens:
> This is caused by [PowerShell/PowerShell#20853](https://github.com/PowerShell/PowerShell/pull/20853) - looks like I'd have to manually copy Atuin's stdout to PowerShell's stdout through the script. Ugh. That's cursed.
> 
> I'll need to think a bit more about it and maybe find a better solution. In any case, it doesn't look like a simple change, so I'm converting this PR to draft for now. 😞

I've tested this on PowerShell 7.5.1 on WSL, and I found a couple of bugs (and a fix for one of them). The two I don't have fixes for are fairly generic, so I'm not sure they would show up for other platforms. I'm also not sure if any of these bugs are the actual issue that was reported to PowerShell, but it seems at least related, so I quoted the previous comment just in case.

#### Bug 1
I have `Escape` mapped to the `RevertLine` function. So when I'm browsing the history with `UpArrow` or `Ctrl+r` and choose not to use a line from the history by pressing `Escape`, it throws the error `Cannot call a method on a null-valued expression`, coming from this line when the user executes the `RevertLine` function before choosing a history item:
```powershell
$suggestion = (Get-Content -Raw $resultFile).Trim()
```
#### Fix 1
I know it's not very DRY, but if the `.Trim()` call gets moved down to both of the `Insert` calls, like this: 
```powershell
[Microsoft.PowerShell.PSConsoleReadLine]::Insert($suggestion.Substring(17).Trim())
```
and:
```powershell
[Microsoft.PowerShell.PSConsoleReadLine]::Insert($suggestion.Trim())
```
(respectively);

The same error also occurs when it's checking for the `__atuin_accept__` string, so adding this line also helps:
```powershell
if ($null -eq $suggestion) { $suggestion = '' } # <- convert $suggestion to empty string instead of $null
if ( $suggestion.StartsWith('__atuin_accept__:'))
```
or for PowerShell 7 only, you can change the line to:
```powershell
if ( ${ $suggestion }?.StartsWith('__atuin_accept__:') ) # <- PS 7 syntax only
```
I prefer not to say "the error goes away" after these changes, but the scriptblock does handle the `$null`'s a bit better when the user chooses **not** to use a line from Atuin's history.

#### Bug 2
Accepting the line from Atuin's history (whether with `Enter` or `Tab`), creates 9-10 lines of whitespace before drawing the prompt again, but only when the `atuin search -i` command causes the console to add lines to the scrolling buffer.

I don't know if this is intended or not, but it's also not a major problem, more like an interesting quirk. I'm not familiar enough with `PSReadLine` to know how to fix it, or if it can be fixed, and it may only be a bug on WSL, I'm not sure yet, but I can test it on my Debian machine later.

#### Bug 3
Pressing `Enter` inserts the line, but it doesn't execute ("Accept") it like I expect it to and/or like it does when using Atuin in Bash or the raw `atuin search -i` command in PowerShell. (e.g. `Tab` and `Enter` effectively do the same thing right now, at least for me in WSL).

Again, this is not a major problem, but it is a slight bother having to hit `Enter` twice.

#### Bug 4
PowerShell history in a Linux or Linux-based environment gets added to the same database as the other shells.
e.g. I see PowerShell commands in my Atuin history/search in Bash.

Not sure if this is fixable, but it would be nice to have the two shells handled separately, if that's possible.

ltrzesniewski:
Hello @Vacant0mens, thank you very much for testing and providing a helpful report!

However, the behavior you are experiencing seems to come from using an Atuin PowerShell module which is *incompatible* with the `atuin` executable in your path, or from an older version. I had experienced some of the behaviors you describe while developing the PR and had fixed them already, so you shouldn't see those anymore.

Could you please run `atuin --version`? At the moment I'm writing this message, the PR is at 84379cd70fa1e708b762f81bfbcc757fc7d8c852 and the output should be `atuin 18.5.0-beta.3`.

If it is something else, you can install the latest version from this page with:

```powershell
cargo install --git https://github.com/ltrzesniewski/atuin.git --branch powershell-pr
atuin init powershell | Out-String | Invoke-Expression
```

After installing it with cargo, `whereis atuin` reports `atuin: /home/lucas/.cargo/bin/atuin` for me.

All right, let's take a closer look.

**Bug 1**

This is very surprising to me, as PowerShell shouldn't execute bindings while Atuin is running. Could you show me the command you used to bind `Escape` to `RevertLine`? I tried `Set-PSReadLineKeyHandler -Chord Escape -Function RevertLine` but couldn't reproduce the problem you describe.

I also tested the following: `Set-PSReadLineKeyHandler -Chord Escape -ScriptBlock { [System.Console]::Beep() }`

It beeps when I press Escape in the shell (the terminal shows a 🔔 icon in the tab at least), but not when I exit Atuin.

image

The line you mention:

```powershell
$suggestion = (Get-Content -Raw $resultFile).Trim()
```

comes from a previous version of this PR. The current version has the following:

```powershell
$suggestion = (Get-Content -Raw $resultFile -Encoding UTF8 | Out-String).Trim()
```

The `Out-String` part should fix the `$null` issue. In any case, this suggests you're not testing the latest version of the PR.

BTW, I can't use the PowerShell 7 syntax, as I want for this feature to work on the PowerShell 5 version which is shipped with Windows.

**Bug 2**

This is something which was very frustrating, and which I have fixed some time ago. Here's the behavior I get:

image

**Bug 3**

Pressing enter should execute the selected line. At least it works for me.

I suppose the `atuin` executable you're using does not recognize the `ATUIN_SHELL_POWERSHELL` environment variable and does not insert the `__atuin_accept__:` prefix.

**Bug 4**

This is default Atuin behavior, but you can use a separate Atuin configuration in your PowerShell session if that's the behavior you'd like to have. You can set a separate `ATUIN_CONFIG_DIR` environment variable in PowerShell.

Vacant0mens:
Ah, darn. My bad. 🤦 
I didn't realize I hadn't installed the right version (I had v18.6). I had installed Atuin earlier using the regular scripted-install method. I was also running the scriptblock manually using the code from the PR rather than using `atuin init powershell` command, which would explain older/already-fixed issues showing up.

That said, for a version of atuin that **wasn't** built with/for the PowerShell additions and for using an older scriptblock, it worked surprisingly well all things considered! 🎉 

I went back and installed the beta version from your branch as you suggested and used the `atuin init powershell` command as you suggested (and as I should've earlier). Other than Bug 4 that you corrected me on (thank, by the way), I didn't experience any of the issues I mentioned previously. 👌 

Good job with this PR! Can't wait for this to get merged and released! 👋

Vacant0mens:
I've also managed to install it just fine on Windows with the same methods.
I don't love that it takes up the whole terminal window when doing the "UpArrow" search, but that seems more like an atuin issue than a powershell issue.

Otherwise it works great!

I haven't had a chance to test it on a Debian machine yet (other than WSL), but I will try to sometime this week.

ltrzesniewski:
> I don't love that it takes up the whole terminal window when doing the "UpArrow" search, but that seems more like an atuin issue than a powershell issue.

That may just be a different config, see if it is related to #2249

lapo-luchini:
I am syncing  my PowerShell and WSL `atuin` to the same server account and I can find all the lines from any PC… I want this.
What I didn't expect is that the WSL and the PowerShell show up in search as the "same HOST", even though `~/.local/share/atuin/host_id` and `$env:USERPROFILE/.local/share/atuin/host_id` have different values.

ltrzesniewski:
@lapo-luchini This should not depend on this PR. I *suppose* you're either calling the same `atuin` executable on both environments, or have an `XDG_DATA_HOME` environment variable defined.

Atuin does the following: if `XDG_DATA_HOME` is defined, it uses it as your home directory path. Otherwise, it uses `USERPROFILE` on Windows and `HOME` on non-Windows *builds* (meaning it depends on the build target, not on the runtime). It then appends `.local/share/atuin/host_id` to the home path.

You may check this with commands such as `Get-Command atuin` on PowerShell, `whereis atuin` on Linux or `where.exe atuin` on Windows, or maybe `atuin info` or `atuin doctor` on both environments and compare their outputs.

lapo-luchini:
> @lapo-luchini This should not depend on this PR. I _suppose_ you're either calling the same `atuin` executable on both environments, or have an `XDG_DATA_HOME` environment variable defined.

Mhh, no, definitely separated folders:

```
% atuin info
Config files:
client config: "/home/lapo/.config/atuin/config.toml"
server config: "/home/lapo/.config/atuin/server.toml"
client db path: "/home/lapo/.local/share/atuin/history.db"
key path: "/home/lapo/.local/share/atuin/key"
session path: "/home/lapo/.local/share/atuin/session"

Env Vars:
ATUIN_CONFIG_DIR = "None"

Version info:
version: 18.6.1
```

vs

```
PS C:\Users\lapo> atuin info
Config files:
client config: "C:\\Users\\lapo\\.config\\atuin\\config.toml"
server config: "C:\\Users\\lapo\\.config\\atuin\\server.toml"
client db path: "C:\\Users\\lapo\\.local\\share\\atuin\\history.db"
key path: "C:\\Users\\lapo\\.local\\share\\atuin\\key"
session path: "C:\\Users\\lapo\\.local\\share\\atuin\\session"

Env Vars:
ATUIN_CONFIG_DIR = "None"

Version info:
version: 18.5.0-beta.3
```

lapo-luchini:
This might be normal tough, as I see that `TABLE history` has only a `hostname`, not a `host_id` column.

lapo-luchini:
Used this a workaround:
```
% cat /etc/wsl.conf
[network]
hostname = hostname-wsl
```

ltrzesniewski:
I rebased the PR on the v18.7.1 release and fixed an issue with quote characters on PS 5.1

/cc @rhinosfly @ajn142

ltrzesniewski:
Well, there has been a new Atuin release today, so I rebased the PR on v18.8.0, and added support for command chaining (#2834).

ltrzesniewski:
I finally took a look at the delay there was when accepting a history item. It's basically instant now, so I suggest updating if you're using this PR. 🙂 

Here are the details for those interested (ping @mickey1700 @the-mentor @God-damnit-all)

The root cause and the fix are here:
- https://github.com/PowerShell/PowerShell/issues/24709
- https://github.com/PowerShell/PowerShell/pull/24711

The fix should be released in PS 7.6.0, but in the meantime, the [following change](https://github.com/ltrzesniewski/atuin/commit/34d1a5a3946948e1984a9645f3613cb9426a96eb) does the trick, and also works on PS 5.1:

```diff
- Start-Process -Wait ...
+ Start-Process -PassThru ... | Wait-Process
```

(There is a small semantic change: `-Wait` also waits for child processes, while `Wait-Process` doesn't, but Atuin doesn't spawn child processes anyway)

worldofgeese:
What needs to be done to merge this?

ajn142:
Just installed the updated version, and it's working great in PS 5.1 and 7.5.3 on Windows 11.

ltrzesniewski:
ℹ️ For those who use multi-line prompts, you shouldn't need to set up `$env:ATUIN_POWERSHELL_PROMPT_OFFSET` anymore with the latest version of this PR.

The value will be initialized automatically the *first time* you run an Atuin search in a given session by executing a prompt in the background and getting its line count. You may also continue to set it manually if needed (though I'm wondering if I should keep the env var or get rid of it entirely).

In other words, no additional config should be needed. As usual, please report any issues you may encounter.

mickey1700:
@ltrzesniewski Thanks for continuing to improve the PR. I've been testing the PR heavily over the past month, and everything works really well. The only feature I wasn't able to get working in the PowerShell version is `atuin scripts`.

ltrzesniewski:
@mickey1700 thanks for testing! This PR isn't related to `atuin scripts`, which is a separate feature.

I took a quick look at it however. There's no docs AFAIK, but I've tried it on Windows with:
```
$ atuin scripts new test --script "C:\Dev\Temp\test.ps1" --shebang 'pwsh'
$ atuin scripts run test
Processing -File 'C:\Users\Lucas\AppData\Local\Temp\.tmpBTLkNp' failed because the file does not have a '.ps1' extension. Specify a valid PowerShell script file name, and then try again.
Script exited with code 64
```

*Sigh.* That's the demanding PowerShell right there. I've tried a few other shebangs such as `--shebang 'pwsh -NoProfile -CommandWithArgs ''Invoke-Expression (Get-Content $args[0])'''`, but this was getting more and more ridiculous...

The following *"worked"* though, if you dare call it that:
```
$ atuin scripts new test --script "C:\Dev\Temp\test.ps1" --shebang 'pwsh C:\Dev\Temp\run.ps1'
$ atuin scripts run test
Hello, world!
```

With `run.ps1` containing: `Get-Content $args[0] | Out-String | Invoke-Expression`. I stopped trying to find a better solution at this point.

But... do you want to know the *funny* part? That first thing I tried on Windows, it works just fine on Linux:

```
$ atuin scripts new test --script /mnt/c/Dev/Temp/test.ps1 --shebang 'pwsh'
$ atuin scripts run test
Hello, world!
```

Because, [of course it does](https://github.com/PowerShell/PowerShell/blob/bd3e3c991caf29a362fcc955a6cb0eace7f5589a/src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs#L1422-L1429)...

In any case, I think it's safe to say that PowerShell isn't really meant to be used this way, but `atuin scripts` *could* be adjusted by adding the `.ps1` suffix to the temporary file when used with `powershell`/`pwsh` shebang on Windows. Someone would need to ask for it, and it wouldn't be part of this PR anyway.

anonygit99:
Great work @ltrzesniewski this PR is feeling pretty solid. I tried out this branch a while ago now, but the changes I pulled in and tested seem to sync & import my pwsh history perfectly OOTB on Windows.

+1 for getting some kind of initial release of a PowerShell integration soon ?

ltrzesniewski:
@anonygit99 and @worldofgeese: As for me, the PR is ready to review. I only added small changes recently (which could have been in a separate PR) and rebased it when it made sense.

You'd need to ask @ellie for the next steps, but given that she released the first open beta of [Atuin Desktop](https://github.com/atuinsh/desktop) yesterday (Congrats BTW! 🎉), I suppose she must be quite busy right now.

ltrzesniewski:
ℹ️ I rebased the PR on Atuin 18.9.0 + 2 more commits (#2922 and #2864), which means this now includes the PowerShell import feature in addition to the changes of the latest Atuin release.

/cc @Vacant0mens

ltrzesniewski:
Now that's a good merge conflict to have 🙂 

image


I'll rebase this later though, there's also `InputAction::AcceptInspecting` to handle and test.

ltrzesniewski:
ℹ️ I rebased the PR on Atuin 18.10.0

ltrzesniewski:
Yay, thank you! 🥳 

Don't worry, I'll be happy to keep maintaining it. I use this every day after all, and it helps me with both Rust and PowerShell. 🙂

Would you reconsider #2714 now that this PR is merged?

PR #2549, opened 2025-01-16
feat(health): add health check endpoint at `/healthz`

enchantednatures:
This adds an endpoint at `/healthz` which returns success. This is useful for both docker-compose and Kubernetes deployments to ensure the server is still.

## Checks

### DIFF
--- a/crates/atuin-server/src/handlers/health.rs
+++ b/crates/atuin-server/src/handlers/health.rs
@@ -1,0 +1,15 @@
+use axum::{http, response::IntoResponse, Json};
+
+use serde::Serialize;
+
+#[derive(Serialize)]
+pub struct HealthResponse {
+    pub status: &'static str,
+}
+
+pub async fn health_check() -> impl IntoResponse {
+    (
+        http::StatusCode::OK,
+        Json(HealthResponse { status: "healthy" }),
+    )
+}
--- a/crates/atuin-server/src/handlers/mod.rs
+++ b/crates/atuin-server/src/handlers/mod.rs
@@ -4,6 +4,7 @@ use axum::{extract::State, http, response::IntoResponse, Json};
 
 use crate::router::AppState;
 
+pub mod health;
 pub mod history;
 pub mod record;
 pub mod status;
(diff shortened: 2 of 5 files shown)


ellie:
It actually makes more sense to use `/` as a healthcheck, as it also makes a query to the database. This is what we do in our helm charts + on the Atuin cloud deployment

enchantednatures:
Thank you for your feedback. I agree that using the root endpoint (/) as a health check is valuable since it tests the deployment's ability to perform a query.

However, in some scenarios it might be desired to have a separate health check for service and database - Kubernetes, for instance, will not route traffic in a load balancer service to a pod which is failing its readiness check, but when a liveness check fails, kubelet will know to restart the container. Due to the different health check types, liveness might make sense to use a lighter weight health check to assist in isolating restarts.

Would it be acceptable to maintain both endpoints—keeping the existing root endpoint for comprehensive health checks while adding a lightweight /healthz endpoint for scenarios where a service-only health check is sufficient?

ellie:
Hmmm you make a good point! In that case, let's have it. Thank you for the PR!

ellie (approved):

mateuscomh:
Question, this mean sense to insert into docker-compose and docker-compose docs?

enchantednatures:
@mateuscomh, I did try to include the health check in the [docker-compose](https://github.com/atuinsh/atuin/pull/2549/files/703f1a8f03b1876eda1607ffdbe0d05f76ad1391#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3R18-R22) file, however as curl is not installed on the image, the health checks could not be completed.

mateuscomh:
@enchantednatures, See if it makes sense to you. In my local container with docker-compose I added this healthcheck for the previous version 18.4.0 since we don't have nc, wget or curl in the default image.
`test: ["CMD", "bash", "-c", "echo > /dev/tcp/localhost/8888"]`
This works for me and show healthy on container
with your contribution on this PR would adding healthz be valid?

Issue #2507, opened 2024-12-29
[Bug]: wrapped subcommand half years seems to be swapped


olekspickle:
### What did you expect to happen?

First of all - thank you for amazing christmas update!

I picked atuin up in the middle of 2024 and wrapped seems to have the year halves swapped, see the next section.

### What happened?

╭────────────────────────────────────╮
│        ATUIN WRAPPED 2024          │
│    Your Year in Shell History      │
╰────────────────────────────────────╯

🎉 In 2024, you typed 7144 commands!
   That's ~19 commands every day

Your Top Commands:
[▮▮▮▮▮▮▮▮▮▮] 1043 ls
[▮▮▮▮▮     ]  567 vim
[▮▮▮▮      ]  515 cd
[▮▮▮       ]  333 gs
[▮▮▮       ]  319 mv
[▮▮        ]  226 cargo run
[▮▮        ]  221 cat
[▮▮        ]  221 rm
[▮▮        ]  209 make
[▮         ]  207 source
Total commands:   7144
Unique commands:  2845

🚀 You're a Navigator! 21.8% of your time was spent navigating directories

📚 Command Vocabulary: You know 2845 unique commands

📦 Package Management: You ran 602 package-related commands

🚨 Error Analysis: Your commands failed 20.6% of the time

🔍 Command Evolution:
  Top Commands in the first half of 2024:
    ls (1041 times)
    cargo (596 times)
    vim (564 times)
  Top Commands in the second half of 2024:

🕘 Most Productive Hour: 17:00 (1026 commands)

### Atuin doctor output

```yaml
{
  "atuin": {
    "version": "18.4.0",
    "sync": null,
    "sqlite_version": "3.46.0"
  },
  "shell": {
    "name": "zsh",
    "default": "zsh",
    "plugins": [
      "atuin"
    ],
    "preexec": "built-in"
  },
  "system": {
    "os": "Ubuntu",
    "arch": "x86_64",
    "version": "24.04",
    "disks": [
      {
        "name": "/dev/mapper/luks-001e5467-103a-4460-9670-18e7e974bb03",
        "filesystem": "ext4"
      },
      {
        "name": "/dev/nvme0n1p2",
        "filesystem": "ext4"
      },
      {
        "name": "/dev/nvme0n1p1",
        "filesystem": "vfat"
      }
    ]
  }
}
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

hetoug:
[...]

PR #2557, opened 2025-01-21
Fix wrapped command if not full year - fix issue #2507

hetoug:
If atuin has not been used the whole year, then the "Command Evolution" will be misleading, as the year is split 182 days after the first command.

This change set the mid-point of the year mid-ways between the first and last command in the year.

## Checks


tessus:
The problem is that it is still wrong.

`in the first half of 2024` means January to June (incl). But if you started to use atuin in July, it will show the commands from July to September (incl) in the first half of the year.

hetoug:
It is better than it is now! Where the second half of the year is 6 months after you started.
`in the first half of 2024` could be construed to mean `the first half of your usage in 2024` -. but probably not.

I couldn't find a short and correct text, so I left it as it was.

PR #2562, opened 2025-01-26
Add the --print0 option to search

offbyone:
-------

This mirrors the addition to `history` from #1274, but with search too.
Ther are history search implementations for shells that are set to
search instead of running the history command.

## Checks


ellie (approved):
lovely! thank you

PR #2430, opened 2024-10-22
feat(client): add filter mode enablement and ordering configuration

pamburus:
Closes #1417
Closes #1654
cc @hco @matthiasbeyer @dasJ 

This is a continuation of the work started in #1655 and #1717.
These pull requests appear to be abandoned, as there has been no activity in them for more than 5 months.

The configuration setting has been renamed according to https://github.com/atuinsh/atuin/pull/1655#issuecomment-1938310766.

## Checks
- [x] I am happy for maintainers to push small adjustments to this PR, to speed up the review cycle
- [x] I have checked that there are no existing pull requests for the same thing

PR #2565, opened 2025-01-28
fix: fixed forced workspace mode enablement

pamburus:
## Checks
- [v] I am happy for maintainers to push small adjustments to this PR, to speed up the review cycle
- [v] I have checked that there are no existing pull requests for the same thing

This pull request fixes backward compatibility break introduced in #2430.


crates/atuin-client/src/settings.rs
```diff
@@ -788,10 +834,6 @@ impl Settings {
             .set_default("daemon.socket_path", socket_path.to_str())?
             .set_default("daemon.systemd_socket", false)?
             .set_default("daemon.tcp_port", 8889)?
-            .set_default(
```
pamburus, line 791: I am not sure about this change.
I noticed duplicate defaults for many parameters here and didn't understand the purpose.

PR #2581, opened 2025-02-11
Add snap packaging

artivis:
This PR adds [snap packaging](https://snapcraft.io/about) together with a ci workflow that builds the snap on pr/merge and publishes it on merge.

The snap is already [available on the Store](https://snapcraft.io/atuin) at the latest release (18.4.0), it is however unlisted at the moment. To give it a shot:

```bash
snap install atuin --beta
```

I've already used it a little while and it seems to work fine. 
But I'm not (yet) an atuin power user and I've only been using it with bash.
However zsh/resh/fish should be supported and I'd love to get some feedback about those other shell.

A couple notes:
- in snaps, `$HOME` is re-mapped and atuin's config files and such are located at `$HOME/snap/atuin/current/`. 
- snaps are sandboxed, therefore to import an existing history the user must connect the appropriate snap interface (e.g. `snap connect atuin:dot-bash-history`). The `import` command in the snap has a wrapper that informs the user if no such connection is established.
- the snap ships [ble.sh](https://github.com/akinomyoga/ble.sh) and [Bash-preexec](https://github.com/rcaloras/bash-preexec) for convenience (e.g. `source /snap/atuin/current/usr/share/blesh/ble.sh`)

I currently own the app on the store and I'd be happy to transfer it to you if you're interested in this pr!

## Checks

### DIFF
--- a/.github/workflows/snap.yaml
+++ b/.github/workflows/snap.yaml
@@ -1,0 +1,70 @@
+name: snap
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    outputs:
+      snap-file: ${{ steps.build-snap.outputs.snap }}
+    steps:
+    - name: Checkout snapcraft.yaml branch
+      uses: actions/checkout@v4
+
+    - name: Build snap
+      uses: snapcore/action-build@v1
+      id: build-snap
+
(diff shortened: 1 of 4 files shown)


ellie:
Hey! Thank you for this, but I'd rather not take this on board. 

We intentionally only build + publish binaries for different operating systems and architectures at the moment, and do not package for individual linux distros or package managers.

Specifically for snap/flatpak/similar, I'd rather avoid having to maintain something that aggressively sandboxes and is likely to have unique problems there

PR #2584, opened 2025-02-14
fix: panic when invoking delete on empty tui

tessus:
If the result set is empty and thus the TUI does not show any entries, a panic occurs when a user invokes 'delete' via `Prefix (Ctrl-A) d` or `Ctrl-D` in the inspector..

image

## Checks


tessus:
Not sure what's up with the macos14 pipeline. It didn't fail. It was cancelled.

ellie (approved):
thank you! 🙏

Issue #2504, opened 2024-12-28
[Bug]: Error unable to open database file


Burnwood1911:
### What did you expect to happen?

it should run correctly

### What happened?

Im on m1 mac latest sequoia
after running the install script it finishes fine when i restart the terminal immediately when i type anything this error shows

```alexp.rossi@IDX-159151INVL ~ % aError: error returned from database: (code: 14) unable to open database file

Caused by:
    (code: 14) unable to open database file

Location:
    crates/atuin/src/command/client.rs:138:18```

### Atuin doctor output

```yaml
i cant even run doctor command
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #2589, opened 2025-02-17
fix: improve broken symlink error handling

jamestrew:
Check atuin setting paths (eg. `db_path`) for broken symlinks on initialization and disable all shell hooks + print error message.

sqlite doesn't create db files even with `.create_if_missing` when the db files are a broken symlink. This would cause sqlite to error and atuin to panic on every single keypress.

Also improves related error handling when calling atuin client commands directly.

closes https://github.com/atuinsh/atuin/issues/2504

## Checks

### DIFF
--- a/crates/atuin-client/src/database.rs
+++ b/crates/atuin-client/src/database.rs
@@ -130,8 +130,12 @@ impl Sqlite {
         let path = path.as_ref();
         debug!("opening sqlite database at {:?}", path);
 
-        let create = !path.exists();
-        if create {
+        if utils::broken_symlink(path) {
+            eprintln!("Atuin: Sqlite db path ({path:?}) is a broken symlink. Unable to read or create replacement.");
+            std::process::exit(1);
+        }
+
+        if !path.exists() {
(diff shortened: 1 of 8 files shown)


crates/atuin-client/src/database.rs
```diff
@@ -118,42 +118,59 @@ pub trait Database: Send + Sync + 'static {
     async fn stats(&self, h: &History) -> Result<HistoryStats>;
 }
 
-// Intended for use on a developer machine and not a sync server.
-// TODO: implement IntoIterator
-#[derive(Debug, Clone)]
-pub struct Sqlite {
-    pub pool: SqlitePool,
-}
+macro_rules! create_sqlite_pool {
```
ellie, line 121: Generally happy with this, thank you! Would it be possible to drop the macro though please?

Seeing as it's only used in two places, I don't think it is necessary. Generally try to keep things simple unless the benefit is really worth it!

ellie (approved):
thank you! appreciate the work here 🙏

PR #2601, opened 2025-03-01
fix: sql files checksums

ltrzesniewski:
Checking out on Windows may cause *.sql files to have CRLF line endings, which will change their checksums and cause migration errors.

Works around https://github.com/launchbadge/sqlx/issues/2659

## Checks


ltrzesniewski:
I also added shell scripts to .gitattributes since `include_str!` doesn't normalize to LF.

ellie (approved):
nice one, thank you!

PR #2605, opened 2025-03-04
fix: don't save empty commands

ltrzesniewski:
I noticed the following after working on #2543:

image

I pressed and held the enter key a few times in order to go to the bottom of the screen to check how Atuin behaved, and now it thinks I love empty commands. 😄 

Even if it shouldn't happen for others that much, I suppose it makes no sense saving these, so this PR removes that.

## Checks


ellie:
Thank you! I'm pretty sure this only ever happened for bash - looks like other shells don't fire hooks for empty commands. 

Great fix to have either way - appreciate the work!

ellie (approved):

Issue #2548, opened 2025-01-16
[Bug]: Ctrl+a not moving cursor to beinning of line


opransjo:
### What did you expect to happen?

Ctrl+a should move cursor to beginning of line, according to https://docs.atuin.sh/configuration/key-binding/#atuin-ui-shortcuts.

### What happened?

The prefix key Ctrl+a conflicts with this, but even changing the prefix key, e.g.

[keys]
prefix = 'q'

does not solve the issue.

The commit "feat: add atuin prefix binding" (https://github.com/atuinsh/atuin/pull/1875), at https://github.com/atuinsh/atuin/commit/b04fc471deb923937b8ef9aab2f39301f8457894, removed `KeyCode::Char('a') if ctrl => self.search.input.start()` from the keybinding handling, which is still not present at https://github.com/atuinsh/atuin/blob/main/crates/atuin/src/command/client/search/interactive.rs#L417 .

### Atuin doctor output

```yaml
{
  "atuin": {
    "version": "18.4.0",
    "sync": null,
    "sqlite_version": "3.46.0"
  },
  "shell": {
    "name": "zsh",
    "default": "zsh",
    "plugins": [
      "atuin"
    ],
    "preexec": "built-in"
  },
  "system": {
    "os": "Arch Linux",
    "arch": "x86_64",
    "version": "rolling",
    "disks": [
      {
        "name": "/dev/nvme1n1p3",
        "filesystem": "ext4"
      },
      {
        "name": "/dev/nvme1n1p1",
        "filesystem": "vfat"
      },
      {
        "name": "/dev/mapper/home",
        "filesystem": "ext4"
      }
    ]
  }
}
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

akinomyoga:
> The prefix key Ctrl+a conflicts with this,

You are supposed to press <kbd>C-a a</kbd>.

> but even changing the prefix key, e.g. [...] does not solve the issue.

I agree that this is an issue.

> The commit "feat: add atuin prefix binding" ([#1875](https://github.com/atuinsh/atuin/pull/1875)), at [b04fc47](https://github.com/atuinsh/atuin/commit/b04fc471deb923937b8ef9aab2f39301f8457894), removed `KeyCode::Char('a') if ctrl => self.search.input.start()`

I think that line can be safely reverted. You should open a pull request.

PR #2626, opened 2025-03-12
feat: Use readline binding for ctrl-a when it is not the prefix

Nelyah:
Previously:
- If the prefix is **not** `ctrl-a`: It will add an `a` character. 
    - expected: Going to the beginning of line

With this fix:
- If the prefix is **not** `ctrl-a`: It will go to the beginning of line

This solves https://github.com/atuinsh/atuin/issues/2548

As I'm understanding it, if `ctrl-a` is the prefix, it will be caught earlier in the function, and if it is not the prefix, then this will act as "go to the beginning of line". 

I have manually tested this on my end and it works on my laptop at least. I didn't see any tests for key inputs specifically in the file. Let me know if there are specific tests that you would like me to add for this!

## Checks


ellie (approved):
looks good to me! thank you

Issue #1578, opened 2024-01-16
Show host and user in the inspector


ellie:
Just opening this so I don't forget

ascheucher:
Just came to the issues to ask for it as a feature request. Given the inspector is the list of commands... :)

dhth:
Hi!

Sending in a [tiny PR](https://github.com/atuinsh/atuin/pull/2634) for this :)
Let me know if the wording needs to change, or you'd prefer to show the user and host info somewhere else in the UI (PR has a screenshot).

PR #2634, opened 2025-03-18
chore: show host and user in inspector

dhth:
## Checks

Changes
---

Fixes https://github.com/atuinsh/atuin/issues/1578.

Adds two new rows for HOST and USER info in the inspector. Looks like this:
image

### DIFF
--- a/crates/atuin/src/command/client/search/inspector.rs
+++ b/crates/atuin/src/command/client/search/inspector.rs
@@ -90,9 +90,11 @@ pub fn draw_stats_table(
 ) {
     let duration = Duration::from_nanos(u64_or_zero(history.duration));
     let avg_duration = Duration::from_nanos(stats.average_duration);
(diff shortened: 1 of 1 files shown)


dhth:
> Thank you! Could we process this field please, and display host and user on separate lines?

Definitely. I've pushed a change for this.

Follow up: if, for whatever reason, `History.hostname` doesn't follow the pattern `HOST:USER`, should we show a fallback string in the TUI? Like `<COULDNT_DETERMINE>` or `<UNKNOWN>`?

ellie:
It always does! The old sync method relied about each install having a unique ID, and we did that by $HOSTNAME:$USER

It's not needed for sync v2, but remains regardless

dhth:
Gotcha. In that case, I'll leave it to be an empty string :)

dhth:
Is there something I need to do to fix the two failing Nix tests?

PR #2638, opened 2025-03-22
fix(build): change atuin-daemon build script .proto paths

sporeventexplosion:
Modify paths specified in the atuin-daemon build script to make tonic-build print the correct cargo:rerun-if-changed paths when the build script is run.

Fixes the atuin-daemon crate being rebuilt unconditionally, even if .proto files are unchanged.

## Checks


ellie (approved):
nice fix, thank you!

Issue #2552, opened 2025-01-18
[ Feature Request ]: Add Snippets Support


stottm:
- Create a Snippet manually or import from history
- TUI prompts for parameters, defaults, common colors, codes, etc.
- Support ASCII / ANSI codes and colors 
- Search / Edit / Tag / Sync Snippets

See [ Pet Snippet Manager ](https://github.com/knqyf263/pet) as inspiration. 

Imagine storing and sync'ing complex commands with multiple parameters, env variables, etc. OpenSSL, LibreSSL, ffmpeg, kubectl, lxc, terraform, etc.

Sure you can write shell functions but integrating with atuin history means you can realize you are repeating frequently and easily build a snippet in a few clicks instead of a function and in my opinion; that is simpler. Plus you could sync the snippets more easily than some dotfile manager such as GNU Stow or even the early dotfile support in atuin. Leave functions do more complex things.

quantumfrost:
+1 for this. Atuin already does most of what makes a snippet manager like pet or navi. Snippets could be implemented as another "filter" context that could be cycled through by pressing Ctrl+R. There could be a shortcut to quickly add a command from the history to snippets, if you realize you've been running it often.

ellie:
I really like this idea, and would find it super useful. Will see what we can do 🙌

alperchoco:
Doesn't Atuin desktop now cover this effectively being larger or smaller snippets?

lasantosr:
I released a while ago [intelli-shell](https://github.com/lasantosr/intelli-shell) based on Pet. It has now reached v1.0.0 in case you want to have a look.

For me, shell history and command templates (or snippets/scripts) are two different things:
https://github.com/lasantosr/intelli-shell?tab=readme-ov-file#intellishell-vs-shell-history

I'm an atuin user myself and i find it incredibly useful, but I developed intelli-shell as a complementary tool

markovejnovic:
Closing since #2644 implements it!

PR #2644, opened 2025-03-24
feat: support storing, syncing and executing scripts

ellie:
ref: #2552

The goal with this PR is to have a totally non-interactive setup. Doesn't really make the most of the rest of the data we have ;)

Follow up with the ability to build scripts via the search tui

## Checks

### DIFF
--- a/crates/atuin-scripts/src/execution.rs
+++ b/crates/atuin-scripts/src/execution.rs
@@ -117,7 +117,7 @@ pub async fn execute_script_interactive(
     // If direct execution fails, try using the interpreter
     if let Err(e) = &child_result {
         debug!("direct execution failed: {}, trying with interpreter", e);
-        
+
         // When falling back to interpreter, remove the shebang from the file
         // Some interpreters don't handle scripts with shebangs well
         debug!("writing script content without shebang for interpreter execution");


crates/atuin/src/command/client/scripts.rs
```diff
@@ -0,0 +1,573 @@
...
+use eyre::{Result, bail};
+use tempfile::NamedTempFile;
+
+use atuin_client::{database::Database, record::sqlite_store::SqliteStore, settings::Settings};
+use tracing::debug;
+
+#[derive(Parser, Debug)]
+pub struct NewScript {
+    #[arg(short, long)]
+    pub name: String,
```
BinaryMuse, line 20: Thoughts on allowing this to be positional? The other commands that take a script name do so, and it's a required parameter.

BinaryMuse (dismissed):
I think this looks excellent. A couple questions, otherwise :shipit:

crates/atuin-scripts/src/execution.rs
```diff
@@ -0,0 +1,282 @@
...
+        if !parts.is_empty() {
+            let mut cmd = tokio::process::Command::new(parts[0]);
+
+            // Add any interpreter args
+            for i in parts.iter().skip(1) {
+                cmd.arg(i);
+            }
+
+            // Add the script path
+            cmd.arg(temp_path.to_str().unwrap());
```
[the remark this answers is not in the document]
ellie: good catch - lots of languages would have been fine but I think some wouldn't deal with it well

crates/atuin/src/command/client/scripts.rs
ellie: yep definitely, that's much better

PR #2646, opened 2025-03-25
feat(zsh): re-enable bracketed paste

lilydjwg:
atuin will reset it so after exiting atuin without executing a command, bracketed paste mode is disabled until a command is executed. This breaks e.g. the bracketed-paste-url-magic widget.

This change will re-enable it if it's enabled; when it's disabled or unavailable, $zle_bracketed_paste[1] will be empty string.

## Checks

### DIFF
--- a/crates/atuin/src/shell/atuin.zsh
+++ b/crates/atuin/src/shell/atuin.zsh
@@ -59,6 +59,8 @@ _atuin_search() {
     output=$(ATUIN_SHELL_ZSH=t ATUIN_LOG=error ATUIN_QUERY=$BUFFER atuin search $* -i 3>&1 1>&2 2>&3)
 
     zle reset-prompt
+    # re-enable bracketed paste
+    echo $zle_bracketed_paste[1] >/dev/tty
 
     if [[ -n $output ]]; then
         RBUFFER=""
--- a/crates/atuin/src/shell/atuin.zsh
+++ b/crates/atuin/src/shell/atuin.zsh
@@ -60,7 +60,8 @@ _atuin_search() {
 
     zle reset-prompt
     # re-enable bracketed paste
-    echo $zle_bracketed_paste[1] >/dev/tty
+    # shellcheck disable=SC2154
+    echo ${zle_bracketed_paste[1]} >/dev/tty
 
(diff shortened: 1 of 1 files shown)


ellie (approved):
thank you!

AnimiVulpis:
For some reason this change resulted in a skewed cursor position in my powerlevel10k based 2-line prompt.

I have set `exit_mode = "return-query"`

Current behaviour (that I dislike):
```bash
❯ ▌ # Using CTRL+R to enter atuin
# Select any line and press TAB (or in my case also Enter)
❯
selected-command-example▌
```
The command will be put below the actual prompt line and there is also no way to get up (other than e.g. CTRL+C)


Behaviour before this change (verified by commenting out `echo ${zle_bracketed_paste[1]} >/dev/tty`:
```bash
❯ ▌ # Using CTRL+R to enter atuin
# Select any line and press TAB (or in my case also Enter)
❯ selected-command-example▌
```

I am using a prompt supplied by https://github.com/romkatv/powerlevel10k/ with the wizard options:
```zsh
# Wizard options: nerdfont-v3 + powerline, large icons, unicode, lean, 24h time,
# 2 lines, solid, no frame, lightest-ornaments, compact, few icons, concise,
# instant_prompt=verbose.
```

Any idea what might be causing this and how to fix this?
(For now I'll keep that line commented out)

_If it help you can check my zsh config [here](https://github.com/AnimiVulpis/dotfiles/blob/main/stow-package/.zshrc)_

lilydjwg:
I can't reproduce this with your .zshrc (also with the other issue). What terminal are you using?

Can you try to replace the added `echo` with `echo -n`?

AnimiVulpis:
That's perfect. `echo -n` fixes the issue. Thank you for looking into this.
Will you create a PR and get it merged?

lilydjwg:
I'll send a pr soon. Did it fix the other issue you mentioned in #1469?

AnimiVulpis:
This change does not fix the issue from #1469. I'm still trying to figure out an easy way to "share" a reproduction environment in order to make debugging possible.
_Maybe a docker image could work_ 🤔

PR #2647, opened 2025-03-25
fix: typeerror in client sync code

### LINKED ISSUE #2645: [Bug]: Build failed installing from cargo

I built from source via `cargo install atuin` on my raspberry pi.
I expected atuin to compile and install successfully.
Machine details:
Raspbian GNU/Linux 11 (bullseye) aarch64
Raspberry Pi 4 Model B Rev 1.2

### What happened?

Compilation failed with the following:
```
   Compiling atuin-client v18.4.0
error[E0283]: type annotations needed
  --> /home/yucjo/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/atuin-client-18.4.0/src/sync.rs:89:52
   |
89 |         if history.len() < remote_status.page_size.try_into().unwrap() {
   |                          -                         ^^^^^^^^
   |                          |
   |                          type must be known at this point
   |
   = note: multiple `impl`s satisfying `usize: PartialOrd<_>` found in the following crates: `core`, `deranged`:
           - impl PartialOrd for usize;
           - impl<MIN, MAX> PartialOrd<deranged::RangedUsize<MIN, MAX>> for usize
             where the constant `MIN` has type `usize`, the constant `MAX` has type `usize`;
help: try using a fully qualified path to specify the expected types
   |
89 |         if history.len() < <i64 as TryInto<T>>::try_into(remote_status.page_size).unwrap() {
   |                            ++++++++++++++++++++++++++++++                       ~
```

### Atuin doctor output

```yaml
Does not apply
```

### Code of Conduct

ellie:
Fixes #2645

This is really weird

1. I have not touched this code in _years_. It has not changed. In recent rust versions, it has a typeerror (see linked issue)
2. This does not occur when running `cargo build`, in release mode or otherwise. It only occurs with `cargo install`
3. I can't find any other occurences of this typeerror online - unsure if it is a compiler regression? The code here is not very complex at all.

## Checks

### DIFF
--- a/crates/atuin-client/src/sync.rs
+++ b/crates/atuin-client/src/sync.rs
@@ -85,8 +85,9 @@ async fn sync_download(
         db.save_bulk(&history).await?;
 
         local_count = db.history_count(true).await?;
+        let remote_page_size = std::cmp::max(remote_status.page_size, 0) as usize;
 
-        if history.len() < remote_status.page_size.try_into().unwrap() {
+        if history.len() < remote_page_size {
             break;
         }
 
--- a/crates/atuin-client/src/sync.rs
+++ b/crates/atuin-client/src/sync.rs
@@ -1,5 +1,4 @@
 use std::collections::HashSet;
-use std::convert::TryInto;
 use std::iter::FromIterator;
 
 use eyre::Result;


alilleybrinker:
Possibly caused by the release of `deranged` version 0.4.1? See: https://github.com/jhpratt/deranged/commit/09108e325a7bedfdb7e285125d7308d3a534e3ec

alilleybrinker:
@ellie Figured it out, [confirmed with jhpratt, creator of `deranged`](https://bsky.app/profile/jhpratt.dev/post/3ll7y4zudm22o). Basically, in 0.4.1, `deranged` added some `PartialOrd` impls that caused breakage here. Rust has a single-impl inference rule where if only one impl is found to exist it'll use that one during inference even if other conflicting impls *could* exist in the future. In this case, `deranged` adding some new impls causing the inference here to become ambiguous.

The reason that happened with `cargo install` and not `cargo build` is that by default `cargo install` ignores the lockfile (you can make it use the lockfile with `cargo install --locked`). This caused the `install` build to pick up the new patch release, and thus to surface the new ambiguity. If you did a `cargo update` to bump the resolved version of `deranged` in the lockfile you'd see the same breakage with `cargo build`.

PR #2648, opened 2025-03-25
chore: don't panic if clipboard cannot be accessed

### LINKED ISSUE #1707: Panic when copying on a system not running X
See more details:

https://forum.atuin.sh/t/error-when-copying-to-clipboard-without-x-server/157

# Replication steps

- Run on a machine without X (eg, SSH into a server)
- ctrl-y on a history item to yank to clipboard
- observe panic

```
thread 'main' panicked at atuin/src/command/client/search/interactive.rs:1013:36:
called `Result::unwrap()` on an `Err` value: XcbConnect(DisplayParsingError)
```

# Expected behaviour

Atuin should do nothing in an environment without a clipboard

dhth:
Fixes https://github.com/atuinsh/atuin/issues/1707.

Summary
---

`atuin` doesn't panic if the system clipboard cannot be accessed for any reason.

Questions
---

- Do we also want to stop panicking if setting text to the clipboard fails?
- Do we also want to stop panicking if the follow up check to see if the value was set fails?
- Should we add an error message to the TUI if any of the steps involved fail?

## Checks

### DIFF
--- a/crates/atuin/src/command/client/search/interactive.rs
+++ b/crates/atuin/src/command/client/search/interactive.rs
@@ -1244,10 +1244,12 @@ pub async fn history(
     any(target_os = "windows", target_os = "macos", target_os = "linux")
 ))]
 fn set_clipboard(s: String) {
-    let mut ctx = arboard::Clipboard::new().unwrap();
-    ctx.set_text(s).unwrap();
-    // Use the clipboard context to make sure it is saved
-    ctx.get_text().unwrap();
+    // Do nothing if clipboard cannot be accessed
+    if let Ok(mut ctx) = arboard::Clipboard::new() {
+        ctx.set_text(s).unwrap();
+        // Use the clipboard context to make sure it is saved
+        ctx.get_text().unwrap();
+    }
 }
 
 #[cfg(not(all(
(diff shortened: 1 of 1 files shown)


markovejnovic:
Hey @dhth, thanks for your PR! I'll chime in a little bit on what I think the best path forwards is:

> - Do we also want to stop panicking if setting text to the clipboard fails?

Yes!

> - Do we also want to stop panicking if the follow up check to see if the value was set fails?

Yes!

> - Should we add an error message to the TUI if any of the steps involved fail?

This one might not be as eager of a _Yes!_ from me. I do think it would be wise to communicate to the user that the clipboard failed, so I lean on a _Yes!_, but there is an argument to be made for throwing a warning in the logs.

I'm relatively new to the atuin codebase, so I have to explore how existing code does it, but if we find that it's sensible to flash a warning to the user and we have existing precedent, I can implement that.

markovejnovic:
Okay, had a look through the code. We pretty much have no way to nag the user -- best we can do is print to stderr for the user to read the message at the end of their interaction with `atuin search`. Seems like that's the path forwards.

markovejnovic:
Thank you for your contribution!

PR #2651, opened 2025-03-27
fix(zsh): fix an error introduced earilier with support for bracketed paste mode

lilydjwg:
An extra newline was written to the terminal and caused issues. See https://github.com/atuinsh/atuin/pull/2646#issuecomment-2757157442.

## Checks


ellie (approved):
thanks!

Issue #503, opened 2022-08-17
Feature: Highlighting search inclusions like fzf do


cab404:
That allows for easier judging on applicability of results and also gives you info whether your query is mistyped or not (which is really important in Atuin due to frequent mistypes, see #471 )

![image](https://user-images.githubusercontent.com/6453661/185037046-8f27361b-6d43-41ac-a233-369175bdaf80.png)

korniychuk:
Really waiting for this feature 👍

PR #2653, opened 2025-03-28
feat: highlight matches in interactive search

frankh:
fixes https://github.com/atuinsh/atuin/issues/503

highlight the matching parts of commands in the search window fzf style

sometimes it takes a while to be able to see why a command has shown up, this highlighting helps

it uses https://github.com/nomad/norm for fzf style matching which supports all the same prefixes we do _except_ regex matches (which just won't highlight)

image
image
image

NOTE: PR is unfinished (need tests, probably a better way of getting the search input and engine into the renderer), but i wanted to make sure people would want this feature first

also, the highlighting is kind of inversed for the selected row, not sure if there's a better way we can do this though

edit: updated the highlighting to change the color on the selected row instead of unbolding:
image

## Checks


frankh:
@ellie pls

ellie:
jusssst holding off on this one - i'd like to get 18.5 stable out this week and would like to give this some proper testing for 18.6! sorry for the delay

frankh:
no worries just wanted to make sure it wasn't forgotten :)

PR #2654, opened 2025-03-28
fix: add redundant clones to clippy and cleanup instances of it

frankh:
clippy::redundant_clone is pretty safe and helps cleanup some useless code, found a few instances of it in the codebase so though its worth adding to CI

## Checks


ellie (approved):
yaya thank you!

PR #2661, opened 2025-04-01
feat: Binaries as subcommands

BinaryMuse:
This PR implements binaries as subcommands, à la Git (and others).

When a user passes a subcommand that Atuin doesn't recognize, we attempt to run an executable called `atuin-<subcommand>`. If it isn't found, we output the standard "unrecognized subcommand" error message. Arguments are passed through to the external binary as expected, and the exit code becomes the exit code for the Atuin process.

```
➜ atuin update
Checking for updates...
Already up to date; not upgrading

➜ atuin update --help
Usage: atuin-update [OPTIONS]

Options:
      --tag <TAG>          Installs the specified tag instead of the latest version
      --version <VERSION>  Installs the specified version instead of the latest version
      --prerelease         Allows prereleases when just updating to "latest"
  -h, --help               Print help

➜ atuin asdf
error: unrecognized subcommand 'asdf' and no executable named 'atuin-asdf' found in your PATH

Usage: atuin <COMMAND>

For more information, try '--help'.
```

## Checks


crates/atuin/src/command/client/external.rs
```diff
@@ -0,0 +1,72 @@
...
+        Err(e) => {
+            println!("{}", e.ansi());
+            process::exit(1);
+        }
+    }
+}
+
+fn render_not_found(subcommand: &str, bin: &str) -> StyledStr {
+    let mut output = StyledStr::new();
+    let styles = Styles::styled();
```
BinaryMuse, line 46: @ellie Do we need to handle disabling colors if the user has them disabled, or does Clap handle that for us?

ellie:
(just tried clicking a new github ai button thing)

crates/atuin/src/command/client/external.rs
ellie: I believe clap should handles that part for us - should be easy to test

ellie (approved):
🚀

Issue #2556, opened 2025-01-21
Help with importing from per-directory-history


TC72:
I've been using per-directory-history and I'd like to bring my history over to Atuin.

I have written some simple python to create a csv which looks like this:
```
cwd, timestamp, command
/Volumes/Users/tc,1675194859,nvim
```

What's the simplest way for me to import something like this into the Atuin database?

I've looked around and see #2431 explains after updating history.db you can use `atuin history init-store` to update records.db.
I could import a csv file into history.db using the TablePlus app on my mac but it wouldn't handle fields like id and session.
(I realise I'd also need to change the timestamp to nanoseconds)

It would be great if Atuin had a simple csv import/export built in. I think it'd make getting started easier. I  also think people might be happier giving Atuin a try if they knew there was an easy way to get their history back if they needed to.

nvm, I see this gets very close to a csv export: `atuin history list --format "{time},{duration},{command},{directory}"`

markovejnovic:
Heyo! Thank you for the issue. We do not plan on adding support for importing CSV and we'd suggest have a look at the `gRPC` interface and import through that. We're discussing whether we want to provide some sort of language-specific bindings or not.

At this moment, we do not plan on adding support for CSV imports themselves, so I'll be closing this issue. I'll leave #2431 open in lieu.

Issue #2431, opened 2024-10-22
Programmatic import


pawamoy:
So I have been using my own [history tool](https://github.com/pawamoy/shell-history) for quite some time now, and have 220K+ commands in a file, in a custom format, that I'd like to import in Atuin :slightly_smiling_face:

How would I go about feeding Atuin programmatically, since surely the `import` subcommand will not support my history file format?

pawamoy:
Hmmm I guess I could simply check the database schema and insert things in it directly ^^

ellie:
It's not _quite_ that simple I'm afraid. We actually have two databases, and it's important to ensure that data is copied across them properly.

1. records.db: this stores your shell history data encrypted, on disk. It's used for syncing cross-machine, and is not at all editable manually
2. history.db: this contains your actual shell history data, in a format we can search and browse

I'd first disable auto-sync if you've already set it up. `auto_sync = false` in your config file.

There is a command that will take the data from history.db, and insert it into records.db. Once you've inserted to history.db, run:

```
atuin history init-store
```

This should be idempotent, and the process may happen automatically at sync time.

Once complete, I'd issue

```
atuin store rebuild history
```

This will wipe out history.db, and rebuild it from the records.db. Just to make sure everything is consistent!

Then, run

```
atuin sync
```

You should be good to go from here, and can re-enable auto_sync.

ellie:
also - love all the graphs from your tool!

pawamoy:
Awesome, thank you so much for the clear instructions! I'll try that and report back :slightly_smiling_face: 

> also - love all the graphs from your tool!

Thanks! I believe you have everything you need to offer similar graphs in Atuin too :smile: I could actually consider contributing if that's something you're interested in :relaxed:

vrozkovec:
I use also something home-baked and there is no easy way to import all records.
What about enabling import of a single command via something like:

`atuin history import --cmd "$command" --cwd "$cwd" --time "@$epoch"`

Then it would be possible for anyone to write their own import logic.
I guess one workaround could be also to convert my data to the history format of one of the shells atuin can import.

PR #2669, opened 2025-04-03
Feature csv import

wyattearp:
this provides a simple csv file import. after reviewing #2556 and #2431, it never seemed like there was an answer to this, so here's my take at a first cut. i will apologize in advance that i'm not super familiar with rust, so feedback is appreciated.

in keeping with the way other parts of the tool operate, we look for a `history.csv` file with the format of:

```csv
timestamp,duration,command
1613322469,10,cargo install atuin
..., ..., ...
```

the test provides an example. note that the class/module is called `csv_importer`, `CsvHistroyImporter` because I ran into conflicts with `csv` and the crate it belongs to, happy to take suggestions.

## Checks

### DIFF
--- a/crates/atuin-client/src/import/csv_history.rs
+++ b/crates/atuin-client/src/import/csv_history.rs
@@ -1,11 +1,11 @@
-use std::path::PathBuf;
 use std::fs::File;
-use std::io::{BufReader, BufRead};
+use std::io::{BufRead, BufReader};
+use std::path::PathBuf;
 
 use async_trait::async_trait;
+use csv::Reader;
 use eyre::{Result, eyre};
 use time::OffsetDateTime;
-use csv::Reader;
 
 use super::{Importer, Loader};
 use crate::history::History;
@@ -20,7 +20,9 @@ fn default_histpath() -> Result<PathBuf> {
     if histpath.exists() {
         Ok(histpath)
     } else {
-        Err(eyre!("Could not find history file. Please create 'history.csv' in the current directory."))
+        Err(eyre!(
+            "Could not find history file. Please create 'history.csv' in the current directory."
+        ))
     }
 }
 
@@ -45,8 +47,12 @@ impl Importer for CsvHistoryImporter {
 
         for result in reader.records() {
             let record = result?;
-            if let (Some(timestamp), Some(duration), Some(command)) = (record.get(0), record.get(1), record.get(2)) {
-                let timestamp = timestamp.parse::<i64>().ok()
+            if let (Some(timestamp), Some(duration), Some(command)) =
+                (record.get(0), record.get(1), record.get(2))
+            {
+                let timestamp = timestamp
+                    .parse::<i64>()
+                    .ok()
                     .and_then(|t| OffsetDateTime::from_unix_timestamp(t).ok())
                     .unwrap_or_else(OffsetDateTime::now_utc);
                 let duration = duration.parse::<i64>().map_or(-1, |t| t * 1_000_000_000);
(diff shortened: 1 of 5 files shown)


wyattearp:
for what it's worth, i've tested this as "works for me":

```
wyatt@lazarus ~/git_repos/atuin (feature-csv-import) $ cat target/release/history.csv 
timestamp,duration,command
112345,10,ls -als
223453,5,echo "hi"
wyatt@lazarus ~/git_repos/atuin (feature-csv-import) $ atuin history list | head -n 3
1970-01-02 03:12:25	ls -als	10s
1970-01-03 10:04:13	echo "hi"	5s
2024-04-03 20:43:52	ls	0s
```

crates/atuin-client/src/import/csv_history.rs
```diff
@@ -0,0 +1,105 @@
...
+
+use super::{Importer, Loader};
+use crate::history::History;
+
+const HISTORY_FILE: &str = "history.csv"; // Global constant for history file
+
+#[derive(Debug)]
+pub struct CsvHistoryImporter;
+
+fn default_histpath() -> Result<PathBuf> {
```
ellie, line 18: The `default_histpath` approach made sense for classic shell history import where there is usually an expected location for data, but I think in this case it isn't needed

Instead, we should make it so that the import csv subcommand takes a mandatory argument of the file to import

crates/atuin/src/command/client/import.rs
```diff
@@ -26,6 +27,8 @@ pub enum Cmd {
     ZshHistDb,
     /// Import history from the bash history file
     Bash,
+    /// Import history from user csv history file
+    CsvHistoryImporter,
```
ellie, line 31:
```suggestion
    Csv,
```

we have `atuin import bash`, `atuin import zsh` etc, so `atuin import zsh` makes more sense here

crates/atuin/src/command/client/import.rs
```diff
@@ -111,6 +114,7 @@ impl Cmd {
             Self::Zsh => import::<Zsh, DB>(db).await,
             Self::ZshHistDb => import::<ZshHistDb, DB>(db).await,
             Self::Bash => import::<Bash, DB>(db).await,
+            Self::CsvHistoryImporter => import::<CsvHistoryImporter, DB>(db).await,
```
ellie, line 117:
```suggestion
            Self::Csv => import::<CsvHistoryImporter, DB>(db).await,
```

see rename change

ellie (changes requested):
hey! thank you for the PR - couple of comments

crates/atuin/src/command/client/import.rs
wyattearp: will update

crates/atuin/src/command/client/import.rs
wyattearp: will update

crates/atuin-client/src/import/csv_history.rs
wyattearp: there's not a facility i'm seeing in any of the other importers to specify the location of a file as part of the command line parsing - am i missing it somewhere?

wyattearp: trying to dig more on this (excuse my rust n00bness) - it looks like it would be either a modification to the main Importer class, which means everything would get an option, but only one of them would support it, or i create a single "one off" function to basically mimic `async fn import` but takes a path.

are there better options i'm not seeing?

wyattearp: @ellie - any suggestions on the above comment?

markovejnovic:
Heyo! Thank you for this PR but since the PR is very old and stale, we do not plan on supporting importing from CSV. Feel free to ping me if you want me to re-open the issue. Thanks!

PR #2670, opened 2025-04-04
fix(zsh): avoid calling user-defined widgets when searching for history position

lilydjwg:
Or those widgets may behave badly when calling them too quickly. We don't need calling them anyway as this is not considered to be user action, and if our calls fail, the history position is unchanged.

Issue introduced in #1469.

## Checks


ellie (approved):

ellie:
thanks for following up!

PR #2673, opened 2025-04-06
[Temporary on v18.4.0 tag]: using tracing for theming debug

philtweir:
Improving alignment of the debug within theming with the rest of atuin.

Originally, I had just added `log` as I was not sure how best to link in, but this uses `tracing`, which is present elsewhere.

As such, you can see theming-related logging by using:

```
ATUIN_LOG=1 atuin search -i
```

For example:

Screenshot From 2025-04-06 18-53-42

Unfortunately, it is a bit hard to read on top of an interactive prompt, so if there is a simple workaround I am missing, please say.

## Checks

### DIFF
--- a/crates/atuin-client/Cargo.toml
+++ b/crates/atuin-client/Cargo.toml
@@ -76,6 +76,7 @@ palette = { version = "0.7.5", features = ["serializing"] }
 lazy_static = "1.4.0"
 strum_macros = "0.26.3"
 strum = { version = "0.26.2", features = ["strum_macros"] }
+tracing.workspace = true
 
 [dev-dependencies]
 tokio = { version = "1", features = ["full"] }
--- a/crates/atuin-client/src/theme.rs
+++ b/crates/atuin-client/src/theme.rs
@@ -1,6 +1,8 @@
 use config::{Config, File as ConfigFile, FileFormat};
 use lazy_static::lazy_static;
 use log;
+use tracing;
+use tracing::instrument;
 use palette::named;
 use serde::{Deserialize, Serialize};
 use serde_json;
@@ -55,6 +57,7 @@ use crossterm::style::{Color, ContentStyle};
 
 // For now, a theme is loaded as a mapping of meanings to colors, but it may be desirable to
 // expand that in the future to general styles, so we populate a Meaning->ContentStyle hashmap.
+#[derive(Debug)]
 pub struct Theme {
     pub name: String,
     pub parent: Option<String>,
@@ -120,6 +123,7 @@ impl Theme {
     // but we do not have this on in general, as it could print unfiltered text to the terminal
     // from a theme TOML file. However, it will always return a theme, falling back to
     // defaults on error, so that a TOML file does not break loading
+    #[instrument(skip(debug))]
     pub fn from_foreground_colors(
         name: String,
         parent: Option<&Theme>,
@@ -133,7 +137,7 @@ impl Theme {
                     *name,
                     StyleFactory::from_fg_string(color).unwrap_or_else(|err| {
                         if debug {
-                            log::warn!(
+                            tracing::warn!(
                                 "Tried to load string as a color unsuccessfully: ({}={}) {}",
                                 name,
                                 color,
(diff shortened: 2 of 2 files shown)


ellie:
btw you can filter tracing by specifying the module, eg

```
ATUIN_LOG=atuin_client=debug
```

to set logs to debug, only for atuin_client

philtweir:
Thanks @ellie ! If this is essentially the right approach, I'll tidy this PR on `main` to fix the theme logging (and maybe add a bit more debug-level output)

markovejnovic:
This seems like the right approach, @philtweir. We've made some large improvements to our logging and use tracing now across the board. I've also added some new profiling utilities and I've written out a guide in `CONTRIBUTING.md`:

### Profiling

The team mostly profiles Atuin through either instrumented profiling or
statistical profiling. For a quick guide on profiling, a solid resource is
[Mathieu Ropert's "The Basics of Profiling"
talk](https://www.youtube.com/watch?v=dToaepIXW4s).

#### Instrumented Profiling

For our instrumented profiling, we use
[`tracing`](https://docs.rs/tracing/latest/tracing/) and the opentelemetry
crates. These export [OTel](https://opentelemetry.io/) spans into
[Jaeger](https://www.jaegertracing.io/).

To get started , you **must** build Atuin from source, with the
`profiling-traced` profile **and** the matching `profiling-traced` feature (which
compiles in the OpenTelemetry exporter). The `cargo build-traced` alias enables
both:

```bash
cargo build-traced
# equivalently:
# cargo build -p atuin --profile profiling-traced --features profiling-traced
```

> [!NOTE]
>
> The `profiling-traced` feature **must** be enabled.

> [!WARNING]
> 
> The `profiling-traced` profile is **necessarily** slower than the published
> `dist` profile.
>
> It builds on the `profiling` profile, described later, and on top of that
> enables `tracing`-level spans, which add latency to function entry and exit.
>
> You will not be lucky in profiling low-level algorithms, memory-alignment,
> etc. this way. If you are optimizing a tight loop, the statistical approach
> (described further down) will be of help.

With this, you can run Jaeger, like so, in one terminal:

```bash
docker run --rm --name jaeger \
  -p 16686:16686 \
  -p 4318:4318 \
  jaegertracing/jaeger:latest
```

You can then, in another terminal, run the `daemon`:

```bash
ATUIN_OTEL=http://localhost:4318 ./target/profiling-traced/atuin daemon start
```

Finally, you can run the client:

```bash
ATUIN_OTEL=http://localhost:4318 ./target/profiling-traced/atuin sync
```

> [!NOTE]
>
> You may need to tweak your exporting rate with the following env values:
>
> ```bash
> OTEL_BSP_MAX_QUEUE_SIZE=262144      # max span count in the queue
> OTEL_BSP_MAX_EXPORT_BATCH_SIZE=8192 # how many spans to push a time
> OTEL_BSP_SCHEDULE_DELAY=500         # how often to publish spans
> ```

You can navigate to `http://localhost:16686` in your local browser and you
should see your traces.

> [!NOTE]
>
> Using OTEL feels heavy here, but there is no simpler approach available
> off-the-shelf. The author wanted to use
> [`tracing-chrome`](https://docs.rs/tracing-chrome/latest/tracing_chrome/),
> but the crate does not gracefully handle `async` futures, so it wasn't
> useful.

#### Statistical Profiling

Statistical profiling should use `cargo build --profile profiling`. Unlike
`profiling-traced`, this will compile-away the `trace`-level events/spans,
which is more representative and closer to the `--profile dist` which is
published.

> You can profile the application however you like. Some of the team members
> use [`Instruments.app`](https://developer.apple.com/tutorials/instruments).
> 
> [!NOTE]
> If you are using `Instruments.app`, the easiest way is to use
> `cargo-instruments`:
> 
> ```bash
> cargo install cargo-instruments
> RUSTFLAGS="-C llvm-args=--inline-threshold=25" \
>   cargo instruments -t time --profile profiling -p atuin --bin atuin -- search "hello world"
> ```

markovejnovic:
I will close this PR since the necessary utilities should already be in mainline. Thank you @philtweir!

PR #2674, opened 2025-04-07
[Project Fluent] Basic internationalization support (would close PR #2553)

philtweir:
Alternative approach to #2553. Not closing that just yet, but keen for feedback either way.

After thinking about the discussion on the [Forum](https://forum.atuin.sh/t/internationalization/733), the biggest challenge is that a lot of small changes brings risk. This is a suggestion:

1. finalize this PR that _only_ adds the machinery for Fluent (and an single "canary" string)
2. this [change](https://github.com/atuinsh/atuin/commit/697001768601d31c5ff4dfce3f281e865b18792d) sits on top to add give MVP coverage of translatable strings (feature/i18n-support-fluent-basic)
3. a much bigger set is here https://github.com/philtweir/fork-atuin/tree/feature/i18n-support-fluent

**Suggestion:** tidy-up, PR and merge this to provide i18n support. Once that is in a release and there is no negative feedback and folks confirm that the single example works OK, then PR 2., which is roughly the minimal set of common translatable strings. Less than that, and there's too much English leaking through to get buy-in, any more and the PR is oversized.

NB: there is a commit in 2. with English translations - note that it covers the full set of changes in 3.  so most of them can be ignored in the near-term (happy to split it in two parts - one for 2. and one for 3.).

*This PR does not add any translatable strings except one*. Changes 2. and 3. only translate the crates `atuin-common` and `atuin-client`, plus `inspector.rs` and `interactive.rs` in the `atuin` crate.

### How to test

This PR has only one translatable string, to make sure it works and breaks nothing, as the main challenge is to get the machinery in without regressions. However, using https://github.com/philtweir/fork-atuin/tree/feature/i18n-support-fluent-basic, you can build and take the interactive search as a minimal test.

In this PR, the only visible change should be:

```
$ LANG=en-US atuin asdf
error: unrecognized subcommand '⁨asdf⁩' and no executable named 'atuin-asdf' found in your PATH

Usage: atuin <COMMAND>

For more information, try '--help'.
```

vs

```
$ LANG=en-GB atuin asdf
error: unrecognised subcommand '⁨asdf⁩' and no executable named 'atuin-asdf' found in your PATH

Usage: atuin <COMMAND>

For more information, try '--help'.
```

### How translatability is implemented

Adding `t!(...)` everywhere appropriate, in some places accounting for static return values, and then progressively adding translations to `*.flt` files as people wish to (separately) PR.

## Notes

* as `t!(...)` can take arguments (e.g. translate `%{time} ago`) it does not return `&'static str`   so, for FilterMode, I used `lazy_static` and elsewhere, switched to `String` or otherwise worked-around.
* originally I leaned to rust-i18n as Project Fluent wants slug-like keys, and that would make PRs much bigger and hard to review, but this approach creates a more traditional gettext-like macro, which is less invasive (as recommended in the `cargo-i18n` docs) - **this means a new `atuin_macro` package** (if this is a hard no, please say and I will try new ideas!)
* I had some difficulty getting the crates to behave consistently independently (which was not previously an issue) but I have now centralized translation strings in a top-level `./i18n` folder (this has a side-benefit/downside that the same string will be translated the same way in all crates)

## Checks

### DIFF
--- a/crates/atuin-common/i18n.toml
+++ b/crates/atuin-common/i18n.toml
@@ -2,7 +2,7 @@
 # source code for gettext system, and the primary fallback language
 # (for which all strings must be present) when using the fluent
 # system.
-fallback_language = "en-US"
+fallback_language = "en-GB"
 
 # Use the fluent localization system.
 [fluent]
--- a/crates/atuin-common/i18n.toml
+++ b/crates/atuin-common/i18n.toml
@@ -1,0 +1,13 @@
+# (Required) The language identifier of the language used in the
+# source code for gettext system, and the primary fallback language
+# (for which all strings must be present) when using the fluent
+# system.
+fallback_language = "en-US"
+
+# Use the fluent localization system.
+[fluent]
+domain = "atuin"
+# (Required) The path to the assets directory.
+# The paths inside the assets directory should be structured like so:
+# `assets_dir/{language}/{domain}.ftl`
+assets_dir = "../../i18n"
--- a/crates/atuin-macro/Cargo.toml
+++ b/crates/atuin-macro/Cargo.toml
@@ -31,7 +31,7 @@ getrandom = "0.2"
 sys-locale = "0.3.2"
 
 lazy_static = "1.4.0"
-i18n-embed = { version = "0.15.3", features = ["fluent", "fluent-system", "tr", "locale_config", "walkdir", "filesystem-assets"] }
+i18n-embed = { version = "0.15.3", features = ["fluent", "fluent-system", "tr", "locale_config", "desktop-requester", "walkdir", "filesystem-assets"] }
 rust-embed = "8"
 i18n-embed-fl = "0.9.3"
 slugify = "0.1.0"
--- a/crates/atuin-macro/Cargo.toml
+++ b/crates/atuin-macro/Cargo.toml
@@ -31,7 +31,7 @@ getrandom = "0.2"
 sys-locale = "0.3.2"
 
 lazy_static = "1.4.0"
-i18n-embed = { version = "0.15.3", features = ["fluent", "fluent-system", "tr", "locale_config", "desktop-requester", "walkdir", "gettext-system", "autoreload", "filesystem-assets", "notify"] }
+i18n-embed = { version = "0.15.3", features = ["fluent", "fluent-system", "tr", "locale_config", "walkdir", "filesystem-assets"] }
 rust-embed = "8"
 i18n-embed-fl = "0.9.3"
 slugify = "0.1.0"
--- a/crates/atuin-macro/Cargo.toml
+++ b/crates/atuin-macro/Cargo.toml
@@ -1,0 +1,41 @@
+[package]
+name = "atuin-macro"
+edition = "2021"
+description = "macro library for atuin"
+
+rust-version = { workspace = true }
+version = { workspace = true }
+authors = { workspace = true }
+license = { workspace = true }
+homepage = { workspace = true }
+repository = { workspace = true }
+
+[lib]
+proc-macro = true
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+time = { workspace = true }
+serde = { workspace = true }
+uuid = { workspace = true }
+typed-builder = { workspace = true }
+eyre = { workspace = true }
+sqlx = { workspace = true }
+semver = { workspace = true }
+thiserror = { workspace = true }
+directories = { workspace = true }
+sysinfo = "0.30.7"
+base64 = { workspace = true }
+getrandom = "0.2"
+sys-locale = "0.3.2"
+
+lazy_static = "1.4.0"
+i18n-embed = { version = "0.15.3", features = ["fluent", "fluent-system", "tr", "locale_config", "desktop-requester", "walkdir", "gettext-system", "autoreload", "filesystem-assets", "notify"] }
+rust-embed = "8"
+i18n-embed-fl = "0.9.3"
+slugify = "0.1.0"
+paste = "1.0.15"
+syn = "2.0.96"
+quote = "1.0.38"
+proc-macro2 = "1.0.93"
--- a/crates/atuin-macro/i18n.toml
+++ b/crates/atuin-macro/i18n.toml
@@ -1,0 +1,13 @@
+# (Required) The language identifier of the language used in the
+# source code for gettext system, and the primary fallback language
+# (for which all strings must be present) when using the fluent
+# system.
+fallback_language = "en-GB"
+
+# Use the fluent localization system.
+[fluent]
+domain = "atuin"
+# (Required) The path to the assets directory.
+# The paths inside the assets directory should be structured like so:
+# `assets_dir/{language}/{domain}.ftl`
+assets_dir = "./tests/i18n"
--- a/crates/atuin-macro/src/lib.rs
+++ b/crates/atuin-macro/src/lib.rs
@@ -2,42 +2,68 @@
 extern crate proc_macro;
 
 use proc_macro::TokenStream;
+use proc_macro2::TokenStream as TokenStream2;
 use quote::quote;
 use slugify::slugify;
 use syn::parse::Parser;
 
+fn literal_to_slug(literal: &syn::ExprLit) -> TokenStream2 {
+    // We pull out the actual text from the literal string expression.
+    let quoted: String = match &literal.lit {
+        syn::Lit::Str(message_id) => message_id.value(),
+        _ => panic!("Message ID must be a literal string"),
+    };
+    // ...and pass it to slugify,
+    let slug = slugify!(quoted.as_str());
+    // ...then turn it back into a literal string.
+    quote!(#slug)
+}
+
 #[proc_macro]
 pub fn tl(tokens: TokenStream) -> TokenStream {
+    // Begin by getting the individual arguments to tl!
     let args = syn::punctuated::Punctuated::<syn::Expr, syn::Token![,]>::parse_terminated
         .parse(tokens)
         .unwrap();
+
     let mut arg_iter = args.iter();
+
+    // The first should always be the fl! macro for fluent (cf. atuin-common/src/i18n.rs)
     let fl = arg_iter.next().unwrap();
+    // atuin-common will send the universal loader as the second argument. This avoids
+    // every translation string having to explicitly pass it.
     let loader = arg_iter.next().unwrap();
+
+    // The third argument should be the message ID. This logic takes the human-readable
+    // string and slugifies it. One of the main benefits of Fluent is that English-language
+    // ASCII is not the de facto reference (and things like gender and plurality can be
+    // encoded even where English makes no grammatical distinction). However, this approach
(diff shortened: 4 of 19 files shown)


crates/atuin/src/command/external.rs
```diff
@@ -54,7 +54,11 @@ fn render_not_found(subcommand: &str, bin: &str) -> StyledStr {
     let _ = write!(output, "{error}error:{error:#} ");
     let _ = write!(
         output,
-        "unrecognized subcommand '{invalid}{subcommand}{invalid:#}' "
+        "{} ",
```
philtweir, line 57: This is an awkward example but maybe helps show where extra steps are required (e.g. a simple dynamic string would be wrapped by `t!(..)` and done)

crates/atuin-macro/Cargo.toml
```diff
@@ -0,0 +1,41 @@
+[package]
+name = "atuin-macro"
+edition = "2021"
+description = "macro library for atuin"
+
+rust-version = { workspace = true }
+version = "18.5.0-beta.2"
```
philtweir, line 7: TODO: this seems to be a cargo-i18n issue that it cannot accept non-literal versions.

That said, running `cargo i18n` doesn't _actually_ seem to be necessary.

crates/atuin-common/i18n.toml
```diff
@@ -0,0 +1,13 @@
+# (Required) The language identifier of the language used in the
+# source code for gettext system, and the primary fallback language
+# (for which all strings must be present) when using the fluent
+# system.
+fallback_language = "en-GB"
```
philtweir: TBC :)

crates/atuin-macro/Cargo.toml
philtweir: But might need an upstream change in cargo-i18n to fix :/

crates/atuin-client/Cargo.toml
```diff
@@ -4,7 +4,7 @@ edition = "2024"
 description = "client library for atuin"
 
 rust-version = { workspace = true }
-version = { workspace = true }
+version = "18.5.0-beta.2"
```
philtweir, line 7: (see below: if it is not easy to address the `cargo-i18n` issue then I'm fairly sure we can ignore it - either way will restore these before merge)

crates/atuin-common/i18n.toml
```diff
@@ -0,0 +1,13 @@
+# (Required) The language identifier of the language used in the
+# source code for gettext system, and the primary fallback language
+# (for which all strings must be present) when using the fluent
+# system.
+fallback_language = "en-US"
```
ellie, line 5: part of me would love to force en-GB 😏

ellie (commented):
generally pretty happy with the direction here - thanks for taking it on!

crates/atuin-common/i18n.toml
philtweir: I'll be honest, it was en-GB until a commit or two ago, so ... :D 

That said, it may be an gentle way to encourage a wide sample, as more people will spot that is falling back to -ise spellings instead of correctly picking up US locales, than the other way around 🤔

crates/atuin-macro/src/lib.rs
```diff
@@ -0,0 +1,52 @@
...
+extern crate proc_macro;
+
+use proc_macro::TokenStream;
+use quote::quote;
+use syn;
+use syn::parse::Parser;
+use slugify::slugify;
+
+#[proc_macro]
+pub fn tl(tokens: TokenStream) -> TokenStream {
```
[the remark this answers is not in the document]
philtweir: Of course - will do! This was the main point against fluent-rs, alongside activity, as their macro (for conceptual reasons) requires slug-strings in the codebase rather than human readable ones, so using it means one of:

1. a wrapper macro to go from human-readable to slugs
2. switching all translatable strings to slugs
3. another idea I had not thought of

2 has the downside of making things like making UI bugs harder to see in reviews, and grepping for error messages harder, as well as more visual change to existing code when making strings translatable, so I thought that was probably too much friction. On the other hand, as mentioned, 1 adds custom magical machinery, a new macro module, and somewhat undermines fluent's own differentiating motivation from gettext (although, of course, it is still possible to use it in the intended way on a case-by-case basis).

However, one other benefit of fluent is that it is a standard and seems to be supported by (for example) [Weblate](https://docs.weblate.org/en/latest/formats/fluent.html) - it might even be possible to get a free instance on their libre-project plan, if desired.

philtweir:
Think that should be a first version - happy to add some more content to docs, etc. but I guess that might be more appropriate for the second PR once there is more than one translated string?

philtweir:
> Think that should be a first version - happy to add some more content to docs, etc. but I guess that might be more appropriate for the second PR once there is more than one translated string?

Just to check in @ellie - happy to look at splitting up further to make review easier, but keen to make sure you are still happy with the overall direction before I rebase onto main?

markovejnovic:
Hey @philtweir! Thank you so much for your PR. I talked with @ellie and we're still uncertain what the exact future of localization will look like. We're interested in finding some docs maintainers first (as that has already diverged), and when we confirm that we can keep our docs up to date, we'll figure out what to do with the app.

Nowadays we have great AI agents that can translate stuff very quickly, but I do not feel comfortable releasing that until we get eyes on the docs.

Either way, regarding the app, I think mainline has sufficiently diverged to the point that we'd need to re-implement i18n from scratch. Sorry to be closing your PR, and I'm very grateful for your work.

markovejnovic:
Reopening this as part of https://forum.atuin.sh/t/internationalization/733/10?u=suis_siva. Will give some direct feedback soon.

@philtweir, would you be open to me rebasing your PR?

philtweir:
Thanks @markovejnovic - appreciate the reconsideration. Will get back on the forum post also but just to acknowledge here.

Happy for you to do so - equally, I should have a chance on Sunday, and it probably would be wise for me to get up to speed on any relevant changes to main since the last rebase, so more than happy to on this end also. That said, if you'd rather crack on, no objections from me :)

markovejnovic:
Would be my pleasure @philtweir, thanks for the response! I'll rebase real fast and then provide some implementation feedback =)

philtweir:
Thanks for the rebase @markovejnovic - if you've got implementation feedback, that would be handy, as I'll take a look at the test fails shortly and do a self-review

markovejnovic:
Sorry for the late response @philtweir, this is a little bit of a back-burner task for me and I've been in a little bit of feature hell, so I haven't been able to get to this sooner. Going to do another rebase. I'll get you some more feedback today.

crates/atuin-macro/src/lib.rs
```diff
@@ -0,0 +1,76 @@
+#![forbid(unsafe_code)]
```
markovejnovic, line 1: I would probably avoid naming this crate `atuin-macro` as I fear it will become a dumping ground of random, unrelated macros.

I would consider renaming this crate to be `atuin-i18n` and have it specifically serve i18n purposes.

markovejnovic, line 1: A fear I have with this implementation is how it will end up working with `thiserror` and clap's documentation parsing.

crates/atuin-macro/Cargo.toml
```diff
@@ -0,0 +1,41 @@
...
+sys-locale = "0.3.2"
+
+lazy_static = "1.4.0"
+i18n-embed = { version = "0.15.3", features = ["fluent", "fluent-system", "tr", "locale_config", "desktop-requester", "walkdir", "filesystem-assets"] }
+rust-embed = "8"
+i18n-embed-fl = "0.9.3"
+slugify = "0.1.0"
+paste = "1.0.15"
+syn = "2.0.96"
+quote = "1.0.38"
```
markovejnovic: The newer pattern we've been establishing is to put all dependencies in the top-level `Cargo.toml` and reference them in nested crates via `{ workspace = true }`

crates/atuin-macro/Cargo.toml
```diff
@@ -0,0 +1,41 @@
...
+proc-macro = true
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+time = { workspace = true }
+serde = { workspace = true }
+uuid = { workspace = true }
+typed-builder = { workspace = true }
+eyre = { workspace = true }
```
markovejnovic: I would avoid adding `eyre`. @taylordotfish and I have been pushing really hard on removing our dependency on `eyre` as we believe there are two large issues with it:

- `eyre` inherently makes it difficult to match on different kinds of errors. Since errors stemming from `eyre` are not algebraic, matching on them is impossible.
- `eyre` promotes opaque and anti-pattern errors that are not defined at the interface of a function. Glancing at a function which returns a `Result<T, E: thiserror::Error>`, I can quickly look at `E` to see what possible errors the function throws. With eyre, the error is opaque at the interface.
- `eyre` has questionable performance characteristics since errors are by necessity `Box`es which means they lose on performance.

crates/atuin-macro/i18n.toml
```diff
@@ -0,0 +1,13 @@
+# (Required) The language identifier of the language used in the
+# source code for gettext system, and the primary fallback language
+# (for which all strings must be present) when using the fluent
+# system.
+fallback_language = "en-GB"
```
markovejnovic, line 5:
```suggestion
fallback_language = "en-US"
```

Would check with @ellie to see if she agrees with this, but Atuin is currently registered in the US and the `en-US` localization is what most software users expect (headcannon).

crates/atuin/src/command/external.rs
markovejnovic: Would it be possible to make a `writet!`? We could use clippy to ban the use of `write!` and replace all the calls with `writet!` instead.

i18n/ga-IE/atuin.ftl
```diff
@@ -0,0 +1,2 @@
+unrecognized-subcommand-subcommand =
+  ordú neamhaitheanta '{ $subcommand }'
```
markovejnovic, line 2: I assume you're a speaker of Gaelic, @philtweir =)?

crates/atuin-macro/tests/basic.rs
```diff
@@ -0,0 +1,123 @@
...
+        let language_loader: FluentLanguageLoader = fluent_language_loader!();
+        let requested_languages = vec!["en-GB".parse().unwrap()];
+
+        let _result = i18n_embed::select(
+            &language_loader, &Localizations, &requested_languages);
+        language_loader
+    };
+}
+
+#[test]
```
markovejnovic, line 24: We've been moving towards using `rstest` and `proptest`. I'd suggest using that in cases here, since it would allow you to simplify a lot of these cases. `rstest` will let you write fixtures and cases that will make it so that a lot of the test setup code ends up being de-duplicated here.

markovejnovic:
I thought about this a little bit and I feel like the crate you've added here is perhaps not really necessary. I've opened a counter-PR at #4233 which directly uses Fluent as a starting point. I find it to be slightly easier and simpler to reason about.

I'll keep your PR open for the time being, but that PR looks like it will land.

markovejnovic (commented):
Posting some of my feedback!

PR #2696, opened 2025-04-12
Updates to config.toml

psalveson:
Commented out an uncommented line, and added store_failed to config.toml

## Checks


crates/atuin-client/config.toml
```diff
@@ -197,6 +172,10 @@ enter_accept = true
 ## Alternatively, set env NO_MOTION=true
 # prefers_reduced_motion = false
 
+## Defaults to true. Configures whether to store commands that failed (those with non-zero 
+## exit status) or not.
```
markovejnovic:
```suggestion
## Defaults to true. Configures whether to store commands that failed (those with non-zero exit status) or not.
```

Would fold this into one line since the convention appears to be `tw=140`.

markovejnovic:
Thank you for your PR!

PR #2697, opened 2025-04-13
feat: delete duplicate history

lilydjwg:
See also #2015.

## Checks
- [x] I am happy for maintainers to push small adjustments to this PR, to speed up the review cycle
- [x] I have checked that there are no existing pull requests for the same thing

Issue #2015, opened 2024-05-13
atuin history gc?


AlJohri:
I seem to have accidentally re-run `atuin import zsh` after already using atuin for a period of time. I installed atuin on a new computer and kept syncing and was getting 0 results. I wasn't sure what was going on until I realized that the new installation was using the sync v2. I have synced upgraded all installations to use sync v2, however as part of trying to figure out what was going on, I had re-run the `atuin import zsh` command and then run `atuin sync`. This appears to have doubled the size of my history from 150k records to 300k records.

Is there anyway I can "garbage collect" down to a list of unique entries and clean up my atuin history?

AlJohri:
It seems like its taking almost 0.5 to 1 second when I press `CTRL-R` (not sure how to check exactly). Would love to figure out some way to trim some of this history down or speed up the TUI so `CTRL-R` is instantaneous.

As a proxy for timing, it takes about 5 seconds to run `atuin history list`:

```
❯ atuin history list | wc -l
366974

❯ time atuin history list &> /dev/null
atuin history list &> /dev/null  4.87s user 2.81s system 172% cpu 4.448 total
```

AlJohri:
Looks like my slowdown is likely related to getting the unique set of commands: See https://github.com/atuinsh/atuin/issues/475

zuzzurro:
I have the same issue. I am actually running "atuin import zsh" multiple times in order to incrementally save my zsh history before I actually really start using atuin and commands are added multiple times. Some sort of deduplication based on the time and command would be extremely appreciated.

ellie:
@AlJohri in response to your issue, I'd suggest trying to craft a `atuin search --delete` that specifically targets your newly-imported data. You should be able to delete all records after a certain date.

If that is not possible, then I'm afraid the only real way forwards would be to delete and start from scratch :/

Due to how the importers work at the moment, it's not that straightforward to correctly identify duplicate imports vs commands that have been ran a few times.

Otherwise, if you're able to profile the startup/searching with your data, I'd really appreciate it! You may find that the prefix search mode performs much faster.

>  I am actually running "atuin import zsh" multiple times in order to incrementally save my zsh history before I actually really start using atuin and commands are added multiple times.

@zuzzurro I'm afraid this use case isn't supported. Importing is for migrating old data to Atuin, and not for keeping Atuin up-to-date. Given that the original data soruce doesn't always have enough information to effectively de-dupe, it's not that straightforward.

[...]

PR #2698, opened 2025-04-13
improve search performance by scanning with the timestamp index iteratively

lilydjwg:
Or else SQLite uses the `command` index and sort. Even with the `timestamp` index, SQLite still needs sort.

This is what atuin uses:
```
select * from history indexed by idx_history_command where deleted_at is null group by command having max(timestamp) order by timestamp desc limit 100;
|--SCAN history USING INDEX idx_history_command
`--USE TEMP B-TREE FOR ORDER BY
```

Try to tell SQLite to use the `timestamp` index. Slightly faster.
```
select * from history indexed by idx_history_timestamp where deleted_at is null group by command having max(timestamp) order by timestamp desc limit 100;
|--SCAN history USING INDEX idx_history_timestamp
|--USE TEMP B-TREE FOR GROUP BY
`--USE TEMP B-TREE FOR ORDER BY
```

The fastest index to use is...not to use an index at all.
```
select * from history not indexed where deleted_at is null group by command having max(timestamp) order by timestamp desc limit 100;
QUERY PLAN
|--SCAN history
|--USE TEMP B-TREE FOR GROUP BY
`--USE TEMP B-TREE FOR ORDER BY
```

The following one is very fast, but it might not fetch enough rows due to duplications. So we do the deduplication ourselves instead and let SQLite just scan using the index and never sort. The downside of this method is that it doesn't perform well when there are too many duplicates, hence #2697.
```
select * from (select * from history where deleted_at is null order by timestamp desc limit 1000) group by command having max(timestamp);
QUERY PLAN
|--CO-ROUTINE (subquery-1)
|  `--SCAN history USING INDEX idx_history_timestamp |--SCAN (subquery-1)
`--USE TEMP B-TREE FOR GROUP BY
```

For the following command, the elapsed times are (lowest among several runs):
```sh
atuin search --search-mode fuzzy --limit 100 --cmd-only > /dev/null
```
```
command index   : 0.154s
timestamp index : 0.139s
no index        : 0.124s
paging          : 0.012s
```

See also #475, #493, #2015.

----

BTW, the `idx_history_command` index is unnecessary since there is `idx_history_command_timestamp`.

The SQL atuin used is something like:

```sql
SELECT * FROM history WHERE (deleted_at IS NULL) GROUP BY command HAVING max(timestamp) ORDER BY timestamp DESC LIMIT 100;
```

The `HAVING max(timestamp)` part doesn't seem right. [`SELECT`](https://sqlite.org/lang_select.html#resultset) says:

> If a HAVING clause is specified, it is evaluated once for each group of rows as a boolean expression. If the result of evaluating the HAVING clause is false, the group is discarded. If the HAVING clause is an aggregate expression, it is evaluated across all rows in the group.

So `HAVING max(timestamp)` is true all the time, and we are not sure which row is outputted. In the following section it says

> Special processing occurs when the aggregate function is either min() or max().

But it refers to the columns being `select`ed, not in the `HAVING` clause.

## Checks

### DIFF
--- a/crates/atuin-client/src/database.rs
+++ b/crates/atuin-client/src/database.rs
@@ -618,10 +618,10 @@ impl Database for Sqlite {
             .map(|before| before.unix_timestamp_nanos() as i64)
         });
         if filter_options.include_duplicates || filter_options.limit.is_none() {
-            return self.search_inner(
-                search_mode, filter, context, query, filter_options,
-                before,
-            ).await;
+            let result = self
+                .search_inner(search_mode, filter, context, query, filter_options, before)
+                .await?;
+            return Ok(ordering::reorder_fuzzy(search_mode, query, result));
         }
 
         let mut result = vec![];
@@ -631,10 +631,16 @@ impl Database for Sqlite {
         filter_options.limit = Some(page_size as i64);
         filter_options.include_duplicates = true;
         while result.len() < limit {
-            let res = self.search_inner(
-                search_mode, filter, context, query, filter_options.clone(),
-                before,
-            ).await?;
+            let res = self
+                .search_inner(
+                    search_mode,
+                    filter,
+                    context,
+                    query,
+                    filter_options.clone(),
+                    before,
+                )
+                .await?;
             let over = res.len() < page_size;
             if let Some(last) = res.iter().last() {
                 before = Some(last.timestamp.unix_timestamp_nanos() as i64);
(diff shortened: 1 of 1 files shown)


ellie:
closed in favour of #3553 - that approach is much simpler, and more obviously an improvement in all cases

lilydjwg:
> that approach is much simpler

My brain can't agree with this: mine has one `SELECT` and 33 instructions but that SQL has four `SELECT`s and 71 instructions, and it uses two indexes! But yay, it is faster due to utilizing both indexes.

Issue #2700, opened 2025-04-14
[Bug]: `[search]` `filters` behaviour inconsistent when used with `workspace` in filter


merc1031:
### What did you expect to happen?

[1]
```
workspaces = true

[search]
filters = ["workspace", "host"]
```
In non git directory, pressing `ctrl+r` expected `host` to come up first 

[2]
```
workspaces = true

[search]
filters = ["workspace", "host", "directory"]
```

In non git directory, pressing `ctrl+r` expected `host` to come up first , then `directory`, then `host`, then `directory` ...

[3]
Works as expected in a git directory

### What happened?

[1] instead `global` comes up, then `host`, then further presses do nothing
[2] instead `global` comes up, then `host`, then `directory` , then `host`, then `directory` ...
[3] pressing `ctrl+r` expected `workspace` to come up first, `host` to come up next , then `directory`, then `worksapce`, then `host` ...

### Atuin doctor output

```yaml
{
  "atuin": {
    "version": "18.5.0",
    "sync": {
      "cloud": false,
      "records": true,
      "auto_sync": true,
      "last_sync": "2025-04-14 20:10:41.509478585 +00:00:00"
    },
    "sqlite_version": "3.46.0"
  },
  "shell": {
    "name": "zsh",
    "default": "zsh",
    "plugins": [
      "atuin"
    ],
    "preexec": "built-in"
  },
  "system": {
    "os": "Ubuntu",
    "arch": "x86_64",
    "version": "24.04",
    "disks": [
      {
        "name": "/dev/mapper/ubuntu--vg-ubuntu--lv",
        "filesystem": "ext4"
      },
      {
        "name": "/dev/nvme0n1p2",
        "filesystem": "ext4"
      },
      {
        "name": "/dev/nvme0n1p1",
        "filesystem": "vfat"
      }
    ]
  }
}
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

merc1031:
This is my first time looking at atuin's source, but ill give a try to fix this.

PR #2703, opened 2025-04-16
fix: More thorough search through search.filters w/ workspaces

merc1031:
Do not just pick first from search.filters, consider all options, filtering based on git-root and workspaces config

atuinsh/atuin#2700

Wrote a few simple unit tests, and tried it by running the client locally. Not sure about which of "[search] filters" vs "workspaces" should be more respected, so i made it so if workspaces = false, regardless of whats in "[search] filters" workspace gets filtered out.

First time looking at atuins code, let me know if this needs any changes.

## Checks
  - #1655 might be related, but it seems to also implement code that is already merged, and has been open for over a year, hope its ok to suggest this

### DIFF
--- a/crates/atuin-client/src/settings.rs
+++ b/crates/atuin-client/src/settings.rs
@@ -719,10 +719,20 @@ impl Settings {
         None
     }
 
-    pub fn default_filter_mode(&self) -> FilterMode {
+    pub fn default_filter_mode(&self, git_root: bool) -> FilterMode {
         self.filter_mode
             .filter(|x| self.search.filters.contains(x))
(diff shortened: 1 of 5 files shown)


merc1031:
@ellie Totally understand if you are busy, but is there any codeowner who can look this over and accept or reject?
I have been using a custom compiled binary and it appears to fix the issue that was shown.

Anyway thanks for the amazing tool

merc1031:
@ellie @BinaryMuse  Just bumping since its been a couple of months. Ive been running this locally since I opened the PR and it appears to fix the issue linked at least.

PR #2706, opened 2025-04-18
fix: selection vs render issue

### LINKED ISSUE #2704: [Bug]: wrong item selected from history

macOS 15.4.1 and atuin 18.5.0

when I press ctrl+r to find history, select item from history wrong item is pasted to the command prompt. 

Can't always reproduce. Usually I do this quickly, ctrl+r, select what I need (by searching, etc) and press enter. 
Few times noticed, wrong command was pasted into prompt from history.  

This happened on at least two occasions just today, which already drove me a bit crazy. When specifically tried, can't reproduce. 

### What happened?

Wrong record was pasted into command prompt from history. let's say I wanted to select line from history where I meant to rerun command to copy file, but instead rm -rf /path which I didn't mean to delete was run. 

### Atuin doctor output

```yaml
$ atuin doctor
Atuin Doctor
Checking for diagnostics


{
  "atuin": {
    "version": "18.5.0",
    "sync": null,
    "sqlite_version": "3.46.0"
  },
  "shell": {
    "name": "zsh",
    "default": "zsh",
    "plugins": [
      "atuin"
    ],
    "preexec": "built-in"
  },
  "system": {
    "os": "Darwin",
    "arch": "arm64",
    "version": "15.4.1",
    "disks": [
      {
        "name": "REDACTED",
        "filesystem": "apfs"
      },
      {
        "name": "REDACTED",
        "filesystem": "apfs"
      },
      {
        "name": "REDACTED",
        "filesystem": "apfs"
      }
    ]
  }
}
```

### Code of Conduct

ellie:
Edit: after testing, I don't think this resolves the issue. HOWEVER I would like to make this change regardless

I can't replicate that issue just yet so i have _no idea_

It could be a classic race condition. When the UI selects an entry, it only stores the index of the selected item, not the actual entry itself. Then later, when it comes time to actually return the command, it goes back to the results array and grabs whatever is at that index. This is fine most of the time...

BUT - if anything changed the results array between selection and execution (like a new search query being processed), the index would still point to the same position, but that position might now contain a completely different history entry! 🫠

There's another potential cause too. We have the main loop, that handles calling draw + also updating the results when needed. But an inner loop, that processes all input events received. If a user sends a lot of events (maybe holds the arrow key?), then there could be a discrepancy between the selected state and the currently rendered state. Ensure we also call draw when we process the Continue event.

## Checks

### DIFF
--- a/crates/atuin/src/command/client/search/interactive.rs
+++ b/crates/atuin/src/command/client/search/interactive.rs
@@ -49,7 +49,7 @@ use ratatui::{
 const TAB_TITLES: [&str; 2] = ["Search", "Inspect"];
 
 pub enum InputAction {
-    Accept(usize),
+    Accept(usize, History),
     Copy(usize),
     Delete(usize),
     ReturnOriginal,
@@ -110,13 +110,14 @@ impl State {
         settings: &Settings,
         input: &Event,
         w: &mut W,
+        results: &[History],
     ) -> Result<InputAction>
     where
         W: Write,
     {
         execute!(w, EnableMouseCapture)?;
         let r = match input {
-            Event::Key(k) => self.handle_key_input(settings, k),
+            Event::Key(k) => self.handle_key_input(settings, k, results),
             Event::Mouse(m) => self.handle_mouse_input(*m),
             Event::Paste(d) => self.handle_paste_input(d),
             _ => InputAction::Continue,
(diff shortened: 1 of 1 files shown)


ellie:
yeah gonna end up with probably 1 more clone than before (only when selecting history, so negligable) and code that makes a bit more sense + is less likely to hide issues

PR #2714, opened 2025-04-21
Add Windows builds to the CI

ltrzesniewski:
This adds Windows builds to the CI, which will be useful for PowerShell support (#2543), but I wanted to submit this PR separately as a few Clippy issues are currently present on Windows and this will help detect them from now on.

I need to point out that this could slow down builds a bit, but the Windows runners just seem slower than the Ubuntu ones, sorry about that.

## Checks


ellie:
Thanks for opening this one, but currently I'm not willing or able to support Windows users. If it builds there + can be used then _amazing_. I'm happy to accept PRs fixing windows issues too, but it is not currently a goal.

ltrzesniewski:
Ok, no worries. Currently everything builds fine on Windows, there are just a few "unused import" warnings which are fixed in the 2nd commit of this PR.

PR #2715, opened 2025-04-21
fix: zsh + omz selecting the wrong item from history

### LINKED ISSUE #2704: [Bug]: wrong item selected from history

ellie:
Resolves #2704 

After seeing a reproduction case that @BinaryMuse found, it seems like the error is occurring in the shell. Opening a new shell stops it from happening

This change reverts all changes to atuin.zsh since 18.3.0

## Checks

### DIFF
--- a/crates/atuin/src/shell/atuin.zsh
+++ b/crates/atuin/src/shell/atuin.zsh
@@ -71,8 +71,6 @@ _atuin_search() {
         then
             LBUFFER=${LBUFFER#__atuin_accept__:}
             zle accept-line
-        else
-            zle .infer-next-history && zle .up-history
         fi
     fi
 }


ellie:
Awesome! I believe that change is part of #2715?

ellie:
rebased so we only have the revert that caused the issue

PR #2719, opened 2025-04-25
feat: sort `atuin store status` output

printfn:
## Checks

I was trying to debug a sync issue and found it quite annoying that the output from `atuin store status` was not sorted.
I also noticed that it was always printing timestamps in UTC so I fixed that as well.


ellie (approved):
looks good to me! thank you

PR #2732, opened 2025-05-02
feat: Implement KV as a write-through cache

BinaryMuse:
We've had reports from users observing long run times from `atuin kv list` and `atuin kv get`; this can happen when the record store has lots of KV entries, even if the actual number of unique keys is not that high (e.g. from automatically creating entries regardless of the current value). Both the aforementioned commands built a map by scanning the entire record store each time.

This PR changes this so that the KV store is a write through cache. During sync (and manually with `atuin kv rebuild`), we iterate over the record store and build up the cache, which we then use to query from. `set` and `delete` calls write to both the cache and the record store.

* Introduce the `atuin-kv` crate
* Introduce a new database file, `kv.db`, the location of which can be overridden with `ATUIN_KV__DB_PATH` or set via `kv.db_path`
* Add new command `atuin kv rebuild` to rebuild the KV cache (also rebuilt during sync)
* Add the ability to pass multiple keys to `atuin kv delete`
* Add `atuin kv ls` as an alias to `atuin kv list`
* Disallow setting a key with an empty namespace via `atuin kv set -n "" -k key value`; the other commands don't perform this check, in case users have existing empty-string-namespace entries they want to use or delete
* Add help messaging for all commands and options

I also tried to allow setting a value via `atuin kv set key value` (without the explicit `--key` argument), but was unable to get this to work while maintaining backward compatibility.

## Checks


ellie (approved):
🚀🚀🚀

Issue #2502, opened 2024-12-28
Add sqlite support for the server


onedr0p:
This was discussed in Discord and seems like Atuin is ready to move forward with this.

b2m:
It seems that the requested feature was introduced in atuin via #2770 in [v18.7.0](https://github.com/atuinsh/atuin/releases/tag/v18.7.0).

See also the updated [server setup documentation](https://docs.atuin.sh/self-hosting/server-setup/).

PR #2770, opened 2025-05-25
Add sqlite server support for self-hosting

szinn:
https://github.com/atuinsh/atuin/issues/2502

This adds support for sqlite for the database in the server rather than the support via https://github.com/conradludgate/atuin-server-sqlite which isn't actively maintained

## Checks


szinn:
Note the migrations are slightly different in this PR so isn't a drop-in replacement for conrad's solution - a new DB is required

Dockerfile
```diff
@@ -20,7 +20,7 @@ FROM debian:bookworm-20250428-slim AS runtime
 
 RUN useradd -c 'atuin user' atuin && mkdir /config && chown atuin:atuin /config
 # Install ca-certificates for webhooks to work
-RUN apt update && apt install ca-certificates -y && rm -rf /var/lib/apt/lists/*
+RUN apt update && apt install ca-certificates sqlite3 -y && rm -rf /var/lib/apt/lists/*
```
ellie: Is this required? We shouldn't be dynamically linking with sqlite

crates/atuin-server-sqlite/migrations/20240108124832_create-total-history-count-user.sql
```diff
@@ -0,0 +1,5 @@
+create table total_history_count_user(
```
ellie: We create this table, and have several queries that read from it, but it's not written to anywhere that I can see

In the postgres implementation, we handle this through triggers. However, for this implementation I think it makes more sense to eliminate the table entirely.

It was originally written so that we could quickly return how much history a user has, at the cost of slower/more expensive writes. I don't expect anyone hosting a sqlite server to also be serving 100s of active users

crates/atuin-server-sqlite/migrations/20230623070418_create-records.sql
```diff
@@ -0,0 +1,15 @@
+create table records (
```
ellie: We can also drop this, as no code uses it

crates/atuin-server-sqlite/src/lib.rs
```diff
@@ -0,0 +1,571 @@
...
+        Ok(())
+    }
+
+    #[instrument(skip_all)]
+    async fn total_history(&self) -> DbResult<i64> {
+        // The cache is new, and the user might not yet have a cache value.
+        // They will have one as soon as they post up some new history, but handle that
+        // edge case.
+
+        let res: (i64,) = sqlx::query_as("select sum(total) from total_history_count_user")
```
ellie:
```suggestion
        let res: (i64,) = sqlx::query_as("select count(1) from history")
```

ellie (commented):
Generally looks good, thank you! Just a bit of feedback wrt some tables that aren't being used in this implementation. 

There are two versions of sync

## v1
Uses an unreliable count-based algorithm, which eventually needed serverside caching on our instance to avoid killing the database

The caching won't be needed here, as 

1. Few people use v1 and
2. There won't be enough users on a sqlite instance for it to be an issue

## v2 
Uses the record store, and does not rely on counting rows. Much more reliable and performant

I suspect that v1 won't actually work at all with this implementation - if we fix the cache table usage it should be fine

onedr0p:
@ellie  I was able to test this and it works. I had to nuke my previous `atuin.db` sqlite file and then run: 

```sh
atuin logout
atuin register -u <username> -e <email>
atuin key
atuin sync
```

ellie:
@onedr0p thanks! Was that v1 or v2? It's v1 I suspect was broken (that's the case with records = false)

will review the changes later today :)

onedr0p:
@ellie I am assuming it's v2 if that's the default, I am using the most recent changes here too.

Dockerfile
onedr0p: This might be for testing and can be removed before merge, is that right @szinn ?

szinn: sqlite3 is so that it is available in the CLI - was nice for testing and seeing what tables had what. Not attached to it being there and can remove it

szinn:
@ellie ping?

szinn:
https://github.com/atuinsh/docs/pull/96

PR #2774, opened 2025-05-27
fix(doctor): mention the required ble.sh version

akinomyoga:
The version requirement for `ble.sh` is not mentioned in the warning message of the `atuin doctor` output. In the Atuin forum, a user reported that the history records are empty even though the user set up both Atuin and `ble.sh` [1], which turned out to be because the user used an old version, ble-0.3.4.

- [1] https://forum.atuin.sh/t/1047

The warning message should include the version requirement for `ble.sh`.

## Checks


akinomyoga:
Thank you!

Issue #2208, opened 2024-06-29
[Bug]: Nushell integration adds a significant delay before each prompt


WieeRd:
### What did you expect to happen?

Atuin integration with Bash/Zsh did not slow down the shell.
I expected the same with Nushell.

### What happened?

There is a noticeable delay after the invoked command's completion and before the new shell prompt shows up.
Even when running a very simple command like `ls` which takes less than a millisecond.

```
~
❯ atuin history list | lines | length
8795

~
❯ timeit do { atuin history end --exit 0 (atuin history start -- ls) }
243ms 217µs 340ns

~
❯ mv ~/.local/share/atuin/ ~/.local/share/atuin.backup

~
❯ atuin history list | lines | length
1

~
❯ timeit do { atuin history end --exit 0 (atuin history start -- ls) }
20ms 317µs 702ns
```

I have 8.8k history records and `atuin history start` + `atuin history end` pair take around 240ms.
After nuking the DB at `~/.local/share/atuin` that time is reduced to 20ms and the delay is no longer noticeable.

The prompt was instantaneous regardless of the history size on the other shells (Bash/Zsh).
So I assume this has to do with Nushell integration specifically.

### Atuin doctor output

```yaml
{
  "atuin": {
    "version": "18.3.0",
    "sync": null,
    "sqlite_version": "3.44.0"
  },
  "shell": {
    "name": "nu",
    "default": "nu",
    "plugins": [
      "atuin"
    ],
    "preexec": "built-in"
  },
  "system": {
    "os": "Arch Linux",
    "arch": "x86_64",
    "version": "unknown",
    "disks": [
      {
        "name": "/dev/nvme0n1p3",
        "filesystem": "btrfs"
      },
      {
        "name": "/dev/nvme0n1p3",
        "filesystem": "btrfs"
      },
      {
        "name": "/dev/nvme0n1p3",
        "filesystem": "btrfs"
      },
      {
        "name": "/dev/nvme0n1p3",
        "filesystem": "btrfs"
      },
      {
        "name": "/dev/nvme0n1p3",
        "filesystem": "btrfs"
      },
      {
        "name": "/dev/nvme0n1p2",
        "filesystem": "ext4"
      },
      {
        "name": "/dev/nvme0n1p1",
        "filesystem": "vfat"
      }
    ]
  }
}
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

WieeRd:
[...]

PR #2778, opened 2025-06-04
feat: nushell init script enchancements

Tyarel8:
Some improvements to the atuin init script for nushell
- Minimum supported version bumped to `0.93.0` (Released `2024-5-1`)
Following #1963, this nushell release is more than a year old and enables doing `if (version).major > ...` without more complications (having to check the version to see if you can check the version more ergonomically is kind of ridiculous). For a program like nushell, I don't think it makes much sense to support even older releases. I have tested this script in both `0.93.0` and latest (`0.104.2`).

- Use `random uuid -v 7` if available.
Nushell has a command for creating uuids in the core library, and it is considerably faster than calling the external `atuin uuid`, saving precious milliseconds at startup. Includes module magic to make it compatible with older versions thanks to @132ikl.
image

- Use `job spawn` for `atuin history end`
Should improve performance of #2208. `atuin history end` can easily be invoked in a job so that it doesn't block the main thread, `atuin history start` is trickier since it needs to be assigned to an `$env`.

## Checks

### DIFF
--- a/crates/atuin/src/shell/atuin.nu
+++ b/crates/atuin/src/shell/atuin.nu
@@ -1,5 +1,14 @@
 # Source this in your ~/.config/nushell/config.nu
-$env.ATUIN_SESSION = (atuin uuid)
+# minimum supported version = 0.93.0
+module compat {
+  export def --wrapped "random uuid -v 7" [...rest] { atuin uuid }
+}
+use (if not (
+    (version).major > 0 or
+    (version).minor >= 103
+) { "compat" }) *
+
+$env.ATUIN_SESSION = (random uuid -v 7 | str replace -a "-" "")
 hide-env -i ATUIN_HISTORY_ID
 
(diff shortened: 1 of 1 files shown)


mirsella:
hello ! when trying to exit nushell after typing at least one command:
```
❯ exit
There are still background jobs running (1).
Running `exit` a second time will kill all of them.
```

edit: doesn't always happens :thinking:

mirsella:
```
❯ ps -l | rg (job list | $in.0.pids.0)
│ 533 │ 102979 │ 102759 │ atuin            │ Sleeping │  0.0000 │  19.6 MB │ 373.4 MB │ /usr/bin/a │ now        │    1000 │       20 │ ... │
```
seems to be running the non-existent /usr/bin/a binary ?
i have a entry in my history with /usr/bin/a a long time ago

Issue #2779, opened 2025-06-05
[Bug]: `could not load client settings` is too verbose


jyn514:
### What did you expect to happen?

errors only happen when i explicitly run an atuin command, or on the first time the shell is loaded. errors can happen in many different ways; a simple one is to run `chmod a-r /home/jyn/.config/atuin/config.toml`. i usually see this when my disk is full, or occasionally when i have not unlocked my home directory (which is encrypted with fscrypt).

### What happened?

atuin outputs an error on every keystroke, making it very difficult to see what i am typing.
```
$  Error: could not load client settings

Caused by:
   0: could not create config file
   1: failed to create file `/home/jyn/.config/atuin/config.toml`
   2: Required key not available (os error 126)

Location:
    atuin-client/src/settings.rs:675:54
 fError: could not load client settings

Caused by:
   0: could not create config file
   1: failed to create file `/home/jyn/.config/atuin/config.toml`
   2: Required key not available (os error 126)

Location:
    atuin-client/src/settings.rs:675:54
 faError: could not load client settings
```

### Atuin doctor output

```yaml
Error: could not load client settings

Caused by:
   0: could not create config file
   1: failed to create file `/home/jyn/.config/atuin/config.toml`
   2: Required key not available (os error 126)

Location:
    atuin-client/src/settings.rs:675:54
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

jyn514:
reading through `atuin init zsh`, it looks like this is coming from `_zsh_autosuggest_strategy_atuin()`, which runs `atuin search --limit 1`. i think any errors from that shell function should be silenced.

jyn514:
yeah ok this works around it:
```sh
_zsh_autosuggest_strategy_atuin-quiet() {
  _zsh_autosuggest_strategy_atuin "$@" 2>/dev/null
}
ZSH_AUTOSUGGEST_STRATEGY=(history completion atuin-quiet)
```

PR #2780, opened 2025-06-05
Don't print errors in `zsh_autosuggest` helper

jyn514:
Previously, this would result in long multi-line errors when typing, making it hard to see the shell prompt:
```
$  Error: could not load client settings

Caused by:
   0: could not create config file
   1: failed to create file `/home/jyn/.config/atuin/config.toml`
   2: Required key not available (os error 126)

Location:
    atuin-client/src/settings.rs:675:54
 fError: could not load client settings

Caused by:
   0: could not create config file
   1: failed to create file `/home/jyn/.config/atuin/config.toml`
   2: Required key not available (os error 126)

Location:
    atuin-client/src/settings.rs:675:54
 faError: could not load client settings
```

Silence these in autosuggestions, such that they only show up when explicitly invoking atuin.

Fixes https://github.com/atuinsh/atuin/issues/2779

## Checks

### DIFF
--- a/crates/atuin/src/shell/atuin.zsh
+++ b/crates/atuin/src/shell/atuin.zsh
@@ -15,7 +15,8 @@ zmodload zsh/datetime 2>/dev/null
 # you'd like to override this, then add your config after the $(atuin init zsh)
 # in your .zshrc
 _zsh_autosuggest_strategy_atuin() {
-    suggestion=$(ATUIN_QUERY="$1" atuin search --cmd-only --limit 1 --search-mode prefix)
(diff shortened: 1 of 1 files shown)


jyn514:
i see there is something equivalent here for bash, but i don't use `ble.sh`, so i can't test any changes i make to that file.

ellie (approved):
nice fix, thank you!

akinomyoga:
> i see there is something equivalent here for bash, but i don't use `ble.sh`, so i can't test any changes i make to that file.

I submitted a PR at #2976.

PR #2781, opened 2025-06-05
feat: Add logging to file

### LINKED ISSUE #87: Log to file
Update our logger to write to a file as well as stdout to help with debugging. (see #79)

thomastaylor312:
This uses tracing appender and combines with the default stdout writer that comes from `fmt::layer`.

Closes #87

### DIFF
--- a/crates/atuin/src/command/client.rs
+++ b/crates/atuin/src/command/client.rs
@@ -6,6 +6,7 @@ use eyre::{Result, WrapErr};
 use atuin_client::{
     database::Sqlite, record::sqlite_store::SqliteStore, settings::Settings, theme,
 };
+use tracing_appender::rolling::{RollingFileAppender, Rotation};
 use tracing_subscriber::{filter::EnvFilter, fmt, prelude::*};
 
 #[cfg(feature = "sync")]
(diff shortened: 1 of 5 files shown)


crates/atuin/src/command/client.rs
```diff
@@ -120,9 +121,20 @@ impl Cmd {
     ) -> Result<()> {
         let filter =
             EnvFilter::from_env("ATUIN_LOG").add_directive("sqlx_sqlite::regexp=off".parse()?);
-
+        let log_dir = atuin_common::utils::state_dir().join("logs");
+        // Ensure the log directory exists
+        tokio::fs::create_dir_all(&log_dir)
+            .await
+            .wrap_err("could not create log directory")?;
+        let appender = RollingFileAppender::builder()
```
thomastaylor312, line 143: This seemed like a decent idea so the log wouldn't grow forever, but we can also just do a plain file as well

crates/atuin/src/command/client.rs
```diff
@@ -120,9 +121,20 @@ impl Cmd {
...
+            .filename_prefix("atuin.log")
+            .rotation(Rotation::DAILY)
+            .max_log_files(7)
```
thomastaylor312: I would probably make this configurable if we want to keep the rotating appender around. If we do, should this just be an env var or also exposed in settings?

thomastaylor312:
In draft so I can get some feedback on the comments above. If there was a different idea in mind, please let me know. This wasn't too difficult so easy enough to throw away and take another approach

crates/atuin/src/command/client.rs
thomastaylor312: I added an env var for now just in case. Easy enough to pop off or modify depending on feedback!

markovejnovic:
Closing this as it appears to be stale and implemented in https://github.com/atuinsh/atuin/pull/3201 and https://github.com/atuinsh/atuin/pull/3622.

Thank you for the PR regardless!

PR #2787, opened 2025-06-10
Nix: various improvements

faukah:
Cleans up the nix code in various ways, the most important ones being:
- Dropping flake-utils, thus reducing abstraction.
- Decreasing the amount of files in the `src` fileset in `atuin.nix` which prevents changes to files like README.md from changing the derivation's output, thus saving rebuils. 
- Adding a sane default formatter.

## Checks

### DIFF
--- a/atuin.nix
+++ b/atuin.nix
@@ -10,7 +10,7 @@
   installShellFiles,
   rustPlatform,
   libiconv,
-  gitRev ? "dirty",
+  gitRev,
 }:
 let
   fs = lib.fileset;
@@ -28,7 +28,7 @@ let
 
 in
 rustPlatform.buildRustPackage {
-  name = "atuin";
+  pname = "atuin";
   version = "${packageVersion}-unstable-${gitRev}";
 
   src = fs.toSource {
--- a/atuin.nix
+++ b/atuin.nix
@@ -1,7 +1,7 @@
 # Atuin package definition
 #
 # This file will be similar to the package definition in nixpkgs:
-#     https://github.com/NixOS/nixpkgs/blob/master/pkgs/by-name/at/atuin/package.nix
+# https://github.com/NixOS/nixpkgs/blob/master/pkgs/by-name/at/atuin/package.nix
 #
 # Helpful documentation: https://github.com/NixOS/nixpkgs/blob/master/doc/languages-frameworks/rust.section.md
 {
@@ -10,11 +10,31 @@
   installShellFiles,
   rustPlatform,
   libiconv,
+  gitRev ? "dirty",
 }:
+let
+  fs = lib.fileset;
+  src = fs.difference (fs.gitTracked ./.) (
+    fs.unions [
+      ./demo.gif
+      ./flake.lock
+      (fs.fileFilter (file: lib.strings.hasInfix ".git" file.name) ./.)
+      (fs.fileFilter (file: file.hasExt "svg") ./.)
+      (fs.fileFilter (file: file.hasExt "md") ./.)
+      (fs.fileFilter (file: file.hasExt "nix") ./.)
+    ]
+  );
+  packageVersion = (builtins.fromTOML (builtins.readFile ./Cargo.toml)).workspace.package.version;
+
+in
 rustPlatform.buildRustPackage {
   name = "atuin";
+  version = "${packageVersion}-unstable-${gitRev}";
 
-  src = lib.cleanSource ./.;
+  src = fs.toSource {
+    root = ./.;
+    fileset = src;
+  };
 
   cargoLock = {
     lockFile = ./Cargo.lock;
@@ -22,9 +42,11 @@ rustPlatform.buildRustPackage {
     allowBuiltinFetchGit = true;
   };
 
-  nativeBuildInputs = [installShellFiles];
+  nativeBuildInputs = [ installShellFiles ];
 
-  buildInputs = lib.optionals stdenv.isDarwin [libiconv];
+  buildInputs = lib.optionals stdenv.isDarwin [
+    libiconv
+  ];
 
   postInstall = ''
     installShellCompletion --cmd atuin \
@@ -35,10 +57,10 @@ rustPlatform.buildRustPackage {
 
   doCheck = false;
 
-  meta = with lib; {
+  meta = {
     description = "Replacement for a shell history which records additional commands context with optional encrypted synchronization between machines";
     homepage = "https://github.com/atuinsh/atuin";
-    license = licenses.mit;
+    license = lib.licenses.mit;
     mainProgram = "atuin";
   };
 }
--- a/flake.nix
+++ b/flake.nix
@@ -1,7 +1,6 @@
 {
   inputs = {
     nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
-    flake-utils.url = "github:numtide/flake-utils";
     flake-compat = {
       url = "github:edolstra/flake-compat";
       flake = false;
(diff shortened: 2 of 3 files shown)


minhtrancccp:
would it be possible for the package to include version as well, as [`home-manager` requires that info to enable daemon](https://github.com/nix-community/home-manager/blob/0cdfcdbb525b77b951c889b6131047bc374f48fe/modules/programs/atuin.nix#L214)?

faukah:
> would it be possible for the package to include version as well, as [`home-manager` requires that info to enable daemon](https://github.com/nix-community/home-manager/blob/0cdfcdbb525b77b951c889b6131047bc374f48fe/modules/programs/atuin.nix#L214)?

I've done that. Should be ready to review now!

ellie:
I'm sorry, we're not in a place to maintain and review this right now - happy to reopen or revisit if @arcuru or someone else who's touched our nix code in the past is willing to give it a once over

shymega:
@ellie I'm experienced with Nix, and contribute to Nixpkgs. If you want to ping me over the weekend, I'll take a look.

markovejnovic:
@shymega, would love to get your feedback! I'll open the PR right up =)

faukah:
I'll update it in a bit, then; it's been a while since I touched this pr :)

shymega:
@faukah Please ping me when you're ready - I'm off $work for the next couple of work days, so I have capacity to look at open-source projects again.

shymega:
@markovejnovic No worries. Yourself and @ellie are welcome to ask for my input/maintenance on Nix in the future :-)

arcuru:
Nix changes bring all the Nix fans out :)

I'll stay out of helping, but my drive-by suggestion is that you keep in mind the goals for the Nix stuff here:

1. Keep `nix run github:atuinsh/atuin` working, so downstream consumers (aka me) can keep running/installing from the repo.
2. A devshell with necessary tools pinned for drive-by development.
3. A dev doc, or whatever else Ellie wants, describing how to update it.
4. Don't put anything else in there.

It's tempting to move everything into Nix (I run [_everything_ inside the Nix sandbox](https://github.com/arcuru/eidetica/)) but that's explicitly not what Atuin needs.

For non-Nix people, I've found that Frontier-level LLMs are quite good at understanding and writing Nix, which should make reviewing easier.

shymega (changes requested):
I feel like this isn't necessary in this MR. It's purely stylistic, and as developers can (and do!) disagree about style, I think this commit should be dropped.

shymega (changes requested):
See: https://github.com/atuinsh/atuin/pull/2787#pullrequestreview-4835197182 regarding formatting.

Do we need this change added to the Flake? Can Atuin build without it? If so, I would argue we don't need stylistic changes.

shymega (changes requested):
See: https://github.com/atuinsh/atuin/pull/2787#pullrequestreview-4835197182 regarding formatting.

Do we need this change added to the Flake? Can Atuin build without it? If so, I would argue we don't need stylistic changes.

atuin.nix
```diff
@@ -10,35 +10,43 @@
...
   rustPlatform,
   libiconv,
-  pkg-config,
-  openssl,
+  gitRev,
 }:
+let
+  fs = lib.fileset;
+  src = fs.difference (fs.gitTracked ./.) (
+    fs.unions [
```
shymega, line 18: You could move `fs = lib.fileset` into a let binding in `src`. No need for global scope.

atuin.nix
```diff
@@ -10,35 +10,43 @@
   installShellFiles,
   rustPlatform,
   libiconv,
-  pkg-config,
-  openssl,
+  gitRev,
 }:
+let
+  fs = lib.fileset;
+  src = fs.difference (fs.gitTracked ./.) (
```
shymega, line 17: Can't we just include only the workspace files, and be more explicit about what's included? I feel like filtering on Markdown, SVGs, and Git metadata, is unnecessary and a maintenance burden.

atuin.nix
```diff
@@ -10,35 +10,43 @@
...
 
-  OPENSSL_NO_VENDOR = 1;
-
-  # native-tls pulls OpenSSL into the sqlx-macros proc-macro, which rustc dlopens at
-  # build time. With OPENSSL_NO_VENDOR it links libssl.so.3 dynamically, so the loader
-  # needs it on LD_LIBRARY_PATH while the proc-macro is loaded.
-  preBuild = ''
-    export LD_LIBRARY_PATH="${lib.makeLibraryPath [ openssl ]}''${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH}"
-  '';
+  buildInputs = lib.optionals stdenv.isDarwin [
```
shymega, line 47: This can be `lib.optional`.

atuin.nix
```diff
@@ -49,10 +57,10 @@ rustPlatform.buildRustPackage {
 
   doCheck = false;
 
-  meta = with lib; {
+  meta = {
     description = "Replacement for a shell history which records additional commands context with optional encrypted synchronization between machines";
     homepage = "https://github.com/atuinsh/atuin";
-    license = licenses.mit;
+    license = lib.licenses.mit;
     mainProgram = "atuin";
```
shymega, line 64: Unnecessary change. The same end-result is going to happen, so no need to change this.

flake.nix
```diff
@@ -12,66 +11,91 @@
...
+      nixpkgs,
+      fenix,
+      ...
+    }:
+    let
+      inherit (nixpkgs) lib;
       systems = [
         "x86_64-linux"
         "aarch64-linux"
         "x86_64-darwin"
```
shymega, line 25: No longer supported upstream in Nixpkgs. But maybe we should address this again later on.

flake.nix
```diff
@@ -12,66 +11,91 @@
...
-            default = atuin;
-          };
+      eachSystem = lib.genAttrs systems;
+      pkgsFor = eachSystem (
+        system:
+        import nixpkgs {
+          localSystem.system = system;
+          overlays = [ self.overlays.atuin ];
+        }
+      );
```
shymega, line 35: I don't like this. I think `flake-utils` did the job just fine, but it does add to the dep graph.

At the very least, merge the two variables into something like `genAttrs systems (system: f (import nixpkgs)` and access the `system` variable from `stdenv.hostPlatform.system`. The way here for iterating over platforms is a bit strange.

flake.nix
```diff
@@ -12,66 +11,91 @@
...
-                echo >&2 "''${ATUIN_DB_PATH} already exists, you might want to double-check that"
-              fi
+          shellHook = ''
+            echo >&2 "Setting development database path"
+            export ATUIN_DB_PATH="/tmp/atuin_dev.db"
+            export ATUIN_RECORD_STORE_PATH="/tmp/atuin_records.db"
 
-              if [ -e "''${ATUIN_RECORD_STORE_PATH}" ]; then
-                echo >&2 "''${ATUIN_RECORD_STORE_PATH} already exists, you might want to double-check that"
-              fi
```
shymega, line 68: Why can't we use `pkgs.mkShell` and `inputsFrom`?

flake.nix
```diff
@@ -65,10 +71,31 @@
...
+              rustc = toolchain;
+            };
+          in
+          {
+            atuin = final.callPackage ./atuin.nix {
+              inherit rustPlatform;
+              gitRev = self.shortRev or "dirty";
+            };
+
+            default = self.overlays.atuin;
```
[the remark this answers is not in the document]
shymega: This should be addresed.

flake.nix
```diff
@@ -12,66 +11,91 @@
...
-              (with pkgs; [
-                cargo-edit
-                rust-analyzer
-              ])
-              ++ [ toolchain ];
+        default = self.packages.${system}.atuin;
+      });
 
-            RUST_SRC_PATH = "${pkgs.rustPlatform.rustLibSrc}";
+      devShells = lib.mapAttrs (system: pkgs: {
```
shymega, line 47: This is unusual for a Flake, and should be less complex.

shymega (changes requested):

shymega:
There's  quite a few things I've flagged, but ultimately I agree with @arcuru on this. There's changes in this PR that are a bit unusual, and I don't want to break compatibility for downstream consumers.

The KISS principle is worth bearing in mind, and we should keep the Nix Flake as simple as we can, so that Ellie and co can focus on what makes Atuin great - rather than  packaging semantics.

markovejnovic:
Still have no clue what I'm looking at but I want to thank @shymega for the work!!

PR #2805, opened 2025-07-01
feat: atuin server kustomize manifests with example overlay

skymoore:
Hello, I deployed my own atuin server and I noticed that you were missing kustomize manifests.

## Checks


ellie:
Thank you for sharing these, but we are not currently able to maintain and support kustomize manifests.

Issue #2727, opened 2025-04-30
feedback on `history dedup` command


benoit-pierre:
- too much output: list the id of each deleted history entry, and on stderr instead of stdout (so I can't use `>/dev/null` to filter those messages and still get errors if any)
- too slow (from a quick look at the code, because it's not using one SQL transaction, but one per entry deleted):
```bash
▹ time atuin history dedup --before now --dupkeep=1 2>/dev/null
Found 38117 duplicates to delete.
atuin history dedup --before now --dupkeep=1 2> /dev/null  13.35s user 12.09s system 40% cpu 1:02.33 total
```
And that's after running the command multiple times before (and aborting it because it was taking too long), so it's only party of my history.

atuin version: 18.6.0-beta.1 (cd5d337)

otaconix:
Here's mine: I just found out about the `history dedup` subcommand, and I happily tried it, as follows:

```
atuin history dedup --before now --dupkeep 0
```

Turns out I'm going to have to restore a backup now 😅. I assumed `--dupkeep n` would keep `n` _duplicates_, where `n == 0` would result in only unique commands remaining. Instead, what I got is an empty database...

Disclaimer: I may be dumb, but it may be a good idea to prevent others from making the same mistake by refusing to perform the operation when `--dupkeep 0` is used, with a message explaining that it would be the same thing as deleting everything (and we already have a command to do that).

gunjubas:
> Here's mine: I just found out about the `history dedup` subcommand, and I happily tried it, as follows:
> 
> ```
> atuin history dedup --before now --dupkeep 0
> ```
> 
> Turns out I'm going to have to restore a backup now 😅. I assumed `--dupkeep n` would keep `n` _duplicates_, where `n == 0` would result in only unique commands remaining. Instead, what I got is an empty database...
> 
> Disclaimer: I may be dumb, but it may be a good idea to prevent others from making the same mistake by refusing to perform the operation when `--dupkeep 0` is used, with a message explaining that it would be the same thing as deleting everything (and we already have a command to do that).

Such an excellent command. Thank you for your comment :) I think there needs to be an alert to avoid this unsafe operation (emptying the whole history)

MatthewFletcher:
Yeah I made the same mistake as gunjubas. I figured it would be a nice way to clean up all of my duplicates. 

Well, it did get rid of the duplicates... but it also got rid of everything else ;-) 

Oh well, nothing terribly important. Just another "warn user before deleting everything". :)

[...]

PR #2807, opened 2025-07-03
fix: refuse "--dupkeep 0"

lilydjwg:
People may think it would keep only one copy, but it didn't work that way.

I'm sorry for people who lost their history due to this, e.g. https://github.com/atuinsh/atuin/issues/2015#issuecomment-3032340594 and #2727.

## Checks

### DIFF
--- a/crates/atuin/src/command/client/history.rs
+++ b/crates/atuin/src/command/client/history.rs
@@ -622,6 +622,13 @@ impl Cmd {
         dupkeep: u32,
         dry_run: bool,
     ) -> Result<()> {
+        if dupkeep == 0 {
+            eprintln!(
+                "\"--dupkeep 0\" would keep 0 copies of duplicate commands and thus delete all of them! Use \"atuin search --delete ...\" if you really want that."
+            );
+            std::process::exit(1);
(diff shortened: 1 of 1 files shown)


benoit-pierre:
Wouldn't it make more sense to have the command actually keep N duplicates? As in always keep the original, remove only duplicates, so `--dupkeep 0` would keep one instance (the original), removing all duplicates.

lilydjwg:
I prefer to rename the option to `--copies-keep` or just `--keep` if a breaking change would be made. Specifying how many copies to keep is more straightforward than specifying "how many plus one" to keep IMO. But I don't know if it is good to do such a rename. @ellie what do you think?

Again, I'm sorry for not looking up in the dictionary before making the option name. Naming things can be hard, especially in a non-native language...

benoit-pierre:
I don't know, I'm not a native speaker, but `--copies-keep` feels similarly confusing. Maybe `--keep <COUNT>  How many recent results to keep`?

benoit-pierre:
Better: `--keep <COUNT>  How many identical recent results to keep`.

PR #2810, opened 2025-07-05
fix(dotfiles): properly escape spaces/quotes in vars

### LINKED ISSUE #2809: [Bug]: dotfile vars don't correctly handle spaces

I'd expect that values containing a space (or perhaps other special characters such as quotes) are properly handled/escaped, such that you can do:

```bash
$ atuin dotfiles var set FOO 'hello there'

# open new shell

$ echo "$FOO"
hello there
```

### What happened?

If I set a dotfiles var with a value containing a space, the resultant value that gets set/exported to my shell is incorrect. The value is stored correctly as shown in `atuin dotfiles var list`, it just doesn't correctly escape values in the shell.

For example:

```bash
$ atuin dotfiles var set FOO 'hello there'
$ atuin dotfiles var list
export FOO=hello there

# open new shell

$ atuin init zsh
# >8 snip
export FOO=hello there
$ echo "$FOO"
hello
```

### Atuin doctor output

```yaml
Atuin Doctor
Checking for diagnostics


{
  "atuin": {
    "version": "18.6.1",
    "sync": {
      "cloud": false,
      "records": true,
      "auto_sync": true,
      "last_sync": "2025-07-05 16:54:01.240302164 +00:00:00"
    },
    "sqlite_version": "3.46.0"
  },
  "shell": {
    "name": "zsh",
    "default": "zsh",
    "plugins": [
      "atuin"
    ],
    "preexec": "built-in"
  },
  "system": {
    "os": "EndeavourOS",
    "arch": "x86_64",
    "version": "rolling",
    "disks": [
      {
        "name": "/dev/nvme0n1p2",
        "filesystem": "btrfs"
      },
      {
        "name": "/dev/nvme0n1p2",
        "filesystem": "btrfs"
      },
      {
        "name": "/dev/nvme0n1p2",
        "filesystem": "btrfs"
      },
      {
        "name": "/dev/nvme0n1p2",
        "filesystem": "btrfs"
      },
      {
        "name": "/dev/nvme0n1p2",
        "filesystem": "btrfs"
      },
      {
        "name": "/dev/nvme0n1p1",
        "filesystem": "vfat"
      }
    ]
  }
}
```

### Code of Conduct

bjschafer:
Previously, setting a dotfiles var whose value contains a space or quote would cause that value to be set incorrectly on shell startup.

This adds escaping for spaces and quotes for all shells so that variables are set correctly on shell startup.

Fix #2809

I'm not real familiar with Rust, so I got help from Claude Code for this. I've still reviewed the generated code and am happy to stand behind it.

## Checks


crates/atuin-dotfiles/src/store/var.rs
```diff
@@ -115,13 +115,59 @@ impl VarStore {
...
+    fn escape_posix_value(value: &str) -> String {
+        // If the value contains no special characters, we can use it unquoted
+        if value
+            .chars()
+            .all(|c| c.is_alphanumeric() || c == '_' || c == '-' || c == '/' || c == '.')
+        {
+            value.to_string()
+        } else {
+            // Otherwise, wrap in double quotes and escape any embedded double quotes
+            format!("\"{}\"", value.replace('\\', "\\\\").replace('"', "\\\""))
```
akinomyoga: `\`, `$`, and `` ` `` inside double-quotes also need to be escaped.

crates/atuin-dotfiles/src/store/var.rs
```diff
@@ -115,13 +115,59 @@ impl VarStore {
...
+    fn escape_fish_value(value: &str) -> String {
+        // If the value contains no special characters, we can use it unquoted
+        if value
+            .chars()
+            .all(|c| c.is_alphanumeric() || c == '_' || c == '-' || c == '/' || c == '.')
+        {
+            value.to_string()
+        } else {
+            // Use single quotes and escape any embedded single quotes
+            format!("'{}'", value.replace('\'', "\\'"))
```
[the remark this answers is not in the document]
bjschafer: Per the [fish docs on quotes](https://fishshell.com/docs/current/language.html#quotes), this is incorrect. To quote:

> The only meaningful escape sequences in single quotes are \', which escapes a single quote and \\, which escapes the backslash symbol.

crates/atuin-dotfiles/src/store/var.rs
bjschafer: Good catch, thanks. Fixed.

PR #2817, opened 2025-07-08
feat: add inline_height_shell_up_key_binding option

### LINKED ISSUE #2612: separate style and inline_height settings for ctrl+r and up keybinding
I would really like to have the "full" style taking up the complete window when I press ctrl+r, but the "compact" style and only 10 lines or so when using the up arrow. 

If creating such a feature wouldn't require huge rewrites then I would love to try implementing it, and would appreciate some pointers about where to start with that.

caio96:
I was looking for option to have a small inline height for the up arrow and full screen for ctrl-r.
This PR adds a separate option to configure the inline height of the up arrow: `inline_height_shell_up_key_binding`.

I followed the same structure as the `search_mode_shell_up_key_binding` option.
Resolves #2612.

## Checks
- [ X] I am happy for maintainers to push small adjustments to this PR, to speed up the review cycle
- [ X] I have checked that there are no existing pull requests for the same thing

### DIFF
--- a/crates/atuin-client/config.toml
+++ b/crates/atuin-client/config.toml
@@ -71,6 +71,11 @@
 ## set it to 0 to always go full screen
 # inline_height = 0
 
+## the maximum number of lines the interface should take up
+## when atuin is invoked from a shell up-key binding
+## the accepted values are identical to those of "inline_height"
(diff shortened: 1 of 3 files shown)


ellie (approved):
neat! thank you!

PR #2832, opened 2025-07-18
fix: Make login exit 1 if already logged in

davidolrik:
To make scripting atuin easier, I've made login exit with 1 when alreading logged in.

## Checks


ellie:
small risk of breaking existing scripts and workflows - but I doubt anyone relies on `login` exiting 0?

PR #2833, opened 2025-07-21
Update indicatif to 0.18.0

musicinmybrain:
The only breaking change in `indicatif` 0.18 is that the `console` dependency is updated to [0.16.0](https://github.com/console-rs/console/releases/tag/0.16.0); the breaking change in that release is that crates that depend on `console` with `default-features = False` may need to explicitly enable the new `std` feature.

## Checks

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -917,15 +917,15 @@ dependencies = [
 
 [[package]]
 name = "console"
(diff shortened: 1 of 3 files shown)


ellie (approved):
Thank you!

PR #2834, opened 2025-07-22
feat: command chaining

ellie:
Allow for smart completion of commands ending in && or ||

## Checks


ltrzesniewski:
~@ellie is it on purpose that `enter_accept` is required for `command_chaining` to be taken into account?~

I suppose it's not intentional, see #2861

PR #2843, opened 2025-07-25
fix: Make status exit 1 if not logged in

davidolrik:
Another small one that makes `atuin status` also exit 1 when not logged in.

## Checks

### DIFF
--- a/crates/atuin/src/command/client/sync/status.rs
+++ b/crates/atuin/src/command/client/sync/status.rs
@@ -3,15 +3,13 @@ use std::path::PathBuf;
 use crate::{SHA, VERSION};
 use atuin_client::{api_client, database::Database, settings::Settings};
 use colored::Colorize;
-use eyre::Result;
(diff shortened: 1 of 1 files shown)


tessus:
bail shows an error and also gives a backtrace.

An informational message:

```
You are not logged in to a sync server - cannot show sync status
```

will become an error message that also shows the backtrace:

```
Error: You are not logged in to a sync server - cannot show sync status

Location:
    crates/atuin/src/.....rs:..:..
```

I am not sure, if this is very user friendly. IMO, the fact that you are not logged in is not an error. Couldn't you just change the exit code without using bail? 

P.S.: I apologize in advance, in case you find my comment rude or hostile.

davidolrik:
> P.S.: I apologize in advance, in case you find my comment rude or hostile.

No need to apologize, I actually agree - but atuin mostly uses bail to exit everywhere else in the code.

PR #2847, opened 2025-07-28
feat: add native mcfly import support

zampierilucas:
## Description

This PR adds native import support for [mcfly](https://github.com/cantino/mcfly) shell history, allowing users to migrate their mcfly history to Atuin with a single command: `atuin import mcfly`

## Implementation

The implementation:
- Executes `mcfly dump` to retrieve history in JSON format
- Parses mcfly's JSON structure (`when_run` timestamp and `cmd` fields)
- Supports both RFC3339 and Unix timestamp formats
- Includes comprehensive tests covering various scenarios

## Testing

- Added unit tests for timestamp parsing
- Added integration tests for the import process
- Tests cover: valid imports, invalid timestamps, empty history

## Future Enhancements

A more advanced approach would be to implement direct SQLite database import similar to `zsh-histdb` importer. This would:
- Avoid the dependency on `mcfly` being in PATH
- Be more efficient for large histories
- Allow importing additional metadata

However, the current implementation using `mcfly dump` is simpler and leverages mcfly's official export functionality.


zampierilucas:
Hi @ellie,
I noticed you were involved in review for other PRs, can you help me figure out I'm missing something to get this Mr reviewed? or should I have opened a 'feature request' first?

ellie:
Apologies - I've been sick and had a lot on recently so have prioritized reviewing + releasing work either from prior contributors, or for PRs that are easier to review. We've had a bunch of issues from importer related PRs in the past, so these tend to need extra scrutiny. 

Anyway. This generally looks good, but I'd prefer that we dropped the requirement on mcfly. If we could directly read the database that would be perfect! I would also accept reading the output of mcfly dump, without automatically running it. Some users might have workflows where they have the data, but do not have the binary in their path.

zampierilucas:
Hi @ellie, hope you are doing better.

I've followed your suggestion a re-did the import from the db, instead of requiring the mcfly binary to dump as json, and It's pretty nice as now it preserves most of the mcfly metadata, and also allow to just specify `--file db.path`.

PR #2851, opened 2025-07-29
feat:Bind Ctrl-v to visually editing the selected command, exactly as per fc

wolf:
It took me a while to get back to this and finally make the needed changes, but here it is at last.

My old workflow was:
  - select the command I want
  - hit tab, so now it is on the command line, atuin is gone, nothing has been executed
  - I use Vim mode on the command line, so Escape puts me in normal mode
  - vv launches my favorite editor
  - edit, write, and quit which makes the updated command execute

My new workflow follows in the footsteps of fc:
  - select the command I want
  - hit Ctrl-v, launches my favorite editor on the text of that command
  - edit, write, and quit which makes the updated command execute and is saved in history

I initially engaged on this topic roughly a year ago. I tried to make this as much like fc as I could because I feel like atuin hides and replaces fc.  Emacs-mode on the command line requires two control keys to jump into the editor.  Vim-mode needs you to get into normal mode, that is NOT inserting, then hit v twice (well ... twice in Zsh, I think just once in actual readline).  Everything else in atuin is a single key, and when stepping through the history list, you're already not in insert mode.  Like fc, I look for editors in this order:
  - the value of $FCEDIT (I know this feels weird, but you are "fixing a command")
  - the value of $VISUAL
  - the value of $EDITOR
  - /usr/bin/editor (Debian/Ubuntu update-alternatives symlink, if present)
  - vim or vi (whichever is found in PATH first)

## Implementation notes

The action is wired through the keymap system as `edit-accept`, bound to `ctrl-v` by default in both emacs and vim-normal keymaps. It can be rebound in `config.toml` like any other action.

The editor command is parsed with `shlex` so values like `VISUAL="nvim -u NORC"` work correctly.

Before spawning the editor, the TUI terminal is fully torn down (raw mode disabled, alternate screen exited). The editor's stdin/stdout/stderr are explicitly redirected to `/dev/tty` so that ncurses-based editors (nano, vim) work correctly — without this, the shell integration's fd-swap for command substitution causes arrow keys to break and terminal escape sequences to leak into the shell's captured output.

I am very open to discussion.  I want to know if this is a desirable feature, the right implementation, in the right place, with the right names.  The functionality is more important to me than exactly how I implemented it.  And atuin staying true to itself is even more important than that.  I originally posted in the CLI channel on the Discord server describing this.

## Checks


crates/atuin/src/command/client/search/interactive.rs
```diff
@@ -84,6 +88,36 @@ struct StyleState {
     inner_width: usize,
 }
 
+fn get_visual_editor() -> String {
+    std::env::var("FCEDIT")
+        .or_else(|_| std::env::var("EDITOR"))
+        .or_else(|_| std::env::var("VISUAL"))
+        .unwrap_or_else(|_| "vim".to_string())
```
jeremycline: Not a maintainer here, just procrastinating on other work I should be doing, so take this suggestion with a grain of salt.

You might consider a different fallback here. Some minimal Fedora derivatives come without `EDITOR` set and with `vim-minimal` installed which doesn't provide `/usr/bin/vim`, just `/usr/bin/vi`. Debian looks like it relies on the [alternatives system](https://www.debian.org/doc/manuals/debian-reference/ch01.en.html#_setting_a_default_text_editor) so calling `editor` will get you the default.

You _could_ probe to see if there's a `/usr/bin/editor` and then fall back to `vi`, or you could turn it into an error instructing the user to set the correct environment variables. That'd look something like this (which I've not tested at all):
```rust
        .or_else(|_| std::env::var("VISUAL"))
        .ok_or_else(|| eyre::eyre!("To edit commands, set the EDITOR environment variable to your preferred editor"))
```

wolf: Using `vi` instead of `vim` sounds right.  In systems with `vim`, `vi` is usually just another name for `vim`.  A thing I was thinking, related to this, is that perhaps `config.toml` could optionally specify an editor, and that would be something to check for first.  Reporting an error seems like user-friendly behavior.  Is that the general way atuin handles situations like this?  With respect to the exact editors used, I feel like I would want to probe for `vi` and issue the error only after that.  That way I maintain the same sequence used by `fc` (modulo any change in `config.toml`).  But maybe my desire to emulate `fc` is not appropriate.

When you say `/usr/bin/editor`, you're just using `editor` as a stand-in, right?  Just double-checking.  You're asking that I make sure no matter what they have specified, even in one of those variables, that their choice actually exists?

Finally, it occurs to me that this is a feature you might want to turn on/off in `config.toml`.  It seems unlikely, at least to me, but it's a thought.  I feel like "off" means "just don't touch Ctrl-v".  Hmm.  And also I should look to see any place else that mentions keys and consequences, e.g., help text, documentation, etc.  A PR that adds a key should update such things as well.

It wasn't immediately obvious to me how to write a test (or a couple of tests) in the test section of `interactive.rs` since running a visual editor requires interactive input.  This is on my mind.

wolf: Oh I see, `editor` is not just a place-holder.  On some OSs, that's an actual thing.  Actually probing (as you suggest) for the existence of `/usr/bin/editor` sounds like good behavior.  I wonder if we also need to check the actual OS, so that we know `editor` is appropriate.  My initial thought is no.

wolf: I think I know how to do this as your are suggesting.  Making sure specific editors exist feels like the code will be bulky compared to what is currently in this PR.  Looking specifically for `/usr/bin/editor`; and maybe adding a key to `config.toml`.  But I also think these are good ideas, regardless of whether you are a maintainer or not.

I will explore implementing this, and update this PR for further review.

jeremycline: Indeed, I wouldn't enumerate `nvim`, `emacs`, etc. As long as the probe for `/usr/bin/editor` happens after checking the `EDITOR` environment variable I don't think you need to worry about checking the OS, either, although that might be necessary if some distribution ships something else as `/usr/bin/editor` and also don't set an `EDITOR` environment variable.

Personally, I feel like we can get away without a `config.toml` setting for the editor since the `EDITOR` environment variable is so widely used and I have a hard time imagining someone who wants to use `nvim` for git commit messages, but `emacs` for editing commands in Atuin.

For whatever it's worth, I lean _slightly_ towards presenting the user with an error over defaulting to `vi` if nothing is set. I expect nearly every Atuin user knows how to use `vi`, but for any that don't it would be a shocking experience to accidentally end up there. I don't have a great sense of whether or not that matches the general Atuin style though.

[There's a table of shortcuts](https://docs.atuin.sh/configuration/key-binding/#atuin-ui-shortcuts), but I'm not sure where the source for that document is.

Finally, for testing maybe you could set `EDITOR` to a shell script that writes a string to the given tempfile so it's not interactive? I've not thought hard about that, though.

wolf: Oh!  A script instead of an actual editor is a great idea!  I don't know why I didn't think of that.  That's exactly how I test all my dotfile shell functions.

wolf:
I can see, after discussion, that some parts of this implementation are naive. Parts of this machinery are already available within the codebase to help me accomplish this. I'm working on it.

wolf:
I've brought everything up-to-date; overhauled the implementation; tested by hand; mentioned in the CLI channel on discord. This work is ready for others to critique, try, review. I hope I've made something here useful to others.

wolf:
Ready for review — @markovejnovic @ellie CI is green and this is now a single clean commit on `main`.

wolf:
pr2851-edit-accept

This is a movie of my PR in action.
* I do a very constrained `git log` on just one file for just one author
* Within the Atuin history picker, I select the earlier `git log` command
* I hit the default key I've given this action: Ctrl-v
* The selected command opens in my chosen editor (Helix for me)
* I change the author from Ellie to Michelle
* I write and quit (much as you would in Vim)
* The command executes
* I re-open the Atuin history picker, you can see both the original command **and** the new command

crates/atuin/src/command/client/search/interactive.rs
```diff
@@ -189,6 +193,71 @@ struct StyleState {
...
+        return Ok("/usr/bin/editor".to_string());
+    }
+
+    // Fall back to vim or vi if present in PATH.
+    let path_var = std::env::var_os("PATH").unwrap_or_default();
+    for name in ["vim", "vi"] {
+        if std::env::split_paths(&path_var).any(|dir| dir.join(name).is_file()) {
+            return Ok(name.to_string());
+        }
+    }
```
markovejnovic, line 263: Is this a safe fallback for us? My guess is that using `nano` as a default would be safer for users who might not be familiar with vim.

crates/atuin/src/command/client/search/interactive.rs
```diff
@@ -189,6 +193,71 @@ struct StyleState {
     inner_width: usize,
 }
 
+fn get_visual_editor() -> Result<String> {
```
markovejnovic: This implementation looks somewhat correct for POSIX, but I can't help but wonder whether it is correct for Windows.

I don't want to block your PR on windows, but I'm curious if you know what the "appropriate" editor in Windows would be. Notepad.exe, I'm guessing? Sounds like the interface for that would be incredibly annoying.

Perhaps a nice solution here would be to spawn a `notepad.exe` and wait until it exits, presenting the user with some sort of message:

> Please finish editing the command in notepad.

I absolutely do not think you need to implement this in your PR, I'm mostly jotting this down for my own sake.

markovejnovic: Perhaps a better signature for this function would be

```rust
#[derive(Debug, Clone, Copy, thiserror::Error)]
pub enum VisualEditorParsingError {
  #[error("failed to parse the $EDITOR variable as a path: {0}")]
  FailedToParseEnv(OsString),
}

fn get_visual_editor() -> Option<Result<PathBuf, VisualEditorParsingError>>;
```

This way you get to achieve two things:
1. You avoid the `shlex::split` in the subsequent function
2. You avoid losing the type information for `std::path::Path::new("/usr/bin/editor")`
3. You avoid this "downcast": `name.to_string()`

I think it would be better to use stricter paths if possible, as it gives us a margin of confidence that the `execve` under `Command::new` will succeed (at least we know the path exists).

crates/atuin/src/command/client/search/interactive.rs
```diff
@@ -2228,6 +2301,11 @@ pub async fn history(
 
     let accept_prefix = "__atuin_accept__:";
 
+    // Restore the terminal before any operation that needs a clean tty (e.g.
+    // launching an external editor). For all other arms this is a no-op — the
+    // Drop would have run on function exit anyway.
+    drop(terminal);
```
markovejnovic: 👍, seems reasonable

crates/atuin/src/command/client/search/interactive.rs
```diff
@@ -189,6 +193,71 @@ struct StyleState {
...
+    let temp_path = temp_file.into_temp_path();
+
+    let editor = get_visual_editor()?;
+    let parts = shlex::split(&editor)
+        .ok_or_else(|| eyre::eyre!("Failed to parse editor command: {editor}"))?;
+    let (program, args) =
+        parts.split_first().ok_or_else(|| eyre::eyre!("Editor command is empty"))?;
+
+    let mut cmd = Command::new(program);
+    cmd.args(args).arg(&temp_path);
```
markovejnovic: Would it be possible to avoid this tempfile by passing the file on `stdin` to the editor? Non-blocking, just curious.

crates/atuin/src/command/client/search/interactive.rs
markovejnovic: The only concern I have here is that this is really fragile -- without any sort of compilation check (ie. through the type system) enforcing the order here, I find it risky that this will get re-ordered in subsequent PRs and that might cause bugginess.

I would do three things:
1. Explicitly convey that this is for the feature you described.
2. Deslop the comment
3. Write a test exercising the behavior you have added here.

crates/atuin/src/command/client/search/interactive.rs
wolf: I tried to mimic `fc` here. On POSIX, many things that open editors fall back to Vim. But that's just how I arrived here, not a defense. If `nano` is desired, that could be the fallback **or** a user who knew they wanted `nano` could set `$FCEDIT` or any of the others... you're really talking about the case where the user doesn't even know those variables exist though. And `nano` is a lot friendlier than `vim` (`nano` tells you what you can do!).

crates/atuin/src/command/client/search/interactive.rs
wolf: I am (not actually) very sorry to say I don't have a Windows machine on which to test. I have seen, on other platforms, launching GUI editors with a flag to make the CLI wait for the GUI editor to exit. I've done this with `$VISUAL` to see, e.g., BBEdit or VS Code do work. I did not try this for this PR. Would you like me to?

wolf: I love using types to make bad things impossible.

crates/atuin/src/command/client/search/interactive.rs
wolf: This is worth exploring.

crates/atuin/src/command/client/search/interactive.rs
wolf: Strong agree.

crates/atuin/src/command/client/search/interactive.rs
wolf: Explored this — I don't think stdin works here, for two reasons.

First, `stdin` is already redirected to `/dev/tty` for the whole editor session so keystrokes behave (see the fd-swap note above) — there's no spare stdin channel left to also carry the initial buffer.

Second, full-screen editors don't have a "write result to stdout" convention — `:w`/^O etc. all target a named file, so write-back needs a real path no matter how the content gets in. (`vim -` can slurp initial stdin, but still needs a filename for `:w`, so it wouldn't save the result either.)

This also matches existing precedent: `scripts.rs::open_editor` solves the identical edit-then-read-back problem with a tempfile, and `fc` itself — the thing this PR is modeled on — does too. Sticking with the tempfile.

crates/atuin/src/command/client/search/interactive.rs
wolf: Addressed in 84b2c2d6. Added a `TtyReleased` token — `visual_edit_command` now requires `&TtyReleased`, obtainable only by consuming (dropping) `terminal` via `TtyReleased::new`. A future reorder that launches the editor before the terminal is released is now a compile error, not a silent regression.

Also did the other two things you asked for: rewrote the comment at the old `drop(terminal)` site to explain the token instead of the old generic note, and pulled the three near-identical `if accept { prefix + command }` blocks into one `with_accept_prefix` function with a test covering the accept x empty/non-empty matrix — including "edited command cleared to empty must not auto-execute," which had no coverage before.

crates/atuin/src/command/client/search/interactive.rs
wolf: Addressed in 86911919, though not literally as `Option<Result<PathBuf, _>>` — a bare `PathBuf` cant represent `EDITOR="code --wait"` (program + args), which is a common, already-supported config, so that would've been a regression.

Went with `EditorCommand { program: PathBuf, args: Vec<String> }` instead, plus a `VisualEditorError` (`NotFound` / `UnparseableCommand`) replacing the ad hoc `eyre::eyre!` calls. The `shlex::split` moved into `get_visual_editor` (a `parse_editor_var` helper) so it happens once instead of being redone in `visual_edit_command`.

One thing worth flagging: `program` isn't guaranteed to exist on disk in all cases — for the `/usr/bin/editor` and PATH-found `vim`/`vi` fallbacks it is (we verify before returning), but for the env-var case it's often just a bare name like `EDITOR=vim` that the OS resolves via `$PATH` at exec time. I left that as-is rather than adding an existence check, since checking would incorrectly reject valid bare-name configs. Also kept the resolved path from the `vim`/`vi` PATH search instead of discarding it (it used to find a match via `.any()` and throw the path away, making `Command::new` re-walk `$PATH`).

crates/atuin/src/command/client/search/interactive.rs
wolf: I don't think I'm entitled to choose here. Vim is a historical default. It was reasonable for me to start there. But the choice of fallback editor must match the "personality" of Atuin. Atuin exists to hold your hand in a way people truly enjoy. `nano` might fit that personality better than `vim`. I can absolutely figure out reasonable technical problems. But I did not choose, nor do I own the personality of Atuin. I don't feel like I have the authority to choose in this situation. What is the right way to decide this?

crates/atuin/src/command/client/search/interactive.rs
wolf: Dug into this — cmd.exe isnt actually in scope: Atuin doesnt support it at all. `atuin init <shell>` only accepts zsh/bash/fish/nu/xonsh/powershell (no `cmd` variant), `Shell` has no `Cmd` variant, and `Shell::default_shell()` literally does `if cfg!(windows) { return Ok(Self::Powershell); }`. Theres no key-binding/history-hook install path for cmd.exe anywhere in the codebase, so theres no "reasonable default editor for cmd.exe" question to answer here — the feature never reaches it.

The real target is PowerShell, and looking at it with that lens turned up two concrete things (neither is `notepad.exe`):

1. `Command::new(program)` wont resolve `.cmd`/`.bat` shims on Windows — a known `std::process::Command` gotcha, since `CreateProcess` needs an exact name match unlike a real shell. VS Code ships as `code.cmd`, so a perfectly normal `$Env:EDITOR = "code --wait"` would silently fail to launch through this code path. Worth fixing or at least flagging on its own — havent verified on an actual Windows box yet, but this is documented Rust-on-Windows behavior.
2. The `vim`/`vi` PATH-probe fallback checks bare names with no `.exe`, so its effectively dead on Windows — falls through to the existing "No editor found" error, which is honest and actionable, not silent breakage. Lower priority than #1.

Also worth checking: the `#[cfg(unix)]` block that reopens `/dev/tty` for the spawned editor has no Windows equivalent. If PowerShells shell integration hits the same fd-swap problem this works around, theres already a working pattern for it in this same file — `TerminalWriter::ConOut` opens `CONOUT$` with a UTF-8 codepage dance, restored on `Drop`.

I dont have a Windows machine to confirm #1 or the fd-swap question, and CI is `ubuntu-latest` only, so none of this is exercised today. @markovejnovic — Id like you to decide whether the `.cmd`/`.bat` issue should gate this PR or get filed as a follow-up issue instead; Im fine either way, but its your projects call, not mine.

PR #2859, opened 2025-08-04
add atuin daemon user service for %{_userunitdir} /usr/lib/systemd/user

13werwolf13:
## Checks

### DIFF
--- a/systemd/atuin-daemon.service
+++ b/systemd/atuin-daemon.service
@@ -1,0 +1,12 @@
+[Unit]
+Description=Atuin Shell History Sync Daemon
+After=network.target
+
+[Service]
+Type=simple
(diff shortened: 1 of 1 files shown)


ellie:
I'm sorry, but we won't be including this. The daemon is intended as a workaround for zfs issues, and will likely be removed in the future. Right now we also leave it entirely up to the users how they wish to run it - using the daemon _should_ be intentional, planned, and not the default.

mcilloni:
note: zfs released version 2.3.4 yesterday, which includes the fix for [this issue](https://github.com/openzfs/zfs/issues/14290). I've been using atuin on this newly released zfs version and the issue immediately stopped appearing as soon as I updated zfs

PR #2861, opened 2025-08-05
fix: dissociate `command_chaining` from `enter_accept`

ltrzesniewski:
I noticed that `command_chaining` requires `enter_accept` to be enabled in order to work. I suppose this wasn't intentional, so this PR dissociates these two features.

Also, xonsh doesn't handle `__atuin_chain_command__:`, so I separated the shell lists for these features.

One question though: is `__atuin_chain_command__:` necessary? Atuin could return the input query with the selected command appended, and that would be compatible with all shells. I suppose `query.join(" ")` could change the whitespace in some cases, but is there another reason?

## Checks

### DIFF
--- a/crates/atuin/src/command/client/search/interactive.rs
+++ b/crates/atuin/src/command/client/search/interactive.rs
@@ -1251,8 +1251,8 @@ pub async fn history(
         InputAction::Accept(index) if index < results.len() => {
             let mut command = results.swap_remove(index).command;
 
-            if is_command_chaining && (utils::is_zsh() || utils::is_fish() || utils::is_bash()) {
-                command = String::from("__atuin_chain_command__:") + &command;
+            if is_command_chaining {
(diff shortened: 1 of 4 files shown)


ellie:
> Atuin could return the input query with the selected command appended, and that would be compatible with all shells. 

You're right; this would be a much better way to do it! Would you be able to make the change here? Thanks for catching this!

ltrzesniewski:
Sure, but that would need to wait for a week, I'm currently AFK.

ltrzesniewski:
@ellie I added the commit you asked for. This reverts the changes made in the scripts in #2834.

I also chose to use `original_query.trim_end()` before appending the space because I ended up too much with output such as `echo foo &&  echo bar` (2 spaces after `&&`) when testing this, as I just add a space by muscle memory after `&&`. Feel free to remove the `trim_end()` if you'd rather keep the old behavior though.

ellie (approved):
thank you!

Issue #2490, opened 2024-12-20
[Feature Request]: Can the config entry for socket_path expand env variables


lespea:
Right now, at least as far as I can tell, there's no way to have `[daemon]`'s `socket_path` resolve to different paths on different machines.  This makes using `systemd_socket` difficult to use if the config is shared between machines since you can't reference `$XDG_RUNTIME_DIR`.  Also maybe it would make sense to change the default `socket_path` if `systemd_socket` is set since most likely the user will want the socket in `$XDG_RUNTIME_DIR`?

PR #2870, opened 2025-08-11
fix: expand path for daemon.socket_path

paulbarton90:
This parses the daemon.socket_path config item to allow it to be set to something like `"${XDG_RUNTIME_DIR}/atuin.sock"`.

Fixes https://github.com/atuinsh/atuin/issues/2490.

## Checks


ellie (approved):
lgtm - thank you!

PR #2878, opened 2025-08-16
Make metrics optional as feature `server-metrics`

nc7s:
First of all, this is more or less a Debian-centric change. We (well, I) didn't feel like packaging the metrics crate, and as many cases have shown, some people using Debian are kinda mad about telemetry, so it's disabled with a patch. However maintaining a patch is more or less a burden. With this upstreamed, I can disable it with a single switch. On a less selfish note, this could benefit other distro maintainers who have similar concerns.

Second, I suppose it's also nice to have for users who just don't want metrics, and like the benefit of reduced code lines and binary size.

Consider this a PoC, some names and the way I no-oped some things (with an empty substitute when `#[cfg(not(feature = "metrics"))]`) can be discussed.

## Checks

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,6 +45,7 @@ tracing = "0.1"
 sql-builder = "3"
 tempfile = { version = "3.19" }
 minijinja = "2.9.0"
+metrics = "0.21.1"
 
 [workspace.dependencies.tracing-subscriber]
 version = "0.3"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -453,7 +453,7 @@ dependencies = [
  "config",
  "eyre",
  "fs-err",
- "metrics",
+ "metrics 0.21.1",
  "metrics-exporter-prometheus",
  "postmark",
  "rand 0.8.5",
@@ -491,7 +491,7 @@ dependencies = [
  "atuin-server-database",
  "eyre",
  "futures-util",
- "metrics",
+ "metrics 0.21.1",
  "rand 0.8.5",
  "serde",
  "sqlx",
@@ -509,7 +509,7 @@ dependencies = [
  "atuin-server-database",
  "eyre",
  "futures-util",
- "metrics",
+ "metrics 0.21.1",
  "serde",
  "sqlx",
  "time",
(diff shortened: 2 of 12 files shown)


nc7s:
CI failures seem to be all due to network conditions in github actions' network. Could they be retried?

ellie:
The CI failures are actually as a result of changing some dependency versions. The newer versions have different ssl dependencies that we still need to fix

Would you be able to make this only change the features please?

> some people using Debian are kinda mad about telemetry, so it's disabled with a patch

This actually has absolutely nothing to do with telemetry. It allows server operators to keep an eye on how their server is performing - query latency, http latency, etc.

nc7s:
> The CI failures are actually as a result of changing some dependency versions. The newer versions have different ssl dependencies that we still need to fix
> 
> Would you be able to make this only change the features please?

Done!

> > some people using Debian are kinda mad about telemetry, so it's disabled with a patch
> 
> This actually has absolutely nothing to do with telemetry. It allows server operators to keep an eye on how their server is performing - query latency, http latency, etc.

I know, it's not "big co telemetry", but some are kinda "paranoid", who don't like anything phoning home, the "home" big or small, heh.

Should probably make a further PR that provides opt-in/out for this feature.

Cargo.toml
```diff
@@ -45,6 +45,7 @@ tracing = "0.1"
 sql-builder = "3"
 tempfile = { version = "3.19" }
 minijinja = "2.9.0"
+metrics = "0.21.1"
```
ellie, line 48: this change is the issue. please don't try and update metrics alongside adding this feature. the metrics update is, unfortunately, far more complex than that.

ellie:
> I know, it's not "big co telemetry", but some are kinda "paranoid", who don't like anything phoning home, the "home" big or small, heh.

I think you've misunderstood; there is no phoning home, of any kind. The metrics feature is literally exporting prometheus metrics, for the server operator

I can't really see much use in disabling that

nc7s:
> > I know, it's not "big co telemetry", but some are kinda "paranoid", who don't like anything phoning home, the "home" big or small, heh.
> 
> I think you've misunderstood; there is no phoning home, of any kind. The metrics feature is literally exporting prometheus metrics, for the server operator

The server is "home" for the _operator_. The _user_ may or may not trust the operator. I guess that's where our understandings part. You could argue the user already trusts the operator when they started using the service, though telemetry is something else.

Anyway. This PR alone seems now less useful as it's in the binary for all the later, unlike something more flexible like a client side option to (temporarily) disable metrics.

ellie:
> The server is "home" for the operator. The user may or may not trust the operator. I guess that's where our understandings part. 

The metrics include things like

- http request timing
- how many records are being uploaded/downloaded, globally
- etc

and nothing that could violate privacy or similar. I understand I can't control what gets packaged on debian, but would appreciate keeping the divergence as small as possible, considering users end up in our issues for support

nc7s:
This has been closed for a while, though I revisited it today when upgrading the Debian package to latest version; after another, deeper read-through I find my previous belief wrong: the metrics code is pure server-side aggregation, while the client doesn't send any "metrics" or "telemetry" per se. There's indeed no privacy concerns. I think we kinda talked pass each other lol.

For later reference.

PR #2885, opened 2025-08-24
build: remove legacy Apple SDK frameworks

EricCrosson:
Thanks for your work on atuin! I've loved using it over the past year.

According to nixos.org[^1], the `darwin.apple_sdk` frameworks have been stubs for some time, and are now removed. Evaluating `atuin` on darwin on `nixpkgs-unstable` results in the following evaluation error:

```
error: darwin.apple_sdk_11_0 has been
  removed as it was a legacy compatibility stub; see
  <https://nixos.org/manual/nixpkgs/stable/#sec-darwin-legacy-frameworks>
  for migration instructions
```

According to those linked docs, the fix is to remove references to these SDKs, which will not have any other effect since they were stubs anyway.

[^1]: https://nixos.org/manual/nixpkgs/stable/#sec-darwin-legacy-frameworks

## Checks

### DIFF
--- a/atuin.nix
+++ b/atuin.nix
@@ -10,9 +10,6 @@
   installShellFiles,
   rustPlatform,
   libiconv,
-  Security,
-  SystemConfiguration,
-  AppKit,
 }:
 rustPlatform.buildRustPackage {
   name = "atuin";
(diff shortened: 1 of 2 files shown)


ellie (approved):
makes sense, thank you for the fix! glad Atuin has been working well for you

Issue #2718, opened 2025-04-24
[Bug]: Up-key bidning not working with Atuin 18.5.0 / fish 4.0 in some terminal emulators


pbek:
### What did you expect to happen?

When pressing, `Up` the glorious Atuin history should appear in terminal emulator Ghostty or Kitty in Linux.

In Atuin 18.4.0 with the fish 4.0 patch this was working great on NixOS with fish 4.0. After updating to Atuin 18.5.0 it stopped working.
Funny thing is it is still working in Konsole or xterm with Atuin 18.5.0!

### What happened?

Up-Key just adds the previous command to the prompt.
The "up" bindings are still present in fish.

```bash
➜ bind | rg atuin
bind ctrl-r _atuin_search
bind -M insert ctrl-r _atuin_search
bind \eOA _atuin_bind_up
bind \e\[A _atuin_bind_up
bind -M insert -k up _atuin_bind_up
bind -M insert \eOA _atuin_bind_up
bind -M insert \e\[A _atuin_bind_up
```

### Atuin doctor output

```yaml
{
  "atuin": {
    "version": "18.5.0",
    "sync": {
      "cloud": false,
      "records": true,
      "auto_sync": true,
      "last_sync": "2025-04-24 9:42:46.782123328 +00:00:00"
    },
    "sqlite_version": "3.46.0"
  },
  "shell": {
    "name": "fish",
    "default": "fish",
    "plugins": [
      "atuin"
    ],
    "preexec": "built-in"
  },
  "system": {
    "os": "NixOS",
    "arch": "x86_64",
    "version": "25.05",
    "disks": [
      {
        "name": "zroot/root",
        "filesystem": "zfs"
      },
      {
        "name": "zroot/root/nix",
        "filesystem": "zfs"
      },
      {
        "name": "zroot/root/nix",
        "filesystem": "zfs"
      },
      {
        "name": "zroot/root/home",
        "filesystem": "zfs"
      },
      {
        "name": "/dev/nvme0n1p1",
        "filesystem": "vfat"
      },
      {
        "name": "/dev/sdc3",
        "filesystem": "ext4"
      }
    ]
  }
}
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

komputerwiz:
I ran into this and decided to investigate more deeply. It's a perfect storm of Fish updates and minor terminfo differences for newer terminal emulators (I'm a Ghostty user myself).

[...]

Issue #2803, opened 2025-06-30
New fish-git doesn't like `bind -k` anymore


xircon:
`atuin init fish | source` contains:
```
bind -M insert -k up _atuin_bind_up 
```
Which generates the error:
```
bind: the -k/--key syntax is no longer supported. See `bind --help` and `bind --key-names`
```
EndeavourOS,  fish 2:4.0.2.r1147.g4d67ca7c5-1,  atuin 18.6.1-1.1.

hogdanish:
Since this hasn't been fixed yet, here's a hacky workaround to silence the error message that should still preserve Atuin's functionality:

```fish
# fish 4.0 deprecates `bind -k`. transform's Atuin's init to drop -k and ensure up-binding works
if type -q atuin
    set -l __atuin_init (atuin init fish | string replace -ra -- 'bind -M ([^ ]+)\s+-k ' 'bind -M $1 ' | string replace -ra -- 'bind\s+-k ' 'bind ')
    if test -n "$__atuin_init"
        printf '%s\n' $__atuin_init | source
        if functions -q _atuin_bind_up
            bind up _atuin_bind_up
            bind -M insert up _atuin_bind_up
        end
    else
        # fallback: source unmodified but silence deprecation noise
        atuin init fish 2>/dev/null | source
    end
end
```

svenjacobs:
Fish [4.1.0](https://github.com/fish-shell/fish-shell/releases/tag/4.1.0) has been released which removed this option. I now get this error every time a start a shell.

ph0sgene:
All i did was remove `-k` and the error message is now gone and atuin works.

part of `atuin init fish` has this at the bottom. so in my `~/.config/fish/config.fish` i changed this
```sh
...
bind \eOA _atuin_bind_up
bind \e\[A _atuin_bind_up
if bind -M insert > /dev/null 2>&1
bind -M insert \cr _atuin_search
bind -M insert -k up _atuin_bind_up  # this is the culprit
bind -M insert \eOA _atuin_bind_up
bind -M insert \e\[A _atuin_bind_up
end
...

```

to this
```sh
...
 bind -M insert up _atuin_bind_up
...
```

<img width="1724" height="741" alt="Image" src="https://github.com/user-attachments/assets/73f816dd-aec8-4b89-ac88-362befd5d09a" />

not-reed:
quick (temp) fix for you fish config:

replace 
```
atuin init fish | source
```

with

```
[...]

PR #2902, opened 2025-09-13
fix: fish up binding

r-vdp:
https://github.com/atuinsh/atuin/pull/2677 introduced logic based on `$FISH_VERSION` to determine how to bind keys in fish.
However, this env var is not exported to child processes, as you can see by running `sh -c 'echo $FISH_VERSION'` from fish.

Instead, we can simply push the if statement into the actual fish code that we generate, since fish is aware of its own version, and then we can decide which binding syntax to use.

## Checks

### DIFF
--- a/crates/atuin/src/command/client/init/fish.rs
+++ b/crates/atuin/src/command/client/init/fish.rs
@@ -10,30 +10,27 @@ pub fn init_static(disable_up_arrow: bool, disable_ctrl_r: bool) {
     // We keep it for compatibility with fish 3.x
     if std::env::var("ATUIN_NOBIND").is_err() {
         const BIND_CTRL_R: &str = r"bind \cr _atuin_search";
+        const BIND_UP_ARROW: &str = r"if string match -q '4.*' $version
+    bind up _atuin_bind_up
+else
+    bind -k up _atuin_bind_up
+end
+bind \eOA _atuin_bind_up
+bind \e\[A _atuin_bind_up";
         const BIND_CTRL_R_INS: &str = r"bind -M insert \cr _atuin_search";
-        const BIND_UP_ARROW_INS: &str = r"bind -M insert -k up _atuin_bind_up
+        const BIND_UP_ARROW_INS: &str = r"if string match -q '4.*' $version
+    bind -M insert up _atuin_bind_up
+else
+    bind -M insert -k up _atuin_bind_up
+end
 bind -M insert \eOA _atuin_bind_up
 bind -M insert \e\[A _atuin_bind_up";
-
-        let bind_up_arrow = match std::env::var("FISH_VERSION") {
-            Ok(ref version) if version.starts_with("4.") => r"bind up _atuin_bind_up",
-            Ok(_) => r"bind -k up _atuin_bind_up",
-
-            // do nothing - we can't panic or error as this could be in use in
-            // non-fish pipelines
-            _ => "",
-        }
-        .to_string();
 
         if !disable_ctrl_r {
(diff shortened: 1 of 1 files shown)


r-vdp:
Additionally, I think that it makes more sense to put this logic inside the generated fish code, since it allows for generating the code from another shell or another fish version than the one where the generated code will be executed.

theodore-s-beers:
Fwiw, I'm a Fish & Atuin user, and my Atuin up binding stopped working a few days ago – I guess with Fish 4.0.6. For the time being, I've fixed this by adding `bind up _atuin_bind_up` to my Fish config. But it will be nice to have the `init` script fixed. Thanks.

anson-vandoren:
Same boat - upgraded to Fish 4.0.6 and up-arrow binding stopped working. But, @r-vdp , please don't forget about us insert-mode users 😂 

```dif
diff --git a/crates/atuin/src/command/client/init/fish.rs b/crates/atuin/src/command/client/init/fish.rs
index 954a7bd6..af076fed 100644
--- a/crates/atuin/src/command/client/init/fish.rs
+++ b/crates/atuin/src/command/client/init/fish.rs
@@ -11,26 +11,16 @@ pub fn init_static(disable_up_arrow: bool, disable_ctrl_r: bool) {
     if std::env::var("ATUIN_NOBIND").is_err() {
         const BIND_CTRL_R: &str = r"bind \cr _atuin_search";
         const BIND_CTRL_R_INS: &str = r"bind -M insert \cr _atuin_search";
-        const BIND_UP_ARROW_INS: &str = r"bind -M insert -k up _atuin_bind_up
+        const BIND_UP_ARROW_INS: &str = r"if string match -q '4.*' $version; bind -M insert up _atuin_bind_up; else; bind -M insert -k up _atuin_bind_up; end
```

Because GitHub is stupid and the line I want to suggest on is too high above the lines you edited, I can't do an inline suggestion... 🤦🏻

r-vdp:
Ah, right! I'll add this to the diff later today

anson-vandoren:
just opened a PR against your PR branch 🎉

r-vdp:
> just opened a PR against your PR branch 🎉

Cool. To avoid a merge commit in this PR, I squashed your commit into mine and added an attribution using `Co-authored-by`, thanks!

r-vdp:
I pushed another commit to (IMHO) clean up a bit the code for the fish key bindings and make it easier to see what's going on.

Let me know if it would be better to submit that in a separate PR after this one.

matthewberryman:
Can this PR be merged (incl. @anson-vandoren's work) ASAP @ellie ? Fish 4.1.0 has been released which removes the -k option entirely!

ellie:
Thanks for the ping - was a little unclear if this one was aactually ready

ellie:
Otherwise - is it standard for fish to introduce breaking changes in a patch release?

matthewberryman:
Thanks @ellie for merging - I had done a code review and tested the resulting binary (now left a comment to that effect, sorry for my tardiness).
The [release notes for 4.1.0](https://github.com/fish-shell/fish-shell/releases/tag/4.1.0) had mentioned the removal (and the prior deprecation). 
Also noting this should fix both https://github.com/atuinsh/atuin/issues/2803 and https://github.com/atuinsh/atuin/issues/2913 (since it uses the internal version and not the env var which isn't available to child processes like atuin). I would have thought it would also fix https://github.com/atuinsh/atuin/issues/2718 but I'm not on Linux (at least not one with a GUI terminal).

Reverier-Xu:
hi, when will you create a release for this fix?

PR #2903, opened 2025-09-14
filter by session with `zsh-autosuggestions` and `nosharehistory`

n0099:
## Checks

When `sharehistory` setopt is not enabled for zsh with plugin `zsh-autosuggestions` whose `ZSH_AUTOSUGGEST_STRATEGY` array contains `atuin` due to https://github.com/zsh-users/zsh-autosuggestions/issues/797, the suggested text will show the result of `atuin search` with default option `--filter-mode global`, but the prompted text after press <kbd>Up</kbd> will comes from `$(history | tail -n1)`.

### DIFF
--- a/crates/atuin/src/shell/atuin.zsh
+++ b/crates/atuin/src/shell/atuin.zsh
@@ -16,11 +16,11 @@ zmodload zsh/datetime 2>/dev/null
 # in your .zshrc
 _zsh_autosuggest_strategy_atuin() {
     if [[ -o nosharehistory ]]; then
-        filter_mode='--filter-mode session'
+        # silence errors, since we don't want to spam the terminal prompt while typing.
+        suggestion=$(ATUIN_QUERY="$1" atuin search --cmd-only --limit 1 --search-mode prefix --filter-mode session 2>/dev/null)
+    else
+        suggestion=$(ATUIN_QUERY="$1" atuin search --cmd-only --limit 1 --search-mode prefix 2>/dev/null)
     fi
-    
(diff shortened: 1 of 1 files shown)


crates/atuin/src/shell/atuin.zsh
```diff
@@ -15,8 +15,12 @@ zmodload zsh/datetime 2>/dev/null
 # you'd like to override this, then add your config after the $(atuin init zsh)
 # in your .zshrc
 _zsh_autosuggest_strategy_atuin() {
+    if [[ -o nosharehistory ]]; then
+        filter_mode='--filter-mode session'
+    fi
+    
     # silence errors, since we don't want to spam the terminal prompt while typing.
-    suggestion=$(ATUIN_QUERY="$1" atuin search --cmd-only --limit 1 --search-mode prefix 2>/dev/null)
+    suggestion=$(ATUIN_QUERY="$1" atuin search --cmd-only --limit 1 --search-mode prefix "$filter_mode" 2>/dev/null)
```
akinomyoga: When `nosharehistory` is enabled, this produces an error `error: unexpected argument '--filter-mode session' found`. When `nosharehistory` is disabled, this passes an empty argument `''` to Atuin and overrides the current command line passed by `ATUIN_QUERY`.

n0099: This option exists: https://github.com/atuinsh/atuin/blob/5614596c8165e6f5e22b5524c7bb127bf55bc48a/crates/atuin/src/command/client/search.rs#L67-L69

PR #2905, opened 2025-09-14
feat: add session-preload filter mode to include global history from before session start

chitao1234:
This mode mimics the default behavior for many shells.

## Checks

### DIFF
--- a/crates/atuin-client/src/database.rs
+++ b/crates/atuin-client/src/database.rs
@@ -20,6 +20,7 @@ use sqlx::{
     },
 };
 use time::OffsetDateTime;
+use uuid::Uuid;
 
 use crate::{
     history::{HistoryId, HistoryStats},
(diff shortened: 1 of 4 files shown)


crates/atuin-client/src/database.rs
```diff
@@ -75,6 +76,16 @@ pub fn current_context() -> Context {
     }
 }
 
+fn get_session_start_time(session_id: &str) -> Option<i64> {
```
ellie, line 79: This is a really neat way of getting the start time 😊

ellie (approved):
Looks good to me, thank you for the work!

arnfaldur:
This feature could use some documentation. I just noticed a new fifth option when cycling the filter modes, which I do frequently. I'd say adding it is a pretty significant change, as the default four are already on the high end of being convenient to cycle through. 
I imagine the available options can be disabled via config, but I think this change merits a question of what the sane defaults are. 
I don't know if the default config has been adjusted somehow, but then considerations for existing users with existing configs are in order.
In any case, some explanations explaining what `SESSION+` does is important. I found nothing on docs.atuin.sh, and nothing on duck duck go about this. I only found this through searching the github repo and using git blame.

If anyone is following my footsteps, I believe `SESSION+` is the same as the `GLOBAL` filter but it excludes any command that was executed after the current session was started. I might be wrong on this.

PR #2911, opened 2025-09-17
bug: Abort install.sh on cURL fails

dkanas:
This PR is an attempt to fix the case, in which the script continues it's execution after the `curl` command for getting the installer script fails. I have opted for removing the `__atuin_install_binary` function in order to simplify the flow of the script. Otherwise, an additional conditional statement would have to be added to check whether the function executed correctly.

## Checks


markovejnovic:
Thank you!

PR #2928, opened 2025-09-30
feat: add various acceptance keys

keithamus:
This change introduces (optional) acceptance keys of Backspace and Left Arrow, when at the start of a line. These two are common muscle memory actions for users.

The configuration defaults to false so as not to disrupt existing user patterns.

This also adds a test that exercises the various acceptance modes, which as it turns out was quite easy to do.

I discussed this on discord where [Ellie suggested I raised an issue](https://discord.com/channels/954121165239115808/1421180955657244703/1422642337481228400), but I felt like a PR would be more tangiable. I've tested this locally and I'm very happy with how these keys work, it fits my needs well.

`exit_past_line_start` and `accept_past_line_start` can technically co-exist. When this happens `accept_past_line_start` takes precedence. Is this okay, or should we reconsider the config? Perhaps `acceptance_keys = []` would be better here? I'm very open to changes here.

## Checks

### DIFF
--- a/crates/atuin-client/src/settings.rs
+++ b/crates/atuin-client/src/settings.rs
@@ -342,6 +342,8 @@ pub struct Keys {
     pub scroll_exits: bool,
     pub exit_past_line_start: bool,
     pub accept_past_line_end: bool,
+    pub accept_past_line_start: bool,
+    pub accept_with_backspace: bool,
     pub prefix: String,
 }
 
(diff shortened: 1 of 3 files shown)


crates/atuin-client/src/settings.rs
```diff
@@ -342,6 +342,10 @@ pub struct Keys {
     pub scroll_exits: bool,
     pub exit_past_line_start: bool,
     pub accept_past_line_end: bool,
+    #[serde(default)]
```
keithamus: For some reason I needed `#[serde(default)]` here or atuin broke without the keys in my config. Perhaps I'm missing something obvious here? Like maybe config is merged with a default one? I couldn't spot that sort of thing (though I didn't look super hard).

keithamus: Nvm found the right place.

PR #2931, opened 2025-10-01
feat: implement database vacuuming.

### LINKED ISSUE #2927: Consider recommending VACUUM for SQLite database maintenance
Hi there,

I noticed that over time, the `~/.local/share/atuin/history.db` file can grow quite large. Out of curiosity, I ran the following command:

```sh
sqlite3 ~/.local/share/atuin/history.db "VACUUM;"
```

Afterward, the database size was significantly reduced.

Would it make sense for Atuin to:

Recommend running VACUUM in the docs as a maintenance step?
Or possibly run it automatically in some scenarios (e.g. after deleting a large amount of history)?

I’m not sure if there are trade-offs (like performance implications or locking concerns), but it seems like it could help keep storage usage efficient.

Thanks for considering this, and for the awesome project.

keithamus:
This change attempts to resolve https://github.com/atuinsh/atuin/issues/2927 by introducing the ability to VACUUM the database by:

 - giving atuin-server-database the `vacuum` method. This simply calls `sqlx::query("VACUUM")`. I also added a test, as we're using sqlx & sqlite this test can be executed in memory. Sadly the same cannot happen for postgres, but the test is kind of trivial.

 - exposing a vacuum endpoint on the server. This handler calls vacuum. This can be useful for users who wish to manually vacuum a db, but this is not exposed via the CLI, instead...

 - adding `auto_vacuum_threshold` option to the server configuration. `auto_vacuum_threshold` is > 0, then the server will increment the `delete_count` atomic counter, and when this counter reached the threshold it will call vacuum and reset the counter.

The default for `auto_vacuum_threshold` is `0` meaning "never vacuum", but users can set this to a value which will come down to their preference, guided by:

 - How long they anticipate their server running for. For example setting it to a high-ish number like 1000 would probably be appropriate for a server running on a machine which has long uptimes. For self hosting for a single user, this might be a little long though, keeping in mind that if the server goes down (or the mahcine goes down) the counter will reset to 0.

 - For users who _really care_ about disk space and don't really care about latency on deletes, setting to `1` will will vacuum on every delete.

I considered adding some kind of "recommended" number but I don't think there is a good number to unilaterally recommend for this.

### DIFF
--- a/crates/atuin-server-database/src/lib.rs
+++ b/crates/atuin-server-database/src/lib.rs
@@ -142,6 +142,8 @@ pub trait Database: Sized + Clone + Send + Sync + 'static {
 
     async fn oldest_history(&self, user: &User) -> DbResult<History>;
 
+    async fn vacuum(&self) -> DbResult<()>;
+
     #[instrument(skip_all)]
     async fn calendar(
         &self,
--- a/crates/atuin-server-postgres/src/lib.rs
+++ b/crates/atuin-server-postgres/src/lib.rs
@@ -672,6 +672,16 @@ impl Database for Postgres {
 
         Ok(status)
     }
+
+    #[instrument(skip_all)]
+    async fn vacuum(&self) -> DbResult<()> {
+        sqlx::query("VACUUM")
+            .execute(&self.pool)
+            .await
+            .map_err(fix_error)?;
+
+        Ok(())
+    }
 }
 
 fn into_utc(x: OffsetDateTime) -> PrimitiveDateTime {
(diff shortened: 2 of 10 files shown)


ellie:
Might we be better off by enabling `PRAGMA auto_vacuum = full` on the server?

keithamus:
Sounds sensible. I'm not a DB expert so I don't know what heuristics `auto_vacuum` uses.

markovejnovic:
relevant post https://www.reddit.com/r/sqlite/comments/104ts8m/is_auto_vacuum_a_good_idea/

markovejnovic:
Thank you for the PR. Unfortunately, the code has changed so severely that managing this change is a little bit difficult. https://github.com/atuinsh/atuin/pull/3634 and https://github.com/atuinsh/atuin/pull/4009 are the two largest offenders for vacuuming.

I will re-visit the original issue when I have time. Feel free to ping me if you think I have unfairly closed your PR.

Issue #2769, opened 2025-05-25
[FR] option to redact secrets rather than delete


huyz:
Regardless of secrets, I'd like to have a complete history of commands in the DB, and I'd like to benefit from the time-saving aspects of being able to look up a command and its arguments as previously used.

Can we redact secrets rather than filter out the entire command line entry?

mhagnumdw:
It makes sense to me.

PR #2932, opened 2025-10-01
feat: more accurately filter secret tokens

keithamus:
This change tries to more accurately capture & filter secret keys by using regular expressions for the key contents, rather than trying to detect the environment variable names. For example rather than filtering matches for `AWS_SECRET_ACCESS_KEY`, this will filter for strings that take the shape of an AWS key. This provides two real benefits & 1 tangential benefit:

- (Biggest benefit): Prior to this change, if a typo is made in the keys name (e.g. `AWS_SECERT_ACCESS_KEY=<my-actual-key>`) the credential was leaked. With this change the key will still be filtered.
- (Small benefit) When trying to test commands with fake tokens (e.g. `AWS_SECRET_ACCESS_KEY=foo my-command`) then these are filtered out, which isn't always exactly desirable, as test commands like these are useful for reference.
- (Tangential benefit) This is a requirement for implementing https://github.com/atuinsh/atuin/issues/2769

## Checks

### DIFF
--- a/crates/atuin-client/src/secrets.rs
+++ b/crates/atuin-client/src/secrets.rs
@@ -17,18 +17,29 @@ pub static SECRET_PATTERNS: &[(&str, &str, TestValue)] = &[
     ),
     (
         "AWS Secret Access Key env var",
-        "AWS_SECRET_ACCESS_KEY",
-        TestValue::Single("AWS_SECRET_ACCESS_KEY=KEYDATA"),
+        "(?:[^A-Za-z0-9/+=])?([A-Za-z0-9/+=]{40})(?:[^A-Za-z0-9/+=])?",
+        TestValue::Multiple(&[
+            "wJalrXUtnFEMI/K7MDENG/bPxRfiCYEXAMPLEKEY", // https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_access-keys.html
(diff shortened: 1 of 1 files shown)


crates/atuin-client/src/secrets.rs
```diff
@@ -17,18 +17,29 @@ pub static SECRET_PATTERNS: &[(&str, &str, TestValue)] = &[
...
+            "AAAAAAAAAAAAAAAA.BBBBBBBBBBBBBBB.CCCCCCCCCCC",
+        ]),
     ),
     (
         "Microsoft Azure secret access key env var",
-        "AZURE_.*_KEY",
-        TestValue::Single("export AZURE_STORAGE_ACCOUNT_KEY=KEYDATA"),
+        "(?:sk-[A-Za-z0-9]{48,}|[A-Za-z0-9+/]{86}={2}|[A-Za-z0-9+/]{87}=|[A-Za-z0-9+/]{88})",
+        TestValue::Multiple(&[
+            "sk-123abc456def789ghi012jkl345mno678pqr901stu234vwx567yz890",
```
keithamus, line 39: This might be wrong? I had difficulty finding examples of these secrets in the Microsoft Azure docs so ended up doing a code search in GitHub and kind of reverse engineering the tags I found.

ellie (approved):
This is great, thank you so much!

PR #2966, opened 2025-10-26
ci: add Windows builds, second try

ltrzesniewski:
Hi @ellie, I don't want to be a burden, but I'd like to ask if you could please reconsider adding Windows builds now that PowerShell support is merged (#2543). 😇

Most people use PowerShell on Windows, so they'd need an artifact in the future releases in order to use it. Otherwise, they'd have to install Rust and compile it themselves, which is not user-friendly.

I'd like to point out that I didn't have issues with building Atuin on Windows when I was rebasing the PowerShell PR (except some initial stuff like `#[cfg]` or a few warnings which were fixed in #2856), so I don't expect this to introduce many problems.

This is essentially an updated version of #2714.

## Checks


ellie:
I'm down to merge this, and need to add something to the readme, but my thoughts are as follows:


---
### Tier 1 support
_MacOS, Linux, what we support rn_

Failing builds and tests block releases, we run a full test suite on PRs + will not merge a PR if it breaks tier 1 platforms

### Tier 2 support
_Windows, potentially others_

We publish builds, we run tests on `main`, but failures do not block + it may end up missing from releases if the build is totally broken

Additionally, I will not be supporting Windows users here, on Discord, or on the Forum. Very happy for others to do so, but it's not something I am able to do.

---

We can fit Windows into tier 2; if @ltrzesniewski and others are willing to fix potential issues then all should be fine. We can do our best to avoid breaking changes, but large feature work/etc will not be delayed due to lack of Windows support.

ltrzesniewski:
I'm fine with that. 👍

I expected something like this as I suggested to first release PowerShell support as a beta in the PR.

There's just one thing that's unclear to me: are you going to disable Windows builds in the PRs? I'd prefer if you keep them so the PR author is aware there's a problem and could fix it if they're able to, or if it's an easy missing `#[cfg(not(windows))]` for instance. You could of course merge a PR which doesn't build on Windows and I'll try to submit a fix afterwards.

I understand you not being able to support Windows by yourself and that's something I'm willing to help you with. Please just redirect people with issues to GitHub so I (and others) can see them.

ellie:
We can run Windows on CI, but I won't set it as a blocking check. Obviously anything that's a trivial fix should be fixed, but it will be considered a "nice to have" more than "essential".

> I understand you not being able to support Windows by yourself and that's something I'm willing to help you with. Please just redirect people with issues to GitHub so I (and others) can see them.

Thank you so much for the support here! I really appreciate you stepping in

ellie (approved):

Issue #2969, opened 2025-10-27
[Bug]: atuin should not try to bind keys when line editor is not available or when up/ctrl-r are disabled


darkfeline:
### What did you expect to happen?

Loading atuin doesn't cause errors in Bash when line editor is disabled

### What happened?

This new snippet causes problems:

```
__atuin_macro_chain='\e[0;0A'
for __atuin_keymap in emacs vi-insert vi-command; do
    bind -m "$__atuin_keymap" "\"$__atuin_macro_chain\": \"\""
done
unset -v __atuin_keymap
```

When Bash is run without line editor (e.g. in a dumb terminal like Emacs shell-mode), then all `bind` commands result in an error:

```
bash: bind: warning: line editing not enabled
```

Previously, this could be circumvented by disabling all atuin bindings using: `atuin init bash --disable-up-arrow --disable-ctrl-r`,  however the above new snippet now causes errors.

There are a couple of solutions to this.  The easiest is just to not run the above code when both up and ctrl-r are disabled.  The ideal but harder solution would be for atuin to detect if the line editor is not available and then skip all bind commands for Bash, although this may be confusing for users who aren't knowledgeable about terminals and Bash's line editor who accidentally encounter this situation.

### Atuin doctor output

```yaml
{
  "atuin": {
    "version": "18.10.0",
    "commit": "NO_GIT",
    "sync": null,
    "sqlite_version": "3.46.0"
  },
  "shell": {
    "name": "bash",
    "default": "bash",
    "plugins": [
      "atuin",
      "bash-preexec"
    ],
    "preexec": "bash-preexec"
  },
  "system": {
    "os": "Arch Linux",
    "arch": "x86_64",
    "version": "rolling",
    "disks": []
  }
}
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #2953, opened 2025-10-18
feat(bash): use Readline's accept-line for enter_accept

akinomyoga:
This PR introduces a mechanism to use Readline's `accept-line` to run the user command properly. The idea is described in [the code comment](https://github.com/atuinsh/atuin/pull/2953/files#diff-57afeb258339de1b14a8dd3fdc88d1a0e192fd186706e570c44c3ef41f7a8c6dR362-R382) in the added code. This naturally fixes #2935 because Readline's `accept-line` also performs the necessary keymap transition.

This PR also fixes the behavior of <kbd>tab</kbd> and <kbd>enter</kbd> with `enter_accept = false` `in Bash <= 3.2. In the previous implementation, the selected command was lost in Bash 3.2, but this PR correctly inserts the selected command into the command line buffer.

This PR adds a utility `atuin-bind` to make it easier to define custom keybindings. The default bindings are also set up by the new function `atuin-bind` now. This new function `atuin-bind` arranges all non-trivial setups to make it possible to call Readline's `accept-line`.

The old mechanism using `__atuin_accept_line` is kept for existing users who set up custom keybindings (without using the new function `atuin-bind`).

## Checks
- [x] I am happy for maintainers to push small adjustments to this PR, to speed up the review cycle
- [x] I have checked that there are no existing pull requests for the same thing

PR #2970, opened 2025-10-27
fix: Defer early Bash binds like macro chain (#2969)

darkfeline:
If the user is not binding any commands, then there is no need to do
the various bind workarounds.  This causes issues if the user
deliberately disabled the bindings because their Bash session does not
have the line editor enabled; attempts to call "bind" will then cause
errors.

## Checks

### DIFF
--- a/crates/atuin/src/shell/atuin.bash
+++ b/crates/atuin/src/shell/atuin.bash
@@ -111,6 +111,19 @@ __atuin_set_ret_value() {
     return ${1:+"$1"}
 }
 
+__atuin_pre_bind_functions=()
+__atuin_run_pre_bind_functions() {
+    # This is used for "helper" binds to work around limitations.
+    # These will only be called if we actually bind anything for
+    # Atuin, to avoid calling "bind" if the user disabled all Atuin
+    # bindings.
+    [[ -v __atuin_pre_bind_done ]] && return
+    for __atuin_func in "${__atuin_pre_bind_functions[@]}"; do
+        $__atuin_func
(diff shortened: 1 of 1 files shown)


darkfeline:
https://github.com/atuinsh/atuin/issues/2969

crates/atuin/src/shell/atuin.bash
akinomyoga, line 1: The call of `__atuin_bash42_bind` on line 481 also sets up keybindings on initialization, so it should also be deferred.

akinomyoga (changes requested):
OK, when preparing #2953, I haven't assumed the case where `bind` will not be used at all.

One anticipated usage of `--disable-up-arrow`, `--disbale-ctrl-r`, or `ATUIN_NOBIND` is that the user disables the default keybindings and defines keybindings manually by themselves. In this usage, the user is expected to call `atuin-bind`, which uses the macro chaining anyway. This PR breaks this expected usage.

Could you modify the PR to perform the initialization on the first call of `atuin-bind`?

crates/atuin/src/shell/atuin.bash
darkfeline: ~I will fix the macro chain first.  There are even more calls below and it seems a little tricky to defer, I would prefer making two separate commits.~

~Would it be okay to have multi commit PR or should they be separate PRs?~

Update: I thought of a good way to do this

darkfeline:
I believe I've addressed the `atuin-bind` defer as requested.

PR #2979, opened 2025-11-05
fix(nushell): reposition cursor back to original prompt after atuin invoked

Jay-Madden:
The nushell integration script has a weird quirk where when keybinding ran `atuin search --interactive` and invoked the tui, the tui would initialize on the same line and replace the prompt text. 

Gif of the prompt being replaced: 
demo

The fix: 
demo

I'm not entirely sure WHY this behavior only seemingly exists in nushell but this does fix it. There is perhaps a better or different way im not aware of. 

## Checks

### DIFF
--- a/crates/atuin/src/shell/atuin.nu
+++ b/crates/atuin/src/shell/atuin.nu
@@ -73,10 +73,14 @@ def _atuin_search_cmd [...flags: string] {
             $ATUIN_KEYBINDING_TOKEN,
             ([
                 `with-env { ATUIN_LOG: error, ATUIN_QUERY: (commandline) } {`,
+                    # Add a newline so that the atuin TUI initializes on the line below the prompt
+                    'print "";',
                     'commandline edit',
                     '(run-external atuin search',
                         ($flags | append [--interactive] | each {|e| $'"($e)"'}),
-                    ' e>| str trim)',
+                    ' e>| str trim);',
+                    # Move the cursor back up two lines to put it on the original prompt after the tui has closed
+                    '$"(ansi --escape "2A")"',
                 `}`,
             ] | flatten | str join ' '),
         ]


ltrzesniewski:
I think this is intentional: when you type something before you press Ctrl+R, Atuin replaces that prompt and uses it as a filter instead of leaving it and having two copies on the screen.



> I'm not entirely sure WHY this behavior only seemingly exists in nushell

It also behaves that way on bash. I didn't test the other shells <sub>(well, besides powershell)</sub>.

Jay-Madden:
Hmm interesting, I feel that my zsh integration has ALWAYS done the "open tui below existing prompt then put it back" 

Example with ZSH: 
image

Maybe this is a factor of the `inline_height`?

I have mine set to `30` https://github.com/Jay-Madden/dots/blob/ad262ac6b20fb1531c1e3f8676b97d146fae8297/atuin/config.toml#L48

I can go ahead and close this and just use my custom config myself if its desired behavior. But i'm not sure why zsh doesnt do it in that case.

ltrzesniewski:
I suppose the shell scripts have been written by different people, so they have slightly different behavior. I'd be interested to know what the "official" behavior should be. Maybe it could even be made configurable.

BTW, is that `$"(ansi --escape "2A")"` part correct for all prompts? You use a 2-line prompt, but other uses could have different heights.

Jay-Madden:
> Maybe it could even be made configurable.

could be worth doing, it was behavior that certainly bothered me enough to spend a few hours digging for how to revert it back to what my zsh looked like. 

> BTW, is that $"(ansi --escape "2A")" part correct for all prompts? You use a 2-line prompt, but other uses could have different heights.

I believe so, the gifs are a one line prompt and it works. I also just tested with a starship two line prompt and it worked fine. 

I think it's "go up 2 from where the tui ended. 

Which is 1 for the new prompt the tui created when it exited, then 1 to go back up to the original line.

PR #2982, opened 2025-11-07
Update the directories dependency to 6.0.0

musicinmybrain:
Update the `directories` dependency to the latest ~~and final~~ version, 6.0.0.

The SemVer-breaking change in `directories` 6 is just that `dirs-sys` is updated to 0.5, which, in turn, is just due to [updating some platform-specific dependencies](https://github.com/dirs-dev/dirs-sys-rs/compare/f369f0904bec9833572f24c988c7e48454173983...8bcd4aa2c35990d57a2cff2953793525fc42709c). There are therefore no API changes we need to worry about here.

~~Note that (as for [`dirs`](https://github.com/dirs-dev/dirs-rs)) the [repository for `directories` is archived](https://github.com/dirs-dev/directories-rs), so it might be time to start thinking about a maintained alternative.~~

## Checks

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1211,11 +1211,11 @@ dependencies = [
 
 [[package]]
 name = "directories"
-version = "5.0.1"
+version = "6.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9a49173b84e034382284f27f1af4dcbbd231ffa358c0fe316541a7337f376a35"
+checksum = "16f5094c54661b38d03bd7e50df373292118db60b585c08a411c6d840017fe7d"
 dependencies = [
- "dirs-sys 0.4.1",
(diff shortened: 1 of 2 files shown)


musicinmybrain:
I notice that this removes older/duplicate versions of `redox_users` and `dirs-sys` from `Cargo.lock`, which is nice.

musicinmybrain:
> Note that (as for [`dirs`](https://github.com/dirs-dev/dirs-rs)) the [repository for `directories` is archived](https://github.com/dirs-dev/directories-rs), so it might be time to start thinking about a maintained alternative.

Just a quick follow-up on this: [`directories`](https://crates.io/crates/directories) *is* still maintained; it has simply moved from https://github.com/dirs-dev/directories-rs to https://codeberg.org/dirs/directories-rs.

PR #3003, opened 2025-11-26
chore: clarify docstring of the enter_accept config key

waldyrious:
While the [docs for this option](https://docs.atuin.sh/configuration/config/#enter_accept) are pretty clear, the explanation in the comment in the config file was slightly ambiguous. I have rephrased it to hopefully be clearer and more explicit.

I've also fixed the comment prefix in a line that should have used `##` instead of a single `#`, which is otherwise only used in lines that can be toggled on and off, i.e. actual configuration values.

## Checks


BinaryMuse (approved):
Thanks very much!

PR #2024, opened 2024-05-16
Support importing from replxx history files

amosbird:
As discussed in Discord, this feature should be useful for managing history of applications which use https://github.com/AmokHuginnsson/replxx . One notable example would be ClickHouse.

I'll try to integrate atuin into clickhouse-client later.

## Checks
- [X] I am happy for maintainers to push small adjustments to this PR, to speed up the review cycle
- [X] I have checked that there are no existing pull requests for the same thing

PR #3005, opened 2025-11-28
feat: support additional history filenames in replxx importer

waldyrious:
The original implementation of the `replxx` importer, from PR #2024, only supported `.histfile` as a filename for the history file, since there is no default filename. However, the replxx codebase does use `replxx_history.txt` as an example history filename (see [here](https://github.com/AmokHuginnsson/replxx/blob/release-0.0.4/examples/c-api.c#L183) and [here](https://github.com/AmokHuginnsson/replxx/blob/release-0.0.4/examples/cxx-api.cxx#L383)), so this patch adds support for that as an alternative filename.

The implementation was modeled after [the one](https://github.com/atuinsh/atuin/blob/v18.10.0/crates/atuin-client/src/import/zsh.rs#L29-L44) currently used for the `zsh` history file importer, which also means the `replxx` importer now produces a human-friendly error if no history file is found in the expected path(s).

## Checks

/cc @amosbird who introduced this importer in #2024.

### DIFF
--- a/crates/atuin-client/src/import/replxx.rs
+++ b/crates/atuin-client/src/import/replxx.rs
@@ -19,8 +19,23 @@ fn default_histpath() -> Result<PathBuf> {
     let home_dir = user_dirs.home_dir();
 
     // There is no default histfile for replxx.
-    // For simplicity let's use the most common one.
-    Ok(home_dir.join(".histfile"))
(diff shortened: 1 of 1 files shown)


ellie (approved):
looks good to me, thank you!

Issue #2663, opened 2025-04-02
[Suggestion]: Don't use `AlertError` for current selection highlight


hschne:
The current theming uses the `AlertError` meaning to highlight the current history selection. I feel that isn't right. The selected item is not in an error state, so why should it be using that color? 

![Image](https://github.com/user-attachments/assets/a2943598-f1a2-424f-8fa3-3e2667e9a563)

It may make sense to use `Base`, or even better, introduce a new meaning. Possibly `BaseSelected` or something like that. The [FZF Color Scheme](https://www.mankier.com/1/fzf#Options-Global_Style) may be interesting to look for as inspiration: 

```
COLOR NAMES:
   fg                    Text
     list-fg             Text in the list section
       selected-fg       Selected line text
     preview-fg          Preview window text
   bg                    Background
     list-bg             List section background
       selected-bg       Selected line background
     preview-bg          Preview window background
     input-bg            Input window background ([--input-border](https://www.mankier.com/1/fzf#--input-border))
     header-bg           Header window background ([--header-border](https://www.mankier.com/1/fzf#--header-border))
   hl                    Highlighted substrings
   ...
```

This is kind of related #1151 which suggests adding a highlight for the current search string.

philtweir:
Thanks! The main reason for using AlertError originally was so that the initial implementation followed the de facto pattern of colour use and for minimal change.

However, I agree - the point of having the `Meaning` enum is that theming keeps semantically aligned (i.e. as far as possible, by default, if a style means X in widget A, it is the same style for the same meaning in widget B, even if it's an unrelated component). Then it is easy to sensibly tweak themes consistently in a few lines to make new ones.

That is a bit different to the fzf approach, but happy to take @ellie 's steer if a wider change would be better. In the current approach, maybe:

```
Meaning.Match
Meaning.Selection
```

as two new terms. `Selection` is to draw the user's attention (such as the current history selection) and `Match` would be for #1151 (for example), which could be bolded by default.

Any thoughts?

hschne:
> That is a bit different to the fzf approach, but happy to take @ellie 's steer if a wider change would be better. In the current approach, maybe:
> 
> ```
> Meaning.Match
> Meaning.Selection
> ```
> 
> as two new terms. `Selection` is to draw the user's attention (such as the current history selection) and `Match` would be for #1151 (for example), which could be bolded by default.
> 
> Any thoughts?

Thanks for providing some context 🙌

Selection and Match sound exactly right!

[...]

PR #3015, opened 2025-12-08
Draft:  Themes v2

philtweir:
Introduces a range of enhancement for themes:

 - new meanings: Match, Highlight, Selection, Failure, Success, Footnote
 - greater styling coverage
 - ability to express attributes, not just colours (also therefore underline and background colours)
 - _partial, restricted_ support for attribute-only fallback definitions (e.g. Highlight=AlertWarn+Bold by default)
 - deprecates `[colors]` in favour of `[styles]` in theme config

Attributes can be expressed as described [in Crossterm's documentation](https://docs.rs/crossterm/latest/crossterm/style/enum.Attribute.html). Bear in mind that, right now, general fallbacks/overrides are whole-style, i.e. if you subtheme you must override everything for that meaning or nothing, so you can say you want to inherit Selection from the parent theme (by omitting it in your theme file), or you want Selection to be blue and bold (by being explicit), but you can't say you want Selection to be bolded with its default colour.

https://github.com/user-attachments/assets/2c3a8bf0-6579-4c17-b52f-dc5590277091

Showing this to demonstrate the [SGR](https://en.wikipedia.org/wiki/ANSI_escape_code#Select_Graphic_Rendition_parameters) parameters work, but for everyone's sake, please don't use blinking text. Also, not all of these work on every terminal, and there is no fallback mechanism (other than decided by the terminal).

I am thinking of including the debug logging fixes that are not yet integrated into this PR, if @ellie reckons that is a good plan? (they are fairly small and I have used the old debug approach for a couple new lines here).

## Checks

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -467,6 +467,7 @@ dependencies = [
  "crossterm 0.29.0",
  "divan",
  "rand 0.8.5",
+ "ratatui-crossterm",
  "serde",
  "time",
  "unicode-segmentation",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -235,7 +235,7 @@ dependencies = [
  "clap_complete",
  "clap_complete_nushell",
  "colored",
- "crossterm",
+ "crossterm 0.29.0",
  "daemonize",
  "eyre",
  "fs-err",
@@ -252,7 +252,7 @@ dependencies = [
  "regex",
  "rpassword",
  "runtime-format",
- "rustix",
+ "rustix 1.1.4",
  "semver",
  "serde",
  "serde_json",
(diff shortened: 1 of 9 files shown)


philtweir:
Links to #2663

nenahp:
Sorry to bother here.. is this feature still moving on? recently atuin support shell sytnax (https://github.com/atuinsh/atuin/commit/29c3dc6888ccb2a318723c9babd2418cbc8b5d99), it's very nice but the color become harder to distinguish, so I want apply "reverse" attribute on match keyword and change the bg of the matched line to make it obvioius.

Currently it looks like this for me:
image


I want archive something like this: 
image

philtweir:
Happy to rebase and progress, including reviewing syntax highlighting implications - any feedback from @ellie first (and i know @markovejnovic is doing a UX sweep too) ?

Issue #2520, opened 2025-01-02
[Bug]: task 5 panicked with message "Client::new(): reqwest::Error { kind: Builder, source: Os { code: 2, kind: NotFound, message: \"No such file or directory\" } }"


takouhai:
### What did you expect to happen?

hit the up key to see my history in atuin after installing it (tried both normal bash install and cargo install)

### What happened?

```
thread 'main' panicked at /Users/[user]/.cargo/registry/src/index.crates.io-6f17d22bba15001f/reqwest-0.11.27/src/async_impl/client.rs:1713:38:                          ⬡ 18.20.4
Client::new(): reqwest::Error { kind: Builder, source: Os { code: 2, kind: NotFound, message: "No such file or directory" } }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Error: task 5 panicked with message "Client::new(): reqwest::Error { kind: Builder, source: Os { code: 2, kind: NotFound, message: \"No such file or directory\" } }"

Location:
    /Users/[user]/.cargo/registry/src/index.crates.io-6f17d22bba15001f/atuin-18.4.0/src/command/client/search/interactive.rs:1153:37
```

### Atuin doctor output

```yaml
Atuin Doctor
Checking for diagnostics


Please include the output below with any bug reports or issues

{
  "atuin": {
    "version": "18.4.0",
    "sync": null,
    "sqlite_version": "3.46.0"
  },
  "shell": {
    "name": "zsh",
    "default": "zsh",
    "plugins": [
      "atuin"
    ],
    "preexec": "built-in"
  },
  "system": {
    "os": "Darwin",
    "arch": "x86_64",
    "version": "14.6.1",
    "disks": [
      {
        "name": "MAC HD",
        "filesystem": "apfs"
      },
      {
        "name": "MAC HD",
        "filesystem": "apfs"
      }
    ]
  }
}
```
```


### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

jhult:
I just bumped into this as well.

```sh
thread 'main' panicked at /Users/brew/Library/Caches/Homebrew/cargo_cache/registry/src/index.crates.io-1949cf8c6b5b557f/reqwest-0.11.27/src/async_impl/client.rs:1713:38:
  Client::new(): reqwest::Error { kind: Builder, source: Os { code: 2, kind: NotFound, message: "No such file or directory" } }
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
[...]

PR #3016, opened 2025-12-09
fix: prevent interactive search crash when update check fails

jhult:
Relates to #2520 - Gracefully handle update check failures instead of crashing the entire interactive search UI.

## Checks


jhult:
This is **not** required to fix #2520. However, it should help in cases where there is a non-SSL related failure check. See my comment [here](https://github.com/atuinsh/atuin/issues/2520#issuecomment-3633343370) for more background.

PR #3032, opened 2025-12-22
fix(powershell): use Replace instead of RevertLine+Insert to set command line

sususu98:
## Summary
- Use `GetBufferState + Replace` to fully replace buffer content, preventing the original query from lingering after search selection
- Add `Set-AtuinCommandLine` helper function with try/catch fallback to `RevertLine + Insert` for PSReadLine 1.x compatibility
- Use explicit `$cursorPos` variable instead of `[ref]$null` for better cross-version compatibility

## Testing
- [x] PowerShell 7.x (Windows)
- [ ] Windows PowerShell 5.1 + PSReadLine 1.x (default)
- [ ] Windows PowerShell 5.1 + PSReadLine 2.x (upgraded)
- [ ] PowerShell 7.x (Linux)
- [ ] PowerShell 7.x (macOS)

**Need reviewer help to test on untested environments.**

### DIFF
--- a/crates/atuin/src/shell/atuin.ps1
+++ b/crates/atuin/src/shell/atuin.ps1
@@ -127,15 +127,33 @@ New-Module -Name Atuin -ScriptBlock {
                 return
             }
 
+            function Set-AtuinCommandLine {
+                param([string]$Text)
+
+                # Replace the entire buffer so the original query does not linger.
+                # Use try/catch to detect Replace method availability (PSReadLine 2.x+).
+                # Note: Missing static methods throw RuntimeException, not MethodException.
+                $currentLine = $null
+                $cursorPos = 0
+                [Microsoft.PowerShell.PSConsoleReadLine]::GetBufferState([ref]$currentLine, [ref]$cursorPos)
+                $length = if ($null -eq $currentLine) { 0 } else { $currentLine.Length }
+
+                try {
+                    [Microsoft.PowerShell.PSConsoleReadLine]::Replace(0, $length, $Text)
+                } catch {
+                    # Fallback for PSReadLine 1.x which lacks Replace method
+                    [Microsoft.PowerShell.PSConsoleReadLine]::RevertLine()
+                    [Microsoft.PowerShell.PSConsoleReadLine]::Insert($Text)
+                }
+            }
+
             $acceptPrefix = "__atuin_accept__:"
 
             if ( $suggestion.StartsWith($acceptPrefix)) {
-                [Microsoft.PowerShell.PSConsoleReadLine]::RevertLine()
-                [Microsoft.PowerShell.PSConsoleReadLine]::Insert($suggestion.Substring($acceptPrefix.Length))
+                Set-AtuinCommandLine $suggestion.Substring($acceptPrefix.Length)
                 [Microsoft.PowerShell.PSConsoleReadLine]::AcceptLine()
             } else {
-                [Microsoft.PowerShell.PSConsoleReadLine]::RevertLine()
-                [Microsoft.PowerShell.PSConsoleReadLine]::Insert($suggestion)
+                Set-AtuinCommandLine $suggestion
             }
         }
         finally {


sususu98:
> I think this tries to solve a problem that doesn't exist.

This issue indeed doesn't exist when PSReadLine is not used. Below is part of my pwsh configuration

```powershell
if ($Host.Name -eq 'ConsoleHost') {
    Import-Module PSReadLine -ErrorAction SilentlyContinue

    $supportsVT = $false
    try { $supportsVT = $Host.UI.SupportsVirtualTerminal } catch { }

    $redirected = $false
    try { $redirected = [Console]::IsOutputRedirected -or [Console]::IsErrorRedirected } catch { }

    if ($supportsVT -and -not $redirected) {
        Set-PSReadLineOption -PredictionSource HistoryAndPlugin
        Set-PSReadLineOption -PredictionViewStyle ListView
    } else {
        Set-PSReadLineOption -PredictionSource None
    }

    Set-PSReadLineOption -EditMode Windows
    Set-PSReadLineKeyHandler -Key UpArrow -Function HistorySearchBackward
    Set-PSReadLineKeyHandler -Key DownArrow -Function HistorySearchForward
    Set-PSReadLineKeyHandler -Key Tab -Function MenuComplete
}
```

1. type 2.6
2. ctrl+r
image
3. enter
image

ltrzesniewski:
Interesting. PSReadLine should be included and enabled by default though. The Atuin PowerShell module won't load if PSReadLine is not available.

Does it still reproduce if you remove the `Import-Module PSReadLine` command? You shouldn't need to import it.

Also, do you `atuin init` before or after the config you showed? Do you still have the problem if you swap the order?

ltrzesniewski:
I can confirm that `Import-Module PSReadLine` is the culprit - it just stops Atuin from recording history.

In a `pwsh -NoProfile` session:
- `atuin init powershell | Out-String | iex`
- `echo foo`
- Ctrl+R: Atuin shows `echo foo` as the last command, so it works fine
- `Import-Module PSReadLine`
- `echo bar`
- Ctrl+R: Atuin shows `Import-Module PSReadLine` as the last command, `echo bar` is missing, so it stopped recording history
- `Remove-Module Atuin; atuin init powershell | Out-String | iex`
- `echo baz`
- Ctrl+R: Atuin shows `echo baz` as the last command, so it works fine again

You should either:
- Remove `Import-Module PSReadLine` from your config, since it's loaded by default (preferred)
- Or run `atuin init powershell | Out-String | Invoke-Expression` at the end of your profile script

sususu98:
I think we are looking at two separate issues here:

1. **History recording issue**: As you observed, running `Import-Module PSReadLine` *after* Atuin init overwrites the hooks. This is indeed a configuration ordering issue.
2. **UI Buffer lingering issue (What this PR fixes)**: Even when history recording is working perfectly (without the `Import-Module` conflict), the UI glitch (`echoecho foo`) still occurs as long as `PredictionViewStyle` is set to `ListView`.

**Here is the reproduction for the UI bug specifically, avoiding the history conflict:**

```powershell
# 1. Start clean session
pwsh -NoProfile

# 2. Initialize Atuin (History recording is working fine here)
atuin init powershell | Out-String | iex

# 3. Only change the prediction style (NO Import-Module called)
Set-PSReadLineOption -PredictionViewStyle ListView

# 4. Reproduce
# Type `echo foo` <Enter> -> Type `echo` -> Ctrl+R -> Tab to select
# Result: "echoecho foo"
```

The root cause is that `RevertLine` + `Insert` fails to properly clear the internal PSReadLine buffer when `ListView` prediction is active.

ltrzesniewski:
Thanks. I can repro this on PSReadLine 2.3.6, and confirm that your fix works.

This is a PSReadLine bug: `RevertLine` is supposed to clear the buffer. :confused:  

~@ellie could you merge this, please? I'll submit a little cleanup PR later.~ See #3040

ltrzesniewski:
I reported this in PSReadLine: https://github.com/PowerShell/PSReadLine/issues/5047

PR #3033, opened 2025-12-22
fix(powershell): run history end in background to avoid blocking prompt

sususu98:
## Summary
- Use `Start-Process` to run `atuin history end` asynchronously, preventing slow sync operations from blocking the prompt
- Redirect output to null device (NUL on Windows, /dev/null on Linux/macOS)
- Use `Get-Variable -ErrorAction SilentlyContinue` to safely check `$IsWindows`, compatible with StrictMode in PS 5.1
- Use splatting to conditionally add `-NoNewWindow` only on Windows for cross-platform compatibility

## Testing
- [x] PowerShell 7.x (Windows)
- [ ] Windows PowerShell 5.1
- [ ] PowerShell 7.x (Linux)
- [ ] PowerShell 7.x (macOS)

**Need reviewer help to test on untested environments.**

### DIFF
--- a/crates/atuin/src/shell/atuin.ps1
+++ b/crates/atuin/src/shell/atuin.ps1
@@ -51,7 +51,24 @@ New-Module -Name Atuin -ScriptBlock {
             $duration = (Get-History -Count 1).Duration.Ticks * 100
             $durationArg = if ($duration) { "--duration=$duration" } else { $null }
 
-            atuin history end --exit=$exitCode $durationArg -- $script:atuinHistoryId | Out-Null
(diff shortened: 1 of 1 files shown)


sususu98:
> This points to a valid problem: `atuin history end` should be run in the background. Oops.
> 
> The PR, however, is AI slop.
> 
> I looked at the issue, and it turns out that PowerShell sucks about running tasks in the background. I'll try to find a better solution than this, however.

I look forward to you providing a better solution.
However, "AI slop" has already solved my problem, so I will close this PR.

PR #3058, opened 2026-01-14
feat: add option to use tmux display-popup

lmBored:
This is a "continuation" of #1177 - which was a draft that used FIFOs (named pipes) to get output from the popup, however this causes popup not being closed properly, so in this PR I use tmpfile to store the result and read after popup closes. @ellie could you review this PR please?

P.S. Thank you @immae for sharing your idea!

## Feature
+ Option to use tmux popup window in `config.toml`
+ Customize window width/height in `config.toml`
+ Tmux display-popup for `zsh, bash, fish`

## Checks

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,15 +3,6 @@
 version = 4
 
 [[package]]
-name = "addr2line"
-version = "0.24.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dfbe277e56a376000877090da837660b4427aad530e3028d44e0bffe4f89a1c1"
-dependencies = [
- "gimli",
-]
-
-[[package]]
 name = "adler2"
 version = "2.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -41,9 +32,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "1.1.3"
+version = "1.1.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8e60d3430d3a69478ad0993f19238d2df97c507009a52b3c10addcd7f6bcb916"
+checksum = "ddd31a130427c27518df266943a5308ed92d4b226cc639f5a8f1002816174301"
 dependencies = [
  "memchr",
 ]
@@ -65,9 +56,9 @@ dependencies = [
 
 [[package]]
 name = "anstream"
-version = "0.6.20"
+version = "0.6.21"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3ae563653d1938f79b1ab1b5e668c87c76a9930414574a6583a7b7e11a8e6192"
+checksum = "43d5b281e737544384e969a5ccad3f1cdd24b48086a0fc1b2a5262a26b8f4f4a"
 dependencies = [
  "anstyle",
  "anstyle-parse",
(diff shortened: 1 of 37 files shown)


ellie:
> @ellie could you review this PR please?

sure, but please at least wait until CI passes/finishes lol 😂

lmBored:
> sure, but please at least wait until CI passes/finishes lol 😂

Oopsie sorry, I forgot about CI... XD

crates/atuin/src/shell/atuin.fish
```diff
@@ -32,10 +73,45 @@ function _atuin_search
...
+            # if mktemp got errors
+            set ATUIN_H (ATUIN_SHELL=fish ATUIN_LOG=error ATUIN_QUERY=(commandline -b) atuin search --keymap-mode=$keymap_mode $argv -i 3>&1 1>&2 2>&3 | string collect)
+        else
+            set -l result_file "$tmpdir/result"
+
+            set -l query (commandline -b | string replace -a "'" "'\\''")
+            set -l escaped_args ""
+            for arg in $argv
+                set escaped_args "$escaped_args '"(string replace -a "'" "'\\''" -- $arg)"'"
+            end
```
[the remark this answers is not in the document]
lmBored: This is likely a false positive. Single quotes are POSIX strong quote, and as variables are wrapped in single quotes within `sh -c`, replacing every internal single quote with `'\''` is sufficient to render all other (special) characters

lmBored:
I believe this PR is ready for reviewing

crates/atuin/src/command/client/init/bash.rs
```diff
@@ -1,7 +1,17 @@
+use atuin_client::settings::Tmux;
 use atuin_dotfiles::store::{AliasStore, var::VarStore};
 use eyre::Result;
 
-pub fn init_static(disable_up_arrow: bool, disable_ctrl_r: bool) {
+fn print_tmux_config(tmux: &Tmux) {
+    if tmux.enabled {
```
ellie, line 6: I'm not super happy about having to set more env vars, but can't see much of a nicer way to get this value out. Maybe at some point we could have a subcommand for fetching individual config values, but not for this PR

lmBored: That's a good point, or we can move tmux popup logic into atuin itself (in `search.rs`) - this requires rewriting popup logic in rust though, and maybe having to add a hidden flag

mattcen:
And this feature is how I discovered this iTerm2/tmux interaction bug! 🙃 
https://github.com/tmux/tmux/pull/4361

Guess I'll disable this for now 🙂

lmBored:
Hm I dont get that issue with iterm2 (I switched to ghostty a long time ago, so my iterm2 version is still at 3.5.11)

https://github.com/user-attachments/assets/be7a73f8-e146-4917-b3ef-cd3f8ea4dcd5

image

mattcen:
> Hm I dont get that issue with iterm2 (I switched to ghostty a long time ago, so my iterm2 version is still at 3.5.11)

You are not using iTerm's tmux integration in this video. The control mode that allowed iTerm integration is what exhibits this issue. https://iterm2.com/documentation-tmux-integration.html

lmBored:
I see, its my first time hearing about this feature. This bug is not going to be fixed anytime soon, so perhaps we should add a note for iTerm users in the docs

mattcen:
Yes potentially; just a note to say "iTerm+tmux users note this new popup feature is enabled by default and doesn't work with iTerm native tmux integration; disable the new atuin popup feature with this config option" would be beneficial.

PR #3063, opened 2026-01-16
docs: Fix punctuation and grammar in basic usage guide

waldyrious:
Just a few punctuation, spelling and grammar fixes, along with a couple of wording changes for clarity.

_Note: This PR is a recreation of https://github.com/atuinsh/docs/pull/110, following the migration of the docs to this repo (https://github.com/atuinsh/docs/pull/112 + #3018)._

## Checks

### DIFF
--- a/docs/docs/guide/basic-usage.md
+++ b/docs/docs/guide/basic-usage.md
@@ -1,6 +1,6 @@
 # Basic Usage
 
-Now that you're all setup and running, here's a quick walkthrough of how you can use Atuin best.
+Now that you're all set up and running, here's a quick walkthrough of how you can use Atuin best.
 
 ## What does Atuin record?
 
--- a/crates/atuin/src/command/client/account/verify.rs
+++ b/crates/atuin/src/command/client/account/verify.rs
@@ -36,7 +36,7 @@ pub async fn run(settings: &Settings, token: Option<String>) -> Result<()> {
 
         (false, false) => {
             println!(
-                "Your Atuin server does not have mail setup. This is not required, though your account cannot be verified. Speak to your admin."
+                "Your Atuin server does not have mail set up. This is not required, though your account cannot be verified. Speak to your admin."
             );
         }
 
--- a/docs/docs/configuration/config.md
+++ b/docs/docs/configuration/config.md
@@ -144,7 +144,7 @@ The default filter to use when searching
 | global (default) | Search history from all hosts, all sessions, all directories |
 | host             | Search history just from this host                           |
 | session          | Search history just from the current session                 |
-| directory        | Search history just from the current directory               |
+| directory        | Search history just from the current directory (global)      |
 | workspace        | Search history just from the current git repository (>17.0)  |
 
 Filter modes can still be toggled via ctrl-r
@@ -266,6 +266,16 @@ Atuin version: >= 18.0
 Default: `true`
 
 Configure whether or not to show tabs for search and inspect.
+
+### `auto_hide_height`
+
+Atuin version: >= 18.4
+
+Default: `8`
+
+Set Atuin to hide lines when a minimum number of rows is subceeded. This has no effect except
+when `compact` style is being used (see `style` above), and currently applies to only the
+interactive search and inspector. It can be turned off entirely by setting to `0`.
 
 ### `exit_mode`
 
@@ -499,6 +509,7 @@ common_subcommands = [
   "git",
   "go",
   "ip",
+  "jj",
   "kubectl",
   "nix",
   "nmcli",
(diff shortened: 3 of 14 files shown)


docs/docs/guide/basic-usage.md
```diff
@@ -1,10 +1,10 @@
 # Basic Usage
 
-Now that you're all setup and running, here's a quick walkthrough of how you can use Atuin best
+Now that you're all setup and running, here's a quick walkthrough of how you can use Atuin best.
 
 ## What does Atuin record?
 
-While you work, Atuin records
+While you work, Atuin records:
```
akinomyoga, line 7: In my understanding, a colon should not be put here because the subsequent list is a part of the sentence. A colon is in general used to rephrase or supplement the previous part. In the present context, the list doesn't try to elaborate on the previous verb "records" nor the previous incomplete clause "Atuin records", but it's an object of the verb "records".

docs/docs/guide/basic-usage.md
```diff
@@ -15,11 +15,11 @@ While you work, Atuin records
 
 ## Opening and using the TUI
 
-At any time, you can open the TUI with the default keybindings of the up arrow, or control-R.
+At any time, you can open the TUI with the default keybindings of the up arrow, or ctrl-r.
 
 Once in the TUI, press enter to immediately execute a command, or press tab to insert it into your shell for editing.
 
-While searching in the TUI, you can adjust the "filter mode" by repeatedly pressing ctrl-r. Atuin can filter by
+While searching in the TUI, you can adjust the "filter mode" by repeatedly pressing ctrl-r. Atuin can filter by:
```
akinomyoga, line 22: ditto

The other additions of colons are fine.

docs/docs/guide/basic-usage.md
```diff
@@ -50,14 +49,14 @@ in your config file.
 
 ### Inline window
 
-If you find the full screen TUI overwhelming or too much, you can adjust it like so
+If you find the full screen TUI overwhelming or too large, you can adjust it like so:
```
akinomyoga, line 52: "too large" seems more specific, but to make it parallel with "overwhelming", I feel "too much" is more suitable.

BinaryMuse:
I think this looks good; I'm not too worried about small punctuation changes and the like. Thanks!

docs/docs/guide/basic-usage.md
waldyrious: Thanks for the careful review! I get your point, but those are not the only uses of the colon. In particular, one of its main uses is to introduce a list of items, which is precisely the case here. That said, the conventions around how colons are used are varied and have evolved over time, so this is more of a style issue. (Which is to say, I'd be fine with removing the colon here, as that would be a matter of preference.)

docs/docs/guide/basic-usage.md
waldyrious: Interesting, I can see what you mean. My reasoning was that "too much" is already pretty much a synonym of "overwhelming" (as you point out), so the "too large" was to add an alternative reason why one might dislike it (after all, it follows an "or").

docs/docs/guide/basic-usage.md
akinomyoga: >  In particular, one of its main uses is to introduce a list of items, which is precisely the case here.

A colon can be used to introduce a list, but it can only be used when the previous part forms a complete clause (and thus the colon essentially elaborates on the clause). I've been following this rule to not put a colon in this type of context in academic papers published in journals. I now googled "colon before a list", but every article (and AI summary that Google shows) says that a colon can only be used when it follows an independent clause.

docs/docs/guide/basic-usage.md
akinomyoga: I understand that "overwhelming or too much" are kind of redundant, but the combination "overwhelming or too large" feels more unnatural to me. I think that's because they belong to different categories of descriptions. It seems a bit weird to connect these with a coordinating conjunction, "or". I asked ChatGPT about which one is more typical: "If you find the full screen TUI overwhelming or too much, you can adjust it." vs "If you find the full screen TUI overwhelming or too large, you can adjust it.", and ChatGPT also says that the former is a natural and common pairing.

docs/docs/guide/basic-usage.md
akinomyoga: Anyway, those are small details. The document has been containing small subtle points, and it just continues to contain less subtle points. I'm fine with it.

docs/docs/guide/basic-usage.md
waldyrious: Oh, yeah, I understand what you mean, though the redundancy isn't great either. Maybe the whole phrase can be reworked differently. Hopefully in the future someone with fresh eyes will take he initiative to add the next brick to this collective building! :grin:  For now, I believe you'll agree that it's been nudged in a positive direction already. :slightly_smiling_face:

PR #3086, opened 2026-01-19
feat(search): Bind `Shift` + `Delete` to directly delete selected entry

mcexit:
Adds `SHIFT` modifier & binds `Shift` + `Delete` to directly delete a selected entry without modifying existing assignments

## Checks


ellie:
I'm afraid this key combo will not work reliably cross-terminal. Many will send the same escape sequence (or none) for Delete regardless of the Shift modifier, meaning that this will behave strangely depending on terminal and setup. We could potentially check if the user has support for the kitty keyboard protocol, where this WOULD work.

We're also very limited in terms of what key combos we can use elsewhere, as we mostly support [readline bindings](https://readline.kablamo.org/emacs.html)

Hence, you can either

1. open the inspector, delete
2. prefix + d (default to ctrl-a, d)

PR #3103, opened 2026-01-27
fix(ui): align cursor when expand column is in the middle

lilydjwg:
Also change the trailing space of each column to be drawn in `render_row` instead so that each column is separated by a space no matter how they are ordered.

## Checks


ellie:
just fixed up the merge conflicts - hope that's ok

PR #3107, opened 2026-01-27
fix: do not set ATUIN_SESSION if it is already set

ellie:
## Checks

### DIFF
--- a/crates/atuin/src/shell/atuin.bash
+++ b/crates/atuin/src/shell/atuin.bash
@@ -12,9 +12,11 @@ else # (include guard) beginning of main content
 #------------------------------------------------------------------------------
 __atuin_initialized=true
 
-ATUIN_SESSION=$(atuin uuid)
+if [[ -z "${ATUIN_SESSION:-}" ]]; then
+    ATUIN_SESSION=$(atuin uuid)
+    export ATUIN_SESSION
+fi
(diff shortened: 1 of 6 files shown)


ellie:
unrelated failure, will followup

akinomyoga:
What is the background of this change?

This ensures that reloading of startup scripts (`~/.bashrc`) doesn't affect `ATUIN_SESSION`. That is good.

However, with this change, all Bash sessions in a terminal multiplexer also share the same `ATUIN_SESSION` as the Bash session that started the terminal-multiplexer instance, because the environment variable `ATUIN_SESSION` is inherited by the terminal multiplexer and Bash started therein. I don't feel that that's useful. Is it intentional? If not, one way to avoid this behavior is to record `SHLVL` in another environment variable and compare `SHLVL` with the recorded one.

ellie:
> This ensures that reloading of startup scripts (~/.bashrc) doesn't affect ATUIN_SESSION. That is good.

This was an issue, as well as a few nested session things I've been doing elsewhere. Essentially wanted to avoid creating new sessions when logically we were in the "same" session (even if zshrc/bashrc has been reloaded)

> I don't feel that that's useful. Is it intentional? 

Not intentional. I'll fix shortly, thank you!

akinomyoga:
Great!

PR #3111, opened 2026-01-28
fix: new session on shlvl change

ellie:
eg, the user uses tmux, runs zsh again as a nested session, etc

## Checks


ellie:
thanks @akinomyoga <3

PR #3112, opened 2026-01-28
feat: move atuin-server to its own binary

ellie:
A combined binary was an early dev decision, when I thought most users would be self hosting. It is now clear that in actual fact, most users do not self host. So let's avoid forcing every user to have a copy of the server literally linked in, and let's stop building server deps over and over.

The deployment for this shouldn't change. `dist` will build a binary for this automatically, and will also add it to the installer. The latter is perhaps something we should explore changing too!

## Checks

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,7 +14,7 @@ COPY --from=planner /app/recipe.json recipe.json
 RUN cargo chef cook --release --recipe-path recipe.json
 
 COPY . .
-RUN cargo build --release --bin atuin
+RUN cargo build --release --bin atuin-server
 
 FROM debian:bookworm-20260112-slim AS runtime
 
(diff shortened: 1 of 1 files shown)


ellie:
Fwiw I'd like to do this with other optional features too, like the daemon

lilydjwg:
Previously the `atuin` binary was 27M in size, now it is 19M, plus a 13M `atuin-server`.

Separating the daemon would again increase the total size, and unlike the server which usually runs on a different system than the client, the daemon needs to run on the same as the client.

ellie:
I think that depends on whether most users use the daemon or not - so far my understanding is that very few do, though we don't have any data there to confirm either way

I'm less sure about the daemon than the server, and will see how this shakes out first

lilydjwg:
I think it's more important that how much file size the daemon feature adds to the binary. I presume that the daemon feature doesn't have a lot of extra dependencies and code, so the overhead it adds to the binary should be small.

marius:
This broke docker containers set up by following the documentation, PR to fix the docs here: https://github.com/atuinsh/atuin/pull/3148

PR #3114, opened 2026-01-29
docs: update the `[keys]` docs

ltrzesniewski:
This is related to:
- #3090
- #1906

---

## No longer relevant:

~I often just hold the backspace key to clear the query line in Atuin, and the change in #3090 makes it no longer possible: it accepts the current history entry instead.~ 🤯

~It may be a muscle memory thing, but it also feels wrong to me for the left key to accept the current entry. I'd prefer for it to do nothing when the cursor is at the start of the line (I've set `exit_past_line_start = false` in my config).~

~This PR makes the new behavior configurable: it adds an `accept_past_empty_line` setting (naming is *hard*, feel free to find a better name) which lets you disable it. It also updates the `[keys]` docs.~

BTW wouldn't setting the `accept_past_line_start` and `accept_with_backspace` settings to `true` (maybe by default) have the same effect as what #3090 introduced?

## Checks

### DIFF
--- a/crates/atuin-client/config.toml
+++ b/crates/atuin-client/config.toml
@@ -231,12 +231,16 @@ enter_accept = true
 [keys]
 # Defaults to true. If disabled, using the up/down key won't exit the TUI when scrolled past the first/last entry.
 # scroll_exits = true
+
 # Defaults to true. The left arrow key will exit the TUI when scrolling before the first character
 # exit_past_line_start = true
+
 # Defaults to true. The right arrow key performs the same functionality as Tab and copies the selected line to the command line to be modified.
 # accept_past_line_end = true
+
 # Defaults to false. The left arrow key performs the same functionality as Tab and copies the selected line to the command line to be modified.
 # accept_past_line_start = false
+
 # Defaults to false. The backspace key performs the same functionality as Tab and copies the selected line to the command line to be modified when at the start of the line.
 # accept_with_backspace = false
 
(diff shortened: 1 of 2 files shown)


ellie:
> It may be a muscle memory thing

Absolutely. I've heard from a lot of people that up-left is something that they commonly do to select the most recent command and then start editing. Personally I haven't felt that, nor has the change been an issue for me.

I'll take this PR, and honestly see if there's something better we can do wrt line editing in the future. All these small muscle memory things for people are a little frustrating.

ellie:
> BTW wouldn't setting the accept_past_line_start and accept_with_backspace settings to true (maybe by default) have the same effect as what https://github.com/atuinsh/atuin/pull/3090 introduced?

You're also correct here, so really I should revert 3090 🤦‍♀️

ellie:
Happy to close this, or if you'd rather just contribute the docs for `accept_past_line_start` and `accept_with_backspace`

ltrzesniewski:
Sure, you can just squash all of this now to keep only the relevant doc changes.

ltrzesniewski:
I kept only the docs commit and renamed the PR.

ellie (approved):
🙌

tedpearson:
> > It may be a muscle memory thing
> 
> Absolutely. I've heard from a lot of people that up-left is something that they commonly do to select the most recent command and then start editing. Personally I haven't felt that, nor has the change been an issue for me.
> 
> I'll take this PR, and honestly see if there's something better we can do wrt line editing in the future. All these small muscle memory things for people are a little frustrating.

@ellie I'm one of those people who do up-left. Is there any way with the new keymap feature or otherwise to end up with the cursor at the start of the selection when accepting with left arrow? I couldn't see a way after reading the documentation, as I only see a single action supported and the `cursor-start` likely only applies to inside the TUI.

PR #3118, opened 2026-01-30
chore(deps): update whoami dependency to v2

musicinmybrain:
In the [2.0.0](https://github.com/ardaku/whoami/releases/tag/v2.0.0) series, `whoami` removed all infallible function variants, and removed the `fallible` module, moving those functions to the root module.

Therefore, I replaced `whoami::fallible::hostname` with `whoami::hostname` (the same function with the same signature, just moved to the root module).

For `whoami::username`, the infallible function that `atuin` was using before is gone, and we must add error handling. I chose to fall back to the string `"unknown-user"` if getting the username fails, just as `"unknown-host"` is already the fallback when getting the hostname fails. This seemed reasonable to me, but it’s worth double-checking if there could be any unintended consequences, especially if `unknown-user` happens to be a real, valid username on the system. The alternatives I can see would be to panic on failure or to amend the signature of `get_username()` and all of its call sites with some kind of more graceful error handling (what?).

## Checks

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -311,7 +311,7 @@ dependencies = [
  "typed-builder",
  "urlencoding",
  "uuid",
- "whoami",
+ "whoami 2.1.0",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


ellie (approved):
> This seemed reasonable to me, but it’s worth double-checking if there could be any unintended consequences, especially if unknown-user happens to be a real, valid username on the system. The alternatives I can see would be to panic on failure or to amend the signature of get_username() and all of its call sites with some kind of more graceful error handling (what?).

I think this is fine - usernames aren't a core part of sync or search, so worst case the user sees some data that isn't quite what they expect

thank you!

PR #3125, opened 2026-02-03
fix: use directories crate for home dir resolution

### LINKED ISSUE #3123: atuin panics when HOME cannot be found
https://github.com/atuinsh/atuin/blob/2ec0cb7fa66ea42cd129396f987dc207c6c5d9d4/crates/atuin-common/src/utils.rs#L70

This line makes atuin panic if there is no `HOME` (which can happen in rare cases with for example `nix develop -i`).

I would expect atuin to simple exit in this case rather than panic.

ellie:
Previously, home_dir() read $HOME directly and panicked if it wasn't set. This could happen in environments like `nix develop -i` which strip the environment.

The directories crate (already a dependency) falls back to getpwuid_r when $HOME is not set, resolving the home directory from /etc/passwd.

Fixes #3123

## Checks


matthiasbeyer:
Amazing, thanks for the fast fix!

PR #3135, opened 2026-02-06
fix: tab behaving like enter, eprintln

### LINKED ISSUE #3134: A few issues in 18.12.0-beta.2
I love the new #3127 keymap feature, thank you @BinaryMuse!

But as you may have guessed, I noticed a few things after some testing:

- When `enter_accept = true` is present in the config, the right key and tab behave like enter, i.e. they immediately accept the item by default (without defining a custom keymap).

- The following warning causes issues:
  > [atuin] warning: [keys] section is ignored when [keymap] is present. Consider migrating your [keys] settings to [keymap].

  - On bash, it causes "execution" which gets you: `[atuin]: command not found`, followed by the actual command:
    Image

  - On PowerShell, the message doesn't show, yet it moves the cursor downwards one line when exiting, which misaligns multi-line prompts or shows lines which were not executed, depending on the order of your actions:
    Image

  I wouldn't be surprised if other issues show up because of this. Are you sure you want to keep using `eprintln!` in interactive mode?
    
  (If yes, I'll fix the behavior in PowerShell. I suppose I'll have to fix it anyway in case any other stderr output is emitted)

I only tested this rapidly, but if I find any other issues, I'll add them here.

ellie:
Resolves #3134 

Otherwise, I would like to figure out a nicer way for both logging, and for surfacing user-facing errors from the tui. perhaps similar to our current method of shell communication (`__atuin_accept`, etc), just providing an error for the shell integration to log

## Checks

### DIFF
--- a/crates/atuin/src/command/client/search/keybindings/defaults.rs
+++ b/crates/atuin/src/command/client/search/keybindings/defaults.rs
@@ -61,18 +61,17 @@ pub struct KeymapSet {
 // ---------------------------------------------------------------------------
 
 /// Add the bindings that are common to all search-tab keymaps:
-/// ctrl-c, ctrl-g, ctrl-o, and tab (respecting `enter_accept`).
+/// ctrl-c, ctrl-g, ctrl-o, and tab.
 ///
 /// Note: `esc`/`ctrl-[` are NOT included here because their behavior differs
 /// between emacs (exit), vim-normal (exit), and vim-insert (enter normal mode).
-fn add_common_bindings(km: &mut Keymap, settings: &Settings) {
+fn add_common_bindings(km: &mut Keymap) {
     km.bind(key("ctrl-c"), Action::ReturnOriginal);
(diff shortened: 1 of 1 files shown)


ltrzesniewski:
Thanks @ellie!

> Otherwise, I would like to figure out a nicer way for both logging, and for surfacing user-facing errors from the tui. perhaps similar to our current method of shell communication (__atuin_accept, etc), just providing an error for the shell integration to log

Adding an `__atuin_error` or similar would be nice, but would require updating all the shell scripts. Maybe just show the messages in the TUI as a separate screen and wait for the user to press enter?

BTW, there are still two more `eprintln!` calls in `defaults.rs`:

```rust
eprintln!("[atuin] warning: invalid key in keymap config: {key_str:?}: {e}");
eprintln!("[atuin] warning: invalid binding for {key_str:?} in keymap config: {e}");
```

PR #3145, opened 2026-02-07
fix: remove invalid IF EXISTS from sqlite drop column migration

Sped0n:
## Checks

## Summary

SQLite doesn't support an `IF NOT EXISTS` on `ALTER TABLE`

> https://www.sqlite.org/lang_altertable.html

```
atuin  | Caused by:
atuin  |     Other(while executing migration 20260127000000: error returned from database: (code: 1) near "exists": syntax error
atuin  |
atuin  |     Caused by:
atuin  |        0: error returned from database: (code: 1) near "exists": syntax error
atuin  |        1: (code: 1) near "exists": syntax error
```

The fix works on my setup

Ref: https://github.com/atuinsh/atuin/pull/3108#pullrequestreview-3713606584

### DIFF
--- a/crates/atuin-server-sqlite/migrations/20260127000000_remove-email-verification.sql
+++ b/crates/atuin-server-sqlite/migrations/20260127000000_remove-email-verification.sql
@@ -1,2 +1,2 @@
 drop table if exists user_verification_token;
-alter table users drop column if exists verified_at;
+alter table users drop column verified_at;


BinaryMuse:
@ellie thoughts on how this would interact with the migration hashing?

Sped0n:
If you're referring to migration checksum mismatch, I don't think that should be the case

According to the [implementation](https://github.com/launchbadge/sqlx/blob/7248f6424426a9035ed28ef7fec7bf03aa767b9b/sqlx-sqlite/src/migrate.rs#L226-L250) (being called from [`Migrator::run_direct`](https://github.com/launchbadge/sqlx/blob/7248f6424426a9035ed28ef7fec7bf03aa767b9b/sqlx-core/src/migrate/migrator.rs#L244)), the migration table (writing the checksum) is only updated after the migration succeeds

But the current situation is that the migration itself just failed (syntax error)

ellie (approved):
thank you! we could also do with getting some better testing in place for the sqlite server - I don't run it anywhere tbh

PR #3148, opened 2026-02-08
Update documentation how to use Docker to self-host

marius:
The server command changed with the switch to a separate server binary.

While here also update the documented postgres version.

## Checks


ellie (approved):
thank you!

mateuscomh:
@marius @ellie is safe update the image docker from postgress 14 to 18 like this change?

ellie:
It is for fresh installs - if you're self hosting and want to upgrade Postgres you'll have to follow their docs for that one! Or use a managed host

Issue #3047, opened 2026-01-07
[Feature] Request for tracking inside singularity and docker


wintervolcano:
Would it be possible to add the command tracking inside singularities and docker containers? Thank you for the amazing CLI!

Issue #1942, opened 2024-04-11
Feature Request: Add option to specify patterns to be ignored


oyarsa:
I don't really want commands like `ls`, `cd`, `git status` and so on to appear when searching the history. It would be good to be able to specify some patterns (globs, regexes, whatever) to filter those out. However, it could be nice to show these if required, so the filtering would be in the UI, not in recording the commands.

This would go along nicely with #1912.

Also, I'd be open to working on this as a PR if I can get some guidance on how to start.

ellie:
> However, it could be nice to show these if required

Have you got an example use case for when you'd want a command like `ls` to be recorded, but not visible in search without toggling a setting?

I'm a bit unclear on the use here, unless it's for stats?

oyarsa:
`cd` would be the motivating example. Most of the time, I don't want it shown in history. But sometimes, when I'm in a complex directory structure with weird names, I'd like to get the latest `cd` commands I used to go back to where I was previously, especially when filtering only commands from the current directory.

To be fair, I can still get this from `history | fzf`, and it might not used frequently enough to warrant a feature.

Issue #1847, opened 2024-03-10
Subprocess commands missing from history


tjarvstrand:
Hi,

First, thanks for a great tool! It's been a joy to use so far.

I routinely do eg. `(cd x; mycommand)` when I only want to run one or two commands inside a directory and I've noticed that these don't end up in the history. Is there any way I can get them to do that?

arcuru:
Interesting, they show up for me using either bash or zsh. What's your environment like?

Note that the whole command is what will be in history, `(cd x; mycommand)`, and the logged directory will be where you ran _that_ command. I'm not sure if you're asking about `mycommand` itself not showing up.

tjarvstrand:
Yeah, `(cd x; mycommand)` is what I would like to be in the history :)

I'm on bash `5.2.15` on MacOS Ventura.

I do get the commands in the history, i.e. I can use the up arrow to run the command, but when I hit ctrl-r, they don't appear in that list.

kenjo:
don't show up for me
GNU bash, version 5.2.15(1)-release (x86_64-pc-linux-gnu)

there also is a problem if I use c-x c-e. I do have 
shopt -s lithist
but removing that has the same result

using c-p I get the correct history in the built in list.

frasty:
> Same here, e.g., `(true)` does not end up in `atuin`'s history
> 
> ```
> atuin:
>   version: 18.1.0
>   sync: null
> shell:
>   name: bash
>   plugins:
>   - atuin
>   - bash-preexec
> system:
>   os: Fedora Linux
>   arch: x86_64
>   version: '38'
> ```
> 
> To be honest, I haven't seen the `atuin` source code, but I'm very curious why there are all sort of such weird "missing in history" issues. Do you do some "parsing" or whatever on history commands? If so, could you explain why? And if so, is there a chance we can have our history intact (optionally, of course, some may not like it), matching exactly what's in our shell history? Thanks.

I definitely second this. Sometimes I am unwilling to search history through atuin because I know that some commands could be missed from history. I'd rather prefer to have my history intact even if that could lead to some privacy issues.

akinomyoga:
> why there are all sort of such weird "missing in history" issues.

@mjf @frasty Those are all the limitations of `bash-preexec`, which Atuin uses to hook into the command execution of Bash.

> Do you do some "parsing" or whatever on history commands?

[...]

PR #3163, opened 2026-02-10
feat(docs): Add Shell Integration and Interoperability docs

### LINKED ISSUE #1545: Env var option to disable atuin or make session history volatile for integrated Terminals (VS Code, IDEs)
I just installed atuin and I like it so far but one thing that is disrupting my workflow is that it also shows up and records history for integrated Terminals in VS Code!

You see I have set VS Code to ignore my shell history and go off the record for whatever I enter in my VS Code Terminals in order to keep my main shell history clean from commands that would only apply to certain projects and not my general usage when I fire up Terminal on my Linux, by adding these in my settings:
```
"terminal.integrated.env.linux": {
        "HISTFILE": ""
    },
```

If you could provide an option to make atuin ignore the main history file and store the current session's history in a temporary file that would go away as soon as the Terminal is terminated, that would be awesome!

### LINKED ISSUE #2983: [QUESTION] How to disable Atuin temporarily?
Hi,

My `~/.zshrc` loads up Atuin, via `atuin init zsh`. How to turn it off?

I’m not sure `atuin uninit` would make sense.

The convention regarding disabling the history of regular shells is to `unset HISTFILE`. So maybe allowing overriding the database path via `ATUIN_DB_PATH` (as is set for the Docker image) would make sense.

That way, disabling Atuin could be achieved via `export ATUIN_DB_PATH=/dev/null` and re-enabling would be as easy as `unset ATUIN_DB_PATH`.

That do you think?

### LINKED ISSUE #3085: [Bug]: atuin not recording history inside the PyCharm terminal

Commands to be recorded in atuin history.

### What happened?

Commands are never recorded if ran in the PyCharm terminal.

### Atuin doctor output

```yaml
Atuin Doctor
Checking for diagnostics


{
  "atuin": {
    "version": "18.10.0",
    "commit": "37e257105ab801b49fd07a927d707a36a2e771a1",
    "sync": {
      "cloud": false,
      "records": true,
      "auto_sync": true,
      "last_sync": "2026-01-16 11:29:24.835216762 +00:00:00"
    },
    "sqlite_version": "3.46.0"
  },
  "shell": {
    "name": "bash",
    "default": "bash",
    "plugins": [
      "atuin",
      "bash-preexec"
    ],
    "preexec": "none"
  },
  "system": {
    "os": "Ubuntu",
    "arch": "x86_64",
    "version": "24.04",
    "disks": [
      {
        "name": "/dev/mapper/vg00-root",
        "filesystem": "ext4"
      },
      {
        "name": "/dev/disk/by-id/dm-uuid-LVM-redacted",
        "filesystem": "ext4"
      },
      {
        "name": "/dev/nvme0n1p2",
        "filesystem": "ext4"
      },
      {
        "name": "/dev/mapper/vg00-tmp",
        "filesystem": "ext4"
      },
      {
        "name": "/dev/mapper/vg00-var",
        "filesystem": "ext4"
      },
      {
        "name": "/dev/mapper/vg00-work",
        "filesystem": "ext4"
      },
      {
        "name": "/dev/nvme0n1p1",
        "filesystem": "vfat"
      }
    ]
  }
}
```

### Code of Conduct

BinaryMuse:
This PR adds a new documentation page called "Shell Integration and Interoperability." It explains how Atuin's shell hooks work, and why they may _not_ work by default in an embedded terminal in another tool. It includes suggestions to ensure the shell is launched in interactive mode, or to use a wrapper script if this isn't possible. It also explains the various mechanisms to use to _exclude_ certain commands, or commands from certain tools, from Atuin's history.

Closes #3085 
Closes #1545 
Closes #2983 

Related to: #1942, #3047, #1847

### DIFF
--- a/.codespellrc
+++ b/.codespellrc
@@ -3,5 +3,5 @@
 skip = .git*,*.lock,.codespellrc,CODE_OF_CONDUCT.md,CONTRIBUTORS
 check-hidden = true
 # ignore-regex = 
-ignore-words-list = crate,ratatui,inbetween
+ignore-words-list = crate,ratatui,inbetween,iterm
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,26 @@
 
 All notable changes to this project will be documented in this file.
 
+## [18.12.1] - 2026-02-11
+
+### Bug Fixes
+
+- *(shell)* Fix ATUIN_SESSION errors in tmux popup ([#3170](https://github.com/atuinsh/atuin/issues/3170))
+- *(tui)* Enter in vim normal mode, shift-tab keybind ([#3158](https://github.com/atuinsh/atuin/issues/3158))
+- Server start commands for Docker. ([#3160](https://github.com/atuinsh/atuin/issues/3160))
+
+
+### Features
+
+- Expand keybinding system with vim motions, media keys, and inspector improvements ([#3161](https://github.com/atuinsh/atuin/issues/3161))
+- Add original-input-empty keybind condition ([#3171](https://github.com/atuinsh/atuin/issues/3171))
+
+
+### Miscellaneous Tasks
+
+- Update changelog
+
+
 ## [18.12.0] - 2026-02-09
 
 ### Bug Fixes
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -211,7 +211,7 @@ checksum = "1505bd5d3d116872e7271a6d4e16d81d0c8570876c8de68093a09ac269d8aac0"
 
 [[package]]
 name = "atuin"
-version = "18.12.0"
+version = "18.12.1"
 dependencies = [
  "arboard",
  "async-trait",
@@ -262,7 +262,7 @@ dependencies = [
 
 [[package]]
 name = "atuin-client"
-version = "18.12.0"
+version = "18.12.1"
 dependencies = [
  "async-trait",
  "atuin-common",
@@ -316,7 +316,7 @@ dependencies = [
 
 [[package]]
 name = "atuin-common"
-version = "18.12.0"
+version = "18.12.1"
 dependencies = [
  "base64",
  "directories",
@@ -336,7 +336,7 @@ dependencies = [
 
 [[package]]
 name = "atuin-daemon"
-version = "18.12.0"
+version = "18.12.1"
 dependencies = [
  "atuin-client",
  "atuin-dotfiles",
@@ -363,7 +363,7 @@ dependencies = [
 
 [[package]]
 name = "atuin-dotfiles"
-version = "18.12.0"
+version = "18.12.1"
 dependencies = [
  "atuin-client",
  "atuin-common",
@@ -377,7 +377,7 @@ dependencies = [
 
 [[package]]
 name = "atuin-history"
-version = "18.12.0"
+version = "18.12.1"
 dependencies = [
  "atuin-client",
  "crossterm",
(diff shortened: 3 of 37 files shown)


docs/docs/guide/shell-integration.md
```diff
@@ -0,0 +1,261 @@
...
+## Embedded Terminals and IDE Integrations
+
+Many development tools include embedded terminals:
+
+- **IDEs**: PyCharm, IntelliJ, VS Code, Cursor, Zed
+- **AI coding assistants**: Claude Code, GitHub Copilot CLI, Aider
+- **Multiplexers and wrappers**: tmux, screen, custom terminal wrappers
+- **Container environments**: Docker, Podman, devcontainers
+
+These tools often spawn shells differently than your regular terminal, which can prevent Atuin from working.
```
akinomyoga, line 40: In what sense do the multiplexers like tmux and screen spawn the shell sessions differently? I don't think they are essentially different from normal terminal emulators with tabs in the perspective of shell settings such as Atuin. The only difference between the terminal multiplexers and normal terminal emulators is how they render the terminal contents; the former renders the contents in another terminal, while the latter renders the contents using GUI.

Rather, some of the normal terminals (kitty, Ghostty, etc.) try to inject shell-terminal integration code using a non-trivial way of spawning shells (e.g. Bash with the POSIX mode turned on, sourcing `~/.bashrc` from their startup file, and modifying `PROMPT_COMMAND`), though I'm not sure if that specifically affects the initialization of Atuin. It probably doesn't affect because we don't receive many issue reports for them.

docs/docs/guide/shell-integration.md
```diff
@@ -0,0 +1,261 @@
...
+
+#### Verifying the Fix
+
+After configuring, open a new terminal in your IDE and run:
+
+```shell
+atuin doctor | grep preexec
+```
+
+You should see `bash-preexec`, `blesh`, or similar—not `none`.
```
akinomyoga:
```suggestion
You should see `built-in`, `bash-preexec`, `blesh`, or similar—not `none`.
```

The possibilities `bash-preexec`, `blesh`, or `none` is only for Bash. In the other shells, they are typically [`built-in`](https://github.com/atuinsh/atuin/blob/47368ac1602376ea1bb6d9da37ade20da600f75c/crates/atuin/src/command/client/doctor.rs#L72).

akinomyoga (commented):
It's a great idea to have a dedicated page for the issues with the shell integration!

docs/docs/guide/shell-integration.md
BinaryMuse: Yeah, I don't think `tmux` or `screen` are concerns; they're mostly included for completeness.

docs/docs/guide/shell-integration.md
```diff
@@ -0,0 +1,261 @@
...
+
+These variables are used internally to track command execution and associate commands with sessions.
+
+## Embedded Terminals and IDE Integrations
+
+Many development tools include embedded terminals:
+
+- **IDEs**: PyCharm, IntelliJ, VS Code, Cursor, Zed
+- **AI coding assistants**: Claude Code, GitHub Copilot CLI, Aider
+- **Multiplexers and wrappers**: tmux, screen, custom terminal wrappers
```
ellie:
```suggestion
```

docs/docs/guide/shell-integration.md
ellie: I just removed that line, I don't want to imply that they might be issues! otherwise this whole page is awesome, thank youuuu!

ellie (dismissed):

PR #3176, opened 2026-02-12
fix systemd Exec for separate server binary

13werwolf13:
## Checks


ellie (approved):
thank you! just a note though, if the system has both atuin and atuin-server installed, it will continue to work as normal

PR #3183, opened 2026-02-13
fix: silent DB failures e.g. when disk is full

jonaylor89:
## Checks

Silent DB failures breaking shell when disk is full

When storage runs out (i.e. disk full), atuin breaks the shell by continuously printing database errors for every character typed. This is a fix to silence the DB errors


ellie:
thank you for the contribution!

PR #1436, opened 2023-12-10
Dissallow deletion if the '--limit' flag is present

sdr135284:
Thanks for writing and supporting atuin!

I ran into a bug where if you combine the --limit and --deletion flags it would delete all of the matching entries instead of some limited number. For example:

```
user@nixos:~/ > cat asdf.txt 
user@nixos:~/ > qwerty1
qwerty1: command not found
user@nixos:~/ > qwerty2
qwerty2: command not found
user@nixos:~/ > qwerty3
qwerty3: command not found
user@nixos:~/ > atuin search qwerty
2023-12-10 19:08:42     qwerty1 39ms
2023-12-10 19:08:47     qwerty2 37ms
2023-12-10 19:08:52     qwerty3 37ms
2023-12-10 19:09:10     atuin search qwerty     0s
user@nixos:~/ > atuin search --delete qwerty --limit 1
deleting 018c5522b1447c778a2261d68cdfc6d4
deleting 018c552226ef71fc8a0efe9a9c376f6c
deleting 018c5521ddec717cbcac7581a229b6c6
deleting 018c5521cc1b79cfb69b6d7431810fbf
deleting 018c5521b90871ffaa304eea46b79649
user@nixos:~/ > atuin search qwerty
2023-12-10 19:10:07     atuin search qwerty     0s
```

Unfortunately I had searched for `"*"` so all my history is gone, luckily on a relatively new computer so not a critical loss.

This is on nixos with atuin version 17.0.1, but the relevant code hasn't meaningfully changed since.

Looking at the code it wasn't obvious what combining "--limit" and "--delete" would even mean, so I put together this change to disallow them from being combined and a comment explaining why.

I tested that the combination of flags does print the message saying they're not compatible and exits.

PR #3184, opened 2026-02-13
Support passing `--limit` to `atuin search --delete`

### LINKED ISSUE #3177: Feature request: searching (deleting) by exact string match
I have a shell alias `oops` which removes the last command from history (implementation: `history delete --exact --case-sensitive -- $history[1]`). I'd like to add Atuin support to it (since having to press ctrl-r, ctrl-o, ctrl-d, esc as well as running `oops` feels a bit much), but I can't find any way to ask Atuin to forget about the last command I ran.

I was hoping to be able to `atuin search --filter-mode session --limit 1 --delete` to delete the last command from this session, but that's not possible: #1436. `atuin search --filter-mode session --search-mode full-text $history[1] --delete` is closer, but it's still not right: the `full-text` search mode is actually a substring search (so deleting a typo'd `j` would delete any command containing a `j`!). `prefix` is similarly bad, and I don't think it's possible to escape an entire command line for the `fuzzy` or `skim` modes in a way that will only match that exact string.

Is there any way to achieve this currently? (Looking at the code, I don't actually really understand why `--delete` and `--limit` _wouldn't_ work together...) If not: could there be?

keysmashes:
Fixes #3177.

As pointed out previously (#1436), there are two plausible ways this could be expected to work:

1. The search is executed across (deduplicated) commands, then every usage of the first N commands in the result is deleted.
2. The search is executed across history entries, then the first N history entries in the result are deleted (and other usages of the same command are left intact).

With option 1, it's nonsensical to pass `--include-duplicates`: if the results for `atuin search --include-duplicates --limit 3` were "foo", "bar", and "foo", then only two commands would actually be deleted. (If the command "baz" was executed before those three, then without `--include-duplicates` it would have been included in the results – so passing that flag would actually have *reduced* the number of commands being deleted!)

With option 2, it's easier to reason about what's happening: every deletion already acts as if `--include-duplicates` was passed, so deletions always operate on individual history entries (not aggregated commands), therefore we don't have to worry about what would happen if `--include-duplicates` was/wasn't passed. To make it obvious that this is how `--limit` works, we require that `--include-duplicates` is *explicitly* passed whenever using `--delete` together with `--limit`.

Therefore this PR implements option 2.

## Checks

### DIFF
--- a/crates/atuin/src/command/client/search.rs
+++ b/crates/atuin/src/command/client/search.rs
@@ -193,19 +193,7 @@ impl Cmd {
             self.query
         };
 
-        if (self.delete_it_all || self.delete) && self.limit.is_some() {
-            // Because of how deletion is implemented, it will always delete all matches
-            // and disregard the limit option. It is also not clear what deletion with a
-            // limit would even mean. Deleting the LIMIT most recent entries that match
-            // the search query would make sense, but that wouldn't match what's displayed
-            // when running the equivalent search, but deleting those entries that are
-            // displayed with the search would leave any duplicates of those lines which may
-            // or may not have been intended to be deleted.
-            eprintln!("\"--limit\" is not compatible with deletion.");
-            return Ok(());
-        }
-
-        if self.delete && query.is_empty() {
+        if self.delete && query.is_empty() && self.limit.is_none() {
             eprintln!(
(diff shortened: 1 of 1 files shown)


keysmashes:
Rebased. (I would appreciate a review!)

PR #3190, opened 2026-02-17
Update config.md to remove NuShell support note

c-git:
Removed unsupported note about NuShell from configuration documentation.

Possibly better alternative would be to say from which version of nushell and atuin support started. Didn't want to spend the time researching that in case it wasn't of value.

## Checks


ellie (approved):
thanks! potentially worth tracking versions, though nushell moves _fast_ and keeping that up to date is unlikely

PR #3203, opened 2026-02-24
Bump prost-related crates

LecrisUT:
Bumping the prost related crates and avoiding as much compat packages as possible. There's still the axum0.7 compat which is introduced with this change, but the update to get rid of that seems more involved

## Checks
- [x] I have checked that there are no existing pull requests for the same thing (https://github.com/atuinsh/atuin/pull/2896)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -153,28 +153,6 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


ellie:
thank you!

Issue #2179, opened 2024-06-23
Store command's stdout/stderr


FlysoftBeta:
First of all, thanks for your great work!

After a few days of using, I found it would be great that we can have an option that allows users to store the output.
This is very useful especially when I'm trying reproducing an error occurred previously.
Also an option can be there to determine how long the history will be kept.

**EDIT: FYI, if you're interested in this, please vote 👍 this post**

ellie:
Thanks for the kind words!

I've been thinking about this a bit lately, so it's nice to get validation it could be useful. We'd have to be careful about disk usage though!

What sort of UX would work for you around it - should output be part of search?

FlysoftBeta:
Thanks for quick reply!

> What sort of UX would work for you around it - should output be part of search?

![graph](https://github.com/atuinsh/atuin/assets/49718840/1c49ac87-6dbb-42c1-b9d7-b6b1c67889e7)

ellie:
Sorry, I should have clarified. Less concerned about how this might work in UI, more concerned about how users would interact with it and what their experience might be like

1. If enabled, do we record everything by default? Or only when specified with something like "atuin record"/etc?
2. Is the search part of the normal TUI (and should it then be fuzzy search and slow, or fts and fast? do users expect search over potentially gigabytes to be rapid?), or is it something separate tailored to this use case?
3. Do outputs get synced across machines? I'd assume yes, but we'd want more configuration flexibility there
4. How do we handle security here? There's likely to be sensitive information in stdout/err 

But yeah more thinking about what sort of interactions you imagine with this, and not so much what the UI might look like

FlysoftBeta:
>     1. If enabled, do we record everything by default? Or only when specified with something like "atuin record"/etc?

Can this be asked during installation? Or it can be included in a command like `atuin setup` to get user started?

We can probably provide different options for user to filter logs (e.g. all/stderr only/regex filtering/etc.)

>     2. Is the search part of the normal TUI (and should it then be fuzzy search and slow, or fts and fast? do users expect search over potentially gigabytes to be rapid?), or is it something separate tailored to this use case?

Yes, as I said before, it can be searched using a keyword. The search can be configured to be fuzzy or precise by user, maybe.

<details>

GitHub's search is a good example:

![Alt](https://github.com/atuinsh/atuin/assets/49718840/285ef30b-13c1-4462-8e1f-d041fa9f6f63)

</details>

[...]

PR #3206, opened 2026-02-25
feat: initial draft of atuin-shell

ellie:
This intentionally does... nothing!

Right now we spawn a PTY running the "default program" - ie, the users shell - and proxy bytes. The user should notice no change at all, while we can run code in the middle.

This middle code can do a bunch of things!

1. We can parse prompts, inputs, etc, and have a way to track commands outside of the shell hook lifecycle
2. We could track outputs! https://github.com/atuinsh/atuin/issues/2179
3. Render UI over the top of the terminal, without clearing space first, and without using alt screen
4. Probs a lot of other cool shit

## Checks


ellie:
@greptile-apps recheck

PR #3212, opened 2026-02-27
chore: Move atuin ai subcommand into core binary

BinaryMuse:
This PR moves the primary invocation for `atuin ai` from a separate binary into the main binary. Still doesn't do anything by default.

closes atu-335


BinaryMuse:
@greptileai

PR #3230, opened 2026-03-03
feat: update script for smoother setup

ellie:
Many users do not seem to read the docs. So, make the install script handle most of the things that we include in the docs quickstart.

Ensures that the install script still functions as expected in non-interactive environments - setup scripts, ci, etc. Offers a flag just in case this check fails, to force non-interactive setup.

## Checks

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -129,7 +129,7 @@ EOF
   case "$sync_answer" in
     [yY]*)
       echo ""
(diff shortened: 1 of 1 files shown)


ellie:
yeah I think greptile is correct, will double check and try it on a few fresh machines! ty

PR #3235, opened 2026-03-05
feat: Allow setting multipliers for frequency, recency, and frecency scores

BinaryMuse:
This PR adds three settings:

* `search.recency_score_multiplier`
* `search.frequency_score_multiplier`
* `search.frecency_score_multiplier`

These settings, which default to `1.0`, specify a multiplier for the recency, frequency, and frecency scoring systems in the daemon-fuzzy search mode, respectively. Values less an `1.0` reduce that elements weight; values greater than `1.0` increase it. `0.0` disables that component of the scoring system entirely. This allows users to customize which aspects of the frecency calculation matter most to them.

This PR also implements a `SettingsWatcher` that uses `notify` and a `tokio::sync::watch` to notify the daemon of changes to the settings file:

```rust
use atuin_client::settings::watcher::global_settings_watcher;

async fn example() -> eyre::Result<()> {
    let watcher = global_settings_watcher()?;
    let mut rx = watcher.subscribe();

    // React to settings changes
    while rx.changed().await.is_ok() {
        let settings = rx.borrow();
        println!("Settings updated!");
    }
    Ok(())
}
```

### DIFF
--- a/crates/atuin-daemon/src/search/index.rs
+++ b/crates/atuin-daemon/src/search/index.rs
@@ -397,9 +397,11 @@ impl SearchIndex {
         let now = OffsetDateTime::now_utc().unix_timestamp();
         let mut frecency_map: HashMap<Arc<str>, u32> = HashMap::new();
 
-        let recency_mul = search_settings.recency_score_multiplier;
-        let frequency_mul = search_settings.frequency_score_multiplier;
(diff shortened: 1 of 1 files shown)


BinaryMuse:
@greptileai recheck

BinaryMuse:
@greptileai

ellie (approved):

PR #3250, opened 2026-03-10
feat: Allow running `atuin search -i` as subcommand on Windows

ltrzesniewski:
This is the equivalent of #3208, but for Windows.

~The rendering performance is noticeably slower in this mode when refreshing a large part of the screen, but it's better than not having the feature at all.~ Fixed 🙂 

The second commit fixes some unrelated warnings.

/cc @BinaryMuse FYI

## Checks

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -272,6 +272,7 @@ dependencies = [
  "tracing-tree",
  "unicode-width 0.2.2",
  "uuid",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


ltrzesniewski:
uh-oh, I noticed some encoding issue, let me fix it.

ltrzesniewski:
This is now ready for review 🙂

I've refactored existing stuff a bit and fixed the issue I noticed right after submitting the PR 🤦  (some of the output was garbled because of the console output encoding).

BinaryMuse:
@greptileai

PR #3255, opened 2026-03-11
feat: Initialize Atuin AI by default with `atuin init`

BinaryMuse:
* Run Atuin AI's `init` during main `init` for bash, zsh, and fish
* Note that logging into Hub will enable sync
* Add instructions for users with existing sync accounts
* Ensure daemon respects `auto_sync` setting
* Update docs on disabling Atuin AI

### DIFF
--- a/crates/atuin-daemon/src/components/sync.rs
+++ b/crates/atuin-daemon/src/components/sync.rs
@@ -25,6 +25,16 @@ enum SyncCommand {
     Stop,
 }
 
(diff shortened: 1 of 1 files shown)


BinaryMuse:
@ellie The `auto_sync` handling was still off — the failure retry mechanism was not compatible with `auto_sync` being disabled. New version should handle properly

ellie (approved):

PR #3266, opened 2026-03-14
fix: Nushell 0.111; future Nushell 0.112 support

stuartcarnie:
Fix issue introduced by #3249. Nushell 0.112 will introduce the breaking change; not 0.111.

The flag for `job spawn` will change, which prevents the script from compiling on Nushell 0.112+

## Checks

### DIFF
--- a/crates/atuin/src/shell/atuin.nu
+++ b/crates/atuin/src/shell/atuin.nu
@@ -36,11 +36,7 @@ let _atuin_pre_prompt = {||
         return
     }
     with-env { ATUIN_LOG: error } {
-        if (version).minor >= 111 or (version).major > 0 {
-            job spawn -d atuin {
-                ^atuin history end $'--exit=($env.LAST_EXIT_CODE)' -- $env.ATUIN_HISTORY_ID | complete
-            } | ignore
-        } else if (version).minor >= 104 or (version).major > 0 {
+        if (version).minor >= 104 or (version).major > 0 {
             job spawn {
                 ^atuin history end $'--exit=($env.LAST_EXIT_CODE)' -- $env.ATUIN_HISTORY_ID | complete
             } | ignore


stevenxxiu:
Does this actually fix it for you? I still get the error `Error: nu::parser::unknown_flag`. Seems like it has to be removed then.

fvsch:
I don't think this fixes the issue because Nushell raises a parser error, not an error at runtime. Just having `job spawn -d foo` in the code to be parsed will crash, even if it's within an `if` block that never executes.

```nushell
~> if false { job spawn -d test { "never" } }
Error: nu::parser::unknown_flag

  × The `job spawn` command doesn't have flag `-d`.
   ╭─[repl_entry #3:1:23]
 1 │ if false { job spawn -d test { "never" } }
   ·                       ┬
   ·                       ╰── unknown flag
   ╰────
  help: Use `--help` to see available flags

```

stevenxxiu:
I think this commit just has to be reverted for now.

FollowTheProcess:
The correct fix is:

```nushell
job spawn --tag "atuin"
```

I've modified my `atuin.nu` locally and this works fine for me on nu `0.111.0`

stevenxxiu:
Yes or when expanded:

```nu
if (version).minor >= 111 or (version).major > 0 {
    job spawn --tag atuin {
        ^atuin history end $'--exit=($env.LAST_EXIT_CODE)' -- $env.ATUIN_HISTORY_ID | complete
    } | ignore
```

t-mart:
If you generate this file on each startup, here's a quick and dirty hack:

```nushell
atuin init nu |
  str replace "job spawn -d atuin" "job spawn --tag atuin" |  # TODO: Remove when https://github.com/atuinsh/atuin/issues/3268 fixed
  save --force ~/.local/share/nushell/vendor/autoload
```

EDIT: This is a workaround for those wanting a solution before a proper fix (this fix?) is released. Sorry for the noise to those working here.

stuartcarnie:
Ah, right. 

I will remove the description completely; it isn't really necessary and will avoid all issues with compatibility.

ellie:
I'd love to get a fix in, but tbh aren't 100% sure with this PR - could you all please confirm exactly what is/isn't working? I can test myself but I am not a nushell users.

Should we merge + release this change?

stuartcarnie:
Hi @ellie – this should be fine to merge. It is both forwards and backwards compatible, as I've removed the flag that named the job, given it might be changing in an up and coming release. We'd have to do it eventually anyway, unless the Nushell team decides to introduce a deprecated version of the flag. 

It's also what @BinaryMuse did in #3272

BinaryMuse:
@stuartcarnie Thanks so much!

stuartcarnie:
No worries – sorry for the issues!

PR #3284, opened 2026-03-16
chore: vendor nucleo-ext + fork, so we can depend on our changes properly

ellie:
We cannot publish to crates.io without specifying a version, and we cannot do that without properly forking nucleo. We're shipping atuin-nucleo, but will likely drop this if we can get our changes upstream. This is highlighted in the README + manifest, and the original author is still included. 

Originally forked here: https://github.com/atuinsh/nucleo-ext

cc @BinaryMuse - this should just be a vendor + restructure, but would appreciate the sanity check

## Checks


LecrisUT:
Is there upstream PR for this?

BinaryMuse:
> Is there upstream PR for this?

I mentioned it in the repo's discussion forum, but haven't heard anything since: https://github.com/helix-editor/nucleo/discussions/99

LecrisUT:
> > Is there upstream PR for this?
> 
> I mentioned it in the repo's discussion forum, but haven't heard anything since: [helix-editor/nucleo#99](https://github.com/helix-editor/nucleo/discussions/99)

Wouldn't it better to open the PR or issue? The discussion board does not seem active

ellie:
The upstream repo is generally not very active, and I suspect our changes are not particularly relevant for them anyway. Nucleo was built for helix

PR #3287, opened 2026-03-17
fix(inspector): use local timezone for day of week stats

### LINKED ISSUE #3251: [Bug]: Incorrect day in inspector

I just set up Atuin today and imported my bash history. The day in the "Runs per day" graph in the inspector is incorrect.

- It's putting everything into a single day. This may be an issue with the data import.
- The day is Tuesday but it reports the commands as happening on Wednesday. Maybe this is a Timezone/UTC issue. I am UTC -5 and it's late, so its Wednesday UTC.

### What happened?

The inspector is showing the incorrect day for usage history.

See the attached gif:

Image

### Atuin doctor output

```yaml
{
  "atuin": {
    "version": "18.12.1",
    "commit": "",
    "sync": null,
    "sqlite_version": "3.46.0"
  },
  "shell": {
    "name": "bash",
    "default": "bash",
    "plugins": [
      "atuin",
      "blesh"
    ],
    "preexec": "blesh-0.4.0-nightly+b99cadb"
  },
  "system": {
    "os": "Darwin",
    "arch": "arm64",
    "version": "15.7.4",
    "disks": [
      {
        "name": "Macintosh HD",
        "filesystem": "apfs"
      },
      {
        "name": "Macintosh HD",
        "filesystem": "apfs"
      }
    ]
  }
}
```

### Code of Conduct

BillionClaw:
## Problem

The inspector's "Runs per day" graph was showing incorrect days for users in non-UTC timezones. Commands run late in the evening local time were being counted as the next day because SQLite's strftime('%w') returns UTC day of week.

For example, a user in UTC-5 running a command at 11 PM local time (Tuesday) would see it counted as Wednesday because that's 4 AM UTC.

## Solution

Changed the stats function to:
1. Accept a timezone parameter (UtcOffset)
2. Fetch raw timestamps from the database
3. Compute day of week in Rust using the user's configured local timezone

This ensures the day of week grouping respects the user's configured timezone setting.

## Changes

- Modified `Database::stats()` trait method to accept `tz: UtcOffset` parameter
- Updated `Sqlite::stats()` implementation to compute day of week in Rust with timezone conversion
- Updated the caller in `interactive.rs` to pass `settings.timezone.0`

Fixes #3251

### DIFF
--- a/crates/atuin-client/src/database.rs
+++ b/crates/atuin-client/src/database.rs
@@ -768,20 +768,23 @@ impl Database for Sqlite {
 
         // Fetch raw timestamps and compute day of week in Rust with proper timezone handling
         // SQLite's strftime('%w') returns UTC day, so we do the conversion here
-        let day_of_week: Vec<(i64, i64)> = sqlx::query_as(
-            "SELECT timestamp, count(1) as count FROM history WHERE command = ?1 GROUP BY timestamp"
+        let timestamps: Vec<(i64,)> = sqlx::query_as(
+            "SELECT timestamp FROM history WHERE command = ?1"
         )
         .bind(&h.command)
         .fetch_all(&self.pool)
         .await?;
 
         // Aggregate by day of week in local timezone
+        // Note: time::Weekday is Monday-based (0=Mon), but num_to_day() expects Sunday-based (0=Sun)
+        // Conversion: (rust_weekday + 1) % 7 = expected
         let mut day_counts: std::collections::HashMap<u8, i64> = std::collections::HashMap::new();
-        for (timestamp_ns, count) in day_of_week {
+        for (timestamp_ns,) in timestamps {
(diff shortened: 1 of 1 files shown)


BillionClaw:
@greptile-apps Thanks for the thorough review! I've addressed both issues:

**P0 - Day label fix:**
- time::Weekday as u8 is Monday-based (0=Mon…6=Sun)
- num_to_day() expects Sunday-based numbering (0=Sun…6=Sat) matching SQLite's strftime('%w')
- Added conversion: (rust_weekday + 1) % 7 to map Monday-based to Sunday-based

**P1 - Performance fix:**
- Removed GROUP BY timestamp which was causing one DB row per command invocation
- Now fetch raw timestamps without grouping, then aggregate by day of week in Rust
- This maintains the timezone-aware behavior while avoiding the performance regression

Changes pushed to the same branch. Ready for re-review.

PR #3288, opened 2026-03-17
chore: Replace atuin-ai rendering with component-oriented system

BinaryMuse:
Prior to this PR, the Atuin AI rendering step was a bit of a mess of height calculations and massive match statements; this PR updates the system to use more of a component-like architecture:

* `VStack` (with scroll support)
* `Spacer`
* `SymbolRow`
* `Separator`
* Separate components for each `Content` type

Each of these components implement a `Component` trait that implements a `render` method as well as a `height` method to calculate the height needed to render its content at a given width.

This PR also cleans up markdown formatting for text in bold or formatted as code: bold gets a bolder color, code gets a new color, and the busy markdown formatting characters are removed:

image

### DIFF
--- a/crates/atuin-ai/src/tui/components.rs
+++ b/crates/atuin-ai/src/tui/components.rs
@@ -332,7 +332,6 @@ impl Component for WarningContent {
 
 /// Animated spinner with status text.
 pub struct SpinnerContent {
-    pub frame: usize,
     pub status_text: String,
 }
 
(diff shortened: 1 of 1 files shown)


ellie (approved):
yesssss 🙌 

would be awesome to do this for the main tui at some point too (no rush tho that's like a WHOLE THING)

PR #3289, opened 2026-03-17
feat: Report distro name with OS for distro-specific commands

BinaryMuse:
We properly send the OS type, but Linux distros are distinct enough that this information can be required to generate the proper commands.


BinaryMuse:
@greptileai

ellie (approved):

PR #3315, opened 2026-03-20
fix: improve non-interactive TTY detection in install.sh

### LINKED ISSUE #3294: [Bug]: unattended install broken

Used to be that I could just have new devices run the following code on setup:

```
if curl -sS --tlsv1.2 'https://setup.atuin.sh' >atuin-install.sh
then
    sh atuin-install.sh
fi
```

e.g. inside a new Ona environment, or whatever else is cool nowadays.

This worked perfectly and everything was magical.

### What happened?

Today I see install seems to hang forever, seems like it's waiting for input, despite there being no terminal.

Note the `cannot open /dev/tty: No such device or address` errors just around the first two `[Y/n]` prompts.

```
 _______  _______  __   __  ___   __    _
|   _   ||       ||  | |  ||   | |  |  | |
|  |_|  ||_     _||  | |  ||   | |   |_| |
|       |  |   |  |  |_|  ||   | |       |
|       |  |   |  |       ||   | |  _    |
|   _   |  |   |  |       ||   | | | |   |
|__| |__|  |___|  |_______||___| |_|  |__|

Magical shell history

Atuin setup
https://github.com/atuinsh/atuin
https://forum.atuin.sh

Please file an issue or reach out on the forum if you encounter any problems!

===============================================================================
downloading atuin 18.13.3 x86_64-unknown-linux-gnu
installing to /home/vscode/.atuin/bin
  atuin
  atuin-update
everything's installed!
To add $HOME/.atuin/bin to your PATH, either restart your shell or run:

    source $HOME/.atuin/bin/env (sh, bash, zsh)
    source $HOME/.atuin/bin/env.fish (fish)
Atuin installed successfully!

Would you like to import your existing shell history into Atuin? [Y/n] Skipping history import. You can always run 'atuin import auto' later.
atuin-install.sh: 98: cannot open /dev/tty: No such device or address
Sync your history across all your machines with Atuin Cloud:

 - End-to-end encrypted — only you can read your data
  - Access your history from any device
  - Never lose your history, even if you wipe a machine
atuin-install.sh: 126: cannot open /dev/tty: No such device or address
Sign up for a sync account? [Y/n]
Already have an account? Log in with 'atuin login'.
You can also run 'atuin register' any time to create one.
> Enable Atuin AI?
  This will enable command generation and other AI features via the question mark key [Y/n]
```

### Atuin doctor output

N/A

### Code of Conduct

mango766:
Fixes #3294

## Checks

## Problem

The install script's non-interactive detection uses `{ true </dev/tty; }` to check if a terminal is available. This only tests whether `/dev/tty` can be *opened*, not whether it is actually a functional terminal. In containers, CI, and headless environments (like the Ona setup described in #3294), `/dev/tty` may exist as a device node that the shell can open, but isn't backed by a real terminal. This causes:

1. The detection to incorrectly conclude the session is interactive
2. The subsequent `read </dev/tty` calls to fail with `cannot open /dev/tty: No such device or address`
3. The final `atuin setup </dev/tty` call to hang indefinitely

## Fix

Two changes, both in `install.sh`:

**1. Stronger TTY detection** — Replace `{ true </dev/tty; }` with `( exec </dev/tty && [ -t 0 ] )`. This opens `/dev/tty` as stdin inside a subshell, then checks it's actually a terminal with `test -t`. The subshell ensures the redirect doesn't affect the parent process.

**2. Suppress stderr on `read` fallbacks** — Add `2>/dev/null` to the two `read </dev/tty` calls. They already have `|| var="n"` fallbacks, but without stderr suppression, failed redirects produce noisy error messages on shells like `dash`.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -11,7 +11,13 @@ for arg in "$@"; do
 done
 
 if [ "$ATUIN_NON_INTERACTIVE" != "yes" ]; then
-  if [ -t 0 ] || { true </dev/tty; } 2>/dev/null; then
+  # Check if we have a usable interactive terminal.
+  # Using `true </dev/tty` is not sufficient: /dev/tty may exist as a device
+  # node that can be opened, but not actually be backed by a real terminal
+  # (e.g. in containers, CI, or headless environments). This causes the later
+  # `read </dev/tty` calls to fail or hang.
+  # Instead, open /dev/tty and verify it is a terminal with `test -t`.
+  if [ -t 0 ] || ( exec </dev/tty && [ -t 0 ] ) 2>/dev/null; then
     ATUIN_NON_INTERACTIVE="no"
   else
     ATUIN_NON_INTERACTIVE="yes"
@@ -95,7 +101,7 @@ echo ""
 if [ "$ATUIN_NON_INTERACTIVE" != "yes" ]; then
 
   printf "Would you like to import your existing shell history into Atuin? [Y/n] "
-  read -r import_answer </dev/tty || import_answer="n"
+  read -r import_answer </dev/tty 2>/dev/null || import_answer="n"
   import_answer="${import_answer:-y}"
 
   case "$import_answer" in
@@ -123,7 +129,7 @@ Sync your history across all your machines with Atuin Cloud:
 EOF
 
   printf "Sign up for a sync account? [Y/n] "
-  read -r sync_answer </dev/tty || sync_answer="n"
+  read -r sync_answer </dev/tty 2>/dev/null || sync_answer="n"
   sync_answer="${sync_answer:-y}"
 
   case "$sync_answer" in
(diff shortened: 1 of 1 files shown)


mentalisttraceur:
Thank so much for taking care of this!

install.sh
```diff
@@ -95,7 +101,7 @@ echo ""
 if [ "$ATUIN_NON_INTERACTIVE" != "yes" ]; then
 
   printf "Would you like to import your existing shell history into Atuin? [Y/n] "
-  read -r import_answer </dev/tty || import_answer="n"
+  read -r import_answer </dev/tty 2>/dev/null || import_answer="n"
```
mentalisttraceur: I recommend **not** suppressing errors from the two reads.

Think about it this way: those error messages were the reason why we were able to diagnose the issue in my bug so quickly.

Without them, it would just be an unexplained hang.

My issue is fixed with a better TTY check, but if we hit some _other_ problem during `read`, the error messages would be helpful.

markovejnovic: agreed! if you can remove these redirects, ill gladly stamp the pr and we can have it merged =)

markovejnovic: oh, nevermind, youre comfortable with maintaners editing the PR -- merging in a jiffy

markovejnovic:
Thank you for your PR!

PR #3317, opened 2026-03-20
chore: Refactor CLI auth flows and token storage

BinaryMuse:
This PR eplaces the binary `is_hub_sync()` auth routing with an explicit `SyncAuth` enum that classifies the client's authentication state at runtime. This fixes a class of bugs where CLI session tokens were silently mis-stored or used with the wrong auth scheme during Hub migration.

### Problem

The client previously used a single boolean (`is_hub_sync()`) to decide all auth behavior. When Hub auto-migration failed (e.g. password too short), Hub returned a CLI session token in the same response shape as a successful Hub login. The client stored this CLI token in the Hub session slot, then sent it as a `Bearer` token — which Hub rejected, silently breaking sync.

Additionally, users in this state couldn't change their password (the `HubAuthClient` would try to use the mis-stored CLI token as a Bearer token, getting a confusing 401), and had no guidance on what went wrong.

### Solution

The PR introduces `SyncAuth`, an enum with four variants that explicitly models each possible auth state:

- **`Legacy`** — self-hosted Rust server, `Token` auth, legacy endpoints
- **`Hub`** — Hub with a valid `atapi_*` token, `Bearer` auth, v0 endpoints
- **`HubViaCli`** — targeting Hub but only has a CLI session token; `Token` auth against compat/record endpoints. Sync works, but user is nudged to upgrade via `atuin login`
- **`NotLoggedIn`** — no valid tokens, with an actionable error message

`resolve_sync_auth()` on `Settings` examines stored tokens and the configured sync target to determine the correct state. It also auto-cleans mis-stored tokens: if a non-`atapi_` token is found in the Hub session slot, it's moved to the CLI session slot (if empty) and the Hub slot is cleared.

### Changes

- **`atuin-common/api.rs`** — `LoginResponse` and `RegisterResponse` now include an `auth: Option<String>` field (`"hub"` or `"cli"`), with `serde(default)` for backward compat with old servers
- **`atuin-server/handlers/user.rs`** — legacy server returns `auth: "cli"` in login/register responses
- **`atuin-client/settings.rs`** — adds `SyncAuth` enum, `resolve_sync_auth()`, and rewrites `sync_auth_token()` to delegate to it
- **`atuin-client/auth.rs`** — `AuthResponse::Success` carries `auth_type`; `HubAuthClient` validates `atapi_` prefix before sending authenticated requests (change_password, delete_account)
- **`login.rs`** — saves tokens to the correct slot based on `auth_type`/prefix; login check uses `resolve_sync_auth()` for richer messages; `HubViaCli` users proceed to re-login instead of being told "already logged in"
- **`register.rs`** — same token-routing pattern; `HubViaCli` users are directed to `atuin login` instead of registering a potentially conflicting new account
- **`doctor.rs`** — replaces `cloud: bool` with `auth_state: String` showing the resolved auth state

### What this does NOT change

- **AI flow (`inline.rs`)** — has its own `ensure_hub_session()` independent of sync auth; unchanged. Hub identity token separation for AI is a planned follow-up.
- **Sync call sites** — continue using `sync_auth_token()`, which now delegates to `resolve_sync_auth().into_auth_token()`. No behavioral change for properly authenticated users.
- **Self-hosted users** — `Legacy` variant is entered when `sync_address` is not an official Atuin address. Behavior is identical to before.

### Companion Hub changes

This PR works with Hub changes that:
- Return the `auth` key in login/register responses
- Add a compat `PATCH /account/password` endpoint
- Reject migrated users on compat endpoints (forcing client upgrade)

### Test plan

- [x] `cargo build` / `cargo test` pass
- [x] Self-hosted config with CLI token resolves `Legacy`, sync works unchanged
- [x] Official config with `atapi_` hub_session resolves `Hub`, sync works via Bearer
- [x] Official config with non-`atapi_` token in hub_session: auto-fixes (moves to CLI slot), resolves `HubViaCli`, sync works via Token auth
- [x] Official config with CLI token only resolves `HubViaCli`, sync works via Token auth
- [x] No tokens at all resolves `NotLoggedIn` with clear error message
- [x] `atuin doctor` shows detailed auth state
- [x] Headless Hub login saves token to correct slot based on `auth` field
- [x] Headless Hub login with migration fallback saves CLI token to session slot (not hub_session), shows warning
- [x] `atuin register` with `HubViaCli` state directs user to `atuin login`

### DIFF
--- a/crates/atuin/src/command/client/account/login.rs
+++ b/crates/atuin/src/command/client/account/login.rs
@@ -125,7 +125,7 @@ impl Cmd {
             tracing::debug!("Could not link CLI account to Hub: {}", e);
         }
 
-        println!("Successfully authenticated with Atuin Hub.");
+        println!("Successfully authenticated.");
         Ok(())
     }
 


BinaryMuse:
@greptileai

PR #3327, opened 2026-03-22
feat(pty-proxy): add --shell flag to override the spawned shell

### LINKED ISSUE #3326: [Bug]: `atuin hex` ignores $SHELL and there is no way to change it

```
> $env.SHELL = 'nu'
> atuin hex
> bash -c 'echo $SHELL'
< 'nu'
```

`atuin hex` should respect $SHELL or have some way to override and not forcefully use login shell.

### What happened?

```
> $env.SHELL = 'nu'
> atuin hex
> bash -c 'echo $SHELL'
< 'zsh'
```

I try to set $SHELL, but it does not work. It *always* chooses to use my login shell.

My setup is to keep zsh as my login shell for simplicity. My tmux has my $SHELL as nu. Trying to use atuin hex forces me to use zsh after running it and there is no way to override it.

### Atuin doctor output

```yaml
{
  "atuin": {
    "version": "18.13.3",
    "commit": "NO_GIT",
    "sync": null,
    "sqlite_version": "3.46.0"
  },
  "shell": {
    "name": "nu",
    "default": "zsh",
    "plugins": [
      "atuin"
    ],
    "preexec": "built-in"
  },
  "system": {
    "os": "NixOS",
    "arch": "x86_64",
    "version": "26.05",
    "disks": [
      {
        "name": "/dev/nvme0n1p2",
        "filesystem": "ext4"
      },
      {
        "name": "/dev/nvme0n1p2",
        "filesystem": "ext4"
      },
      {
        "name": "/dev/nvme0n1p1",
        "filesystem": "vfat"
      }
    ]
  }
}
```

### Code of Conduct

xav-ie:
Greetings, I am open to an alternative solution like subcommand support or something else which I am happy to accomodate.

Fixes #3326.

## Checks

### DIFF
--- a/crates/atuin-pty-proxy/src/lib.rs
+++ b/crates/atuin-pty-proxy/src/lib.rs
@@ -27,6 +27,11 @@ mod unsupported {
         #[arg(long)]
         debug_osc133: bool,
 
+        /// Path to the shell binary that atuin pty-proxy should spawn.
+        /// Defaults to the system login shell. Only valid when no subcommand is given.
+        #[arg(long, value_name = "PATH")]
+        shell: Option<std::path::PathBuf>,
+
         #[command(subcommand)]
         cmd: Option<Cmd>,
     }
--- a/crates/atuin-pty-proxy/src/pty_proxy.rs
+++ b/crates/atuin-pty-proxy/src/pty_proxy.rs
@@ -1,3 +1,5 @@
+use std::path::PathBuf;
+
 use clap::{Args, Subcommand, ValueEnum};
 
 use crate::{CommandCaptureSink, runtime};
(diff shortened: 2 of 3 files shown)


ellie (changes requested):
Thank you for the work here! I'm not sure this makes sense to override it - typically, $SHELL is set to the login shell, and while it could make sense to allow the user to override it to change what hex launches, it may then affect anything running within hex that might expect that to remain true

In this case, I think we'd be better off allowing the user to provide `--shell` to hex, and pass in a path to launch

xav-ie:
That sounds good to me. Let me try to update this PR

limonkufu:
Hey any progress on this or a workaround for nushell?

xav-ie:
Pushed: switched to a `--shell <PATH>` flag per your feedback. Defaults to login shell when omitted :)

xav-ie:
I have updated my code to fix the P2 of the greptile review.

xav-ie:
@greptileai retrigger

xav-ie:
Rebased and force-pushed: squashed onto the latest `main` and ported onto the renamed `atuin-pty-proxy` crate (formerly `atuin-hex`, renamed in #3473), so this now applies cleanly again.

This is focused purely on the `--shell <PATH>` flag, as requested — each shell's init snippet passes its own interpreter path (`$BASH`, `${ZSH_ARGZERO:-$(command -v zsh)}`, `(status fish-path)`, `$nu.current-exe`) so the proxy re-execs the same binary that sourced the init. Passing `--shell` with a subcommand is rejected with exit code 2.

Ready for another look 🙏

ellie (dismissed):

ellie (dismissed):
looks great! thank you for getting this wrapped up 🙏

Issue #3329, opened 2026-03-22
Maybe atuin hex init should be separate setup step?


xav-ie:
Hello, I am trying to implement `atuin hex init nu` support. 

The problem I am facing is that the current docs say that you should pregenerate and save file for regular init. https://docs.atuin.sh/cli/guide/installation/#__tabbed_2_4:
> Run in Nushell:
> `mkdir ~/.local/share/atuin/`
> `atuin init nu | save ~/.local/share/atuin/init.nu`
> Add to config.nu:
> `source ~/.local/share/atuin/init.nu`

This makes sense. Nushell does not support dynamic eval on purpose.

For `atuin hex init`, the general shape of the emitted script is:
```
1. If not "hexed" enter hexed (proxied) shell with `atuin hex`:
   if (!ATUIN_HEX_ACTIVE) {
     ATUIN_HEX_ACTIVE=1
     atuin hex
   }
   This causes you to re-proceed to step 1, if already set env vars, this step is skipped!
2. Run `eval $(atuin hex init)`
```

For clarity here is the bash `atuin hex init`:
<summary> 
The bash init script. (expand to see)
<details>

```bash
if [[ "$-" == *i* ]] && [[ -t 0 ]] && [[ -t 1 ]]; then
  _atuin_hex_tmux_current="${TMUX:-}"
  _atuin_hex_tmux_previous="${ATUIN_HEX_TMUX:-}"

  if [[ -z "${ATUIN_HEX_ACTIVE:-}" ]] || [[ "$_atuin_hex_tmux_current" != "$_atuin_hex_tmux_previous" ]]; then
    export ATUIN_HEX_ACTIVE=1
    export ATUIN_HEX_TMUX="$_atuin_hex_tmux_current"
    exec atuin hex
  fi

  unset _atuin_hex_tmux_current _atuin_hex_tmux_previous
fi

eval "$(atuin init bash)"
```
</details>
</summary>

If you try to naively replicate this in nushell, you face issue because the `atuin hex init` pattern is to generate *more* init script and eval dynamically.

I am thinking that maybe the hex init pattern could maybe become a separate step and should not call init generation on behalf of the user. Here is why I think this:

1. To enter the proxy shell, users might have run any number of init before reaching the atuin initialization. This is wasted time that could have been saved if they put the hexing first. 
For nushell, it is particularly bad. Nushell in my experience has terrible init time and paying this cost would be mostly avoided if we could separate the hex proxying from the rest of init.

[...]

Issue #3323, opened 2026-03-22
Combine `atuin hex init` and `atuin init --disable-*`


bcspragu:
For a given shell, what's the correct way to set Atuin to use the new hex/pty proxy mode, but also configure settings like `--disable-up-arrow` at the same time? Trying something like `atuin hex init fish -- --disable-up-arrow` doesn't work. And the initialization script output from `atuin hex init` uses the default init options. E.g. here's the fish shell output:

```fish
if status is-interactive; and test -t 0; and test -t 1
  ...
end

atuin init fish | source # No way to pass --disable-up-arrow
```

Happy to contribute something here if folks have a sense for how they'd like it integrated.

praxxis:
I fixed this in [a local fork](https://github.com/praxxis/atuin/tree/hex-init-options) so I can use hex until a proper solution is figured out. My guess is moving the flags into proper settings would be the best solution, to avoid having to duplicate the args logic.

PR #3330, opened 2026-03-22
feat: hex init nu

xav-ie:
- **feat(hex): add nushell support for `atuin hex init`**
- **docs(hex): add nushell hex setup instructions**

This adds setup for nushell. It breaks the pattern of calling `eval $(atuin init)` on behalf of the user because nushell simply cannot do this. I tried to source the atuin.nu file and add the preamble to it, but it is part of the atuin package, so it made things too difficult.

I think settling for separate init is ok.

Partially addresses #3329.

Please see #3323 as well.

I was able to get it working and I am using these changes in my dotfiles:
https://github.com/xav-ie/dots/compare/b1a8cf96b58f802396ac5103f0925e1434fc473c...696dbf31008395587353e3071f5296c459685a17

Basically, I just do as the new docs say and make sure to add hex init before regular init and source it.

## Checks

### DIFF
--- a/docs/docs/guide/installation.md
+++ b/docs/docs/guide/installation.md
@@ -260,6 +260,27 @@ After installing, remember to restart your shell.
     source ~/.local/share/atuin/init.nu
     ```
 
+    ??? tip "Optional: Atuin Hex"
+        Hex is a lightweight pty proxy that renders the Atuin popup over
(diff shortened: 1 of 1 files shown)


ellie (approved):
this is great, thank you so much!! 🙌

PR #3334, opened 2026-03-23
fix: Don't run 'atuin init' in 'atuin hex init' — each must be initialized separately

### LINKED ISSUE #3323: Combine `atuin hex init` and `atuin init --disable-*`

### LINKED ISSUE #3329: Maybe atuin hex init should be separate setup step?

BinaryMuse:
This PR separates `atuin init` from `atuin hex init`, allowing for configuration flags to be passed to `atuin init`.

Fixes #3329 
Fixes #3323

### DIFF
--- a/docs/docs/reference/hex.md
+++ b/docs/docs/reference/hex.md
@@ -1,6 +1,6 @@
 # hex
 
-Atuin Hex is an experimental lightweight PTY proxy, providing new features without needing to replace your existing shell. Atuin Hex currently supports bash, zsh, fish, and nu.
+Atuin Hex is an experimental lightweight PTY proxy, providing new features without needing to replace your existing terminal or shell. Atuin Hex currently supports bash, zsh, fish, and nu.
 
 ## TUI Rendering
 


docs/docs/reference/hex.md
```diff
@@ -0,0 +1,65 @@
+# hex
+
+Atuin Hex is an experimental lightweight PTY proxy, providing new features without needing to replace your existing shell. Atuin Hex currently supports bash, zsh, fish, and nu.
```
ellie:
```suggestion
Atuin Hex is an experimental lightweight PTY proxy, providing new features without needing to replace your existing terminal or shell. Atuin Hex currently supports bash, zsh, fish, and nu.
```

ellie (approved):
looks good and solves our current issues but i REALLY want the hex init to be much much easier than this. no ifs or thinking required for the user to get it running

PR #3351, opened 2026-03-26
fix: set WorkingDirectory in PowerShell Invoke-AtuinSearch

### LINKED ISSUE #3350: [Bug]:  Interactive search directory filter uses wrong directory on PowerShell

Switching to directory filter mode in interactive search (up arrow / ctrl+r) should show history for my current directory.

### What happened?

Directory mode in interactive search shows results for the wrong directory. It seems to use whatever directory PowerShell was started in rather than where I've actually cd'd to. So if I open a terminal in C:\Users\MY_USER then cd to a project, directory mode still shows history for C:\Users\MY_USER .

Using `atuin search --filter-mode directory` from the CLI returns the correct results, so the history data itself seems recorded correctly.

### Atuin doctor output

```yaml
Atuin Doctor
Checking for diagnostics


{
  "atuin": {
    "version": "18.13.3",
    "sync": null,
    "sqlite_version": "3.46.0"
  },
  "shell": {
    "name": "pwsh.exe",
    "default": "powershell",
    "plugins": [
      "atuin"
    ],
    "preexec": "built-in"
  },
  "system": {
    "os": "Windows",
    "arch": "x86_64",
    "version": "11 (26200)",
    "disks": [
      {
        "name": "",
        "filesystem": "NTFS"
      }
    ]
  }
}
```

### Code of Conduct

jamesbrooks:
Fixes #3350

PowerShell's Set-Location (cd) doesn't update the process-level working directory, so the spawned search process sees a stale cwd and directory filter mode matches against the wrong directory. Setting `WorkingDirectory` on the `ProcessStartInfo` ensures the correct cwd is passed through.

### DIFF
--- a/crates/atuin/src/shell/atuin.ps1
+++ b/crates/atuin/src/shell/atuin.ps1
@@ -150,6 +150,8 @@ New-Module -Name Atuin -ScriptBlock {
             $process.StartInfo.StandardErrorEncoding = [System.Text.Encoding]::UTF8
             $process.StartInfo.EnvironmentVariables["ATUIN_SHELL"] = "powershell"
             $process.StartInfo.EnvironmentVariables["ATUIN_QUERY"] = Get-CommandLine
+            # PowerShell's Set-Location (cd) doesn't update the process-level working directory, set it explicitly
+            $process.StartInfo.WorkingDirectory = (Get-Location -PSProvider FileSystem).ProviderPath
 
             try {
                 $process.Start() | Out-Null


ltrzesniewski:
FML 🤦 

image

Thanks for fixing this!

Could you just please replace `$PWD.ProviderPath` with `(Get-Location).ProviderPath`, since `$PWD` can be overwritten? Thanks!

Note: I didn't check if that works, and obviously PowerShell is full of surprises...

jamesbrooks:
> replace `$PWD.ProviderPath` with `(Get-Location).ProviderPath`, since `$PWD` can be overwritten?

Yep that makes perfect sense, updated!

> PowerShell is full of surprises

My daily driver is a mac so I'm only occasionally on PowerShell. I was excited to see atuin available on Windows, then saddened when the directory filter didn't work the way I knew and loved ;)

ltrzesniewski:
Sorry to bother you again, but I was in a hurry when I wrote my previous comment. According to [this SO answer](https://stackoverflow.com/a/76996334/3764814), the most robust way to get the filesystem path is: `(Get-Location -PSProvider FileSystem).ProviderPath`

I would never had thought that `cd` didn't change the process' working directory when using a filesystem drive... 🤦

~The Atuin module doesn't like being in a non-filesystem location, but that's a different problem I can take a look at later.~ Fixed in #3353

jamesbrooks:
No worries, updated! :) thanks.

PR #3360, opened 2026-03-29
fix(ui): make preview line breaking algorithm aware of CJK double-width characters

lilydjwg:
## Checks

Or part of the command may be missing and the user may be misled to run a wrong command.


ellie (approved):
nice one, thank you! :shipit:

PR #3367, opened 2026-03-30
feat: opt-in to sharing last command with ai

ellie:
This enables it to perform more effectively and give better suggestions. Same as send_cwd, disabled by default, opt in.

## Checks

### DIFF
--- a/crates/atuin-client/src/settings.rs
+++ b/crates/atuin-client/src/settings.rs
@@ -676,10 +676,10 @@ pub struct Ai {
 #[derive(Default, Clone, Debug, Deserialize, Serialize)]
 pub struct AiOpening {
     /// Whether or not to send the current working directory to the AI endpoint.
-    pub send_cwd: bool,
+    pub send_cwd: Option<bool>,
 
     /// Whether or not to send the last command as context in the opening AI request.
-    pub send_last_command: bool,
+    pub send_last_command: Option<bool>,
 }
 
 impl Default for Preview {
--- a/crates/atuin-ai/src/commands/inline.rs
+++ b/crates/atuin-ai/src/commands/inline.rs
@@ -401,9 +401,10 @@ async fn run_inline_tui(
         .build()?;
 
     // Support both legacy [ai] send_cwd and new [ai.opening] send_cwd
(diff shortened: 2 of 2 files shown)


crates/atuin-client/src/settings.rs
```diff
@@ -666,6 +666,10 @@ pub struct Ai {
 
     /// Whether or not to send the current working directory to the AI endpoint.
     pub send_cwd: bool,
+
+    /// Whether or not to allow the AI to read shell history.
+    /// When enabled, the previous command is sent as context in the initial request.
+    pub read_history: Option<bool>,
```
BinaryMuse: Do we want separate settings for "is the AI allowed to even ask to read my history" and "do you want to automatically send the last command"?

ellie: I wasn't sure that they were functionally different enough, but I guess one is potentially "scarier" than the other 🤔 

I guess we could have

1. `read_history`, global, can the ai see anything about your shell history
2. `send_last_command` obvious
3. `search_history` obvious

if you just have (1) set, then (2) and (3) inherit it

if you set (1), and then explicitly set either (2) or (3), then they override (1)

I don't really want us to have a billion checkboxes for everything, but do want to make sure we have enough granularity

alternatively scratch that and just have 2 and 3

BinaryMuse: The capability stuff defaulting to `None` will fall back to the permission system, so those probably won't be used as much. But I'm loathe to automatically send information in the opening context without express permission.

ellie: > But I'm loathe to automatically send information in the opening context without express permission.

Yeah totally. Honestly could be a few things that we want to configure for the opening context in this way. I'd be happy to just split them into two, or

```toml
[ai.opening]
send_cwd = true
send_last_command = true

[ai.tools]
search_history = true
```

could maybe work? as they're kinda different and for tools it can v clearly ask the user

BinaryMuse: Yeah that makes sense to me

ellie:
@greptile-apps one more time my dude

PR #3368, opened 2026-03-30
feat: Add 'atuin config' subcommand for reading and setting config values

BinaryMuse:
Adds a new `atuin config` command with three subcommands for inspecting and modifying `config.toml` without opening an editor.

### `atuin config get <key>`

Read a single value from the config file, the resolved config (after defaults + env overrides), or both.

```
$ atuin config get search_mode
fuzzy

$ atuin config get enter_accept
(not set in config file)

$ atuin config get enter_accept --resolved
false

$ atuin config get logs --resolved
logs.enabled = true
logs.dir = /home/user/.local/share/atuin/logs
logs.level = info
logs.search.file = search.log
...

$ atuin config get enter_accept --verbose
Config file:
  (not set in config file)

Resolved:
  false
```

### `atuin config set <key> <value>`

Write a value to `config.toml`, preserving existing formatting and comments (via `toml_edit`).

```
$ atuin config set search_mode fuzzy
$ atuin config set daemon.enabled true
$ atuin config set sync_frequency 600 --type string
```

Type detection: when the key already exists in the file, the existing TOML type is matched automatically (so setting `sync_frequency` to `600` keeps it as a string `"600"` if that's what was there before). For new keys, the value is auto-detected as bool/int/float/string. `--type` overrides everything.

Guards against accidentally overwriting a table with a scalar value.

### `atuin config print [key]`

Dump the config file (or a section of it) in TOML format.

```
$ atuin config print daemon
[daemon]
enabled = true
socket_path = "/tmp/atuin_daemon.sock"
pidfile_path = "/tmp/atuin_daemon.pid"
autostart = false
```

### Changes

- **`crates/atuin/src/command/client/config.rs`** — new subcommand module
- **`crates/atuin/src/command/client.rs`** — wire up `Config` variant
- **`crates/atuin-client/src/settings.rs`** — `get_config_value` now handles tables and arrays (via `ValueKind` matching), not just strings
- **`docs/docs/reference/config.md`** — command reference

### DIFF
--- a/crates/atuin/src/command/client/config.rs
+++ b/crates/atuin/src/command/client/config.rs
@@ -235,7 +235,7 @@ impl PrintCmd {
                             .as_table_like()
                             .expect("is_table()/is_inline_table() but no table"),
                         "",
-                        &mut vec![key.to_string()],
+                        &mut vec![key.clone()],
                     )?;
                 } else {
                     println!("{}", current.to_string().trim().trim_matches('"'));


ellie:
oh fuck yeah i love this

PR #3370, opened 2026-03-31
feat: Client-tool execution + permission system

BinaryMuse:
## Summary

Adds client-side tool execution to Atuin AI, starting with `atuin_history`. The server can request tool calls, which are executed locally with a permission system, and results are sent back to continue the conversation.

### Architecture

- **Modular extraction**: Decomposed the monolithic `inline.rs` into `stream.rs` (SSE framing), `dispatch.rs` (event handling), `state.rs` (domain state), `tools/` (tool types + execution), and `permissions/` (rule resolution)
- **ToolTracker**: Single source of truth for tool execution lifecycle — replaces the previous split between `pending_tool_calls` and `ConversationEvent` preview fields. Tracks each tool from permission check through completion, including cached previews for shell command output
- **Permission system**: File-based permission rules with tree-sitter shell command parsing for scoped matching (e.g. `git *` allows all git subcommands). Supports auto-allow, deny, and interactive ask with an inline selection UI
- **Shell command execution**: VT100 terminal emulator for live output preview in a `Viewport` widget, with interrupt support (Ctrl+C)

### Ships with

- `atuin_history` tool active by default — lets the AI search your command history
- Other tools (read_file, execute_shell_command, str_replace, file_create, file_insert) are defined but gated behind capability flags
- Permission denied flows surface errors inline on the tool call rather than as separate system messages
- Markdown list rendering fix (items no longer break across lines)
- Integrates main's new Hub auth flow and `ai.opening` settings (send_cwd, send_last_command)
- New `ai.capabilities` settings to control the capabilities advertised to the server
- If `ai.enabled` is `None`, app will prompt to enable, disable, or cancel. Disabling releases `?` keybind.

### DIFF
--- a/crates/atuin-client/src/settings.rs
+++ b/crates/atuin-client/src/settings.rs
@@ -679,7 +679,7 @@ pub struct Ai {
 
 #[derive(Default, Clone, Debug, Deserialize, Serialize)]
 pub struct AiCapabilities {
-    /// Whether the AI can request to search Atuin history. `None` = unset (defaults to enabled).
+    /// Whether the AI can request to search Atuin history. `None` = unset (defaults to enabled, and the ai will ask for permission).
     pub enable_history_search: Option<bool>,
 }
 


crates/atuin-ai/src/permissions/check.rs
```diff
@@ -0,0 +1,80 @@
...
+        Self { files }
+    }
+
+    pub async fn check<'t>(
+        &self,
+        request: &'t PermissionRequest<'t>,
+    ) -> Result<PermissionResponse> {
+        // Files are in order from deepest to shallowest, so we can stop at the first match.
+        // Within a file, deny rules take precedence over ask and allow rules.
+        // Ask rules take precedence over allow rules.
```
BinaryMuse: I find myself uncertain on this point. Maybe "Ask" should be highest priority? What do you think?

BinaryMuse: (@ellie)

BinaryMuse:
@greptileai Please review this draft

crates/atuin-ai/src/permissions/check.rs
ellie: Hmmm. I think we want to gate against accidental misconfiguration, with doing something dangerous being the worst thing

So, both ask and deny being higher priority than deny makes sense. In that case, ask and deny are both equally safe - neither will run a dangerous command without permission, for instance

However, if we also factor in being clear and not confusing, I think prioritizing ask over deny makes more sense.

crates/atuin-ai/src/commands/inline.rs
```diff
@@ -21,7 +21,7 @@ pub(crate) async fn run(
     if !settings.ai.enabled.unwrap_or(false) {
         emit_shell_result(
             Action::Print(
-                "Atuin AI is not enabled. Please enable it in your settings or run `atuin setup`."
+                "Atuin AI is not enabled. Please enable it in your settings with `atuin config set ai.enabled true` or run `atuin setup`.\nRun `atuin config set ai.enabled false` to disable the ? keybinding."
```
ellie: I think making it `atuin ai enable/disable` would be better. I'd like to make these prompts much less wordy!

.kilo/plans/table-rendering-for-markdown.md
```diff
@@ -0,0 +1,246 @@
+# Plan: Add Table Rendering to atuin-ai Markdown
```
ellie: we probably do not want to check this in

.opencode/plans/lively-nibbling-fog.md
```diff
@@ -0,0 +1,374 @@
+# Plan: Client-Side Shell Command Execution with Preview Window
```
ellie: or this haha

crates/atuin-ai/src/tui/components/atuin_ai.rs
```diff
@@ -55,15 +56,24 @@ fn atuin_ai(
             return EventResult::Ignored;
         };
 
-        // Ctrl+C always exits
+        // Ctrl+C — interrupt executing command or exit
```
ellie, line 59: do we also bind esc to this? people will probs be used to that

BinaryMuse: Not yet, but plan to - deferred the rest of the bash work

crates/atuin-ai/src/tui/events.rs
```diff
@@ -20,8 +27,18 @@ pub enum AiTuiEvent {
...
     /// Retry after error
     Retry,
     /// Exit the application
     Exit,
 }
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub(crate) enum PermissionResult {
+    Allow,
+    AlwaysAllowInDir,
```
ellie, line 41: thought for later, but maybe we want the same for workspaces too? not sure. less options is usaully better

BinaryMuse: This is poorly named; if in a git workspace, it'll say "in this workspace" instead of "in this directory" and use the workspace root.

crates/atuin-ai/Cargo.toml
```diff
@@ -39,9 +43,21 @@ async-stream = "0.3"
...
+glob-match = { workspace = true }
+regex = { workspace = true }
+time = { workspace = true }
+toml = "1.1"
+toml_edit = { workspace = true }
+tree-sitter-lib = { package = "tree-sitter", version = "0.26.8", optional = true }
+tree-sitter-bash = { version = "0.25.1", optional = true }
+tree-sitter-fish = { version = "3.6.0", optional = true }
+typed-builder = { workspace = true }
+vt100 = "0.16"
```
ellie: probably want to use the same version as Hex and enforce this via the cargo workspace

crates/atuin-client/src/settings.rs
```diff
@@ -671,6 +671,16 @@ pub struct Ai {
...
     pub opening: AiOpening,
+
+    /// Tool capability flags.
+    #[serde(default)]
+    pub capabilities: AiCapabilities,
+}
+
+#[derive(Default, Clone, Debug, Deserialize, Serialize)]
+pub struct AiCapabilities {
+    /// Whether the AI can request to search Atuin history. `None` = unset (defaults to enabled).
```
ellie:
```suggestion
    /// Whether the AI can request to search Atuin history. `None` = unset (defaults to enabled, and the ai will ask for permission).
```

crates/atuin-ai/Cargo.toml
BinaryMuse: ✅ ee1c5311799821d6e6d198bcb3d0e8c80f95c825

ellie (approved):
a few comments but nothing major, I love this! eye-declare code is nice to read, using tree sitter is neat, awesome work 🙏 

otherwise generally want to make sure we're really clear to users what is/is not shared, in our interface, code, and docs

crates/atuin-ai/src/tui/events.rs
ellie: oh awesome!!!!

crates/atuin-ai/src/tui/components/atuin_ai.rs
ellie: sounds goooood!

PR #3372, opened 2026-04-02
feat: option to disable mouse support

lilydjwg:
## Checks

This adds an option to disable mouse support, so that the terminal handles mouse events, making it easy to select text. I've found myself needing to copy from history frequently recently and holding Shift is not convenient.

However, this makes mouse scrolling not work as expected, e.g. in my case it scrolls six lines at a time.

Also see #1209.

### DIFF
--- a/crates/atuin-client/src/settings.rs
+++ b/crates/atuin-client/src/settings.rs
@@ -1094,6 +1094,7 @@ pub struct Settings {
     pub history_format: String,
     pub prefers_reduced_motion: bool,
     pub store_failed: bool,
+    pub no_mouse: bool,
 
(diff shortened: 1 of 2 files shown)


ellie (approved):
thank you!

xidiot:
@lilydjwg , if I add `no_mouse = true` in my case it scrolls few lines at a time. The same like you pointed in your initial comment. Am I missing something, do you have any idea?

lilydjwg:
It is expected. Setting `no_mouse = true` means that atuin doesn't handle mouse events and your terminal decides what to do with them, usually the terminal will convert one wheel event to several arrow key events.

PR #3407, opened 2026-04-14
feat: Allow resuming previous AI sessions

BinaryMuse:
This PR introduces session continuation to Atuin AI.

* Conversations with Atuin AI are stored in a local SQLite database
* Upon startup, Atuin AI tries to find a session to resume based on its directory/workspace and the time since the last event
* If found, Atuin AI will show a note that the session has been resumed, and an event is added to help the LLM know where the invocation boundaries are
* If not, Atuin AI will create a new conversation
* The user can create a new conversation with `/new`
* The new setting `ai.session_continue_minutes`, which defaults to `60`, controls how old the last event in a session can be before it's no longer considered for automatic resuming.

image

## Architecture

A new `SessionService` trait defines an API contract for a service that can manage session data. `LocalSessionService` implements this, with `DaemonSessionService` a possible future extension point.

`SessionManager` owns a `dyn SessionService` and delegates as appropriate.


BinaryMuse:
@greptileai do ur thing pls

crates/atuin-ai/migrations/20260413000000_create_ai_sessions.sql
```diff
@@ -0,0 +1,32 @@
...
+    git_root            TEXT,
+    created_at          INTEGER NOT NULL,
+    updated_at          INTEGER NOT NULL,
+    archived_at         INTEGER
+);
+
+CREATE INDEX idx_sessions_directory  ON sessions(directory);
+CREATE INDEX idx_sessions_git_root   ON sessions(git_root);
+CREATE INDEX idx_sessions_updated_at ON sessions(updated_at);
+CREATE INDEX idx_sessions_created_at ON sessions(created_at);
```
ellie, line 15: do we definitely want an index per column or would an index with several make more sense? unsure on the query pattern rn but will dig in. if we're mostly querying for a single column at a time then this makes sense, but otherwise might want to combine them

BinaryMuse: Ah yes, this was written early in the feature's development. `find_resumable_session` and `load_events` are probably the most interesting methods in `crates/atuin-ai/src/store.rs`

PR #3413, opened 2026-04-14
Enable atuin hex for illumos

papertigers:
This bumps the `portable-pty` dep which itself has a dependency on a newer termios that knows about illumos. Unless this was pinned for a particular reason I think this is fine.

With this dep bumped we can enable `atuin hex` on illumos.

## Checks

### DIFF
--- a/crates/atuin-hex/Cargo.toml
+++ b/crates/atuin-hex/Cargo.toml
@@ -13,9 +13,9 @@ repository = { workspace = true }
 [dependencies]
 clap = { workspace = true }
 
-[target.'cfg(all(unix, not(target_os = "illumos")))'.dependencies]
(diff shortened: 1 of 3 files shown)


ellie (approved):
yay!!! thank you!!

crates/atuin-hex/Cargo.toml
```diff
@@ -13,9 +13,9 @@ repository = { workspace = true }
 [dependencies]
 clap = { workspace = true }
 
-[target.'cfg(all(unix, not(target_os = "illumos")))'.dependencies]
+# [target.'cfg(all(unix, not(target_os = "illumos")))'.dependencies]
```
papertigers, line 16: Oops this can totally be removed. Sorry about letting this slip through!

PR #3415, opened 2026-04-15
fix: loss of loading spinners + tokio panic on exit

BinaryMuse:
PR fixes two issues:

1. Loading spinner for agent turn doesn't appear until server request is received - broken in 0cd40fe1875074b0fc33b783a1a8c20ada74d9f4. PR changes it so it renders but doesn't send to the API.
2. Occasional tokio panic on exit - due to a `tx` being dropped when shutting down; changed to drain the dispatch loop before shutting down the TUI loop

### DIFF
--- a/crates/atuin-ai/src/tui/dispatch.rs
+++ b/crates/atuin-ai/src/tui/dispatch.rs
@@ -84,7 +84,10 @@ fn persist_session(handle: &Handle<Session>, session_mgr: &mut SessionManager) {
         return;
     };
 
-    let rt = tokio::runtime::Handle::current();
+    let Ok(rt) = tokio::runtime::Handle::try_current() else {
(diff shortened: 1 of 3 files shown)


crates/atuin-ai/src/tui/dispatch.rs
```diff
@@ -84,7 +84,10 @@ fn persist_session(handle: &Handle<Session>, session_mgr: &mut SessionManager) {
         return;
     };
 
-    let rt = tokio::runtime::Handle::current();
+    let Ok(rt) = tokio::runtime::Handle::try_current() else {
+        tracing::debug!("tokio runtime gone, skipping session persist");
```
ellie: can we get a comment explaining this? semi scary comment without context

ellie (dismissed):

PR #3423, opened 2026-04-17
feat: AI tool rendering overhaul + edit_file tool

BinaryMuse:
Overhaul of how AI tool calls are modeled, rendered, and displayed in the Atuin AI TUI. Fixes bugs in shell command output capture, implements the `edit_file` tool with full safety infrastructure, and adds a diff preview for edits.

## VT100 / shell output fixes

- **CRLF normalization**: Piped subprocess output sends bare `\n` (no PTY), but the VT100 spec treats LF as "cursor down" without returning to column 0. Added `normalize_newlines_for_vt100` to insert `\r` before bare `\n`, matching the kernel terminal driver's ONLCR behavior. Fixes garbled/wrapped output in shell previews.

- **Trailing blank stripping**: The VT100 screen is 10 rows but the Viewport is 5. `vt100_screen_lines` now strips trailing blank rows so the Viewport's tail-mode display shows actual content instead of padding.

- **LLM row estimate**: `strip_ansi_via_vt100` estimated VT100 screen height from `byte_count / 120`, which severely undercounted short-line output (4 lines in 130 bytes → 2 rows → lost lines). Now estimates from newline count + a wrapping term.

## Tool rendering architecture

Replaced the monolithic tool-call rendering branch in `agent_turn_view` with a structured pipeline:

```
ToolTracker → TurnBuilder (build_render_data) → ToolRenderData / ToolGroup → per-tool view functions
```

- **`ToolRenderData` enum** (in `turn.rs`): Shell, FileRead, FileEdit, FileWrite, HistorySearch, Remote. Each variant carries tool-type-specific data (command string, file path, query + filter modes, preview, etc).

- **`ToolGroup`**: Consecutive FileRead or HistorySearch calls collapse into a single group event. TurnBuilder handles the grouping pass alongside the existing remote→ToolSummary collapse.

- **Per-tool view functions**: `shell_tool_view`, `file_edit_tool_view`, `file_write_tool_view`, `file_read_group_view`, `history_search_group_view`. Adding a new tool type = add a ToolRenderData variant + populate in build_render_data + write a view function.

## UI improvements

- **Shell tool view**: Status line with `hide_checkmark` spinner, output in `└`-indented HStack with dynamically-sized Viewport (shrinks to fit, caps at 5 lines).

- **File read groups**: "Read N files" header + indented list with `└` tree markers, `✓`/`✗` per-row status, smart path formatting (relative to cwd / `~/` / absolute).

- **History search groups**: "Searched Atuin history:" header + query list with `(filter_mode)` labels, "recent commands" in dim italic for empty queries.

- **Event spacing**: Single blank line between all agent turn events via explicit spacer. Empty/whitespace-only text chunks filtered in TurnBuilder to prevent phantom spacers and broken grouping.

- **Per-tool permission options**: Permission options are now tool-configurable. Edit tools show [Allow, Allow this file for this session, Always allow, Deny] while other tools keep the existing workspace-scoped options.

- **Edit diff preview**: Completed edits show a unified diff with dual-track line numbers (old-file numbers for removed lines, new-file for added/context), color-coded by operation (red/green/gray). Tabular layout with fixed-width gutter prevents wrapped text from spilling into line numbers.

## edit_file tool

Client-side implementation of `str_replace`-based file editing (matches Anthropic's trained format). Full safety infrastructure:

- **Freshness tracking** (`FileReadTracker`): Records content hash + mtime on file reads. Edit attempts check freshness — file must have been read first, and must not have been modified externally since. Uses Claude Code-compatible error messages for LLM recovery.

- **Snapshot system** (`SnapshotStore`): Backs up files before the first edit in a session. Snapshots stored at `~/.local/share/atuin/ai/snapshots/<session_id>/` with percent-encoded filenames and a manifest.json mapping back to original paths.

- **Atomic writes**: All file mutations use `tempfile` + fsync + rename. Preserves permissions from original file.

- **Match semantics**: `old_string` must be an exact match. Default requires unique match (error with count if ambiguous). `replace_all` opt-in replaces every occurrence.

- **Diff computation** (`imara-diff`): Histogram algorithm with indentation-aware postprocessing. Structured output (Context/Added/Removed lines with hunk merging) stored on `TrackedTool` for rendering.

- **Session-scoped edit permissions** (`EditPermissionCache`): "Allow this file for this session" caches a per-file grant with 1-hour TTL. Persisted in session metadata across CLI invocations.

- **Unified permission hierarchy**: Edit and Write share the `"Write"` rule name. A Write permission also implies Read — approving a file for editing automatically permits reading it.

- **Session metadata persistence**: Generic key-value table in SQLite (`session_metadata`) for file tracker state, edit permission grants, and future session-scoped data. Survives across CLI invocations within the same logical session.

## Capability alignment

- Fixed capability string mismatches between client and server (`client_v1_read` → `client_v1_read_file`, etc.)
- Added `enable_file_tools`, `enable_command_execution` to `AiCapabilities` settings
- Tilde expansion via `shellexpand` for all file paths from API tool calls

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -285,6 +285,7 @@ dependencies = [
  "fs-err",
  "futures",
  "glob-match",
+ "imara-diff",
  "pretty_assertions",
  "pulldown-cmark",
  "ratatui",
@@ -2256,6 +2257,16 @@ checksum = "3acae9609540aa318d1bc588455225fb2085b9ed0c4f6bd0d9d5bcd86f1a0344"
 dependencies = [
  "icu_normalizer",
  "icu_properties",
+]
+
+[[package]]
+name = "imara-diff"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2f01d462f766df78ab820dd06f5eb700233c51f0f4c2e846520eaf4ba6aa5c5c"
+dependencies = [
+ "hashbrown 0.15.5",
+ "memchr",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -314,6 +314,7 @@ dependencies = [
  "unicode-width 0.2.2",
  "uuid",
  "vt100",
+ "xxhash-rust",
 ]
 
 [[package]]
@@ -6721,6 +6722,12 @@ name = "x11rb-protocol"
 version = "0.13.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ea6fc2961e4ef194dcbfe56bb845534d0dc8098940c7e5c012a258bfec6701bd"
+
+[[package]]
+name = "xxhash-rust"
+version = "0.8.15"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fdd20c5420375476fbd4394763288da7eb0cc0b8c11deed431a91562af7335d3"
 
 [[package]]
 name = "yansi"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -294,6 +294,7 @@ dependencies = [
  "reqwest",
  "serde",
  "serde_json",
+ "shellexpand",
  "sqlx",
  "tempfile",
  "thiserror 2.0.18",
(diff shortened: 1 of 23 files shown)


BinaryMuse:
@greptileai Please re-review, but note that the write_tool will be implemented in a follow-up PR. Not a concern for this one.

PR #3432, opened 2026-04-20
feat: Implement write_file tool with overwrite safety

BinaryMuse:
## Summary

Implements the `write_file` client-side tool — creates new files or overwrites existing ones with an explicit `overwrite` flag for safety.

- **Overwrite flag**: Writing to an existing file without `overwrite: true` returns an error directing the LLM to set the flag or use `edit_file` for targeted changes. Prevents accidental overwrites.
- **Snapshots**: Existing files are backed up before overwriting (same infrastructure as `edit_file`).
- **Content preview**: Completed writes show the first 10 lines in gray with line numbers, plus "+ N more lines" for longer files.
- **Atomic writes**: Uses `tempfile` + fsync + rename (same as `edit_file`).
- **File tracker update**: After writing, the file is registered in the tracker so subsequent `edit_file` calls work without a separate read.
- **Permission**: Shares the `"Write"` rule with `edit_file` — one permission covers both tools.

Also includes fixes from the edit_file PR review:
- **xxh3 hash**: Replaced non-deterministic `DefaultHasher` with `xxhash-rust` for file freshness tracking across process restarts
- **Windows fixes**: `sanitize_path` now encodes `\` and strips drive prefixes; tests use `TempDir` instead of `NamedTempFile` to avoid open-handle rename failures
- **Unified permissions**: Edit and Write share `"Write"` rule name; Write implies Read

## Test plan

- [x] 5 new write_file execution tests (create, exists-error, overwrite, parent dirs, is-directory)
- [x] 137 total tests passing
- [x] Server-side tool definition with tests (separate hub PR)
- [x] Manual test: create new file via Atuin AI
- [x] Manual test: overwrite existing file (verify snapshot created)
- [x] Manual test: attempt write without overwrite flag (verify error message)

### DIFF
--- a/crates/atuin-ai/src/diff.rs
+++ b/crates/atuin-ai/src/diff.rs
@@ -101,6 +101,40 @@ impl EditPreview {
     }
 }
 
+/// Maximum lines to show in a write preview.
+const WRITE_PREVIEW_LINES: usize = 20;
+
+/// A content preview for a write_file operation.
+///
+/// Shows the first N lines of the written content plus a count of
+/// remaining lines if truncated.
+#[derive(Debug, Clone)]
+pub(crate) struct WritePreview {
+    /// First lines of content (up to WRITE_PREVIEW_LINES).
+    pub lines: Vec<String>,
+    /// Total number of lines in the written file.
+    pub total_lines: usize,
+}
+
+impl WritePreview {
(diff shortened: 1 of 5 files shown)


BinaryMuse:
@greptileai

ellie (approved):
:shipit:

PR #3440, opened 2026-04-22
fix: require all subcommands covered for shell allow rules

BinaryMuse:
## Summary

- Compound shell commands (e.g. `git add . && npm test`) now require **every** subcommand to be individually covered by at least one `allow` rule. Previously, a single matching subcommand was enough to allow the entire compound command — so `Shell(git *)` would silently permit `git add . && rm -rf /`.
- Deny and ask rules retain the existing any-match semantics: if **any** subcommand hits a deny/ask rule, the whole command is denied/asked.
- Adds `all_covered_by` to the `PermissableToolCall` trait with a default that preserves existing behavior for non-compound tools (Read, Write, AtuinHistory).

## Test plan

- [x] 7 new unit tests covering simple commands, compound all-covered, compound partially-covered, unscoped/wildcard rules, non-shell default, and `matches_rule` any-semantics for deny/ask
- [x] All 174 existing tests pass


BinaryMuse:
@greptileai Please review this draft

PR #3442, opened 2026-04-22
fix: bound daemon history RPCs so a wedged daemon doesn't hang the shell

truffle-dev:
Fixes #3382.

The daemon-enabled `atuin history start` path calls into gRPC with no per-RPC deadline. When the socket is live but the daemon never replies (crash-restart mid-migration, or systemd holding the listen socket while the unit is in StartLimitBurst), the shell hook's preexec command substitution blocks on `atuin` forever. The reporter recovered by SSHing in without a TTY.

Fix: wrap the blocking RPC paths in `tokio::time::timeout` keyed to `settings.local_timeout` (2.0s default, floored at 500ms). Applies to `start_history`, `end_history`, and `probe`, plus the retry calls that follow `restart_daemon`. `handle_daemon_start`/`end` already swallow daemon errors and fall back to a locally-generated history id to avoid breaking the shell, so a timed-out call just routes through that existing fallback.

### Regression evidence

Listener that accepts but never responds (full script in the issue comment):

```
# before:  real 10.004s, exit 124 (external timeout killed it, shell would have hung forever)
# after:   real  2.015s, exit 0   (timeout hit, fallback returned id, shell continued)
```

Stock binary hangs to the external `timeout 10` cap with no output. Patched binary returns a local id in ~2s matching the `local_timeout` cap, which is what the shell hook expects.

### Why `local_timeout`

Reusing the existing setting keeps the user-facing surface area unchanged. If you'd rather have a dedicated `daemon.request_timeout`, I'm happy to plumb it through as a follow-up. Flagged that choice on the issue.

### Not included here

- Timeouts on `SearchClient`, `ControlClient`, and `tail_history`. Search happens from the interactive UI where hangs are recoverable (Ctrl-C). Control and tail are not called from the shell's critical path. Narrowed scope to the two RPCs that actually wedge the shell.
- A regression test. The hang requires a real accept-but-silent UnixListener, which means spawning a listener task in a test that doesn't touch the daemon binary. Happy to add one if you'd like; the repro in #3382 serves as the manual check.

### DIFF
--- a/crates/atuin/src/command/client/daemon.rs
+++ b/crates/atuin/src/command/client/daemon.rs
@@ -16,7 +16,7 @@ use clap::Subcommand;
 use daemonize::Daemonize;
 use eyre::{Result, WrapErr, bail, eyre};
 use fs4::fs_std::FileExt;
-use tokio::time::sleep;
+use tokio::time::{sleep, timeout};
 
 #[derive(clap::Args, Debug)]
 pub struct Cmd {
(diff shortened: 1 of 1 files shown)


truffle-dev:
Checking in — anything else needed before this can merge?

ellie:
Not super keen to merge code that 

1. has probably never been tested
2. has not been reviewed before the PR is opened

Appreciate what you're trying to do here, but this is also part of why GitHub has been under crippling load.

PR #3453, opened 2026-04-26
fix: atuin update on windows

### LINKED ISSUE #3451: [Bug]: Windows PowerShell - Update Failing with "it is being used by another process"

Not sure when this started, but the last few builds.

Seems I have to reboot in order to update now. :(

Process Explorer does NOT show atuin open or another powershell / cmd.

### What happened?

```
atuin update
Checking for updates...
downloading atuin 18.15.2 (x86_64-pc-windows-msvc)
installing to C:\Users\Will\.atuin\bin
We encountered an error trying to perform the installation;
please review the error messages below.

The process cannot access the file 'C:\Users\Will\.atuin\bin\atuin.exe' because it is being used by another process.
  x The installation failed. Output from the installer:
```

### Atuin doctor output

```yaml
Atuin Doctor
Checking for diagnostics


{
  "atuin": {
    "version": "18.11.0",
    "commit": "63de4501a4a2c4af98e520689fa74cd314d7f643",
    "sync": null,
    "sqlite_version": "3.46.0"
  },
  "shell": {
    "name": "pwsh.exe",
    "default": "powershell",
    "plugins": [
      "atuin"
    ],
    "preexec": "built-in"
  },
  "system": {
    "os": "Windows",
    "arch": "x86_64",
    "version": "11 (28000)",
    "disks": [
      {
        "name": "",
        "filesystem": "NTFS"
      },
      {
        "name": "",
        "filesystem": "NTFS"
      },
      {
        "name": "",
        "filesystem": "exFAT"
      }
    ]
  }
}
```

### Code of Conduct

ltrzesniewski:
This fixes the `atuin update` command on Windows.

Windows doesn't let you overwrite a running exe, but it lets you rename it. This PR special-cases the official `update` plugin by renaming the running `atuin.exe` to `atuin.old` before the update, and rolling it back if the update fails.

Note that the `atuin.old` file is left behind on success, which shouldn't be a problem in practice: it will be overwritten on the next call to `atuin update` (also deleted if there's no update available), and is located in `~/.atuin/bin`, which is an isolated location specific to Atuin.

Fixes #3451

## Checks

### DIFF
--- a/crates/atuin-client/src/plugin.rs
+++ b/crates/atuin-client/src/plugin.rs
@@ -42,7 +42,7 @@ impl OfficialPluginRegistry {
                 "Update atuin to the latest version",
                 "The 'atuin update' command is provided by the atuin-update plugin.\n\
                  It is only installed if you used the install script\n  \
-                 If you used a package manager (brew, apt, etc), please continue to use it for updates"
+                 If you used a package manager (brew, apt, etc), please continue to use it for updates",
             ),
         );
     }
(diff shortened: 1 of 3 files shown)


ltrzesniewski:
@greptileai - I like turtles 🐢

ltrzesniewski:
@greptileai

ellie (approved):
nice one! thanks!

PR #3461, opened 2026-05-01
feat(pty-proxy): mirror inner shell cwd via OSC 7

### LINKED ISSUE #3296: [Bug]: When `hex` is enabled, the working directory is not passed to the Ghostty

When opening an new terminal window, the current working directory of the existing terminal should be used for the new window.

I'm using Ghostty. I can't replicate this issue with macOS Terminal.app

### What happened?

1. Run `atuin hex` in a terminal window in `directory-a/`
2. Change directory to something else
3. Open a new terminal tab/window
4. Observe that the previous directory (`directory-a/`) is used for the new terminal window

### Atuin doctor output

```yaml
{
  "atuin": {
    "version": "18.13.3",
    "sync": null,
    "sqlite_version": "3.46.0"
  },
  "shell": {
    "name": "zsh",
    "default": "zsh",
    "plugins": [
      "atuin"
    ],
    "preexec": "built-in"
  },
  "system": {
    "os": "Darwin",
    "arch": "arm64",
    "version": "26.3",
    "disks": [
      {
        "name": "Macintosh HD",
        "filesystem": "apfs"
      },
      {
        "name": "Macintosh HD",
        "filesystem": "apfs"
      }
    ]
  }
}
```

### Code of Conduct

xav-ie:
Fixes #3296.
Partially addresses #3337 (cwd-tracking portion; shell-integration hooks are a separate problem).

## The bug

Terminals and multiplexers (tmux's `pane_current_path`, Ghostty / Kitty / WezTerm "open in same dir", etc.) read a pane's cwd by inspecting the foreground process via `proc_pidinfo` (macOS) or `/proc/<pid>/cwd` (Linux). When `atuin pty-proxy` is active, the foreground process is the proxy — *not* the inner shell where the user actually runs `cd`. The inner shell lives on a private PTY invisible to the multiplexer.

Result: terminal features that depend on knowing the current directory always see the proxy's startup directory, not wherever the user `cd`-ed to.

## The fix

Inner shells emit OSC 7 (`\e]7;file://<host><path>\e\\`) on every prompt change. The proxy parses the wire payload from the inner-PTY byte stream and `chdir`s itself to match. Subsequent `proc_pidinfo` / `/proc` lookups by the terminal then see the correct cwd.

All path encoding lives in a new `atuin pty-proxy emit-osc7` subcommand — each shell's init only needs a one-line hook that calls it. This avoids per-shell URL encoders (4 implementations with subtly different RFC 3986 edge cases) in favor of one well-tested Rust source of truth, while keeping the wire format strictly RFC 3986–compliant for any other terminal listening to OSC 7.

Per-shell hooks:
- **bash / zsh:** `_atuin_pty_proxy_osc7` runs in `precmd` (zsh) or `PROMPT_COMMAND` (bash), gated on `$PWD` change.
- **fish:** `--on-variable PWD` event handler.
- **nushell:** `env_change.PWD` hook.

## Per-prompt cost

The hooks fork+exec `atuin pty-proxy emit-osc7` per `cd` (gated on `$PWD` change). The atuin shell integration **already** synchronously forks+execs `atuin history start` on every preexec, plus a backgrounded `atuin history end` per command. The OSC 7 emission piggybacks on this existing budget — backgrounded, gated on PWD change — so it adds at most one extra `atuin` invocation per directory change, strictly less than the existing per-command cost.

## Trust boundary

The proxy trusts the OSC 7 wire payload and `chdir`s based on it. Anyone who can write to the inner PTY can drive the proxy's cwd to any path the user can already `chdir` to themselves. This is the same trust model as every other OSC sequence terminals honor (OSC 8 hyperlinks, OSC 52 clipboard, etc.) — if attacker-controlled output (e.g. `cat /random/file`) can spew bytes into your terminal, you've already lost.

Defense-in-depth in the parser:
- Reject non-absolute paths (OSC 7 paths are absolute by definition; relative = malformed).
- Reject paths containing `..` components (defends against `cat /attacker/file` redirecting cwd via traversal).
- 4096-byte cap on parameter buffer (drop, no panic, on overflow).

## Alternative considered

Kernel-side cwd lookup (`proc_pidinfo` / `/proc/<inner-shell-pid>/cwd`) instead of trusting the wire payload would eliminate the trust boundary entirely — the proxy would query the kernel directly and the wire payload would be a decorative refresh trigger. I explored this on a separate branch and it works, but it adds OS-specific code paths (libproc on macOS, procfs on Linux) and the security delta is small relative to the existing OSC trust model. Happy to revisit if you'd prefer that approach.

## Testing

- 11 unit tests in `crates/atuin-pty-proxy/src/osc7.rs`, including: streaming chunk boundaries, BEL vs ESC \\ terminators, percent-encoded multibyte UTF-8, malformed URI rejection, relative-path rejection, `..`-traversal rejection, over-long payload drop.
- Manually verified end-to-end on macOS with tmux + nushell: `cd` in the inner shell, then `tmux split-window` opens a new pane in the right directory.
- Module gated on `#[cfg(unix)]` so Windows builds (`atuin` with default features including `pty-proxy`) remain unaffected.

## Checks

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -510,6 +510,7 @@ dependencies = [
  "clap",
  "crossterm",
  "eyre",
+ "percent-encoding",
  "portable-pty",
  "signal-hook",
  "vt100",
--- a/crates/atuin-pty-proxy/Cargo.toml
+++ b/crates/atuin-pty-proxy/Cargo.toml
@@ -16,6 +16,7 @@ clap = { workspace = true }
 [target.'cfg(unix)'.dependencies]
 crossterm = { workspace = true }
 eyre = { workspace = true }
+percent-encoding = "2.3"
 portable-pty = "0.9"
 signal-hook = "0.3"
 vt100 = { workspace = true }
--- a/crates/atuin-pty-proxy/src/lib.rs
+++ b/crates/atuin-pty-proxy/src/lib.rs
@@ -5,6 +5,8 @@ mod debug;
 #[cfg(unix)]
 mod osc133;
 #[cfg(unix)]
+mod osc7;
+#[cfg(unix)]
 mod pty_proxy;
 #[cfg(unix)]
 mod runtime;
--- a/crates/atuin-pty-proxy/src/osc7.rs
+++ b/crates/atuin-pty-proxy/src/osc7.rs
@@ -1,0 +1,282 @@
+//! Streaming parser for OSC 7 (current working directory) escape sequences.
+//!
+//! OSC 7 is the de-facto standard for shells to inform the terminal of their
+//! current working directory:
+//!
+//! ```text
+//! ESC ] 7 ; file://<hostname>/<path> ST
+//! ```
+//!
+//! ST is either BEL (0x07) or ESC \ (0x1B 0x5C).  The path is percent-encoded
+//! per RFC 3986 (e.g. spaces as %20).
+//!
+//! atuin pty-proxy parses this from the inner shell's output and updates its
+//! own cwd to match, so that terminals/multiplexers (tmux, Ghostty, Kitty,
+//! etc.) that read pane cwd via process introspection see the inner shell's
+//! directory rather than the proxy's startup directory.
+//!
+//! # Design goals
+//!
+//! Mirrors `osc133.rs`: streaming, transparent (caller still forwards bytes).
+//! The parameter buffer is pre-sized to its hard cap so steady-state parsing
+//! does not allocate.
+
+use percent_encoding::{AsciiSet, NON_ALPHANUMERIC, percent_decode};
+use std::ffi::OsString;
+use std::os::unix::ffi::OsStringExt;
+use std::path::PathBuf;
+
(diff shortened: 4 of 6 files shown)


xav-ie:
Rebased and force-pushed: ported onto the renamed `atuin-pty-proxy` crate (formerly `atuin-hex`, renamed in #3473), so this applies cleanly to current `main` again.

Scoped down to **just** the OSC 7 feature — this branch was previously stacked on the `--shell` work (now its own PR, #3327), so those changes are no longer included here.

The OSC 7 streaming parser (`osc7.rs`), the `emit-osc7` subcommand, the per-shell PWD hooks (bash `PROMPT_COMMAND` / zsh `precmd` / fish `--on-variable PWD` / nu `env_change.PWD`), and the cwd-mirroring in the pty loop all carried over. Relative-path and `..`-traversal rejection preserved.

BlackDex:
Is there a reason for this to be still in draft?
I have not tested it, but would be nice if OSC7 would work again :).

markovejnovic:
Closing this since the mentioned issue is fixed.

Issue #3456, opened 2026-04-27
[Bug]: Ubuntu 26.04 Install causing bash gibberish, history not updating


Just-Some-Guy:
### What did you expect to happen?

Every command I do in Terminal gives gibberish before the results.  I am not seeing atuin history not being updated

[Atuin Ubuntu 26.04 Problem Notes.pdf](https://github.com/user-attachments/files/27135312/Atuin.Ubuntu.26.04.Problem.Notes.pdf)

### What happened?

After I installed it, I edited the .toml file to change the sync server to self-hosted server.   When I saved and exited nano bash spit out the entire script and now is not functioning and prefixing any results with gibberish.

See notes PDF above attached with results pasted in

### Atuin doctor output

```yaml
{
  "atuin": {
    "version": "18.15.2",
    "commit": "1b4eb4901d256637c75849f0e4f82e476934b922",
    "sync": {
      "auth_state": "Hub (authenticated)",
      "records": true,
      "auto_sync": true,
      "last_sync": "2026-04-27 17:22:04.365233092 +00:00:00"
    },
    "sqlite_version": "3.46.0"
  },
  "shell": {
    "name": "bash",
    "default": "bash",
    "plugins": [
      "atuin",
      "bash-preexec"
    ],
    "preexec": "none"
  },
  "system": {
    "os": "Ubuntu",
    "arch": "x86_64",
    "version": "26.04",
    "disks": [
      {
        "name": "/dev/sda4",
        "filesystem": "ext4"
      },
      {
        "name": "/dev/sda3",
        "filesystem": "vfat"
      },
      {
        "name": "/dev/sdc1",
        "filesystem": "exfat"
      },
      {
        "name": "/dev/sda2",
        "filesystem": "ntfs3"
      }
    ]
  }
}
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

pabloh:
I'm experiencing a similar issue (history does not update), when using `atuin` with `preexec` `0.6` in Ubuntu 26.04

truffle-dev:
The gibberish (`__bp_invoke_preexec_from_ps0 "<arg>" >&2; }` before each command's output) is the signature of bash-preexec's PS0 hook from rcaloras/bash-preexec#166, which lives only in master, not the 0.6.0 release.

**Install path.** `install.sh` line 65 fetches bash-preexec from master, not a release tag:
```sh
[...]

PR #3469, opened 2026-05-10
fix(install): pin bash-preexec to 0.6.0 release

ChrisJr404:
## What

`install.sh` and the bash install docs fetch `bash-preexec.sh` from `master` instead of a release tag. This PR pins all three call sites to the `0.6.0` tag.

## Why

A commit landed on `master` after the `0.6.0` release that adds a PS0 based preexec hook for bash 5.3+ using the new "command substitution that has no value" syntax:

```sh
PS0=${PS0-}'${ __bp_invoke_preexec_from_ps0 "$_" >&2; }'
```

Ubuntu 26.04 ships bash 5.3.9, which activates that branch. Under some shell environments the funsub body is printed verbatim before each command instead of being suppressed cleanly, producing the `__bp_invoke_preexec_from_ps0 "..." >&2; }` text the issue reports.

I can reproduce the literal leak locally with bash 5.3.9 by sourcing the `master` file with `promptvars` off:

```
$ echo HELLO
${ __bp_invoke_preexec_from_ps0 "$_" >&2; }HELLO
```

The reporters in #3456 have `promptvars on`, so something else in their `~/.bashrc` flips the funsub from "fires silently" to "prints body", and I haven't pinned down the exact environmental trigger from the issue alone. What's clear is that `0.6.0` doesn't set `PS0` at all, so pinning to that tag sidesteps the whole code path regardless of which environment trigger is at play.

I'm not 100% sure this matches every report in the thread (one user noted `ble.sh` fixed the history symptom but not the gibberish), but the gibberish signature is unambiguously the PS0 hook from `master`, and pinning a release tag is the right move either way for a fetched dependency.

## History capture

`0.6.0` still installs the DEBUG-trap preexec path, which fires `__atuin_preexec` on every command, so atuin history capture continues to work on bash 5.3.9.

## Files changed

- `install.sh` line 65 (the actual install path users hit)
- `docs/docs/guide/installation.md` (English docs `bash-preexec` manual setup instructions)
- `docs-i18n/zh-CN/README.md` (Simplified Chinese mirror of the same)

## Long term

The underlying bug belongs upstream in `bash-preexec`. Pinning here protects new atuin installs in the meantime; a future bump to a new `bash-preexec` release that fixes the PS0 funsub is a small follow-up.

Refs #3456

### DIFF
--- a/docs-i18n/zh-CN/README.md
+++ b/docs-i18n/zh-CN/README.md
@@ -200,7 +200,7 @@ zinit load ellie/atuin
 我们需要设置一些钩子（hooks）, 所以首先需要安装 bash-preexec :
 
 ```
-curl https://raw.githubusercontent.com/rcaloras/bash-preexec/master/bash-preexec.sh -o ~/.bash-preexec.sh
+curl https://raw.githubusercontent.com/rcaloras/bash-preexec/0.6.0/bash-preexec.sh -o ~/.bash-preexec.sh
 echo '[[ -f ~/.bash-preexec.sh ]] && source ~/.bash-preexec.sh' >> ~/.bashrc
 ```
 
--- a/docs/docs/guide/installation.md
+++ b/docs/docs/guide/installation.md
@@ -225,7 +225,7 @@ After installing, remember to restart your shell.
         To use bash-preexec, download and initialize it
 
         ```shell
-        curl https://raw.githubusercontent.com/rcaloras/bash-preexec/master/bash-preexec.sh -o ~/.bash-preexec.sh
+        curl https://raw.githubusercontent.com/rcaloras/bash-preexec/0.6.0/bash-preexec.sh -o ~/.bash-preexec.sh
         echo '[[ -f ~/.bash-preexec.sh ]] && source ~/.bash-preexec.sh' >> ~/.bashrc
         ```
 
--- a/install.sh
+++ b/install.sh
@@ -62,7 +62,7 @@ fi
 # shellcheck disable=SC2016
 
 if ! grep -q "atuin init bash" ~/.bashrc; then
-  curl --proto '=https' --tlsv1.2 -LsSf https://raw.githubusercontent.com/rcaloras/bash-preexec/master/bash-preexec.sh -o ~/.bash-preexec.sh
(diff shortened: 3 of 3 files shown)


akinomyoga:
This was also reported at https://github.com/rcaloras/bash-preexec/issues/186.

> The reporters in #3456 have `promptvars on`, so something else in their `~/.bashrc` flips the funsub from "fires silently" to "prints body", and I haven't pinned down the exact environmental trigger from the issue alone.

The problem is caused by `/etc/profile.d/vte.sh`, which leaves an unclosed backslash escape as `PS0='<something>\'`. Then, when any other frameworks append any expansions such as `PS0+='$var'`, the beginning `$` is escaped and the entire expansion is suppressed.

> I'm not 100% sure this matches every report in the thread (one user noted `ble.sh` fixed the history symptom but not the gibberish), but the gibberish signature is unambiguously the PS0 hook from `master`,

The user should have loaded both `ble.sh` and `bash-preexec` in Bash sessions, because `ble.sh` doesn't include any strings like `${ __bp_invoke_preexec_from_ps0 "..." >&2; }`. It should be noted that, even if the user doesn't explicitly load `bash-preexec.sh`, a certain version of `bash-preexec` may be included in other scripts, such as the terminal's shell integration scripts (by iTerm2, kitty, Ghostty, VS Code Terminal, etc.), injected on the startup of Bash, or by a framework such as Bash-it.

akinomyoga:
Ah, sorry. I intended to reply to the original issue #3456 (instead of this PR #3469). I'll also reply to the original issue.

akinomyoga:
We've submitted a patch to the upstream: https://gitlab.gnome.org/GNOME/vte/-/work_items/2951.

taylordotfish:
Closing this as the [upstream issue in VTE](https://gitlab.gnome.org/GNOME/vte/-/work_items/2951) has been fixed; thank you @akinomyoga for your help with this!

For anyone who needs an urgent fix for this issue before their version of VTE is updated, @akinomyoga's original patch should work. In `/etc/profile.d/vte.sh`, replace this line:

```sh
    PS0=$(__vte_termprop_signal "vte.shell.preexec")
```

with these five lines:

```sh
    if shopt -q promptvars; then
        PS0=$(__vte_termprop_signal "vte.shell.preexec" | command sed 's/\\/\\\\\\\\/g')
    else
        PS0=$(__vte_termprop_signal "vte.shell.preexec" | command sed 's/\\/\\\\/g')
    fi
```

This is different from the final fix that landed in VTE, but it should work just as well.

PR #3480, opened 2026-05-14
fix(atuin-nucleo): duplicated "the" in utf32_str/lib doc-comments

otjdiepluong:
Two one-line typo fixes for duplicated "the" in atuin-nucleo doc-comments:
- `crates/atuin-nucleo/matcher/src/utf32_str.rs` — "/// times (making the the up-front cost more worth it). That means that its" → "...making the up-front cost more worth it..."
- `crates/atuin-nucleo/src/lib.rs` — "/// Restart the the item stream. Removes all items and disconnects all" → "/// Restart the item stream. Removes all items and disconnects all"

No code/behavior change.


ellie:
bot

PR #3487, opened 2026-05-14
refactor: pull `fn into_utc` into atuin-server-database crate

oxo42:
The `fn into_utc` was defined in both `atuin-server-postgres` and `atuin-server-sqlite`, with tests being in the postgres crate.  This pulls the function into the `atuin-server-database` crate along with the tests and references it from both crates.

## Checks


ellie (approved):
thank you!!

PR #3489, opened 2026-05-18
Highlight `Ctrl-r` keybinding on docs page

mrnugget:
I had to look 5 times and I think highlighting it like this might've reduced it to 2 :)

## Checks

### DIFF
--- a/docs/docs/guide/basic-usage.md
+++ b/docs/docs/guide/basic-usage.md
@@ -15,11 +15,11 @@ While you work, Atuin records:
 
 ## Opening and using the TUI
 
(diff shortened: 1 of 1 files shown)


ellie (approved):
Good idea, thank you!

Is there anywhere else you looked before finding it? Maybe we could put it in a few places 🤔

mrnugget:
Yeah, I looked here and expected it to have a list of all keybindings: https://docs.atuin.sh/cli/configuration/key-binding/

And then I looked here and also couldn't find a list of default keybindings: https://docs.atuin.sh/cli/configuration/advanced-key-binding/

Ideally there'd be a list with the 5 most common keybindings that I need to know.

PR #3496, opened 2026-05-20
Add agent history filter mode

dhruv-anand-aintech:
## Summary
- add an agent filter mode to interactive history search
- filter user vs known-agent history consistently across DB, daemon fallback, and skim paths
- document the new filter mode

## Tests
- cargo fmt --check
- cargo test -p atuin-history sort

### DIFF
--- a/crates/atuin-client/src/database.rs
+++ b/crates/atuin-client/src/database.rs
@@ -404,6 +404,7 @@ impl Database for Sqlite {
                 }
                 FilterMode::Directory => query.and_where_eq("cwd", quote(&context.cwd)),
                 FilterMode::Workspace => query.and_where_like_left("cwd", &git_root),
+                FilterMode::Agent => &mut query,
             };
         }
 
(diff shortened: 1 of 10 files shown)


ellie:
Hey! Thanks for contributing, but I think this needs some discussion first; as a human, what do you find useful about being able to interactively search agent history? Happy to reopen if we get this figured out, but in my experience most agent history is an awful lot of noise

dhruv-anand-aintech:
Hi @ellie, thanks for having a look.
The main usecase is to relaunch servers, curl commands,  in the same way that the agent just ran on its own. Often this info gets lost in the chat transcripts and you need another chat turn to ask it to give you a clean command. I'd rather just scroll 1 command up in the history and hit enter.
I'll just keep these changes on my fork of your repo for now.

PS: I'm building a tool https://github.com/dhruv-anand-aintech/agent-rules-sync that keeps all agent-specific context synced, and I thought adding a log of terminal commands run by each agent would be a good addition (not for sync, but to view as a list).

PR #3502, opened 2026-05-26
fix: Atuin hangs when attempting to spawn daemon from Ctrl+R invocation

### LINKED ISSUE #3499: [Bug]: Ctrl+R -> Tab -> Ctrl+R -> Esc causes Atuin to freeze indefinitely if the daemon is enabled but not running

I expected Atuin to display history, edit the most recent command, display search results for that command, and then exit back to the shell.

### What happened?

Atuin behaves as expected until Esc is pressed. Then, the terminal is blank except for the cursor and does not respond to any keypresses, even Ctrl+C.

I'm able to reproduce the issue consistently as follows:

1. In terminal A, start `sh` or otherwise temporarily disable Atuin in that terminal only, to avoid automatically restarting the daemon.
2. In terminal A, kill the Atuin daemon, e.g., with `pkill -x atuin`.
3. In terminal B, press Ctrl+R, then Tab, then Ctrl+R, then Esc.

I also experienced this issue immediately after installing Atuin, without ever killing the daemon, so it seems like the relevant factor is more whether the daemon is running rather than whether it has been killed per se.

Here's a screen recording of the issue:

https://github.com/user-attachments/assets/6e9f0b25-03b2-46c0-8ce6-be775d8d42e4

### Atuin doctor output

```yaml
{
  "atuin": {
    "version": "18.16.1",
    "commit": "671f96b60dac49d1d2de73cc0812986a5e22ce7b",
    "sync": {
      "records": true,
      "auto_sync": true,
      "last_sync": "2026-05-25 9:09:43.888177193 +00:00:00"
    },
    "sqlite_version": "3.46.0"
  },
  "shell": {
    "name": "bash",
    "default": "bash",
    "plugins": [
      "atuin",
      "bash-preexec"
    ],
    "preexec": "bash-preexec"
  },
  "system": {
    "os": "Debian GNU/Linux",
    "arch": "x86_64",
    "version": "13",
    "disks": [
      {
        "name": "(redacted)",
        "filesystem": "ext4"
      },
      {
        "name": "(redacted)",
        "filesystem": "ext4"
      }
    ]
  }
}
```

### Code of Conduct

BinaryMuse:
This PR fixes a shell hang when daemon autostart happens from the interactive search widget.

The bash/zsh/fish integrations run `atuin search -i` under command substitution and swap stdout/stderr so the TUI can draw to the terminal while the selected command is captured:

```sh
3>&1 1>&2 2>&3
```

This leaves fd 3 open in the atuin search process. If search autostarts the daemon, the spawned long-running `atuin daemon start --daemonize` inherits fd 3, the command-substitution pipe, so the shell keeps waiting for EOF until the daemon is killed.

The fix: close fd 3 after the swap in the non-tmux bash/zsh/fish paths:

```sh
3>&1 1>&2 2>&3 3>&-
```

Tested on zsh; I still need to confirm for bash and fish. Near as I can tell, the other shell integrations don't need this change.

Fixes #3499

### DIFF
--- a/crates/atuin/src/command/client/search/interactive.rs
+++ b/crates/atuin/src/command/client/search/interactive.rs
@@ -1536,16 +1536,16 @@ impl Drop for Stdout {
             tracing::error!(?e, "Failed to pop keyboard enhancement flags");
         }
 
-        if !self.inline_mode {
-            if let Err(e) = execute!(self.writer, terminal::LeaveAlternateScreen) {
-                tracing::error!(?e, "Failed to leave alt screen mode");
-            }
+        if !self.inline_mode
+            && let Err(e) = execute!(self.writer, terminal::LeaveAlternateScreen)
+        {
+            tracing::error!(?e, "Failed to leave alt screen mode");
         }
 
-        if !self.no_mouse {
-            if let Err(e) = execute!(self.writer, event::DisableMouseCapture) {
-                tracing::error!(?e, "Failed to disable mouse capture");
-            }
+        if !self.no_mouse
+            && let Err(e) = execute!(self.writer, event::DisableMouseCapture)
+        {
+            tracing::error!(?e, "Failed to disable mouse capture");
         }
 
         if let Err(e) = execute!(self.writer, event::DisableBracketedPaste) {
--- a/crates/atuin/src/command/client/search/interactive.rs
+++ b/crates/atuin/src/command/client/search/interactive.rs
@@ -1532,17 +1532,17 @@ impl Stdout {
 impl Drop for Stdout {
     fn drop(&mut self) {
         #[cfg(not(target_os = "windows"))]
-        execute!(self.writer, PopKeyboardEnhancementFlags).unwrap();
+        let _ = execute!(self.writer, PopKeyboardEnhancementFlags);
 
         if !self.inline_mode {
-            execute!(self.writer, terminal::LeaveAlternateScreen).unwrap();
+            let _ = execute!(self.writer, terminal::LeaveAlternateScreen);
         }
         if !self.no_mouse {
-            execute!(self.writer, event::DisableMouseCapture).unwrap();
+            let _ = execute!(self.writer, event::DisableMouseCapture);
         }
-        execute!(self.writer, event::DisableBracketedPaste).unwrap();
+        let _ = execute!(self.writer, event::DisableBracketedPaste);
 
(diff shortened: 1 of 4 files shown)


crates/atuin/src/command/client/search/interactive.rs
```diff
@@ -1532,17 +1532,17 @@ impl Stdout {
 impl Drop for Stdout {
     fn drop(&mut self) {
         #[cfg(not(target_os = "windows"))]
-        execute!(self.writer, PopKeyboardEnhancementFlags).unwrap();
+        let _ = execute!(self.writer, PopKeyboardEnhancementFlags);
```
ellie: why are we swallowing these errors?

unwrap is probably not ideal, but at least is loud of this operation fails. we should confirm

1. _Can_ this return an error under any normal circumstance?
2. If so, we should return this to the user in a way that makes sense, rather than leave the terminal in an unknown state
3. If not, we should leave a not to describe why it is OK to swallow the error message

crates/atuin/src/shell/atuin.zsh
```diff
@@ -119,15 +119,21 @@ _atuin_search() {
...
     zle reset-prompt
     # re-enable bracketed paste
     # shellcheck disable=SC2154
     echo -n ${zle_bracketed_paste[1]} >/dev/tty
 
+    if (( __atuin_status != 0 )); then
+        [[ -n $output ]] && print -r -- "$output" >/dev/tty
+        return $__atuin_status
+    fi
+
```
ellie, line 136: 3>&- is all that was needed for the fix. this is probably nice to have, assuming included while you were in the space?

crates/atuin/src/command/client/search/interactive.rs
BinaryMuse: The thought was that if one errors, we never disable raw mode. But if these fail, I suppose it's likely that disabling raw mode will fail as well?

crates/atuin/src/shell/atuin.zsh
BinaryMuse: Yes - added this in the process of debugging the original issue.

crates/atuin/src/command/client/search/interactive.rs
BinaryMuse: Moved to `if let Err(...)` pattern as discussed.

ellie (approved):

PR #3508, opened 2026-05-28
feat: add command aliases for stats

### LINKED ISSUE #2458: [Feature Request] ability to define aliases for stats
it would be nice to be able to merge the stats for commands that have aliases. Eg: i have `vim` aliased to `v`; rather than having separate entries for both, I'd like to optionally have them merged as a single entry

erics118:
add to config like this:

```toml
[stats.command_aliases]
"g" = "git"
"dune te" = "dune runtest"
"dune test" = "dune runtest"
```

## Checks

closes #2458


erics118:
@greptileai

erics118:
nix fails due to an old version of nixpkgs not using a custom UA to prevent ratelimiting from crates.io. running `nix flake update nixpkgs` fixes the issue.

markovejnovic:
Hey! I see the desire, but I don't think I like this implementation. If we add this feature to the config, we expect users to manually add their aliases to the config.

I'd rather query the relevant shell for aliases. The next release of atuin will store information on the shell, so we can ask bash "hey what aliases do you currently have".

Unfortunately, even the way I present it here it's not so easy to implement. Aliases change all the time -- what we should rather do is have a mechanism to track the aliases as they change, and hold this state as time progresses.

I wonder what @ellie and @BinaryMuse think.

markovejnovic:
Talked with @ellie briefly over Slack and I think we both agree that the best path forwards is to query the user's `alias`es at runtime.

We have an implementation already which has some rudimentary support for this, but it's wrong, for the most part. I am working on this and should hopefully have a PR up soon~ish.

erics118:
A concern for this is that some things are not actually aliases, or should not be considered as aliases:
- I have `ls` aliased to `eza --icons auto --color auto -F always` but I would prefer if history (and atuin) displayed it as `ls`
- `dune te`, `dune tes`, `dune test`, `dune runtest` all do the same thing, and this is an internal mechanism provided by `dune` that is not an alias exposed to the shell
- `zmv` is aliased to `noglob zmv`, but this should always be shown as `zmv` without ever displaying the `noglob`

Maybe this could complement your current alias implementation, by providing specific overrides?

markovejnovic:
I'm not sure I like the specific overrides just yet. This causes the user to have to spend time managing their aliases when a part of that information is available at runtime.

I hear what you're saying, though. My gut feeling is that there's some good middle ground. Give me a day or so to think about the user experience we could build out here.

PR #3509, opened 2026-05-29
chore: update flake.nix

erics118:
- switches from `flake-utils` to the more modern `flake-parts`
- update `flake.lock`
- add `rust-analyzer`
- removes explicitly including `clippy` and `rustfmt` becase they are included with the toolchain
- adds `nixfmt` as formatter for the flake itself

## Checks


erics118:
nix CI will fail until this PR gets added to unstable (probs a few days): https://github.com/NixOS/nixpkgs/pull/525067

phanirithvij:
Please don't remove flake-compat, flakes are not infact "pretty much the standard" yet for a reason.

erics118:
Added back the default.nix

markovejnovic:
Would you mind sharing the motivation with us a little bit? I'm not a `nix` user (maybe I'll become one) so I'd love to know more as to why we should merge this! Thanks!

erics118:
This is a mostly chore pr, but it provides some nice changes. Mainly:

we switch from `flake-utils` to `flake-parts`, as `flake-utils` has some big downsides. [this blog](https://ayats.org/blog/no-flake-utils) explains it, but its quite technical, but tldr it lacks type safety and you could easily write broken things.

`rust-analyzer` needs to be added to the flake so that we can actually have the lsp in the dev environment. (b/c with nix, these tools are scoped to per-project rather than globally installed or installed via cargo)

adds `nixfmt` which is just a nice utility to format the nix files, but not mandatory

markovejnovic:
Got it. Thank you. Let me add `nixfmt` as part of our CI and make one more review and then we can merge!

PR #3514, opened 2026-06-04
feature: Create database integration tests for atuin-server

oxo42:
## Why?
I am trying to write a mysql database driver for atuin-server, and after doing mostly copy-pasta on the sql queries, I need to get them working properly.  I found it was taking a long time to spin up a server and go through and try to exercise all the functions. So enter integration tests for a database.

By default this one test will use the `sqlite::memory:` DB URI and will go through a number of things that I think work most of the sqlite crate.  You can also set the DB URI with an environment variable, e.g.

```shell
ATUIN_TEST_DB_URI=postgres://postgres:pg@localhost/atuin_test_ cargo test
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.47s
     Running tests/db_story.rs (/Users/joxley/projects/atuin/target/debug/deps/db_story-035ff5d9f7ad8b5e)

running 1 test
test test_full_db_story ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.37s
```

## Questions to reviewer
1. Should I care about history?
2. Is it worth running these in an integration test at build time? If so, I can do another PR that adds a Github workflow
3. Right now there is a single "story" test.  This is GoodEnough™️ for me to build the MySQL driver. Do you want me to make this better and if so, how?

## Checks


ellie (approved):
This looks good to me, thank you!

> Should I care about history?

Not sure what you mean here - the history api? It has been 100% replaced by the store api, and only very old clients still use it. For atuin hub, we've dropped support. Feel free to ignore

 > Is it worth running these in an integration test at build time? If so, I can do another PR that adds a Github workflow

A github workflow would be great!

> Right now there is a single "story" test. This is GoodEnough™️ for me to build the MySQL driver. Do you want me to make this better and if so, how?

Good enough for me. Thanks for the work!

PR #3521, opened 2026-06-08
Fix/zsh count multiline

### LINKED ISSUE #3008: [Bug]: Import/ UI Count. Numbers mismatch

I expect to see ALL my zsh created history entries to be imported and visible in the atuin search.

### What happened?

I can see atuin imports all entries, all 6064.
But when i fire up atuin it only shows a smaller number: 5026.

Image

When i search for the very first entry, it actually brings it up as a result.
that is great.
So just wondering about that number mismatch?

Does anyone have noticed this, too?

ps: Also when i scroll up.. i can only scroll till 6 months back in time. Is that intentional?

### Atuin doctor output

```yaml
{
  "atuin": {
    "version": "18.10.0",
    "commit": "",
    "sync": null,
    "sqlite_version": "3.46.0"
  },
  "shell": {
    "name": "zsh",
    "default": "zsh",
    "plugins": [
      "atuin"
    ],
    "preexec": "built-in"
  },
  "system": {
    "os": "Darwin",
    "arch": "arm64",
    "version": "15.6.1",
    "disks": [
      {
        "name": "Macintosh HD",
        "filesystem": "apfs"
      },
      {
        "name": "Macintosh HD",
        "filesystem": "apfs"
      },
      {
        "name": "patze_BAKKUP",
        "filesystem": "apfs"
      }
    ]
  }
}
```

### Code of Conduct

Dilyxs:
### Problem
when importing zsh_history, if there are multi-line commands, the progress bar displays an incorrect count. This happens because entries() counts physical lines via count_lines and not the commands. 

### Fix  
.entries() now counts joined commands instead of lines.Same complexity,  no extra traversal (it already walked the whole file)

### Test plan
test_parse_file updated and passing; new test (test_count_lines_multiline  does this check more explicitly)

Closes #3008 

## Checks

### DIFF
--- a/crates/atuin-client/src/import/zsh.rs
+++ b/crates/atuin-client/src/import/zsh.rs
@@ -198,6 +198,19 @@ mod test {
     }
 
     #[tokio::test]
+    async fn test_count_entries_multilines() {
+        let bytes = r"ls
+    pwd
+    date
+    export ATUIN_RECORD_STORE_PATH=/tmp/atuin_records.db # path to primary record store \
+    export ATUIN_META__DB_PATH=/tmp/atuin_meta.db        # path to meta database
+    "
+        .as_bytes()
+        .to_owned();
+        let mut zsh = Zsh { bytes };
+        assert_eq!(zsh.entries().await.unwrap(), 4);
+    }
+    #[tokio::test]
     async fn test_parse_file() {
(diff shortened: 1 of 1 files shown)


Dilyxs:
fix already implemented

PR #3523, opened 2026-06-10
Review agent history filter mode

dhruv-anand-aintech:
## Summary
- adds an agent history filter mode and config/docs wiring
- carries the filter through daemon, DB, skim, and shell history paths
- fixes agent history filter paths in the client database layer

## Validation
- git diff --check origin/main...HEAD

## Not run
- cargo test / builds, per this staged review pass constraint to avoid builds/deploys

## Notes
- Local branch is currently 6 commits behind upstream main after fetch; leaving rebase out of this pass.


dhruv-anand-aintech:
Closing because this branch is intended to live only in my fork, not as an upstream PR.

PR #3526, opened 2026-06-13
feat: record shell per history entry

### LINKED ISSUE #608: Record the shell used for each command

sridhar-3009:
Closes #608.

## Changes

Adds a nullable `shell` column to history so each command carries the shell it was run in.

### Database (`atuin-client`)
- New migration `20260604000000_history_shell.sql`: `ALTER TABLE history ADD COLUMN shell TEXT`
- `history.rs`: `shell: Option<String>` field on `History`; `new()` wires it through
- `history/builder.rs`: `shell` in all four builders (`from_db`, `from_db_with_prefix`, `new`, `import`)
- `database.rs`: `shell` column in `INSERT` / `UPDATE` / `query_history` `SELECT`

### CLI (`atuin`)
- `atuin history start --shell <SHELL>` flag added; threaded through `handle_start` → `apply_start_metadata` → `History::new`

## Example

```sh
atuin history start --shell zsh "git status"
```

The stored entry will have `shell = "zsh"`. Legacy entries without the column default to `NULL`.

### DIFF
--- a/crates/atuin-ai/src/history_format.rs
+++ b/crates/atuin-ai/src/history_format.rs
@@ -98,6 +98,7 @@ mod tests {
             hostname: String::new(),
             author: String::new(),
             intent: None,
+            shell: None,
             deleted_at: None,
         }
     }
--- a/crates/atuin-client/migrations/20260604000000_history_shell.sql
+++ b/crates/atuin-client/migrations/20260604000000_history_shell.sql
@@ -1,0 +1,1 @@
+ALTER TABLE history ADD COLUMN shell TEXT;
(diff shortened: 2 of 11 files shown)


ellie:
I'm really sorry but this was so far from being ready and I am not in a place to coach someone (or their agent) through completing the work.

PR #3529, opened 2026-06-16
fix(pty-proxy): forward terminal pixel size to the child pty

xav-ie:
## What

Forward the controlling terminal's pixel dimensions to the child pty in
`atuin pty-proxy`.

The pty was created (and resized on `SIGWINCH`) with `pixel_width: 0` /
`pixel_height: 0`, because `crossterm::terminal::size()` only reports
columns/rows. With zero pixel dimensions, programs inside the proxy that query
`TIOCGWINSZ` (e.g. `image.nvim`) fall back to an assumed ~8×16 cell, so inline
images render far too small on HiDPI terminals.

This switches to `terminal::window_size()`, which includes `ws_xpixel` /
`ws_ypixel`, and forwards those to the child pty both at startup and on every
resize. It falls back to cols/rows (pixels `0`) on terminals that don't report
pixel size.

## Why

Zero pixel dimensions break inline image rendering for terminal programs that
rely on the kernel-reported cell pixel size.

## Testing

- `cargo build` / `cargo test` / `cargo clippy` for `atuin-pty-proxy` (all pass).
- Manually verified inline images (image.nvim) render at the correct size inside
  `atuin pty-proxy` on a HiDPI terminal.

## Checks

### DIFF
--- a/crates/atuin-pty-proxy/src/runtime.rs
+++ b/crates/atuin-pty-proxy/src/runtime.rs
@@ -19,16 +19,31 @@ pub(crate) fn main(options: RuntimeOptions) {
     }
 }
 
+/// Terminal size including pixel dimensions: `(cols, rows, pixel_width, pixel_height)`.
+///
+/// We forward the controlling terminal's pixel width/height to the child pty so
+/// programs inside (e.g. image.nvim) can derive the cell size in pixels via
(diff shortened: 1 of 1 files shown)


markovejnovic:
Thank you for your PR!

PR #3552, opened 2026-06-28
fix: Double input on arrow keys in AI setup prompt on Windows

lxl66566:
## Checks

## Description

When pressing the Up/Down arrow keys in the Atuin AI setup prompt (`?`) on Windows (e.g., Git Bash / msys2), the cursor moves twice per physical keystroke.

This prevents the user from selecting the `Disable ? Keybind` option, which is located in the middle of the list.

```sh
~

  Atuin AI is not yet configured.

  > Enable Atuin AI
    Disable ? Keybind
    Cancel
```

## To Reproduce

1. Open Git Bash / msys2 on Windows.
2. Press `?` to trigger the AI setup prompt.
3. Press the Down arrow key once.
4. The cursor jumps two lines instead of one.

## Fix

This is a [known `crossterm` behavior on Windows](https://github.com/crossterm-rs/crossterm/issues/797). Since `crossterm` 0.26, it emits both `KeyEventKind::Press` and `KeyEventKind::Release` events on Windows. 

In `prompt_ai_setup`, the event loop ignores `KeyEventKind`. Both the press and release events trigger the cursor movement.

So the fix is: Filter for `KeyEventKind::Press` in the event loop.

### DIFF
--- a/crates/atuin-ai/src/commands/inline.rs
+++ b/crates/atuin-ai/src/commands/inline.rs
@@ -426,7 +426,7 @@ enum SetupChoice {
 fn prompt_ai_setup() -> Result<SetupChoice> {
     use crossterm::{
         cursor,
-        event::{self, Event, KeyCode},
+        event::{self, Event, KeyCode, KeyEventKind},
         terminal,
     };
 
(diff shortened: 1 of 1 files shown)


ellie (approved):
thank you! 🙏

PR #3553, opened 2026-06-28
perf(search): scan history by recency until N unique

jbellis:
Inverts the ctrl-r search from "materialize all unique commands, then pick N most recent" to "scan commands by recency until we have N unique ones". Basically the entire change is the SQL query, the rest is scaffolding so we can put filters in the new subquery where they belong.

Measured on a 557k-row history (≈369k distinct commands), empty-query search goes from *~0.78s warm / ~18s cold* to *~0.01s warm / ~0.3s cold*. 

Related PRs I saw:
1. https://github.com/atuinsh/atuin/pull/3549: directed at the same problem but breaks semantics as noted by the bot reviewer
2. https://github.com/atuinsh/atuin/pull/2698: more code churn, looks like main idea is doing the slow path page-at-a-time, which is better than all-of-history but still worse in principle; corner cases are still bad

Both of the above also split the logic between a SQL query and Rust-side filtering; this one keeps the logic in SQL so the query engine can do its job efficiently.

## Checks

### DIFF
--- a/crates/atuin-client/src/database.rs
+++ b/crates/atuin-client/src/database.rs
@@ -492,25 +492,11 @@ impl Database for Sqlite {
         query: &str,
         filter_options: OptFilters,
     ) -> Result<Vec<History>> {
+        // Build the inner query holding all of the user's filters (filter mode,
+        // fuzzy/regex command matches, exit/cwd/date filters, author, deleted_at).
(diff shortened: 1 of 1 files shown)


ellie (approved):
looks good, thank you!! nice and neat 🙏

PR #3567, opened 2026-07-06
Exclude AI agent commands from zsh-autosuggestions

brianhelba:
Without this, commands from AI agents will pollute the suggestions passed to `zsh-autosuggestions` (which is used by live users).

This touches the same code as #2903, but this PR should be a less controversial simple bugfix.

## Checks


ellie (approved):
ah yes, thank you!!

PR #3569, opened 2026-07-07
fix(sync): skip records that fail to decrypt or decode instead of failing the whole store

ellie:
Previously, we chose to fail loudly when there was a key mismatch in a users store. However, in the past, we did not check the key at login, so users could easily end up in a state where they inserted an invalid key, and then broke their store. 

This PR ensures that even with a partially invalid store, the user can still sync their data. 

This does not mean that users will be able to introduce new invalid keys, and simply makes things better for older users (pre key-checking). We still need to figure out a more straightforward way for a user to repair their entire record store if they have somehow mixed up keys, other than the `purge` + `push/pull --force` steps.

## Checks

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -445,6 +445,7 @@ dependencies = [
  "rmp",
  "serde",
  "tokio",
+ "tracing",
 ]
 
 [[package]]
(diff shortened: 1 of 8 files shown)


ellie:
@greptile-apps plz sir may i have one review 🙏

BinaryMuse (approved):
This seems reasonable. Do we have a good way to surface to the user that something went wrong other than them looking in the logs?

davidolrik:
> We still need to figure out a more straightforward way for a user to repair their entire record store if they have somehow mixed up keys, other than the `purge` + `push/pull --force` steps.

@ellie: I have a solution for this in #3429, which adds `atuin store repair`. - It has just been rebased on top of  `main`

PR #3572, opened 2026-07-07
feat(doctor): add whether daemon is enabled to `doctor` output

taylordotfish:
Add whether the daemon is enabled to the output of `atuin doctor`.


ellie (approved):
yaya ty! :shipit:

PR #3573, opened 2026-07-07
chore: adopt derive_more to reduce boilerplate

markovejnovic:
This PR adds [derive_more](https://docs.rs/derive_more/latest/derive_more/) with the explicit aim of removing boilerplate from the code.

There is no obligation to merge this PR -- it's more-so just a suggestion. AI-generated, hand-reviewed.

## Checks


crates/atuin-common/src/shell.rs
```diff
@@ -4,37 +4,26 @@ use serde::Serialize;
...
-            Shell::Xonsh => "xonsh",
-            Shell::Sh => "sh",
-            Shell::Powershell => "powershell",
-
-            Shell::Unknown => "unknown",
-        };
-
-        write!(f, "{shell}")
-    }
-}
```
markovejnovic, line 36: This was what inspired me for `derive_more`

crates/atuin-ai/src/fsm/tools.rs
```diff
@@ -62,7 +62,7 @@ pub(crate) struct TrackedTool {
 impl TrackedTool {
     /// Whether this tool has reached a terminal state.
     pub fn is_resolved(&self) -> bool {
-        matches!(self.state, ToolState::Completed | ToolState::Denied)
+        self.state.is_completed() || self.state.is_denied()
```
taylordotfish: I would hope this gets optimized the same as a single match, but I'm not so sure

crates/atuin-daemon/src/daemon.rs
```diff
@@ -71,8 +71,10 @@ pub struct DaemonState {
 /// // Access database
 /// let history = handle.history_db().load(id).await?;
 /// ```
-#[derive(Clone)]
+#[derive(Clone, derive_more::Debug)]
+#[debug("DaemonHandle {{ .. }}")]
```
taylordotfish: Are the `fmt::Formatter<'_>` methods guaranteed to produce the same output across Rust versions? If not, replacing the calls with a static string could cause the output to get out of sync.

crates/atuin-ai/src/permissions/check.rs
```diff
@@ -2,31 +2,23 @@ use eyre::Result;
 
 use crate::{permissions::file::RuleFile, tools::PermissibleToolCall};
 
+#[derive(derive_more::Constructor)]
```
taylordotfish: I'm a little conflicted on the constructor derivations; if I saw a call to `PermissionRequest::new` and was trying to find the definition I think it would take me a while to figure out what was going on. It would be nice if at the very least the word `new` appeared somewhere.

crates/atuin-common/src/shell.rs
taylordotfish: Yes, this is definitely an improvement

crates/atuin-dotfiles/src/store.rs
```diff
@@ -123,23 +123,14 @@ impl AliasRecord {
...
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, derive_more::Constructor)]
 pub struct AliasStore {
     pub store: SqliteStore,
     pub host_id: HostId,
     pub encryption_key: [u8; 32],
 }
 
 impl AliasStore {
-    // will want to init the actual kv store when that is done
```
taylordotfish, line 134: Let's not lose this comment. Actually, this comment seems to suggest we may want to add something more to the function body so it may be best to leave this one alone

crates/atuin-ai/src/fsm/mod.rs
```diff
@@ -541,7 +541,7 @@ impl AgentFsm {
                 let mut effects = Vec::new();
 
                 // Abort stream if still active
-                if !matches!(stream, StreamPhase::Done) {
+                if !stream.is_done() {
```
taylordotfish: I'm not sold on the `IsVariant` derivations either. If I saw this code, I'd look for the function definition to see how `is_done` is being calculated (is it a complex boolean expression? could it even potentially make a system call?), and then struggle a bit before realizing what was going on. With the call to `matches!` it's obvious.

taylordotfish (commented):
The cases where `derive_more` is being used to derive a trait like `From` or `Display` generally look good to me, but some of the `Constructor` and `IsVariant` uses are a little too magic for me and I could see them causing confusion.

crates/atuin-ai/src/fsm/mod.rs
markovejnovic: fair. i use `derive_more` quite frequently so its part of my `grep` mental head cannon, going to revert the `IsVariant`

crates/atuin-ai/src/permissions/check.rs
markovejnovic: same as my comment with `IsVariant` =)

crates/atuin-ai/src/fsm/tools.rs
markovejnovic: I'd guess it's the same, though you nerd sniped me, going to research while claude tears it out from the PR

markovejnovic:
```rust
use derive_more::IsVariant;

#[derive(IsVariant)]
pub enum ToolState {
    Pending,
    Running,
    Completed,
    Denied,
}

#[unsafe(no_mangle)]
#[inline(never)]
pub fn check_matches(state: &ToolState) -> bool {
    matches!(state, ToolState::Completed | ToolState::Denied)
}

#[unsafe(no_mangle)]
#[inline(never)]
pub fn check_is_variant(state: &ToolState) -> bool {
    state.is_completed() || state.is_denied()
}

fn main() {
    let s = ToolState::Completed;
    println!("{} {}", check_matches(&s), check_is_variant(&s));
}
```

compiles to 

```asm
Disassembly of section __TEXT,__text:

0000000100000934 <_check_is_variant>:
0000000100000934 <_check_matches>:
100000934: 39400008     ldrb    w8, [x0]
100000938: 7100051f     cmp     w8, #0x1
10000093c: 1a9f97e0     cset    w0, hi
100000940: d65f03c0     ret
```

so it looks like it's the same =)

crates/atuin-daemon/src/daemon.rs
markovejnovic: No. I think you're correct in your assumption of it not being equivalent. Although I think the functional impact is probably the same either way, might be safer to leave it as it was before

crates/atuin-client/src/settings.rs
```diff
@@ -29,6 +29,8 @@ pub(crate) mod meta;
 mod scripts;
 pub mod watcher;
 
+#[derive(derive_more::AsRef)]
+#[as_ref(forward)]
```
taylordotfish: If I'm understanding the docs right, `as_ref(forward)` generates a blanket `AsRef<T>` implementation where `String: AsRef<T>`. That could potentially conflict with `AsRef` implementations added later or lead to worse error messages. It looks like we could do `as_ref(str)` instead though to only generate the `AsRef<str>` impl?

crates/atuin-server-database/src/lib.rs
```diff
@@ -19,30 +15,15 @@ use serde::{Deserialize, Serialize};
...
 
-#[derive(Debug)]
+#[derive(Debug, derive_more::Display, derive_more::Error, derive_more::From)]
+#[display("{self:?}")]
 pub enum DbError {
+    #[from(skip)]
     NotFound,
+    #[from(time::error::ComponentRange, time::error::Error)]
     Other(eyre::Report),
 }
```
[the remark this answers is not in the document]
taylordotfish: I don't see it being a problem that `<DbError as Error>::source` can now return `Some` but would like someone more experienced to weigh in on it. @ellie @BinaryMuse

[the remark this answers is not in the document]
ellie: no, I don't think this is an issue! thanks for flagging though

markovejnovic:
ty for your pr @taylordotfish

taylordotfish (approved):
looks good, thanks!

PR #3577, opened 2026-07-07
perf: add synchronous(Normal) + optimize_on_close to record store SQLite

markovejnovic:
This PR adds two commits:

- The first commit adds benchmarking support to `crates/atuin-client`.
- The second commit adds a tiny performance optimization to `record/sqlite_store.rs` which requests `sqlite` `pragma optimize` on close as well as prefers the WAL rather than immediate `fsync`.

Hand-rolled PR (I hated claude's output).

## Benchmarking

I used [divan](https://github.com/nvzqz/divan) since this was the existing precedent. To run benchmarks:

```bash
cargo bench -p atuin-client
```

I think this is a little bit naive, and ideally we'd have rich benchmarking in CI, but I think that could be considered a future improvement.

If you think it's worth dropping the benchmarking addition from this PR (in lieu of a future, better, CI-integrated addition), I'm totally down to do that =)

Had the directory structure mirror the source tree

## Optimization

There is a semantic change that occurred here -- in case the user's machine loses power, we risk losing the WAL on their machine, but:
  - for the remote case, it's fine, since we'll pick up on those changes in the next sync with the remote. 
  - the local case _does_ risk losing data, but this seems to be in-line with the history's database.
  
### Results

Tested on an M4 mac.

At `171f7643d4337165ab96d15c010fb6638a3dfcb9` (before):

```
record_store         fastest       │ slowest       │ median        │ mean          │ samples │ iters
╰─ record                          │               │               │               │         │
   ╰─ sqlite_store                 │               │               │               │         │
      ╰─ push_batch                │               │               │               │         │
         ├─ 1        138.6 µs      │ 5.801 ms      │ 247.8 µs      │ 280.1 µs      │ 2504    │ 2504
         ├─ 10       177.7 µs      │ 5.204 ms      │ 200.4 µs      │ 215.3 µs      │ 3842    │ 3842
         ╰─ 100      617.7 µs      │ 8.771 ms      │ 718.1 µs      │ 779 µs        │ 2430    │ 2430

```

At `4f30c909f459676857e564b338cfd2622794714f` (after):

```
record_store         fastest       │ slowest       │ median        │ mean          │ samples │ iters
╰─ record                          │               │               │               │         │
   ╰─ sqlite_store                 │               │               │               │         │
      ╰─ push_batch                │               │               │               │         │
         ├─ 1        98.83 µs      │ 5.41 ms       │ 183.7 µs      │ 189.3 µs      │ 4770    │ 4770
         ├─ 10       151.6 µs      │ 6.139 ms      │ 170.2 µs      │ 186 µs        │ 4736    │ 4736
         ╰─ 100      578.1 µs      │ 8.483 ms      │ 661 µs        │ 704.7 µs      │ 2890    │ 2890
```

I had claude generate a comparison table:

```
  ┌────────────────┬─────────────────┬────────────────┬────────┐
  │   Benchmark    │ Before (median) │ After (median) │ Change │
  ├────────────────┼─────────────────┼────────────────┼────────┤
  │ push_batch/1   │ 247.8 µs        │ 183.7 µs       │ -26%   │
  ├────────────────┼─────────────────┼────────────────┼────────┤
  │ push_batch/10  │ 200.4 µs        │ 170.2 µs       │ -15%   │
  ├────────────────┼─────────────────┼────────────────┼────────┤
  │ push_batch/100 │ 718.1 µs        │ 661.0 µs       │ -8%    │
  └────────────────┴─────────────────┴────────────────┴────────┘
```

Worth noting that my machine appears to be very noisy. I think less noisy benchmarks require we have a hosted bare-metal machine for this, specifically designed for benchmarking, but I think this is sufficient proof that the flags I've added are valuable.

## Checks


ellie (approved):
looks good, thank you!

I think definitely a nice starting place for benchmarks for us, but in the long term we need a stable CI run so that we can track things properly - as you suggested!

PR #3583, opened 2026-07-08
docs: make commented-out lines in config.toml match real defaults

taylordotfish:
Make the commented-out lines in the example config.toml match the real default values. This results in less surprising behavior and can help users make incremental changes; for example, a user who wants to adjust the inline UI height by only a few lines will have an easier time if they can see that the default height is 40.


ellie (approved):
🙏

PR #3584, opened 2026-07-08
perf: increase sync page size from 100 to 1000 records

markovejnovic:
## Summary

- Add divan benchmark harness with `push_batch` benchmark covering batch sizes 1, 10, 100, and 1000
- Increase record sync page size from 100 to 1000 (`record/sync.rs`), reducing HTTP round-trips per sync by 10×
- Server default `page_size` is already 1100 (`atuin-server/src/settings.rs:79`), so 1000 is well within the accepted limit

## Test plan

- [ ] `cargo check -p atuin-client` compiles clean
- [ ] `cargo test -p atuin-client` passes (sync tests don't exercise live sync)
- [ ] `cargo bench -p atuin-client --bench benchmarks -- push_batch` runs all 4 batch sizes
- [ ] Verify server accepts 1000-record batches in a real sync


markovejnovic:
going to rebuild this from scratch

PR #3586, opened 2026-07-08
chore(warnings): fix compile warnings with latest dependencies

taylordotfish:
`time::format_description::parse` is deprecated in the latest version of `time` (used when the lockfile is ignored, as with `cargo install atuin`); replace with the recommended alternative. This change is compatible with the older version of `time` in the lockfile.


ellie (approved):
:shipit:

PR #3587, opened 2026-07-08
fix(nu): suppress error when ATUIN_HISTORY_ID is missing in pre_prompt hook

### LINKED ISSUE #3520: [Bug]: Environment variable 'ATUIN_HISTORY_ID' not found while using nushell overlay

No errors upon using the nushell overlay.

### What happened?

Nushell doesn't support dynamic path overlays, so for things like python venvs I use a workaround like `nu -e $"overlay use ($path)/bin/activate.nu"`.
This subshell doesn't capture ATUIN_HISTORY_ID, which leads to the error in `hide-env ATUIN_HISTORY_ID` (line 48 on 18.16.1 and [line 72](https://github.com/atuinsh/atuin/blob/main/crates/atuin/src/shell/atuin.nu#L72) on main).

The [-i flag](https://github.com/atuinsh/atuin/blob/main/crates/atuin/src/shell/atuin.nu#L15) is used for an earlier hide-env call to suppress errors, I wonder if it's safe to do the same here?

### Atuin doctor output

```yaml
Atuin Doctor
Checking for diagnostics


{
  "atuin": {
    "version": "18.16.1",
    "commit": "NO_GIT",
    "sync": null,
    "sqlite_version": "3.53.2"
  },
  "shell": {
    "name": "nu",
    "default": "nu",
    "plugins": [
      "atuin"
    ],
    "preexec": "built-in"
  },
  "system": {
    "os": "Arch Linux",
    "arch": "x86_64",
    "version": "rolling",
    "disks": [
      {
        "name": "/dev/nvme0n1p5",
        "filesystem": "ext4"
      },
      {
        "name": "/dev/nvme0n1p3",
        "filesystem": "vfat"
      },
      {
        "name": "/dev/nvme0n1p2",
        "filesystem": "ext4"
      },
      {
        "name": "/dev/sda1",
        "filesystem": "ext4"
      },
      {
        "name": "/dev/sda2",
        "filesystem": "ext4"
      }
    ]
  }
}
Error: nu::shell::env_variable_not_found

  × Environment variable 'ATUIN_HISTORY_ID' not found
    ╭─[/home/aman/.local/share/atuin/init.nu:48:14]
 47 │     }
 48 │     hide-env ATUIN_HISTORY_ID
    ·              ────────┬───────
    ·                      ╰── environment variable not found
 49 │ }
    ╰────
```

### Code of Conduct

AndrianBalanescu:
## What

Fixes a crash in the nushell integration when `ATUIN_HISTORY_ID` is not set in the environment.

The `_atuin_pre_prompt` hook calls `hide-env ATUIN_HISTORY_ID` at the end of every prompt. If the env var doesn't exist (e.g. when using nushell overlays/subshells that don't capture it, like `nu -e "overlay use ..."`), this throws an unhandled `env_variable_not_found` error.

## How

The fix mirrors what the script already does at line 15 — use the `-i` (ignore-if-missing) flag:

```nu
// Before
hide-env ATUIN_HISTORY_ID

// After
hide-env -i ATUIN_HISTORY_ID
```

This is the exact same pattern already used in the initialization section (line 15: `hide-env -i ATUIN_HISTORY_ID`), so it's consistent with the existing code. The reporter of #3520 also suggested this approach. (◕‿◕)

One file, one line changed. No behavior change for the normal case (var exists → it gets hidden). Only difference: missing var no longer crashes.

Closes #3520.

### DIFF
--- a/crates/atuin/src/shell/atuin.nu
+++ b/crates/atuin/src/shell/atuin.nu
@@ -69,7 +69,7 @@ let _atuin_pre_prompt = {||
         }
 
     }
-    hide-env ATUIN_HISTORY_ID
+    hide-env -i ATUIN_HISTORY_ID
 }
 
 def _atuin_search_cmd [...flags: string] {


ellie (approved):
thank you!

PR #3590, opened 2026-07-08
fix: strip NUL bytes from history commands (closes #3589)

### LINKED ISSUE #3589: [Bug]: NUL bytes in history

(As below.)

### What happened?

Thoughts?

### Atuin doctor output

```yaml
{
  "atuin": {
    "version": "18.16.1",
    "sync": {
      "records": true,
      "auto_sync": true,
    },
    "sqlite_version": "3.46.0",
    "daemon_enabled": false
  },
  "shell": {
    "name": "bash",
    "default": "bash",
    "plugins": [
      "atuin",
      "bash-preexec"
    ],
    "preexec": "bash-preexec"
  },
  "system": {
    "os": "Ubuntu",
    "arch": "x86_64",
    "version": "26.04",
    "disks": [
      {
        "name": "/dev/sdh",
        "filesystem": "ext4"
      },
      {
        "name": "drivers",
        "filesystem": "9p"
      },
      {
        "name": "none",
        "filesystem": "overlay"
      },
      {
        "name": "/dev/sdh",
        "filesystem": "ext4"
      },
      {
        "name": "none",
        "filesystem": "overlay"
      },
      {
        "name": "none",
        "filesystem": "overlay"
      },
      {
        "name": "none",
        "filesystem": "overlay"
      },
      {
        "name": "/dev/sdg2",
        "filesystem": "ext4"
      },
      {
        "name": "/dev/md0",
        "filesystem": "ext4"
      },
      {
        "name": "/dev/sdf1",
        "filesystem": "ext4"
      },
      {
        "name": "C:\\134",
        "filesystem": "9p"
      },
      {
        "name": "D:\\134",
        "filesystem": "9p"
      },
      {
        "name": "E:\\134",
        "filesystem": "9p"
      },
      {
        "name": "G:\\134",
        "filesystem": "9p"
      }
    ]
  }
}
```

### Code of Conduct

AndrianBalanescu:
## Summary

External tools — in particular AI agents that mis-escape values — can write raw NUL bytes (`\0`) into the shell history. For example, writing `start=\0` instead of `start=0` lands a literal NUL byte in the command string.

NUL bytes break everything downstream that expects valid UTF-8 text: `atuin search` output needs to be piped through `tr -d \x27\\000\x27` or it corrupts the consuming process, messagepack serialisation chokes, and even simple terminal output is garbled.

## Fix

Rather than reject the whole entry (which loses an otherwise useful command), this PR strips NUL bytes at the single point where **all** new history entries are constructed — `History::new`. This covers every entry path (import, capture, daemon capture) with one change.

The new `sanitize_command` helper has a fast-path: it checks `command.contains(\x27\\0\x27)` first and returns the string unchanged when clean, so there is zero allocation overhead for the overwhelmingly common case of a normal command.

## Testing

Added two unit tests in `history::tests`:

- `test_sanitize_strips_nul_bytes` — verifies `start=\0 foo\0bar` becomes `start= foobar`
- `test_sanitize_preserves_clean_command` — verifies a normal command is untouched

All 12 existing `history::` tests continue to pass:

```
running 12 tests
test history::store::tests::test_serialize_deserialize_delete ... ok
test history::store::tests::test_serialize_deserialize_create ... ok
test history::tests::disable_secrets ... ok
test history::tests::known_agents_include_pi ... ok
test history::tests::test_sanitize_preserves_clean_command ... ok
test history::tests::test_sanitize_strips_nul_bytes ... ok
test history::tests::test_serialize_deserialize ... ok
test history::tests::test_serialize_deserialize_deleted ... ok
test history::tests::test_serialize_deserialize_version ... ok
test history::tests::test_serialize_deserialize_with_author_and_intent ... ok
test history::store::tests::test_history_skips_corrupt_records ... ok
test history::tests::privacy_test ... ok
test result: ok. 12 passed; 0 failed; 0 ignored; 0 measured; 82 filtered out
```

Closes #3589.

### DIFF
--- a/crates/atuin-client/src/history.rs
+++ b/crates/atuin-client/src/history.rs
@@ -163,6 +163,21 @@ impl History {
             .map_or_else(|| hostname.to_owned(), |(_, user)| user.to_owned())
     }
 
+    /// Strip NUL bytes from a command string.
+    ///
+    /// NUL bytes can enter the history when an external tool writes a raw NUL
+    /// byte into the command. They break downstream consumers that expect
+    /// valid UTF-8 text. Rather than reject the whole entry, we silently
+    /// strip the offending bytes. See atuinsh/atuin#3589.
+    fn sanitize_command(command: String) -> String {
+        if !command.contains('\0') {
+            // fast path: most commands are clean
+            command
+        } else {
+            command.replace('\0', "")
+        }
+    }
+
     #[allow(clippy::too_many_arguments)]
     fn new(
         timestamp: OffsetDateTime,
@@ -191,7 +206,7 @@ impl History {
         Self {
             id: uuid_v7().as_simple().to_string().into(),
             timestamp,
-            command,
+            command: Self::sanitize_command(command),
             cwd,
             exit,
             duration,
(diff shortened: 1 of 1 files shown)


markovejnovic:
This should be superseded by https://github.com/atuinsh/atuin/pull/3660

PR #3591, opened 2026-07-08
feat: Add status bar with model and usage information

BinaryMuse:
This PR adds a status bar that includes the currently selected model on the left and the user's overall AI usage on the right, if it's over a certain threshold, as well as a short string (either in whole number days, hours, or minutes) describing when the limit will reset.

image

The threshold is currently set to 50%, and it changes color as usage gets higher.

image

image

The percentage is based on either input or output usage, whichever is higher; e.g., if a user has 40% input usage and 80% output usage, the widget will show 80%.

Usage is cached for 60 seconds in the AI database, and if stale, is updated in the background when the TUI launches. Each turn ends in an event that sends updated usage information, so the widget will remain correct over the course of a session.

### DIFF
--- a/crates/atuin-ai/src/session.rs
+++ b/crates/atuin-ai/src/session.rs
@@ -152,13 +152,7 @@ impl SessionService for LocalSessionService {
     }
 
     async fn get_cached_usage(&self, user_key: &str) -> Result<Option<CachedUsageSnapshot>> {
-        match self.store.get_usage(user_key).await? {
-            Some((json, written_at)) => Ok(Some(CachedUsageSnapshot {
-                snapshot: serde_json::from_str(&json)?,
-                written_at,
-            })),
-            None => return Ok(None),
-        }
+        self.store.get_usage(user_key).await
     }
 
     async fn set_cached_usage(&self, user_key: &str, snapshot: &UsageSnapshot) -> Result<()> {
--- a/crates/atuin-ai/src/session.rs
+++ b/crates/atuin-ai/src/session.rs
@@ -156,8 +156,7 @@ impl SessionService for LocalSessionService {
     }
 
     async fn set_cached_usage(&self, user_key: &str, snapshot: &UsageSnapshot) -> Result<()> {
-        let snapshot_json = serde_json::to_string(snapshot)?;
-        self.store.set_usage(user_key, &snapshot_json).await
+        self.store.set_usage(user_key, snapshot).await
     }
 }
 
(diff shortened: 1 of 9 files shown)


crates/atuin-ai/src/session.rs
```diff
@@ -139,6 +144,14 @@ impl SessionService for LocalSessionService {
     async fn set_metadata(&self, session_id: &str, key: &str, value: &str) -> Result<()> {
         self.store.set_metadata(session_id, key, value).await
     }
+
+    async fn get_cached_usage(&self, user_key: &str) -> Result<Option<(String, i64)>> {
```
taylordotfish: Is there a good reason to return a JSON string here instead of deserializing it and returning that? Then we won't have to deserialize it at every call site, if we add more call sites later.

It might also be a good idea to make this a struct with named fields instead of a tuple.

BinaryMuse: Updated in ef16e0966c012df89dd5afa916fc0ff576038db5

crates/atuin-ai/src/store.rs
```diff
@@ -331,6 +331,35 @@ impl AiSessionStore {
         .await?;
         Ok(())
     }
+
+    // ── Usage cache (server credit totals, one row per user key) ──
+
+    /// Read the cached usage snapshot for a user key. Returns the snapshot
+    /// JSON and the unix timestamp it was written at.
+    pub async fn get_usage(&self, user_key: &str) -> Result<Option<(String, i64)>> {
```
taylordotfish: The tuple return here makes me slightly nervous, but if this method is only going to be called from one place it's probably fine.

taylordotfish (dismissed):
Looks good overall, minor comment about another tuple return but I'll leave it to you to decide if it's worth changing.

crates/atuin-ai/src/store.rs
BinaryMuse: Went ahead and converted that one too

crates/atuin-ai/src/store.rs
```diff
@@ -331,11 +333,50 @@ impl AiSessionStore {
...
+
+        let snapshot = serde_json::from_str(&snapshot)?;
+        Ok(Some(CachedUsageSnapshot {
+            snapshot,
+            written_at,
+        }))
+    }
+
+    /// Write the usage snapshot for a user key (upsert).
+    pub async fn set_usage(&self, user_key: &str, snapshot_json: &str) -> Result<()> {
```
taylordotfish: Just for symmetry, if `get_usage` returns a `CachedUsageSnapshot`, I feel like `set_usage` should take a `CachedUsageSnapshot` and do the serialization here.

BinaryMuse: derp 🙈

taylordotfish (approved):

PR #3602, opened 2026-07-10
feat(tui): truncate long commands from middle to show start...end

### LINKED ISSUE #3596: [Feature] Shrink commands / show start and end of a command
I really would like to be able to show the start and end of a command when the terminal width is not enough to show the full command.

Image

How hstr is handling this:

Image

No matter how small the terminal width is it is always showing the start and end of the command:

Image

Image

I know that this can partially be achieved by adjusting the `max_preview_height`, but this only shows 1 command at a time, however the other implementation is really useful since I see multiple commands at a glance:

Image

AndrianBalanescu:
## Summary

Resolves #3596

When a command is too long to fit in the available terminal width, truncate it from the middle showing both the beginning and end of the command separated by `...` instead of cutting off at the end.

This makes it easier to identify long commands at a glance in narrow terminals, similar to how `hstr` handles command display.

## Changes

- **`history_list.rs`**: Modified `command()` render method to use middle-truncation when the command doesn't fit the available width
- **`truncate_middle()`**: Extracted the truncation logic as a pure, testable function that:
  - Returns the display string and a byte-position mapping (for search highlight alignment)
  - Handles Unicode correctly (char-based, not byte-based)
  - Returns the original string as-is when it fits or when the width is too small (< 7 chars)
- **8 unit tests** covering: no truncation when fits, exact fit, too-small width, start/end display, byte mapping preservation, Unicode, minimum width, odd remaining

## Behavior

| Before | After |
|--------|-------|
| `docker run --rm -it -v /home/user/long/path:/data ubuntu:22.04 bash -c` (cut off) | `docker run --rm -it -v /home/u...:22.04 bash -c` |

The prefix gets slightly more space than the suffix (ceiling division) since the beginning of a command is usually more identifying.

## Testing

- `cargo test --bin atuin` — 222 passed (214 existing + 8 new)
- `cargo clippy -- -D warnings` — clean
- `cargo fmt --check` — clean

### DIFF
--- a/crates/atuin/src/command/client/search/history_list.rs
+++ b/crates/atuin/src/command/client/search/history_list.rs
@@ -222,7 +222,7 @@ impl DrawState<'_> {
                 UiColumnType::Host => self.host(h, width),
                 UiColumnType::User => self.user(h, width),
                 UiColumnType::Exit => self.exit_code(h, width),
-                UiColumnType::Command => self.command(h),
+                UiColumnType::Command => self.command(h, width),
             }
         }
     }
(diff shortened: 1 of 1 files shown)


ellie (approved):
looks great, thank you!

PR #3605, opened 2026-07-10
feat: add auto shell detection to `atuin init`

### LINKED ISSUE #3604: Auto-detect shell in `atuin init`
**Is your feature request related to a problem? Please describe.**

I shouldn't have to do `atuin init zsh` (or bash, etc.) when setting up Atuin. Especially since `atuin ai init` already auto-detects the shell.

**Describe the solution you'd like**

It would be great if `atuin init` with no arguments (or `atuin init auto`) could automatically detect the current shell, just like `atuin ai init` does. The explicit shell names would still work for cases where you need them.

**Describe alternatives you've considered**

- Keep things as-is — minor inconvenience
- Document the auto-detect pattern in the install guide

**Additional context**

Atuin already has the shell detection infrastructure in place (used by `atuin ai init` and `atuin doctor`), so this would mostly be wiring it up to the init command.

vegerot:
Add `atuin init` (no arguments) and `atuin init auto` which auto-detect
the current shell using the parent process name, matching the behavior
already available in `atuin ai init`. Explicit shell names continue to
work as before.

Also update bash and zsh docs to recommend `atuin init` (auto-detect)
alongside the explicit shell forms.

Closes #3604

### DIFF
--- a/crates/atuin/src/command/client/init.rs
+++ b/crates/atuin/src/command/client/init.rs
@@ -5,6 +5,7 @@ use atuin_client::{
     record::sqlite_store::SqliteStore,
     settings::{Settings, Tmux},
 };
+use atuin_common::shell::Shell as CommonShell;
 use atuin_dotfiles::store::{AliasStore, var::VarStore};
 use clap::{Parser, ValueEnum};
 use eyre::{Result, WrapErr};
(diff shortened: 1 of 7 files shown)


vegerot:
cc @ellie @markovejnovic please review 🙂

Issue #3606, opened 2026-07-11
[Bug]: `atuin pty-proxy init` exports `SHELL=zsh` instead of the spawned shell path


hangrymuppet:
### What did you expect to happen?

When Atuin pty-proxy launches a proxied zsh session, `$SHELL` should remain an executable path for the shell that launched Atuin.

For example, on a macOS system whose terminal launches the login shell as `/bin/zsh`, the proxied session should keep:

```text
SHELL=/bin/zsh
```

This matters because many tools correctly expect `$SHELL` to refer to an executable path.

### What happened?

`atuin pty-proxy init` can launch the proxied zsh with:

```text
SHELL=zsh
```

### Atuin doctor output

```yaml
{
  "atuin": {
    "version": "18.17.0",
    "commit": "",
    "sync": {
      "auth_state": "Hub (legacy token — run 'atuin login' to upgrade)",
      "records": true,
      "auto_sync": true,
      "last_sync": "REDACTED"
    },
    "sqlite_version": "3.46.0",
    "daemon_enabled": true
  },
  "shell": {
    "name": "zsh",
    "default": "zsh",
    "plugins": [
      "atuin"
    ],
    "preexec": "built-in"
  },
  "system": {
    "os": "Darwin",
    "arch": "arm64",
    "version": "26.5.2",
    "disks": [
      {
        "name": "Macintosh HD",
        "filesystem": "apfs"
      },
      {
        "name": "Macintosh HD",
        "filesystem": "apfs"
      }
    ]
  }
}
```

### Shell setup

The shell startup runs Atuin pty-proxy early:

```zsh
if [[ "$-" == *i* ]] && command -v atuin >/dev/null 2>&1; then
  eval "$(atuin pty-proxy init)"
fi
```

The generated zsh init code currently uses `ZSH_ARGZERO` directly after
stripping a leading login-shell dash:

```zsh
_atuin_pty_proxy_zsh="${ZSH_ARGZERO:-$(command -v zsh)}"
exec atuin pty-proxy --shell "${_atuin_pty_proxy_zsh#-}"
```

In the affected proxied session:

```text
ATUIN_PTY_PROXY_ACTIVE=1
ATUIN_PTY_PROXY_SOCKET=/var/folders/.../T/atuin-pty-proxy-70454.sock
SHELL=zsh                    # <-- not an executable path
argv0=zsh
ZSH_ARGZERO=zsh
command -v zsh=/opt/homebrew/bin/zsh
[...]

PR #3607, opened 2026-07-11
fix(pty-proxy): resolve SHELL path in proxied zsh sessions

hangrymuppet:
Fixes self-reported #3606; regression introduced after #3548 

Thank you for maintaining this awesome piece of work!

## Checks

### DIFF
--- a/crates/atuin-pty-proxy/src/pty_proxy.rs
+++ b/crates/atuin-pty-proxy/src/pty_proxy.rs
@@ -147,29 +147,75 @@ fn env_flag(name: &str) -> bool {
     })
 }
 
-fn render_init(shell: Shell) -> &'static str {
+fn get_parent_shell_path(target_shell: Shell) -> Option<String> {
+    #[cfg(unix)]
+    {
+        let ppid = unsafe { libc::getppid() };
+
+        #[cfg(target_os = "linux")]
+        let path = std::fs::read_link(format!("/proc/{ppid}/exe")).ok();
+
+        #[cfg(target_os = "macos")]
+        let path = {
+            let mut buf = [0u8; 4096];
+            let res = unsafe {
+                libc::proc_pidpath(
+                    ppid,
+                    buf.as_mut_ptr() as *mut std::ffi::c_void,
+                    buf.len() as u32,
+                )
+            };
+            if res > 0 {
+                let bytes = &buf[..res as usize];
+                let len = bytes.iter().position(|&b| b == 0).unwrap_or(bytes.len());
+                std::str::from_utf8(&bytes[..len]).ok().map(PathBuf::from)
+            } else {
+                None
+            }
+        };
+
+        #[cfg(any(target_os = "illumos", target_os = "solaris"))]
+        let path = std::fs::read_link(format!("/proc/{ppid}/path/a.out")).ok();
+
+        #[cfg(not(any(
+            target_os = "linux",
+            target_os = "macos",
+            target_os = "illumos",
+            target_os = "solaris"
+        )))]
+        let path: Option<PathBuf> = None;
+
+        if let Some(path) = path
+            && let Some(detected_shell) = shell_from_name(&path.to_string_lossy())
+            && detected_shell == target_shell
+        {
+            return Some(path.to_string_lossy().into_owned());
+        }
+    }
+    None
+}
+
+fn render_init(shell: Shell) -> String {
     // Each shell embeds its own interpreter path in the `--shell` argument so
     // `atuin pty-proxy` spawns the same binary that sourced the init, rather
     // than resolving via $PATH (which can pick the wrong installation when the
     // user has, for instance, both /usr/bin/bash and /opt/homebrew/bin/bash).
     match shell {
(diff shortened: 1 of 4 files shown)


crates/atuin-pty-proxy/src/pty_proxy.rs
```diff
@@ -164,12 +164,22 @@ fn render_init(shell: Shell) -> &'static str {
...
+      # ZSH_ARGZERO preserves argv0, including the leading dash used by login
+      # shells, so normalize it before choosing a shell path.
+      _atuin_pty_proxy_zsh_argv0="${ZSH_ARGZERO:-zsh}"
+      _atuin_pty_proxy_zsh_argv0="${_atuin_pty_proxy_zsh_argv0#-}"
+      if [[ "$_atuin_pty_proxy_zsh_argv0" == */* ]]; then
+        # argv0 already contains a path, so it identifies the spawned zsh.
+        _atuin_pty_proxy_shell="$_atuin_pty_proxy_zsh_argv0"
+      elif [[ -f "${SHELL:-}" ]] && [[ -x "${SHELL:-}" ]] && [[ "${SHELL##*/}" == "$_atuin_pty_proxy_zsh_argv0" ]]; then
+        # Bare argv0: prefer the matching login shell path over PATH lookup.
+        _atuin_pty_proxy_shell="$SHELL"
```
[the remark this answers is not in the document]
hangrymuppet: Agreed; this is brittle. Refactored to find the parent shell in rust and pass it to the script.

crates/atuin-pty-proxy/src/runtime.rs
```diff
@@ -35,7 +35,26 @@ fn run(options: RuntimeOptions) -> eyre::Result<()> {
...
+                // Fallback if not found in $PATH
+                .or_else(|| Some(path.clone()))
+        } else {
+            // No $PATH variable
+            Some(path.clone())
+        }
+    } else {
+        // No shell specified
+        None
+    };
```
poliorcetics:
```suggestion
    let shell_path = match options.shell {
        Some(ref path) => if path.is_absolute() {
                Some(path.clone())
            } else if let Ok(paths) = std::env::var("PATH") {
                // Resolve via $PATH
                std::env::split_paths(&paths)
                    .map(|p| p.join(path))
                    .find(|p| p.is_file())
                    // Fallback if not found in $PATH
                    .or_else(|| Some(path.clone()))
            } else {
                // No $PATH variable
                Some(path.clone())
            }
        // No shell specified
        None => None,
    };
```

Or even use `options.shell.as_ref().map(|path| { /* ... */ })` maybe ?

crates/atuin-pty-proxy/src/pty_proxy.rs
```diff
@@ -147,29 +147,75 @@ fn env_flag(name: &str) -> bool {
     })
 }
 
-fn render_init(shell: Shell) -> &'static str {
+fn get_parent_shell_path(target_shell: Shell) -> Option<String> {
+    #[cfg(unix)]
+    {
+        let ppid = unsafe { libc::getppid() };
```
poliorcetics:
```suggestion
        let ppid = std::os::unix::process::parent_id();
```

crates/atuin-pty-proxy/src/pty_proxy.rs
```diff
@@ -147,29 +147,75 @@ fn env_flag(name: &str) -> bool {
...
+
+        #[cfg(target_os = "linux")]
+        let path = std::fs::read_link(format!("/proc/{ppid}/exe")).ok();
+
+        #[cfg(target_os = "macos")]
+        let path = {
+            let mut buf = [0u8; 4096];
+            let res = unsafe {
```
poliorcetics: Could you add `// Safety:` comments on every `unsafe` block ?

crates/atuin-pty-proxy/src/pty_proxy.rs
```diff
@@ -147,29 +147,75 @@ fn env_flag(name: &str) -> bool {
...
+            target_os = "solaris"
+        )))]
+        let path: Option<PathBuf> = None;
+
+        if let Some(path) = path
+            && let Some(detected_shell) = shell_from_name(&path.to_string_lossy())
+            && detected_shell == target_shell
+        {
+            return Some(path.to_string_lossy().into_owned());
+        }
```
poliorcetics, line 203:
```suggestion
        if let Some(path) = path
            && let path = path.to_string_lossy()
            && let Some(detected_shell) = shell_from_name(&path)
            && detected_shell == target_shell
        {
            return Some(path.into_owned());
        }
```

Avoids doing the `to_string_lossy()` twice (which does a UTF-8 check each time)

crates/atuin-pty-proxy/src/pty_proxy.rs
poliorcetics: With this change it's also possible to make the `libc` dep macos-only, see https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#platform-specific-dependencies

PR #3612, opened 2026-07-12
fix(cli): support well-defined but read-only databases

### LINKED ISSUE #3550: [Bug]: Atuin hard-fails when the DB is unwritable — propose read-only fallback

When the history database is unwritable for **any** reason, Atuin should degrade gracefully to **read-only** history search — continuing to serve `__atuin_history` search and up-arrow recall from the existing DB, skipping new-history recording until the DB is writable again. The history already in the DB is fully available for read; a write failure should not make it inaccessible.

### What happened?

When the DB cannot be opened read-write, Atuin emits a hard error and the history UI is completely inaccessible. One concrete trigger (observed here):

```
Error: error returned from database: (code: 4874) disk I/O error

Caused by:
    (code: 4874) disk I/O error

Location:
    crates/atuin/src/command/client.rs:354:18
```

The shell itself keeps working, but command history is locked behind the very database that can't be written — even though its existing contents are still readable. Any DB-writability failure (not just this one) produces the same lockout.

### Related / prior issues

Same root condition reported in two closed issues, framed here as a feature request for **read-only degradation** rather than a bug in the error path:

- #2334 — same `(code: 4874) disk I/O error`, closed as "expected behaviour" (maintainer: no way to function correctly with no disk space).
- #1886 — "Out of disk space makes atuin unusable", closed as out of scope. A Feb 2026 comment by @linuxissuper describes the recovery use case: *"I want to open my shell history and look for a command that frees some space, but because atuin cannot function in read only mode, I cannot look for the command."*

The maintainer's prior position was that failing with an error is acceptable and that a pre-write disk-space check on every operation is out of scope. This issue does **not** ask for a disk-space check. It asks for read-only fallback: if a DB write fails (for any reason — I/O error, read-only mount, lock contention, network drop), continue serving read-only queries from the existing DB rather than hard-failing all subsequent history access.

### Proposal: read-only fallback

When the DB write path fails, open the DB read-only (`SQLITE_OPEN_READONLY`) and keep serving `__atuin_history` search and up-arrow recall. Skip new-history recording until the DB is writable again (warn once, then suppress). SQLite already supports this — the history in the DB is readable regardless of writability.

### Situations where read-only history access is genuinely useful

**Generic unwritable / filesystem**
1. **Immutable / read-only filesystems** — atuin DB on a read-only mount, NixOS read-only root, ephemeral containers, forensics/audit hosts where the history store is intentionally mounted read-only.

**Trust & security**
2. **Agent sandboxes** — when a semi-trusted agent injects commands into the shell, the user may not want those commands persisted to history (risk of accidentally re-running an agent's exploratory command later). Read-only mode lets the user browse history without atuin recording the agent's injected commands, and without trusting the agent's commands as future-history the user might blindly run.
3. **History-poisoning defense** — when running *untrusted* code (downloaded scripts, inspected binaries, curl|bash): a distinct threat model from the semi-trusted-agent case. Read-only mode ensures untrusted code cannot inject malicious or misleading commands into the user's permanent history that they might later recall and run.
4. **Audit / compliance mode** — regulated environments where the shell user must not mutate history; a central process owns writes, users get read-only access to the shared history.

**ENOSPC & recovery**
5. **ENOSPC recovery** — the case in #1886/#2334: the recovery commands are *in* the history, unreachable because the DB won't open read-write.
6. **Network-attached DB flapping** — DB on NFS/Syncthing/etc. where the mount drops; read-only fallback lets the user keep working with the last-good history rather than hard-failing every keystroke.
7. **DB-locked / WAL-checkpoint-failed states** — SQLite `SQLITE_LOCKED` or a failed WAL checkpoint can make writes fail while reads still work; degrading to read-only keeps history usable during recovery.

**Resource & shared**
8. **Shared history stores** — multiple machines reading one shared DB (NFS, Syncthing) where only one is the designated writer; others should be read-only to avoid SQLite multi-writer corruption. Currently there's no supported read-only-client mode.
9. **Tmpfs / tiny `/home`** — dev laptops and CI shells where the user wants history *access* but not the write amplification of recording every command.

### Atuin doctor output

```yaml
{
  "atuin": {
    "version": "18.16.1",
    "commit": "NO_GIT",
    "sync": null,
    "sqlite_version": "3.53.1"
  },
  "shell": {
    "name": "zsh",
    "default": "zsh",
    "plugins": ["atuin"],
    "preexec": "built-in"
  },
  "system": {
    "os": "Manjaro Linux",
    "arch": "x86_64",
    "version": "26.1.0-pre"
  }
}
# `system.disks` omitted: atuin doctor emits ~60 duplicate btrfs subvolume
# entries for the same backing device (`/dev/mapper/vg_svelte-home`,
# filesystem `btrfs`). Full output available on request.
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

poliorcetics:
As described in #3550, `atuin` failed on a read-only database even when the operation was just "read existing history".

I added a fallback that attempts to detect a read-only database and gracefully handles write operations in that case.

Note that read operations will still fail if migrations are missing or have been only partially applied, we can't exactly do much about this, we just hope for the best.

I did not add a user-visible warning or notice, I'm not sure where I should do or even _if_ I should do so.

Closes #3550.

## Checks

### DIFF
--- a/crates/atuin-client/src/database.rs
+++ b/crates/atuin-client/src/database.rs
@@ -204,6 +204,7 @@ pub trait Database: Send + Sync + 'static {
 #[derive(Debug, Clone)]
 pub struct Sqlite {
     pub pool: SqlitePool,
+    pub read_only: bool,
 }
 
 impl Sqlite {
@@ -224,9 +225,16 @@ impl Sqlite {
             fs::create_dir_all(dir)?;
         }
 
+        // Best effort, we can likely improve this detection if more exotic cases are reported.
+        let read_only = path.exists() && fs::metadata(path)?.permissions().readonly();
+        if read_only {
+            debug!("database is read-only: write operations will be ignored");
+        }
+
         let opts = SqliteConnectOptions::from_str(path.as_os_str().to_str().unwrap())?
             .journal_mode(SqliteJournalMode::Wal)
-            .optimize_on_close(true, None)
+            .optimize_on_close(!read_only, None)
+            .read_only(read_only)
             .synchronous(SqliteSynchronous::Normal)
             .with_regexp()
             .create_if_missing(true);
@@ -236,8 +244,11 @@ impl Sqlite {
             .connect_with(opts)
             .await?;
 
-        Self::setup_db(&pool).await?;
-        Ok(Self { pool })
+        // We can't migrate if the database is readonly, in which case we hope for the best.
+        if !read_only {
+            Self::setup_db(&pool).await?;
+        }
+        Ok(Self { pool, read_only })
     }
 
     pub async fn sqlite_version(&self) -> Result<String> {
@@ -324,6 +335,12 @@ impl Sqlite {
 impl Database for Sqlite {
     async fn save(&self, h: &History) -> Result<()> {
         debug!("saving history to sqlite");
+
+        if self.read_only {
+            debug!("database is read-only, skipping save");
+            return Ok(());
+        }
+
         let mut tx = self.pool.begin().await?;
         Self::save_raw(&mut tx, h).await?;
         tx.commit().await?;
@@ -333,6 +350,11 @@ impl Database for Sqlite {
 
     async fn save_bulk(&self, h: &[History]) -> Result<()> {
         debug!("saving history to sqlite");
+
+        if self.read_only {
+            debug!("database is read-only, skipping bulk save");
+            return Ok(());
+        }
 
         let mut tx = self.pool.begin().await?;
 
@@ -359,6 +381,11 @@ impl Database for Sqlite {
 
     async fn update(&self, h: &History) -> Result<()> {
         debug!("updating sqlite history");
+
+        if self.read_only {
+            debug!("database is read-only, skipping update");
+            return Ok(());
+        }
 
         sqlx::query(
             "update history
@@ -745,6 +772,11 @@ impl Database for Sqlite {
     // deleted_at doesn't mean the actual time that the user deleted it,
     // but the time that the system marks it as deleted
     async fn delete(&self, mut h: History) -> Result<()> {
+        if self.read_only {
+            debug!("database is read-only, skipping delete");
+            return Ok(());
+        }
+
         let now = OffsetDateTime::now_utc();
         h.command = rand::thread_rng()
             .sample_iter(&Alphanumeric)
@@ -759,6 +791,11 @@ impl Database for Sqlite {
     }
 
     async fn delete_rows(&self, ids: &[HistoryId]) -> Result<()> {
+        if self.read_only {
+            debug!("database is read-only, skipping rows deletion");
+            return Ok(());
+        }
+
         let mut tx = self.pool.begin().await?;
 
         for id in ids {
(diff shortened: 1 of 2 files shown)


crates/atuin-client/src/database.rs
```diff
@@ -224,9 +225,15 @@ impl Sqlite {
             fs::create_dir_all(dir)?;
         }
 
+        let read_only = path.exists() && fs::metadata(path)?.permissions().readonly();
```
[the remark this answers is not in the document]
markovejnovic: Greptile is unfortunately right. This doesn't work on windows, unfortunately. `fs::metadata` invokes [`GetFileInformationByHandle`](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getfileinformationbyhandle) which has no system to answer whether a file can be read or not.

Additionally, this doesn't work on Linux either due to groups. Please see [`Permissions::readonly`](https://doc.rust-lang.org/std/fs/struct.Permissions.html#method.readonly) for more info if you're curious.

I would suggest we design a system in `atuin-common` which enables us to answer the question of whether the current user can read the file or not. I'm not 100% sure what that looks like right now.

crates/atuin-client/src/database.rs
```diff
@@ -224,9 +225,16 @@ impl Sqlite {
             fs::create_dir_all(dir)?;
         }
 
+        // Best effort, we can likely improve this detection if more exotic cases are reported.
+        let read_only = path.exists() && fs::metadata(path)?.permissions().readonly();
+        if read_only {
+            debug!("database is read-only: write operations will be ignored");
```
markovejnovic, line 231:
```suggestion
            warn!("database is read-only: write operations will be ignored");
```

crates/atuin-client/src/database.rs
```diff
@@ -224,9 +225,15 @@ impl Sqlite {
...
 
+        let read_only = path.exists() && fs::metadata(path)?.permissions().readonly();
+        if read_only {
+            debug!("database is read-only: write operations will be ignored");
+        }
+
         let opts = SqliteConnectOptions::from_str(path.as_os_str().to_str().unwrap())?
             .journal_mode(SqliteJournalMode::Wal)
             .optimize_on_close(true, None)
+            .read_only(read_only)
```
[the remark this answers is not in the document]
markovejnovic: Yikes, I didn't even think about this, and greptile is correct. I think this case is really rare, so I wouldn't worry about it, but I would at least leave a comment.

crates/atuin-client/src/database.rs
```diff
@@ -236,8 +244,11 @@ impl Sqlite {
             .connect_with(opts)
             .await?;
 
-        Self::setup_db(&pool).await?;
-        Ok(Self { pool })
+        // We can't migrate if the database is readonly, in which case we hope for the best.
+        if !read_only {
+            Self::setup_db(&pool).await?;
+        }
```
markovejnovic, line 250:
```suggestion
        if !read_only {
            Self::setup_db(&pool).await?;
        } else {
            warn!("...");
        }
```

crates/atuin-client/src/database.rs
```diff
@@ -324,6 +335,12 @@ impl Sqlite {
 impl Database for Sqlite {
     async fn save(&self, h: &History) -> Result<()> {
         debug!("saving history to sqlite");
+
+        if self.read_only {
+            debug!("database is read-only, skipping save");
+            return Ok(());
+        }
```
markovejnovic, line 342: I really dislike this personally... It's quite bug-prone to have these checks at the start of every function. I would suggest something like this:

```rust
trait ReadsDb {
  async fn load(&self, id: &str) -> Result<Option<History>>;
  // ...
}

trait WritesDb {
  async fn save(&self, h: &History) -> Result<()>;
}

struct NopDbWriter;
impl WritesDb for NopDbWriter {
  async fn save(&self, h: &History) -> Result<()> {
    warn!("i cannot write!");
  }
}

struct DbWriter;
impl WritesDb for DbWriter {
  async fn save(&self, h: &History) -> Result<()> {
    sqlx::query("update ...");
  }
}
```

and then compose these

```rust
struct Database {
  reader: Box<ReadsDb>,
  writer: Box<WritesDb>,
}

impl Database {
  fn reader(&self) -> Box<ReadsDb> {
    self.reader
  }
  fn writer(&self) -> Box<WritesDb> {
    self.writer
  }
}
```

So the calling convention becomes

```rust
db.writer().save(h);
```

and each implementation worries about itself. Worth noting that the `Box` means that cache coherency isn't great. Something something fat pointer/vtable. Perhaps there can be an `enum` facade which composes all the possible writers/readers to solve this problem:

```rust
enum Writer {
  Nop(NopDbReader),
  Standard(DbWriter),
}

impl WritesDb for Writer {
  async fn save(&self, h: &History) -> Result<()> {
    match self {
      x => x.save()
    }
  }
}

struct Database {
  reader: DbReader,
  writer: Box<WritesDb>,
}

impl Database {
  fn reader(&self) -> &impl ReadsDb {
    self.reader
  }
  fn writer(&self) -> &impl WritesDb {
    self.writer
  }
}
```

This avoids the `Box` but is so darn verbose. I wish there was a crate (maybe there is, I didn't do my research).

poliorcetics:
@markovejnovic

Thanks for the review, sadly this is entirely too much work for me to invest in this MR.

Feel free to take it up or give it to someone else, but seeing any LLM reviewing or contributing kills my motivation like nothing else so I'll not be continuing it.

markovejnovic:
No worries, thank you for your contribution regardless! The review is not LLM generated (I wouldn't do that to anyone), fully human written =), but I totally understand!

No worries, sorry for causing you pain. Thanks regardless for your PR!

poliorcetics:
> The review is not LLM generated

Oh no sorry, I was not talking about _your_ review but about the LLM review bot, sorry if that was not clear!

markovejnovic:
No worries! We'll try to implement this at some point =)

PR #3616, opened 2026-07-13
fix: Various prefix mode bugs

BinaryMuse:
1. `is_insert_mode` now requires `tab_index == 0`, so unmatched keys on the inspector tab no longer fall through to a hidden search input
2. `default_inspector_keymap` now binds `ctrl+[prefix]` to `EnterPrefixMode`, same as the other keymaps
3. While prefix mode is armed (e.g. prefix chord pressed but waiting on the next key), the mode bracket shows `PREFIX`

### DIFF
--- a/crates/atuin/src/command/client/search/interactive.rs
+++ b/crates/atuin/src/command/client/search/interactive.rs
@@ -276,11 +276,14 @@ impl State {
     }
 
     /// Whether the current mode supports character insertion on unmatched keys.
+    /// The inspector tab has no text input, so unmatched keys are dropped there
+    /// rather than leaking into the (hidden) search input.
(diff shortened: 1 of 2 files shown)


crates/atuin/src/command/client/search/interactive.rs
```diff
@@ -1197,7 +1200,9 @@ impl State {
     }
 
     fn build_input(&self, style: StyleState, prefix_width: u16) -> Paragraph<'_> {
-        let (pref, mode) = if self.switched_search_mode {
+        let (pref, mode) = if self.prefix {
+            ("", "PREFIX")
```
ellie, line 1204: this could be confused for the prefix search mode, but tbh I can't think of anything better

PR #3617, opened 2026-07-13
chore: update to rust 1.97

ellie:
Updates across the board, and also fix the clippy issues that came up

## Checks


develop7:
[openSUSE package](https://build.opensuse.org/package/show/utilities/atuin) maintainer here: I don't think it's a good idea to bump MSRV in a _patch_ release. To the one affected to a major issue, at that — https://github.com/rust-lang/rust/issues/159035

PR #3622, opened 2026-07-14
chore(logging): refactor logging

taylordotfish:
* Write logs with level `warn` and above to stderr when a log file isn't being used. Some parts of the code call `tracing::warn!` with user-facing messages; we want to make sure the user actually sees them.
* Add `--no-logs` option to `history start` and `history end` to skip initializing logging; used by shell hooks to avoid corrupting the terminal and to minimize the amount of time the commands take to run.
* Initialize logs only in one place (`atuin/src/logs.rs`); remove large amounts of duplication.
* Don't set `ATUIN_LOG` in shell hooks; some of these occurrences were useless (we didn't log anything from `atuin history` prior to this commit) and all of them removed the ability to override the log level (e.g., to `debug`).
* Don't exit or crash the program if logging fails to initialize. This is more important now that logging is initialized in all code paths.
* Logs from `atuin daemon --show-logs` now match the logs written to `daemon.log` (different log levels were sometimes used previously). Note that these logs are now written to stderr instead of stdout.
* Logs written to stderr don't show the target or time by default (overridden for `daemon --show-logs`).
* Logs written to stderr don't use ANSI escape sequences when stderr is not a terminal.

### DIFF
--- a/crates/atuin-common/src/logs.rs
+++ b/crates/atuin-common/src/logs.rs
@@ -1,0 +1,83 @@
+use serde::{Deserialize, Serialize};
+use std::ffi::OsStr;
+use std::path::{Path, PathBuf};
+
+/// Log level for file logging. Maps to tracing's LevelFilter.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, Deserialize, Serialize)]
+#[serde(rename_all = "lowercase")]
+pub enum LogLevel {
+    Trace,
+    Debug,
+    #[default]
+    Info,
+    Warn,
+    Error,
+}
+
+impl LogLevel {
+    #[cfg(feature = "tracing")]
+    pub fn to_tracing(&self) -> tracing::Level {
+        use tracing::Level;
+        match self {
+            LogLevel::Trace => Level::TRACE,
+            LogLevel::Debug => Level::DEBUG,
+            LogLevel::Info => Level::INFO,
+            LogLevel::Warn => Level::WARN,
+            LogLevel::Error => Level::ERROR,
+        }
+    }
+}
+
+#[derive(Debug)]
+pub struct FileConfig {
+    pub path: PathBuf,
+    pub level: LogLevel,
+    pub retention_days: u64,
+}
+
+impl FileConfig {
+    pub fn directory(&self) -> &Path {
+        self.path.parent().unwrap_or_else(|| Path::new(""))
+    }
+
+    pub fn name(&self) -> &OsStr {
+        self.path.file_name().unwrap_or_else(|| OsStr::new(""))
+    }
+}
+
+#[derive(Debug, Default)]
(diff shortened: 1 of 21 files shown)


markovejnovic:
looking

crates/atuin-common/src/logs.rs
```diff
@@ -0,0 +1,83 @@
...
+        use tracing::Level;
+        match self {
+            LogLevel::Trace => Level::TRACE,
+            LogLevel::Debug => Level::DEBUG,
+            LogLevel::Info => Level::INFO,
+            LogLevel::Warn => Level::WARN,
+            LogLevel::Error => Level::ERROR,
+        }
+    }
+}
```
markovejnovic, line 29: Why not `impl From<LogLevel> for tracing::Level`

crates/atuin-common/src/logs.rs
```diff
@@ -0,0 +1,83 @@
...
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, Deserialize, Serialize)]
+#[serde(rename_all = "lowercase")]
+pub enum LogLevel {
+    Trace,
+    Debug,
+    #[default]
+    Info,
+    Warn,
+    Error,
+}
```
markovejnovic, line 15: I vote in favor of removing this if possible, since I think it's equivalent to `tracing::Level`. If anything, I'd rather we implement an extension for deserializing it, if a FromStr doesn't already exist.

markovejnovic: It looks like this would have to be a `newtype` pattern, would suggest we do that.

ellie (approved):

crates/atuin/src/logs.rs
```diff
@@ -0,0 +1,166 @@
...
+pub fn init_logging(config: &LogConfig) {
+    // We have to dispatch the time config statically; see
+    // https://github.com/tokio-rs/tracing/issues/3180
+    match &config.stderr {
+        Some(StderrConfig {
+            show_time: false, ..
+        }) => with_stderr_time::<()>(config),
+        _ => with_stderr_time::<fmt::time::SystemTime>(config),
+    }
+}
```
markovejnovic, line 20: I would prefer if this was a `lazy_static!` structure that was initialized on-demand, and we could even do a `debug_assert!` that this isn't re-initialized twice.

crates/atuin-common/src/logs.rs
taylordotfish: Absolutely open to a PR if you can figure out a good way to do this that's less code and doesn't allocate

crates/atuin-common/src/logs.rs
taylordotfish: Fair point, mostly historical because it was replacing another method called `as_directive`

PR #3623, opened 2026-07-14
docs(ai): Add self hosting + slash command docs

BinaryMuse:
Adds docs for:

- Atuin AI self hosting
- Atuin AI slash commands


ellie (dismissed):
looks great! probs worth linking to from the intro on the main ai page too

PR #3624, opened 2026-07-14
fix(search): correct fuzzy result ranking (#3603)

### LINKED ISSUE #3603: Should Atuin prefer one or the other of these two selections?
I am recalling the line `nix-shell -p irssi screen --run "screen -r"` using `^R`.

Upon typing `^Rscreen`, I am presented with:

Image

and upon pressing space, such that I've typed `^Rscreen `, I am presented with:

Image

I would think that `^Rscreen ` should still prefer the first result, since there's an actual space in that line, and the space is actually after `screen`.

sshine:
- Fixes #3603

Fuzzy search ranked results wrongly whenever the query handed to the span scorer didn't match the characters SQL actually matched on. Most visibly: adding a trailing space to a query pushed an unrelated row to the top.

- **Non-matches outranked real matches.** `reorder_fuzzy` scored an unmatched row with the sentinel `(0, results.len())`. With few results that number is tiny and could beat a genuine match's span. Unmatched rows now map to `usize::MAX`, so they always sort last.
- **The ranking query carried characters SQL never matched.** SQL matches on tokens with whitespace and operators (`^ $ ' ! |`, `r/.../`) stripped, but the raw query went to the span scorer. Any space or operator failed the subsequence check, so every row scored as a non-match and fell back to recency. The ranking query is now rebuilt from the tokenizer's plain terms, so it always equals what SQL matched.

Each fix ships a regression test isolating it from the others, plus a characterisation test that plain and multi-word queries rank unchanged.

### DIFF
--- a/crates/atuin-client/src/database.rs
+++ b/crates/atuin-client/src/database.rs
@@ -1796,11 +1796,14 @@ mod test {
     }
 
     // Reproduces the trailing-space ranking bug (atuinsh/atuin#3603).
+    #[rstest]
     #[tokio::test(flavor = "multi_thread")]
-    async fn test_search_fuzzy_trailing_space() {
-        let mut db = Sqlite::new("sqlite::memory:", test_local_timeout())
-            .await
-            .unwrap();
+    async fn test_search_fuzzy_trailing_space(
+        #[future(awt)]
+        #[from(empty_db)]
+        db: Sqlite,
+    ) {
+        let mut db = db;
 
         let now = OffsetDateTime::now_utc();
         let irssi = "screen irssi";
--- a/crates/atuin-client/src/database.rs
+++ b/crates/atuin-client/src/database.rs
@@ -806,9 +806,16 @@ impl Database for Sqlite {
             .fetch_all(&self.pool)
             .await?;
 
-        // Rank against the same characters SQL matched: the tokenizer drops spaces, so
-        // strip them here too rather than scoring them as part of the fuzzy subsequence.
-        let reorder_query: String = orig_query.split_whitespace().collect();
+        // Rank against the same characters SQL matched: drop spaces and operators.
+        let reorder_query: String = QueryTokenizer::new(orig_query)
+            .filter_map(|token| match token {
+                QueryToken::Match(term, _)
+                | QueryToken::MatchStart(term, _)
+                | QueryToken::MatchEnd(term, _)
+                | QueryToken::MatchFull(term, _) => Some(term),
+                QueryToken::Or | QueryToken::Regex(_) => None,
+            })
+            .collect();
         Ok(ordering::reorder_fuzzy(search_mode, &reorder_query, res))
     }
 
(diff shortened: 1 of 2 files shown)


sshine:
Note: the fossier/greptile-generated summaries only review the test commit that existed when the PR was created, so it does not mention the 3 fix commits.

crates/atuin-client/src/database.rs
```diff
@@ -806,7 +806,17 @@ impl Database for Sqlite {
             .fetch_all(&self.pool)
             .await?;
 
-        Ok(ordering::reorder_fuzzy(search_mode, orig_query, res))
+        // Rank against the same characters SQL matched: drop spaces and operators.
+        let reorder_query: String = QueryTokenizer::new(orig_query)
+            .filter_map(|token| match token {
```
taylordotfish, line 812: If we don't exclude inverse tokens, any query containing them will cause all commands to have the same fuzzy score of `usize::MAX`.

```suggestion
        let reorder_query: String = QueryTokenizer::new(orig_query)
            .filter(|token| !token.is_inverse())
            .filter_map(|token| match token {
```

crates/atuin-client/src/ordering.rs
```diff
@@ -14,19 +14,48 @@ where
...
-            // this is a little unfortunate: when we are asked to match a query that is found nowhere,
-            // we don't want to return a None, as the comparison behaviour would put the worst matches
-            // at the front. therefore, we'll return a set of indices that are one larger than the longest
-            // possible legitimate match. This is meaningless except as a comparison.
-            None => (0, len),
-        };
-        1 + to - from
+        // A non-matching row sorts last rather than by a sentinel that can undercut a real match.
+        minspan::span(qvec, &(f(h).chars().collect()))
+            .map_or(usize::MAX, |(from, to)| 1 + to - from)
```
taylordotfish, line 22: Thank you for this! The old fallback length was just completely wrong; we were using the number of results, but it looks like we meant to use the length of the longest result. But `usize::MAX` should work just as well.

taylordotfish (commented):
Looks good, just one comment about removing inverse terms from the stripped query.

crates/atuin-client/src/ordering.rs
```diff
@@ -14,19 +14,48 @@ where
...
+    }
+
+    // The unchanged match path: a tight match outranks a loose one.
+    #[test]
+    fn reorder_matches_by_span() {
+        let res = vec![hist("central urllib"), hist("curl")];
+        let out = reorder_fuzzy(DbSearchMode::Fuzzy, "curl", res);
+        assert_eq!(commands(out), vec!["curl", "central urllib"]);
+    }
+}
```
markovejnovic, line 57: we're. moving towards https://rstest.rs/, would you mind using rstest here?

crates/atuin-client/src/database.rs
sshine: Thanks for making me aware of inverse tokens. Looking further at this, I think that a `.filter(|token| !token.is_inverse())` will do the job of not including inverse tokens, but I think there is more work to be done in ranking them properly.

crates/atuin-client/src/ordering.rs
sshine: Ah yes, I had mindlessly rebased without noticing the migration.

sshine:
Thanks, @taylordotfish and @markovejnovic. I've addressed your feedback; I have refrained from squashing my fixup commits for ease of review, but I'll squash them any time.

I think there is some improvements to be made wrt. operators and inverses, but it's a little out of scope perhaps.

markovejnovic:
Thank you for your PR @sshine!

PR #3625, opened 2026-07-14
AI WIP | refactor(client): migrate search builder to sqlx bind parameters

markovejnovic:
## Summary

Migrates the client-side SQLite search builder (`Sqlite::search`) from **inlining escaped values** into the SQL string (via `sql_builder`'s `quote()`/`esc()`) to **real sqlx bind parameters** (`?` placeholders + `.bind()`). Query correctness no longer depends on a human remembering the `quote()`-for-`_eq` vs raw-for-`_like` escaping contract on every edit.

This started as a defense-in-depth refactor — the original search code was **not** actually injectable (the escaping was correct). Along the way an audit turned up one genuinely fragile site elsewhere (`daemon.rs`), which is also fixed here.

## What changed

- **`crates/atuin-client/src/database.rs`**
  - `build_inner_query` builds the inner `SELECT` with `?` placeholders and collects values, in placeholder order, into a `Vec<SqlValue>` (`Int(i64)`/`Text(String)`). No user value is interpolated into the SQL text.
  - `assemble_search_query` returns `(String, usize)` — the assembled SQL and how many times the inner query was embedded (once for `--dupes`, twice for the `NOT EXISTS` dedup wrapper). `search` binds `args` once per embed, driven by that returned count rather than an independently-recomputed one. A `debug_assert_eq!` guards bind arity (sqlx-sqlite binds `NULL` *silently* on a mismatch — no error, just zero rows).
  - `fuzzy_condition` and `apply_author_filter` migrated to placeholders + args.
  - `list` / `stats` / `Paged` / `all_with_count` are **deliberately left** on `quote()`-based inlining (correct as-is); the split is now called out in a code comment so nobody copies a bare `"?"` into them without a matching `.bind()`.
- **`crates/atuin/src/command/client/search/engines/daemon.rs`** — `hydrate_from_db` interpolated history ids into `WHERE id IN (...)` with `format!("'{id}'")` and **no escaping**. Safe today only because its one caller feeds it clean 32-hex UUIDs, but `HistoryId` is an unvalidated `String` newtype fed from a gRPC field, and `query_history` takes raw SQL with no bind support. Now every id is normalised through `Uuid::parse_str().as_simple()`; non-UUID input is dropped, so nothing but 32 hex chars can reach the SQL.
- **`SQL_INJECTION_AUDIT.md`** — full-workspace audit documenting what's parameterized, the `sql_builder` escaping contract, and the remaining documented-not-fixed items (below).

## Not fixed here (documented in the audit, deferred)

- `interactive.rs` `DeleteAllMatching` — hand-rolled `'` escaping. Safe, but not bound.
- `atuin-daemon/src/components/search.rs` — interpolates hyphenated *record* UUIDs unquoted against a column that stores 32-hex *history* UUIDs; looks like it can never match. Suspected pre-existing bug, filed separately.

## Testing

- `cargo test -p atuin-client`: 98 lib tests + doctests pass, including new unit tests asserting the `?`-count == `args.len()` invariant, that user input is never inlined, and end-to-end quote handling.
- `cargo test -p atuin --features client daemon::`: 9 pass, including two new tests that throw injection strings at `hydrate_from_db` and confirm only valid UUID rows survive.
- `cargo clippy --all-targets -- -D warnings` clean on both crates.
- Manual smoke test against a real ~51k-row history DB: plain terms, `OR` tokens, regex tokens, a bare `'`, and a `'; DROP TABLE history; --` payload all execute without error; history table intact.

## Review notes

The load-bearing correctness property is that values are pushed to `args` in the same order the `?` placeholders appear left-to-right in the generated SQL. This holds because `sql_builder` stores WHERE conditions in insertion order and `or_where` appends to the last element — traced against the crate source for every filter-mode × search-mode × filter-option combination during review.


markovejnovic:
although the intent of this PR is correct, the refactor is involved, difficult and introduces too much risk to be valuable imo. closing

PR #3626, opened 2026-07-14
feat(common): add storage-generic CommandStr/CommandString newtype

markovejnovic:
## What

Adds a storage-generic newtype for shell command strings in `atuin-common`: `crates/atuin-common/src/command_str.rs`.

`Command<S>` is a `#[repr(transparent)]` wrapper generic over the *storage* `S` that holds the command text, specialised by type alias:

| Alias | Storage | Plays the role of |
|---|---|---|
| `CommandStr<'a>` | `&'a str` | `&'a str` (borrowed, `Copy`) |
| `CommandString` | `String` | `String` (owned) |
| `CommandCow<'a>` | `Cow<'a, str>` | `Cow<'a, str>` |

Every string-facing impl is written once, bounded uniformly on `S: AsRef<str>`, so any other string storage (`Arc<str>`, `Box<str>`, an interned handle, a compact-string type) becomes a supported command flavour with no new code.

## Design notes

- **Minimal behaviour, one validated invariant.** The wrapper stores its string verbatim — no normalisation, shell splitting, escaping, quoting, trimming, or truncation. `len`/`is_empty`/`contains`/etc. come free via `Deref<Target = str>`. The single structural invariant it enforces is at the validated-construction boundary (below).
- **Validated construction: `CommandString::try_from(&str)`.** Rejects any command containing a **NUL byte**, returning the `InvalidCommand` error (with the offending byte index). NUL is the one thing that can't legitimately appear in a shell command: it can't arrive through argv/env capture (both NUL-terminated), it's silently truncated by the shell's `$(...)` substitution on replay, and it confuses SQLite's NUL-terminated string functions used in history search. **Newlines (multi-line commands), tabs, other control characters and the empty string all remain valid** and are not rejected.
- **Cross-storage comparison.** `PartialEq`/`PartialOrd` are hand-written and compare across storages (`CommandStr == CommandString` when the text matches); `Eq`/`Ord`/`Hash` are hand-written too, all forwarding to `as_str()`. Not derived, because deriving would bound on `S: PartialEq`/`S: Hash` and leave the specialisations mutually incomparable.
- **`Borrow<str>` is contract-correct.** `Hash`/`Eq`/`Ord` all match `str`'s own, so a `HashMap<CommandString, _>` can be probed with a plain `&str`.
- **Transparent serde.** `#[serde(transparent)]` — a command serialises as a bare string, byte-identical to the `String` it wraps, so a future migration of an existing `String` field to `CommandString` never changes the wire format. Mirrors the UUID newtypes in `lib.rs`.
- **`CommandStr<'a>` is a sized value type, not an unsized `str`-like DST** — a true DST would require `unsafe` (the crate is `#![deny(unsafe_code)]`) and could not be generic over storage. Consequence: `&CommandString` does not auto-deref into a `CommandStr`; call `.as_command_str()`.
- Std derives are used rather than the repo's usual `derive_more`, because the impls need a uniform `S: AsRef<str>` bound and content-based comparison that derives cannot express.

## On the `TryFrom` shape (a coherence note)

Validation is exposed as a real `TryFrom<&str> for CommandString`, which required **removing the infallible `From<&str> for CommandString`** (so `"x".into()` no longer produces a `CommandString` — use `CommandString::try_from("x")?`).

A literal `TryFrom<&str> for CommandStr` (borrowed) is **not possible**: it collides (E0119) with core's blanket `impl<T, U: Into<T>> TryFrom<U> for T`, because the reflexive `From<S> for Command<S>` — which is the fundamental generic constructor and can't be removed — already provides `&str: Into<CommandStr>`. So validated construction targets the owned `CommandString`; borrow it back down with `.as_command_str()` if you need a `CommandStr`.

Because the reflexive `From`/`Command::new` stay infallible, **validation is opt-in at the raw-`&str` boundary, not an airtight gate** — `CommandString::new(s)` and `String::into()` still bypass it. Enforcing the invariant unconditionally would mean removing those infallible constructors, a much larger design change left out of scope here.

## Not in scope

The type is **not yet adopted** anywhere — `History::command` and all other call sites are untouched. Adoption is separate follow-up work.

## Testing

- 20 inline unit tests + 1 doctest, covering multi-storage wrapping, cross-storage equality/ordering/hashing, `HashMap` lookup by `&str`, the `From` lattice (with zero-copy `Cow` preservation), `Display` width/alignment, transparent + zero-copy-borrowed serde, and validated construction (accepts newlines/tabs/empty, rejects NUL and reports its byte index).
- Full workspace green: `cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings -D clippy::redundant_clone`, `cargo nextest run -p atuin-common` (32 passing), `cargo test --doc -p atuin-common`, and a `--locked` build.

## Follow-ups to weigh at adoption time

- `impl PartialEq<str>`/`PartialEq<&str>` so `cmd == "literal"` compiles directly (today: `cmd.as_str() == "..."`).
- `Into<String>` for smoother interop with `String`-taking APIs (today: `.into_inner()`).
- If the NUL invariant should be *enforced* rather than opt-in, revisit whether the infallible `Command::new`/reflexive `From` should remain.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -397,6 +397,7 @@ dependencies = [
  "rustls",
  "semver",
  "serde",
+ "serde_json",
  "sqlx",
  "sysinfo",
  "thiserror 2.0.18",
--- a/crates/atuin-common/Cargo.toml
+++ b/crates/atuin-common/Cargo.toml
@@ -36,3 +36,4 @@ tracing-subscriber = { workspace = true, optional = true }
 
 [dev-dependencies]
 pretty_assertions = { workspace = true }
+serde_json = { workspace = true }
--- a/crates/atuin-common/src/command_str.rs
+++ b/crates/atuin-common/src/command_str.rs
@@ -13,6 +13,24 @@
 //! extra code: `Command<Arc<str>>` is a command too.
 //!
 //! The wrapper adds no behaviour to the string it holds.
+//!
+//! ```
+//! use std::borrow::Cow;
+//!
+//! use atuin_common::command_str::{CommandCow, CommandStr, CommandString};
+//!
+//! let borrowed: CommandStr<'_> = CommandStr::new("cargo test");
+//! let owned: CommandString = borrowed.to_command_string();
+//! let cow: CommandCow<'_> = CommandCow::new(Cow::Borrowed("cargo test"));
+//!
+//! // The specialisations compare by command text, whatever they are stored in.
+//! assert_eq!(borrowed, owned);
+//! assert_eq!(owned, cow);
+//!
+//! // Borrow any of them back down to a `CommandStr`, or read the text directly.
+//! assert_eq!(owned.as_command_str(), borrowed);
+//! assert_eq!(cow.as_str(), "cargo test");
+//! ```
 
 use std::{
     borrow::{Borrow, Cow},
--- a/crates/atuin-common/src/command_str.rs
+++ b/crates/atuin-common/src/command_str.rs
@@ -12,7 +12,9 @@
 //! Any other storage that is `AsRef<str>` (`Arc<str>`, `Box<str>`, ...) works with no
 //! extra code: `Command<Arc<str>>` is a command too.
 //!
-//! The wrapper adds no behaviour to the string it holds.
+//! The wrapper stores its string verbatim. The one structural invariant it can
+//! check is that a command contains no NUL byte — use [`CommandString::try_from`]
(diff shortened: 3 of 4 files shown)


markovejnovic:
Closing this. As per my comment in https://github.com/atuinsh/atuin/issues/3589#issuecomment-4975523991, I think a better approach here would be _not_ to filter out commands, but to rather render dirty characters with some sort of human-readable replacements.

PR #3628, opened 2026-07-14
feat(daemon)!: ATU-538 — daemon-only client (daemonless code removed)

markovejnovic:
## What

ATU-538: make the daemon the sole owner of local storage and **completely remove** the daemonless code paths. Investigation in `docs/dev/atu-538-ephemeral-daemon-investigation.md`.

## Result: complete

The CLI is now a pure daemon client — **zero direct storage access, zero daemonless code**. There is no longer any way to build a daemonless variant.

**History `Database` axis** — all reads (search, stats, wrapped, history subcommands, scripts, import, mcp, ai TUI) and writes (start/end) go through the daemon. Linchpin: blanket `impl Database for Box<dyn Database>`.

**Record `Store` axis** — `Store` made object-safe; the five typed stores hold `Arc<dyn Store>`; new `StorageStore` gRPC service + `DaemonStore` proxy (encrypted-blob model, crypto stays client-side). All record-store commands (sync, account/login rekey, kv, dotfiles, store rebuild/rekey/verify/purge/push/pull, scripts, wrapped, init) routed through it.

**Daemonless code deleted (Part B)** — `atuin-daemon` is now a mandatory dependency; the `daemon` cargo feature is removed from `atuin`/`atuin-client`/`atuin-ai`. ~45 `#[cfg(feature="daemon")]` gates un-gated, all `#[cfg(not(feature="daemon"))]` arms deleted (`handle_start`/`handle_end`, no-daemon branches and fallbacks, the `DaemonFuzzy` fallback, two tests).

The `daemon` subcommand keeps the real `Sqlite`/`SqliteStore` (it *is* the owner). `settings.daemon.enabled`/`autostart` are now vestigial (removable in a follow-up).

## Verification

`default` and `--no-default-features` both build (the latter now includes the daemon); `clippy` clean; 224 bin tests + workspace lib tests pass; smoke-tested end-to-end (history write/search, stats, kv set/get, dotfiles, store status, daemon lifecycle) — all served by one spawn-on-demand daemon.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -426,6 +426,7 @@ dependencies = [
  "prost-types",
  "protox",
  "rand 0.8.5",
+ "sqlx",
  "tempfile",
  "time",
  "tokio",
--- a/crates/atuin-ai/Cargo.toml
+++ b/crates/atuin-ai/Cargo.toml
@@ -14,7 +14,6 @@ repository = { workspace = true }
 
 [features]
 default = []
-daemon = []
 tree-sitter = ["dep:tree-sitter-lib", "dep:tree-sitter-bash", "dep:tree-sitter-fish"]
 
 [dependencies]
--- a/crates/atuin-ai/src/commands/inline.rs
+++ b/crates/atuin-ai/src/commands/inline.rs
@@ -1,5 +1,3 @@
-#[cfg(not(feature = "daemon"))]
-use std::path::PathBuf;
 use std::sync::mpsc;
 
 use crate::context::{AppContext, ClientContext};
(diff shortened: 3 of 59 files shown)


ltrzesniewski:
err... you're sure you want to do that? that would insta-kill windows support, as the daemon has not been ported there.

EDIT: oh, ok, I've taken a look at the doc, you'll apparently take care of it by spawning a daemon process. Thanks 🙏 
I initially thought about making a windows _service_ out of the daemon, but this should be a simpler solution.

markovejnovic:
Sorry! This is mostly an experimental slop PR. We're kind of spending some time to evaluate the feasability of this idea, and the PR, as-is, will definitely not merge. I'm just mostly evaluating whether it works or not.

I'll refrain from making slop PRs in the future and just do the reviews of these kinds of experimental things locally (or on a branch without a PR)! Sorry about that!

markovejnovic:
slop pr, mostly used to just evaluate the feasability of this; looks like we're not there yet

PR #3630, opened 2026-07-15
docs: advise users to pass `--locked` when installing from Cargo

taylordotfish:
The binaries published on GitHub are built using the lockfile; when installing from Cargo, we should suggest users pass `--locked` so they arrive at a similar binary.


ellie:
let's rewrite that whole section to run from the root of the repo, no cd

ellie (approved):
:shipit:

markovejnovic (approved):

PR #3636, opened 2026-07-15
feat: add `shell` field to history entries

taylordotfish:
Add a `shell` field to history entries with the shell that was used to run the command. Future PRs will make use of this data; for now, we just track it.

* Bump history format to v2. V2 is designed to allow adding fields without bumping the version.
* Add a database migration to add a `shell` column to the database
* Add `shell` fields to the relevant daemon messages
* Set `ATUIN_SHELL` when running `atuin history start` in shell hooks
* Create new `rmp` module in `atuin_client::utils`, with improved error types and support for encoding and decoding optional types and reading owned strings
* Coalesce disparate deserialization implementations for history items into a single function with a version parameter
* Add a structured `Version` type for representing a history version
* Remove unnecessary clones in `normalize_optional_field` and `handle_start` and reduce a clone in `handle_daemon_start` from the entire `History` struct to just the ID
* Simplify creation of history items, removing the need for `apply_start_metadata`
* Reduce duplication in many of the history importers
* Test deserialization of every version into every other version
* Ensure result of deserialization is as expected
* Note: there is technically a behavior change with `ATUIN_HISTORY_INTENT=<some-string> atuin history start --intent ""`; previously this would set the intent to `None`, but now it sets it to `<some-string>`. This new behavior matches the existing behavior for `--author`.


BinaryMuse (dismissed):
Looks good — I'm least confident about the changes in `crates/atuin-client/src/history.rs` but they seem okay. @ellie may have additional thoughts.

crates/atuin-client/src/history.rs
```diff
@@ -161,21 +200,26 @@ impl History {
...
+    /// Serializes a history entry in the V2 format.
+    ///
+    /// Differences from V1:
+    ///
+    /// * `intent` is always written; if `None`, nil is written to the output.
+    /// * Added new field `shell`.
+    ///
+    /// V2 is designed to allow new fields to be added without incrementing the version. V1 cannot
+    /// accommodate this because its deserialization routine errors if more than 11 fields are
+    /// provided.
```
ellie, line 216: Nice change

ellie (approved):
looks great! thank you!

taylordotfish:
This PR is the first part of #608

PR #3638, opened 2026-07-15
refactor(hook): typed protocol for agent hook events

### LINKED ISSUE #3589: [Bug]: NUL bytes in history

markovejnovic:
This PR introduces a structured type for the hook event claude/codex/pi send us. The goal of this is to enable an easier way to have custom serde on the event type.

There are functional changes:
- Historically `tool_response.exitCode: 3.5` (float instead of int) would result in an `End(exit=0)` event being record. Now it is skipped.
- Historically `tool_input.description: 42` (int instead of string) would result in an `Start(intent=None)`. Now it is skipped.

Working on a PR stacked on top of this one which fixes https://github.com/atuinsh/atuin/issues/3589 which is the original motivation for all of this anyways.


crates/atuin/src/command/client/hook.rs
```diff
@@ -325,10 +265,14 @@ fn add_hook_entries(hooks: &mut Value, agent: &Agent) -> Result<()> {
...
-                    .any(|hook| hook["command"].as_str() == Some(hook_command))
-            })
+            entry
+                .get("hooks")
+                .and_then(Value::as_array)
+                .is_some_and(|hooks| {
+                    hooks.iter().any(|hook| {
+                        hook.get("command").and_then(Value::as_str) == Some(hook_command)
+                    })
+                })
```
markovejnovic, line 275: the legacy implementation would panic! claude caught this one

crates/atuin/src/command/client/hook/event.rs
```diff
@@ -0,0 +1,387 @@
...
+    #[case::unknown_event_skipped(
+        json!({
+            "hook_event_name": "SomeFutureEvent",
+            "tool_name": "Bash",
+            "tool_input": {"command": "ls"},
+            "tool_use_id": "toolu_abc123"
+        }),
+        None
+    )]
+    // A missing event name is ignored.
```
ellie, line 257: I know I had issues reading these in your other PR, but in this instance they are much better

markovejnovic: lol im the opposite -- i find these to contain too much whitespace

[uiua my beloved](https://www.uiua.org/)

ellie (approved):
nice one! 🚀

PR #3639, opened 2026-07-15
feat(string): EllipsizeExt -- General-purpose utility for adding ellipses to strings

markovejnovic:
This PR adds an `EllipsizeExt` string extension to help commonize and remove some of the duplicated code we've had with ellipses. The main goal is to be slightly tidier, with a common helper utility that's generic, well-tested, etc.

You can now do

```rust
use EllipsizeExt as _;

fn your_fxn() {
  return "hello-world".ellipsize(Budget::Columns(5), Pos::Middle, Indicator::UNICODE); // "he…ld"
}
```

### Impact

This actually fixes a tiny bug according to claude:

>   1. Wide-character column misalignment (real, user-visible). This is the actual fix.
  The old directory/host/user renderers in the interactive search list budgeted
  truncation in chars().count(). A CJK glyph or emoji is 1 char but 2 terminal 
  columns, so a column sized for width N could render into up to 2N cells —
  overflowing its column and shoving every column to its right out of alignment
  whenever your history contained e.g. a /世界/… path or an emoji. The new code
  measures in display columns (unicode-width), so those truncate to actually fit. If
  you have any wide-char history entries, the search UI was visibly misaligned before
  and isn't now. (Two smaller correctness bumps ride along: the old host/user had an
  off-by-one that under-filled their column by one cell vs directory, now unified; and
  at very narrow widths w < 4 the old code fell through to a non-truncating format!
  that overflowed, where the new code hard-truncates.)
  
Also, it uses unicode ellipses now in the history list!

### Testing

There's a bunch of unit tests, and I also tested it manually:

```
$ ./target/debug/atuin search -i
```

Before:

image

After:

image

Looks like a nice improvement to me!

### DIFF
--- a/crates/atuin-common/src/string/ellipsis.rs
+++ b/crates/atuin-common/src/string/ellipsis.rs
@@ -46,12 +46,10 @@ impl Default for Indicator<'_> {
 /// How much room to truncate into, and the unit it is measured in.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum Budget {
-    /// A UTF-8 byte budget. Cuts on `char` boundaries, so the result is valid
-    /// UTF-8 and its byte length never exceeds the budget. Use for storage or
-    /// prompt byte budgets.
+    /// A UTF-8 byte budget.
     Bytes(usize),
-    /// A display-column budget via `unicode-width` — a double-width glyph such
-    /// as `世` or `🦀` counts as two. Use for TUI column layout.
+    /// A display-column budget via `unicode-width` - a double-width glyph such
+    /// as `世` or `🦀` counts as two. Use for presentation.
     Columns(usize),
 }
 
@@ -75,7 +73,7 @@ impl Budget {
 pub trait EllipsizeExt: AsRef<str> {
     /// Truncate this string to fit within `budget`, splicing in `indicator` on
     /// `side` if any content had to be dropped. Returns a lazy [`Ellipsized`]
-    /// view — no allocation until you ask for an owned string.
+    /// view - no allocation until you ask for an owned string.
     fn ellipsize<'a>(
         &'a self,
         budget: Budget,
@@ -88,7 +86,7 @@ pub trait EllipsizeExt: AsRef<str> {
         // The boundary helpers only need a per-grapheme cost, not the budget.
         let cost = |seg: &str| budget.cost(seg);
 
-        // Fast path: already fits — a single contiguous slice.
+        // Fast path: already fits - a single contiguous slice.
         if cost(s) <= amount {
             return Ellipsized::contiguous(s, 0);
         }
@@ -124,13 +122,11 @@ pub trait EllipsizeExt: AsRef<str> {
 
 impl<T: AsRef<str>> EllipsizeExt for T {}
 
-/// A budget-truncated view of a source string — either a single contiguous
+/// A budget-truncated view of a source string - either a single contiguous
 /// slice (it fit, or there was no room even for the indicator) or a head +
-/// indicator + tail with content elided between them. Cheap (`Copy`); `Display`
-/// writes the pieces straight to the formatter with no intermediate `String`.
+/// indicator + tail with content elided between them.
 ///
-/// The spliced form always carries an indicator, so a gap between head and tail
-/// can never be rendered without a marker — that state is unrepresentable.
+/// Cheap `Copy`. `Display` writes the pieces straight to the formatter.
 #[derive(Debug, Clone, Copy)]
 pub struct Ellipsized<'a>(Repr<'a>);
 
@@ -139,7 +135,7 @@ enum Repr<'a> {
     /// The whole result is one contiguous slice of the source; `source_offset`
     /// is the slice's byte offset in the source.
     Contiguous { text: &'a str, source_offset: usize },
-    /// A head slice, an indicator, and a tail slice — always with a marker.
+    /// A head slice, an indicator, and a tail slice - always with a marker.
     Spliced {
         head: &'a str,
         indicator: Indicator<'a>,
@@ -157,8 +153,6 @@ impl<'a> Ellipsized<'a> {
         })
     }
 
-    /// Build a spliced view from the source and the head/tail byte boundaries;
-    /// the slicing happens once here.
     fn spliced(
         source: &'a str,
         head_end: usize,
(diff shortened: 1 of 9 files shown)


crates/atuin/src/command/client/search/history_list.rs
```diff
@@ -355,53 +359,48 @@ impl DrawState<'_> {
...
-        // Truncate from the left with "..." if too long, plus trailing space
-        // Use character count for comparison and skip for UTF-8 safety
-        let display = if char_count > w && w >= 4 {
-            let truncated: String = cwd.chars().skip(char_count - (w - 3)).collect();
-            format!("...{truncated}")
+        // Elide from the left with "..." so the leaf directory stays visible;
+        // pad to the column width when it already fits.
+        let display = if cwd.width() > w {
+            cwd.ellipsize(Budget::Columns(w), Pos::Left, Indicator::ASCII)
+                .to_string()
```
markovejnovic, line 367: This is actually unnecessary -- the caller _could_ be modified to take `Display` rather than `&str`, but I want to avoid making too many paper cuts in this PR.

crates/atuin/src/command/client/search/history_list.rs
```diff
@@ -310,28 +313,27 @@ impl DrawState<'_> {
...
+            if highlighted {
                 if row_highlighted {
                     char_style = self.theme.as_style(Meaning::AlertWarn);
                 }
                 char_style.attributes.set(style::Attribute::Bold);
             }
-
-            let s = ch.to_string();
-            self.draw(&s, Style::from_crossterm(char_style));
+            self.draw(&ch.to_string(), Style::from_crossterm(char_style));
```
markovejnovic, line 336: Again, same deal here -- we actually allocate many many times, but this is a performance problem in `self.draw` not being able to accept `Display` objects, which I think it should. Will follow-up and get a nice performance speedup =)

crates/atuin-common/src/string/ellipsis.rs
```diff
@@ -0,0 +1,576 @@
...
+
+/// Which side of the string to elide when it does not fit.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Pos {
+    /// Keep the tail, elide the head: `…orld`.
+    Left,
+    /// Keep both ends, elide the middle: `he…ld`.
+    Middle,
+    /// Keep the head, elide the tail: `hel…`.
+    Right,
```
taylordotfish:
```suggestion
    Start,
    /// Keep both ends, elide the middle: `he…ld`.
    Middle,
    /// Keep the head, elide the tail: `hel…`.
    End,
```

crates/atuin-common/src/string/ellipsis.rs
```diff
@@ -0,0 +1,576 @@
...
+
+/// Byte end index of the longest prefix of `s` whose cost (in `budget`'s unit)
+/// is at most `max`.
+///
+/// Iterates whole [`units`], so it never splits one and the returned index is
+/// always a valid slice point. Summing per-unit costs matches the whole-string
+/// cost [`Budget::cost`] reports — `unicode-width`'s context effects (a
+/// variation selector or combining mark adjusting the width of the character it
+/// follows) stay within a single grapheme cluster.
+fn prefix_within(s: &str, max: usize, budget: Budget) -> usize {
```
taylordotfish: `prefix_boundary`, `suffix_boundary` instead of `_within`? At first it sounds like we're searching for a prefix “within” the string, but a prefix can only be in one place

crates/atuin-common/src/string/ellipsis.rs
```diff
@@ -0,0 +1,576 @@
...
+
+impl<'a> From<Ellipsized<'a>> for Cow<'a, str> {
+    /// Borrowed when nothing was spliced (contiguous); owned only when an
+    /// indicator sits between head and tail.
+    fn from(ellipsized: Ellipsized<'a>) -> Self {
+        match ellipsized.indicator {
+            None if ellipsized.tail_start == ellipsized.source.len() => {
+                Cow::Borrowed(ellipsized.head())
+            }
+            None => Cow::Borrowed(ellipsized.tail()),
```
taylordotfish: Might be worth noting that we never create an `Ellipsized` that has the middle part elided and a `None` indicator? It would be nice if the type enforced this somehow

crates/atuin-common/src/string/ellipsis.rs
```diff
@@ -0,0 +1,576 @@
...
+impl<T: AsRef<str>> EllipsizeExt for T {}
+
+/// A budget-truncated view of a source string: a head slice, an optional
+/// indicator, and a tail slice — none concatenated. Cheap (`Copy`); `Display`
+/// writes the pieces straight to the formatter with no intermediate `String`.
+#[derive(Debug, Clone, Copy)]
+pub struct Ellipsized<'a> {
+    source: &'a str,
+    head_end: usize,
+    tail_start: usize,
```
taylordotfish: What about making these string slices instead of offsets? It would make the display logic simpler and more obvious that we've handled all the cases, and also potentially slightly faster due to avoiding checked index operations

crates/atuin-common/src/string/ellipsis.rs
```diff
@@ -0,0 +1,555 @@
...
+    /// Borrowed for a contiguous slice; owned only when an indicator is spliced.
+    fn from(ellipsized: Ellipsized<'a>) -> Self {
+        match ellipsized.0 {
+            Repr::Contiguous { text, .. } => Cow::Borrowed(text),
+            Repr::Spliced { .. } => Cow::Owned(ellipsized.to_string()),
+        }
+    }
+}
+
+impl PartialEq<&str> for Ellipsized<'_> {
```
taylordotfish: This should be `PartialEq<str>`

crates/atuin-common/src/string/ellipsis.rs
```diff
@@ -0,0 +1,555 @@
...
+
+/// Byte end index of the longest prefix of `s` whose cost (in `budget`'s unit)
+/// is at most `max`.
+///
+/// Iterates whole grapheme clusters, so it never splits one and the returned
+/// index is always a valid slice point. Summing per-cluster costs matches the
+/// whole-string cost [`Budget::cost`] reports — `unicode-width`'s context
+/// effects (a variation selector or combining mark adjusting the width of the
+/// character it follows) stay within a single grapheme cluster.
+fn prefix_boundary(s: &str, max: usize, budget: Budget) -> usize {
```
taylordotfish: Instead of passing `max: usize` plus a separate `budget` whose value we don't care about, can we just pass a single `Budget`?

crates/atuin-common/src/string/ellipsis.rs
```diff
@@ -0,0 +1,562 @@
...
+    fn ellipsize<'a>(
+        &'a self,
+        budget: Budget,
+        side: Pos,
+        indicator: Indicator<'a>,
+    ) -> Ellipsized<'a> {
+        let s = self.as_ref();
+        let amount = budget.amount();
+        let len = s.len();
+        // The boundary helpers only need a per-grapheme cost, not the budget.
```
taylordotfish: slop comment

crates/atuin-common/src/string/ellipsis.rs
```diff
@@ -0,0 +1,562 @@
...
+    pub const UNICODE: Self = Self("…");
+
+    /// Wrap an arbitrary marker string.
+    pub const fn new(marker: &'a str) -> Self {
+        Self(marker)
+    }
+}
+
+impl Default for Indicator<'_> {
+    /// The Unicode ellipsis `…`.
```
taylordotfish, line 40:
```suggestion
```

markovejnovic: i find this to be useful

taylordotfish: it says nothing that isn't obvious from the code

ellie (approved):

crates/atuin-common/src/string/ellipsis.rs
```diff
@@ -0,0 +1,514 @@
...
+impl PartialEq<&str> for Ellipsized<'_> {
+    fn eq(&self, other: &&str) -> bool {
+        self == *other
+    }
+}
+
+/// Total byte length of the longest leading run of `graphemes` whose summed
+/// `cost` is at most `max`.
+fn fitting_bytes<'a>(
+    graphemes: impl Iterator<Item = &'a str>,
```
taylordotfish, line 245: Useless lifetime? What is `'a` doing here

markovejnovic: It communicates that the iterator's item lifetime is scoped to the function, no?

PR #3641, opened 2026-07-15
fix(daemon): index synced history directly instead of malformed SQL (#3627)

### LINKED ISSUE #3627: Daemon search index: RecordsAdded handler queries history.id with record UUIDs (malformed SQL + wrong id-space), silently indexes nothing
### Summary

In the daemon search component, the `RecordsAdded` event handler builds a SQL query that (a) is malformed and (b) queries the wrong id-space, so it almost always returns nothing. The error is swallowed by `.unwrap_or_default()`, so newly-synced history is silently **not** added to the in-memory daemon search index — it only appears after the periodic frecency rebuild or a daemon restart.

Discovered during the SQL bind-parameter audit in #3625; filing separately since it's a pre-existing functional bug, not part of that refactor.

### Location

`crates/atuin-daemon/src/components/search.rs` (~L192–206):

```rust
let histories: Vec<_> = handle
    .history_db()
    .query_history(
        format!(
            "select * from history where id in ({})",
            records
                .iter()
                .map(|record| record.0.to_string())
                .collect::<Vec<_>>()
                .join(",")
        )
        .as_str(),
    )
    .await
    .unwrap_or_default();
```

`records: &Vec<RecordId>` comes from `DaemonEvent::RecordsAdded` (`events.rs:30`), emitted with `downloaded_records` from `sync.rs:262`.

### Two independent defects

1. **Malformed SQL.** `RecordId` is a UUID newtype (`new_uuid!(RecordId)`), so `record.0.to_string()` renders the **hyphenated** form and it is interpolated **unquoted**:
   `select * from history where id in (018cd4fe-8175-7cd2-aee6-5cd7861f9c81, ...)`.
   That's not a list of string literals — SQLite parses each as an arithmetic expression over bare tokens, not a match against the text `id` column. The query errors or returns nothing.

2. **Wrong id-space.** Even with correct quoting/format, these are **record-store `RecordId`s**, but the query filters `history.id`, which holds the **`HistoryId`** of the decrypted payload (stored as `as_simple()` 32-hex text, e.g. `018cd4fe81757cd2aee65cd7861f9c81`). A record's id is not its history's id (`history/store.rs` builds each `Record` with its own `RecordId(uuid_v7())`, distinct from `History.id`). So `history.id IN (record_ids)` cannot match.

### Effect

`RecordsAdded` (fired after every sync that downloads records) indexes nothing into the daemon's in-memory search index. Freshly synced history is missing from daemon-mode search until the periodic `rebuild_frecency` pass or a restart. `.unwrap_or_default()` hides the failure entirely — no log, no error.

### Suggested fix

The handler already has the record ids; it should resolve them to the history rows those records built (the sibling `HistoryStore::incremental_build(history_db, &downloaded_records)` at `sync.rs` writes exactly those rows and knows the mapping), then index by the correct `HistoryId`. Whatever the lookup key, the ids should be bound (or normalised through `Uuid::as_simple()`) rather than interpolated raw, and the result should not be `.unwrap_or_default()`-swallowed — a failure here should at least be logged.

### Environment

Found on `main` (v18.17.1).

markovejnovic:
Fixes #3627.

## Bug

#3627 was a bug wherein `search.rs:handle_event` would run the rough equivalent of:

```rust
format!(
  "select * from history where id in ({})",
  (records: Vec<RecordId>).map(|r| r.to_string()).join(",")
)
```

There are two bugs:
- This would result, in a query such as `"select * from history where id in (019f6bee-4786-78e0-98f3-344eeb32ac92, 019f6bee-668a-79c9-a6f2-181babba68fd)"`. These record IDs are not escaped, so, effectively, this would throw a SQL exception.
- We are querying the `history` table with the `RecordId` PK. That's obviously wrong. The only reason this never failed is is because we used `.unwrap_or_default()` which would completely swallow the error.

## Fix

There are two ways to fix this problem:

- ~~Naively fix this~~:
  * Escape things correctly
  * ~~Map `RecordId` somehow to `HistoryId` and then use those to pick and prod in the database.~~
    - This doesn't work because there's no way to actually do this -- mapping `RecordId` to `History` means decrypting data.
- Implement it as I have here -- changing the emitted `DaemonEvent` to be `HistorySynced(Arc<[RecordId]>)` rather than `RecordsAdded(Vec<RecordId>)`.
  * Note the use of `Arc<[RecordId]>`. The daemon has an event bus (of which `HistorySynced` is one possible message) and calling `DaemonEvent::emit` causes a spmc broadcast. Naively using `Vec` would cause a huge amount of copies.
  * Note that `incremental_build` was already doing the `RecordId -> History` decryption and mapping, so this helps avoid the double decryption risk.
  * Note also that the payload size has significantly increased now. That `Arc<[History]>` now contains 2M+ of `History` entries (depending on how long the history is). Antagonistically, I estimate that the `History` size could be 
  ```rust
  struct History {
    command:  5985
    id:       32
    session:  32
    hostname: 256
    cwd:      256
  }
  // -------- 6330
  ```
  
  Naively, I initially had `incremental_build` return a `Vec<History>`, but I did some math: If a user is syncing 2M+ history entries, we might see a memory usage of about 12GB. Not exactly great. Looking at _my_ particular database, a sync would spike up to 1GB RAM usage. I don't have that kind of RAM to afford, so we have to do better.
    - So I stopped doing this in favor of having `incremental_build` return a stream which is lazily processed.
    - ~~The other alternative was to pass a callback into `incremental_build` that it would call as required, but I felt like a stream _felt_ like a better abstraction.~~
    
## Testing

Here's some slop: [test-behavior.sh](https://github.com/user-attachments/files/30107752/test-behavior.sh). Drop this in the root of the repo, review it  and run it. It will:

- Spin up the actual server
- Creates a client A, which records two CLI commands.
- Creates a client B which queries the server for CLI commands.

On main, the last query returns 0, on this branch, it returns 2. Slop output:

```
== preflight ==
  ✓ real history.db has no 'shell' column (unmigrated, as expected)
== scaffolding (throwaway) ==
  ✓ wrote 2 examples + dev-dependencies
== build (a few minutes on a cold cache) ==
  ✓ built
== server (real atuin server, sqlite backend) ==
  ✓ listening on 127.0.0.1:54139 (pid 29873)
== client A: record 2 commands, sync up ==
  ✓ 2 commands uploaded
== client B: THIS BRANCH ==
  ✓ db rows=2  index hits=2
== client C: CONTROL on main (this is what makes a pass mean anything) ==
  ✓ db rows=2  index hits=0
== result ==
                         db rows      index hits
  main (broken)          2            0
  this branch (fixed)    2            2

  PASS — main leaves synced history out of the index; this branch indexes it.
== cleanup ==
  ✓ stopped pid 29873
  ✓ stopped pid 29909
  ✓ stopped pid 36325
  ✓ removed scaffolding
  ✓ removed worktree
  ✓ your real history.db untouched
```


crates/atuin-daemon/src/components/search.rs
```diff
@@ -181,36 +181,17 @@ impl Component for SearchComponent {
...
+                    "Indexing synced history entries into search index"
                 );
 
-                let handle_guard = self.handle.read().await;
-                if let Some(handle) = handle_guard.as_ref() {
-                    let histories: Vec<_> = handle
-                        .history_db()
-                        .query_history(
-                            format!(
-                                "select * from history where id in ({})",
```
markovejnovic, line 196: This was one bug:

When a new record was added, we would query: `"SELECT * from history where id in (019f6845-737e-7939-99bd-15198ba992b2, 019f6846-75ef-7548-81ed-c78c90f6c483)"`

Oops, that's malformed SQL. The hyphens scare sqlite.

crates/atuin-daemon/src/components/search.rs
```diff
@@ -181,36 +181,17 @@ impl Component for SearchComponent {
...
-                                records
-                                    .iter()
-                                    .map(|record| record.0.to_string())
-                                    .collect::<Vec<_>>()
-                                    .join(",")
```
markovejnovic, line 201: The other bug here is that we're using the record ID searching in the history table. That's always empty.

crates/atuin-daemon/src/components/sync.rs
```diff
@@ -250,17 +250,20 @@ async fn do_sync_tick(
...
                 .incremental_build(handle.history_db(), &downloaded_records)
                 .await
             {
-                tracing::error!("failed to build history from downloaded records: {e}");
+                Ok(histories) => {
+                    // Emit the synced history so the search component can index it.
+                    handle.emit(DaemonEvent::HistorySynced(histories));
+                }
+                Err(e) => {
+                    tracing::error!("failed to build history from downloaded records: {e}");
```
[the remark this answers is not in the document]
markovejnovic: reviewing

BinaryMuse (approved):
Awesome :shipit:

Issue #2455, opened 2024-11-23
[Bug]: Unexpectedly getting a 404 when registering to my server


axelson:
### What did you expect to happen?

I expected that after setting the IP address of my server where atuin is running that I would be able to login and register from my clients.

### What happened?

```
> atuin register
Registering for an Atuin Sync account
Please enter username: jason
Please enter email: jason@myemail.com
Please enter password:
Error: There was an error with the atuin sync service: Status 404.
If the problem persists, contact the host

Location:
    /Users/runner/work/atuin/atuin/crates/atuin-client/src/api_client.rs:162:9
```

### Atuin doctor output

```yaml
{
  "atuin": {
    "version": "18.3.0",
    "sync": null,
    "sqlite_version": "3.46.0"
  },
  "shell": {
    "name": "fish",
    "default": "zsh",
    "plugins": [
      "atuin"
    ],
    "preexec": "built-in"
  },
  "system": {
    "os": "Darwin",
    "arch": "arm64",
    "version": "14.6.1",
    "disks": [
      {
        "name": "Macintosh HD",
        "filesystem": "apfs"
      },
      {
        "name": "Macintosh HD",
        "filesystem": "apfs"
      }
    ]
  }
}
```
```


### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

axelson:
I've tracked this down. The issue is related to configuration. Specifically:
- From https://docs.atuin.sh/self-hosting/server-setup/ I can see that atuin uses a default port of `8888`
- From the self-hosting usage page https://docs.atuin.sh/self-hosting/usage/ I followed the "config" link to https://docs.atuin.sh/configuration/config/#sync_address
- That told me "The address of the server to sync with!" and gave the example of `sync_address = "https://api.atuin.sh"`
- So I set my local server in my `~/.config/atuin/config.toml` as `sync_address = "http://192.168.1.2"`
- However then when I run `atuin register` I get the 404 mentioned above
- In troubleshooting this I tried setting `sync_address = "http://192.168.1.2:8888"` and then registering and syncing worked

So I think the bug is either:
- The docs should be updated to include the port number (specifically the https://docs.atuin.sh/configuration/config/#sync_address page)
- The default port is incorrect/doesn't match between server and client (it's not clear what default port my client is using)
  - But after using `strace` it appears that the default client port is actually `80`, not `8888`

Additionally I think it would be helpful if:
[...]

Issue #1099, opened 2023-07-12
Self-hosting atuin with apache as reverse proxy


Flowdalic:
This issue has two concerns: 1. It is about self-hosting autin with  apache as reverse proxy 2. it is about "Error: error decoding response body: expected value at line 1 column 1" when trying to login

I am not sure if 2. is caused by me misconfiguring 1., but in any case, the error message in 2. could maybe improved. In fact I am stuck with 1. and miss general tooling, command line switches and techniques to diagnose the root cause. Of course, please do not rule out that I am just incompetent.

So here is what I did

1. Atuin server, configured to listen on localhost:8888. The server seems to be running fine and has access to postgres
2. Apache virtual host, with

```
ProxyPass "/atuin" "http://localhost:8888/"
ProxyPassReverse "/atuin" "http://localhost:8888/"
```

in its config. And, in fact, accessing https://mydomain.org/atuin, returns

```json
{"homage":"\"Through the fathomless deeps of space swims the star turtle Great A'Tuin, bearing on its back the four giant elephants who carry on their shoulders the mass of the Discworld.\" -- Sir Terry Pratchett","version":"15.0.0"}
```

But I am not sure if this is supposed to happen. I only suspect that this is the atuin server talking because the version number. I am also not sure if I should set `path = "/autin"` in atuin's `server.toml`.

And upon login, I get
```bash
$ atuin login                                                                                                                                                                                                                                           
Please enter username: flow
Please enter encryption key [blank to use existing key file]:    
Please enter password: 
Error: error decoding response body: expected value at line 1 column 1

Caused by:
    expected value at line 1 column 1

Location:
    /data-scratch/var-tmp/portage/app-shells/atuin-15.0.0/work/atuin-15.0.0/atuin-client/src/api_client.rs:83:21
```

I would appreciate any help setting up my atuin server. Thanks in advance. :)

Flowdalic:
Same behavior with atuin 16.0. Any help would be appreciated.

forquare:
It's been a long while since I used Apache, but I got this working using nginx today with the following configuration:

```
server {
	listen 443 ssl;

	ssl_certificate /usr/local/etc/ssl/domain/fullchain.pem;
	ssl_certificate_key /usr/local/etc/ssl/domain/privkey.pem;

	server_name atuin.domain.co.uk;

	location / {
		proxy_pass http://172.16.0.5:8888/;
		proxy_set_header Host $host;
[...]

Issue #190, opened 2021-09-27
new user registration failed


kaosbeat:
Tried the sync server, it fails for me

```
atuin register -u <I did insert a username here>  -e <i did insert my email here>  -p <not posting any passwords on git but it was here>
Error: failed to register user

Location:
    atuin-client/src/api_client.rs:56:20
```
when I try again I get
```
atuin register -u kaotec -e kasper.jordaens@gmail.com -p Mypassword
Error: username already in use

Location:
    atuin-client/src/api_client.rs:44:20

```
but when I try to sync

```

~$ atuin sync
Error: failed to get count (are you logged in?)

Location:
    /home/runner/work/atuin/atuin/atuin-client/src/api_client.rs:103:24

```

If I try to login supplying the key I got from `atuin key` I get `Error: invalid login details`

could it have been "special characters" in the password?

ellie:
I've edited your password out of your comment, but I think because it ends with a ! and you didn't wrap it with ''

conradludgate:
> I've edited your password out of your comment

Editing comments on github preserves history :grimacing:

conradludgate:
On this. I think it might be wise to read the password from stdin rather than from args. Someone suggested it before and it would solve this kind of problem

kaosbeat:
so the problem probably was my password contained a "!" and now I cannot register a new name with the same email address I guess?

ellie:
> so the problem probably was my password contained a "!" and now I cannot register a new name with the same email address I guess?

Well, it's more an issue with how shells work, special characters have to be either properly escaped or in quotes 

I'm on holiday at the moment but I can sort the account when I'm home

Alternatively:

1) put a plus in the email "foo+bar@baz.com"
2) use another email
3) run your own server

kaosbeat:
allright! 

1. was good enough as a workaround
Thanks

tfriedel:
I had the same error when trying to use a custom server. The reason it failed was because the URL in the client config had a trailing '/' (i.e. http://myserver.com:8888/ ). After removing the trailing slash registering a user worked. I'm writing this in case someone else gets stuck with this error.

conradludgate:
@tfriedel do you mind creating a new issue for this? I believe this is something we should be able to fix

PR #3644, opened 2026-07-15
refactor: use typed Url instead of hand-built string URLs

markovejnovic:
This PR replaces a large number of `String`, `&str` usages we've had in the codebase with `Url` and `&Url`.

This is motivated by a number of bugs that have been introduced due to our relatively loose use of URLs:

- #389
- #190
- #2455
- #1099

The following new user-facing features are supported:
- **Usernames can now safely contain special characters** (and will be gracefully URL-encoded)
- **Host names can now contain weird characters** and nothing will break
- **Query parameters are safely encoded** in `crates/atuin-client/src/api_client.rs`.
- **Ports and trailing slashes** no longer misclassify in `crates/atuin-client/src/settings.rs:is_official_address`.
- **Malformed URLs in the config abort early** -- we fail on bad configs.
- **Tiny breaking change**: `atuin sync status` now prints `https://api.atuin.sh/` (with the trailing slash, this catch courtesy of Claude)

Internally, this creates a new useful `UrlAppendExt` trait extension which allows you to quite gracefully:

```rust
use atuin_common::url::UrlAppendExt;

let base = url::Url::parse("https://host.example/atuin").unwrap();

assert_eq!(base.append(["foo/bar"])?.as_str(), "https://host.example/atuin/foo%2Fbar");
assert_eq!(base.append(["foo", "bar"])?.as_str(), "https://host.example/atuin/foo/bar");

// Url::join behavior:
assert_eq!(base.join("foo/bar").unwrap().as_str(), "https://host.example/foo/bar");

let base = url::Url::parse("https://host.example/atuin").unwrap();
assert_eq!(
    base.append_path("api/v0/me")?.as_str(),
    "https://host.example/atuin/api/v0/me",
);
```

Which handles the legacy `make_url` functions we had repeated and sprinkled around.

### DIFF
--- a/crates/atuin-client/src/settings.rs
+++ b/crates/atuin-client/src/settings.rs
@@ -12,9 +12,16 @@ use regex::RegexSet;
 use semver::Version;
 use serde::{Deserialize, Serialize};
 use serde_with::DeserializeFromStr;
-use std::{collections::HashMap, io::prelude::*, path::PathBuf, str::FromStr, sync::OnceLock};
+use std::{
+    collections::HashMap,
+    io::prelude::*,
+    path::PathBuf,
+    str::FromStr,
+    sync::{LazyLock, OnceLock},
+};
 use time::{OffsetDateTime, UtcOffset, format_description::FormatItem, macros::format_description};
 use tokio::sync::OnceCell;
+use url::Url;
 
 static EXAMPLE_CONFIG: &str = include_str!("../config.toml");
 
@@ -28,21 +35,13 @@ pub(crate) mod meta;
 mod scripts;
 pub mod watcher;
 
-#[derive(derive_more::AsRef)]
-#[as_ref(str)]
-pub struct HubEndpoint(String);
+/// Default sync address for Atuin's hosted service, parsed once.
+pub static DEFAULT_SYNC_URL: LazyLock<Url> =
+    LazyLock::new(|| Url::parse("https://api.atuin.sh").expect("default sync address is valid"));
 
-/// Default sync address for Atuin's hosted service
-pub const DEFAULT_SYNC_ADDRESS: &str = "https://api.atuin.sh";
-
-/// Default Hub web/API endpoint for Atuin's hosted service
-pub const DEFAULT_HUB_ENDPOINT: &str = "https://hub.atuin.sh";
-
-impl Default for HubEndpoint {
-    fn default() -> Self {
-        HubEndpoint(DEFAULT_HUB_ENDPOINT.to_string())
-    }
-}
+/// Default Hub web/API endpoint for Atuin's hosted service, parsed once.
+pub static DEFAULT_HUB_URL: LazyLock<Url> =
+    LazyLock::new(|| Url::parse("https://hub.atuin.sh").expect("default hub endpoint is valid"));
 
 #[derive(Clone, Debug, Deserialize, Copy, ValueEnum, PartialEq, Serialize)]
 pub enum SearchMode {
@@ -649,7 +648,7 @@ pub struct Ai {
 
     /// The address of the Atuin AI endpoint. Used for AI features like command generation.
     /// Only necessary for custom AI endpoints.
-    pub endpoint: Option<String>,
+    pub endpoint: Option<Url>,
 
     /// How to talk to `endpoint`. See [`AiEndpointProtocol`].
     #[serde(default)]
@@ -1001,7 +1000,7 @@ pub struct Settings {
     pub update_check: bool,
 
     /// The sync address for atuin.
-    pub sync_address: String,
+    pub sync_address: Url,
 
     /// Sync protocol for authentication. When set to "auto" (default), the protocol
     /// is inferred from sync_address. Set to "hub" to force Hub auth with a custom
@@ -1191,16 +1190,10 @@ impl Settings {
         }
     }
 
-    /// Normalize a URL for comparison by trimming trailing slashes
-    fn normalize_url(url: &str) -> &str {
-        url.trim_end_matches('/')
-    }
-
-    /// Check if a URL matches one of Atuin's official hosted addresses
-    fn is_official_address(url: &str) -> bool {
-        let normalized = Self::normalize_url(url);
-        normalized == Self::normalize_url(DEFAULT_SYNC_ADDRESS)
-            || normalized == Self::normalize_url(DEFAULT_HUB_ENDPOINT)
+    /// Check if a URL matches one of Atuin's official hosted addresses.
+    fn is_official_address(url: &Url) -> bool {
+        let origin = url.origin();
+        origin == DEFAULT_SYNC_URL.origin() || origin == DEFAULT_HUB_URL.origin()
     }
 
     /// Returns whether this configuration uses Hub-style sync.
@@ -1224,7 +1217,7 @@ impl Settings {
     /// `endpoint` is the resolved AI endpoint — the `--api-endpoint` flag or
     /// `ai.endpoint` setting, after defaults are applied — which is why it's a
     /// parameter rather than read from `self.ai.endpoint`.
-    pub fn is_hub_ai_endpoint(&self, endpoint: &str) -> bool {
+    pub fn is_hub_ai_endpoint(&self, endpoint: &Url) -> bool {
         match self.ai.endpoint_protocol {
             AiEndpointProtocol::Hub => true,
             AiEndpointProtocol::Oss => false,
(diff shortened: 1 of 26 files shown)


crates/atuin-common/src/url.rs
```diff
@@ -0,0 +1,159 @@
...
+    where
+        I: IntoIterator<Item = S>,
+        S: AsRef<str>,
+    {
+        let mut url = self.clone();
+        url.path_segments_mut()
+            .map_err(|()| NonSplittableUrlError)?
+            // A base like `https://h.example/atuin/` ends in an empty segment. Drop it.
+            .pop_if_empty()
+            .extend(segments);
```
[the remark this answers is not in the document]
markovejnovic: This is true, but I don't think it is the responsibility of this piece of code to validate that the username isn't `.` nor `..`. Additionally, I don't think this is a very common/interesting username, but I'll check the code anyways.

crates/atuin-client/src/settings.rs
```diff
@@ -1191,16 +1190,10 @@ impl Settings {
...
-    /// Check if a URL matches one of Atuin's official hosted addresses
-    fn is_official_address(url: &str) -> bool {
-        let normalized = Self::normalize_url(url);
-        normalized == Self::normalize_url(DEFAULT_SYNC_ADDRESS)
-            || normalized == Self::normalize_url(DEFAULT_HUB_ENDPOINT)
+    /// Check if a URL matches one of Atuin's official hosted addresses.
+    fn is_official_address(url: &Url) -> bool {
+        let origin = url.origin();
+        origin == DEFAULT_SYNC_URL.origin() || origin == DEFAULT_HUB_URL.origin()
     }
```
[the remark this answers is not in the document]
markovejnovic: I think it makes sense to handle it this way. In all existing use-cases, official atuin addresses are considered as _official_, but now this function also supports Urls which have some paths/args/whatever associated with them.

BinaryMuse (approved):
Heck yeah :shipit: 

Thank you!

crates/atuin-common/src/url.rs
```diff
@@ -0,0 +1,232 @@
+//! General-purpose URL utilities.
+
+use ::url::Url;
```
taylordotfish: very minor but is the leading `::` necessary here? this module doesn't contain an item called `url`

crates/atuin-common/src/url.rs
```diff
@@ -0,0 +1,232 @@
...
+                path.push(segment);
+            }
+        }
+
+        Ok(url)
+    }
+
+    fn append_path(&self, path: &'static str) -> Result<Url, UrlAppendError> {
+        self.append(path.split('/').filter(|s| !s.is_empty()))
+    }
```
taylordotfish: can we move this to be a default method impl on the trait itself rather than this impl? if we ever impl UrlAppendExt for another type we'll get append_path for free

crates/atuin-common/src/url.rs
```diff
@@ -0,0 +1,232 @@
...
+                path.pop_if_empty();
+            }
+
+            for segment in segments {
+                let segment = segment.as_ref();
+                if matches!(segment, "." | "..") {
+                    return Err(UrlAppendError::DotSegment(segment.to_owned()));
+                }
+                path.push(segment);
+            }
```
taylordotfish: why is this in a block?

crates/atuin-common/src/url.rs
```diff
@@ -0,0 +1,232 @@
...
+#[derive(Debug, Clone, PartialEq, Eq, Error)]
+pub enum UrlAppendError {
+    /// The URL given cannot be a base URL (e.g. `mailto:` or `data:`).
+    ///
+    /// These are usually URLs which do not support file-path-like semantics.
+    #[error("URL cannot be a base: it has no hierarchical path to append segments to")]
+    NonSplittable,
+
+    #[error("segment {0:?} is a dot segment, which cannot be a literal path segment")]
+    DotSegment(String),
```
taylordotfish: it feels a little unbalanced that in the case of `DotSegment` the error tells us what the exact bad segment was, but for `NonSplittable` it doesn't tell us what the bad base url was.

I would either make `DotSegment` a unit variant and have the error message be `path segments cannot be . or ..` (which would avoid an allocation, albeit in an error path) or make `NonSplittable` contain the invalid base url

crates/atuin-common/src/url.rs
markovejnovic:
```
error[E0505]: cannot move out of `url` because it is borrowed
  --> crates/atuin-common/src/url.rs:97:12
   |
80 |         let mut url = self.clone();
   |             ------- binding `url` declared here
81 |         let mut path = url
   |                        --- borrow of `url` occurs here
...
97 |         Ok(url)
   |            ^^^ move out of `url` occurs here
98 |     }
   |     - borrow might be used here, when `path` is dropped and runs the `Drop` code for type `PathSegmentsMut`
```

taylordotfish: oh can we try just inserting `drop(path)` above `Ok(url)` instead of the block?

markovejnovic:
```
marko@bee:~/Desktop/atuin-5$ cargo build -p atuin-common --features unicode
   Compiling atuin-common v18.17.1 (/Users/marko/Desktop/atuin-5/crates/atuin-common)
error[E0505]: cannot move out of `url` because it is borrowed
   --> crates/atuin-common/src/url.rs:98:14
    |
 80 |         let mut url = self.clone();
    |             ------- binding `url` declared here
 81 |
 82 |         let mut path = url
    |                        --- borrow of `url` occurs here
...
 98 |         drop(url);
    |              ^^^ move out of `url` occurs here
 99 |         Ok(url)
100 |     }
    |     - borrow might be used here, when `path` is dropped and runs the `Drop` code for type `PathSegmentsMut`

error[E0505]: cannot move out of `url` because it is borrowed
   --> crates/atuin-common/src/url.rs:99:12
    |
 80 |         let mut url = self.clone();
    |             ------- binding `url` declared here
 81 |
 82 |         let mut path = url
    |                        --- borrow of `url` occurs here
...
 99 |         Ok(url)
    |            ^^^ move out of `url` occurs here
100 |     }
    |     - borrow might be used here, when `path` is dropped and runs the `Drop` code for type `PathSegmentsMut`

error[E0382]: use of moved value: `url`
  --> crates/atuin-common/src/url.rs:99:12
   |
80 |         let mut url = self.clone();
   |             ------- move occurs because `url` has type `Url`, which does not implement the `Copy` trait
...
98 |         drop(url);
   |              --- value moved here
99 |         Ok(url)
   |            ^^^ value used here after move
   |
help: consider cloning the value if the performance cost is acceptable
   |
98 |         drop(url.clone());
   |                 ++++++++

Some errors have detailed explanations: E0382, E0505.
For more information about an error, try `rustc --explain E0382`.
error: could not compile `atuin-common` (lib) due to 3 previous errors
```

Not sure if this is worth it, I kind of don't really care about the block

taylordotfish: you didn't try what I said

markovejnovic: ok let me give it an earnest shot

markovejnovic: it does work, but i hate it

markovejnovic: i prefer the scoped block frankly. this `drop` feels comparatively procedural

taylordotfish: the only thing the block is doing is forcing `path` to be dropped before returning `url`. I'd rather have that be explicit with a call to `drop` which also reduces nesting

markovejnovic: yeah but scope > `drop`

markovejnovic: `drop` reminds me of `defer please_delete`

markovejnovic: or `free(some_ptr)`

taylordotfish: I won't block this pr on getting rid of the block but if you want to keep it, could you add a comment explaining why the block is necessary?

markovejnovic: lexical scoping is a little bit more intuitive to grasp imo

markovejnovic: i added it

crates/atuin-common/src/url.rs
```diff
@@ -0,0 +1,229 @@
+//! General-purpose URL utilities.
+
+use thiserror::Error;
+use url::Url;
+
+#[derive(Debug, Clone, PartialEq, Eq, Error)]
```
taylordotfish:
```suggestion
#[derive(Debug, Clone, Copy, PartialEq, Eq, Error)]
```

taylordotfish (approved):

PR #3645, opened 2026-07-16
fix(init): fix include guard in Bash init script; chore(init): improve inclusion of shell init scripts in Rust

taylordotfish:
* Ensure the entire Bash init script is covered by the include guard. `client/init/bash.rs` outputs code dynamically but it wasn't covered by the include guard.
* Add `atuin::shell` module with all init scripts instead of calling `include_str!` from files in `command/client/init`
* Strip trailing whitespace from the shell init scripts; they are all printed using `println!` which will add a newline


markovejnovic (approved):
talked in person and i would prefer if these were jinja-like templates since it would help prevent this whole class of bugs, but for now lgtm

PR #3646, opened 2026-07-16
fix(theme): fix low-contrast annotation color in default and autumn themes

### LINKED ISSUE #3539: [Bug]: directory and user columns invisible with default and marine themes

I want to configure the history popup so that it shows the directory -- but nothing is shown. Same for the user and host columns. So I have this in config.toml 

columns = ["duration", "time", "directory", "user", "host", { type = "command", expand = false }] 

I expected to see all those columns.

I did *not* -- but should have -- seen a header at the top of the history with this text:

```
 Atuin v18.16.1     <esc>: exit, <tab>: edit, <enter>: edit, <ctrl-o>: insphistory count: 1448
```

### What happened?

Whatl I see in the history popup is:

    147ms   13h ago                                 the_host      ls bin
    178ms   13h ago                                 the_host      cd ..
    119ms   13h ago                                 the_host      echo $PATH
  

and with no header line like the above.

I thought those were just missing, but when I select the text, I see that the text is there! It's just invisible in my terminal. 

If I change the theme to "marine", I can see the text (and the header), but the default and autumn themes show invisible text.

I actually didn't even realize that there was a header because it was always invisible before now.  I found atuin rather hard to use because the keybindings were mysterious. But they were being shown the whole time!

I'm using the `foot` terminal with the sway tiling desktop. I think I have a dark solarized theme. 

Can those themes get updated to be visible? Or is this some idiosyncratic configuration on my end?

### Atuin doctor output

```yaml
Atuin Doctor
Checking for diagnostics


{
  "atuin": {
    "version": "18.16.1",
    "commit": "671f96b60dac49d1d2de73cc0812986a5e22ce7b",
    "sync": null,
    "sqlite_version": "3.46.0"
  },
  "shell": {
    "name": "zsh",
    "default": "zsh",
    "plugins": [
      "atuin"
    ],
    "preexec": "built-in"
  },
  "system": {
    "os": "Ubuntu",
    "arch": "x86_64",
    "version": "26.04",
    "disks": [
      {
        "name": "/dev/sda2",
        "filesystem": "ext4"
      },
      {
        "name": "/dev/sda3",
        "filesystem": "ext4"
      },
      {
        "name": "/dev/sda1",
        "filesystem": "vfat"
      },
    ]
  }
}
```

### Code of Conduct

hexonal:
Fixes #3539

The `default` and `autumn` themes both set `Meaning::Annotation` to `Color::DarkGrey` in `crates/atuin-client/src/theme.rs`. This meaning styles the directory/user/host columns and header in the history TUI popup (`history_list.rs`, `interactive.rs`, `inspector.rs`), and on many dark-background terminals `DarkGrey` renders as nearly the same color as the background, making that text unreadable.

The `marine` theme doesn't have this problem because it already uses `"steelblue"` for `Annotation`. This PR changes `default` and `autumn` to use the same `"steelblue"` value, which keeps reasonable contrast on both light and dark terminal backgrounds.

Added a test asserting `default`, `autumn`, and `marine` all resolve `Annotation` to the same color, so a future revert to `DarkGrey` would fail CI.


hexonal:
Rebased onto current main — this had gone into conflict.

The production change was untouched; the collision was in the test module. `test: adopt rstest across the workspace test suite` (#3785) and `test: parametrize tests with rstest across 8 crates` (#3653) converted the surrounding tests to `#[rstest]` with a `manager` fixture, and my new test sat right on that boundary. It now uses the same fixture as its neighbours rather than constructing a `ThemeManager` itself:

```rust
#[rstest]
fn test_annotation_has_contrast_in_builtin_themes(mut manager: ThemeManager) {
    for name in ["default", "autumn", "marine"] {
```

Re-verified after the rebase: `cargo test -p atuin-client --lib theme_tests` gives 25/25, `cargo fmt --check` and `cargo clippy -p atuin-client --lib --all-features` are both clean, and reverting just the two `Annotation` colour lines makes the new test fail — so it's still pinning the actual behaviour rather than passing incidentally.


---

On the red `CodSpeed Performance Analysis` — that isn't this change. All four flagged regressions are in `crates/atuin-nucleo/matcher/benches/fuzzy_match.rs` (`match_list[...]`), a different crate with no call path from a theme colour constant; the diff here only touches `crates/atuin-client/src/theme.rs`. The one benchmark it reports as *improved* (+14.74%) is a third unrelated place, `atuin-client/benches/record/sqlite_store.rs::push_batch[10]`.

CodSpeed flags the reason itself in the same report: "Unknown Walltime execution environment detected — Using the Walltime instrument on standard Hosted Runners will lead to inconsistent data." Consistent with that, the check currently fails on several `atuinsh`-owned PRs too (e.g. #3798, #3796) while passing on others, so it isn't a fork-permissions thing either. Happy to acknowledge it on the CodSpeed dashboard if that's the usual handling here.

philtweir:
Thanks for the theme work @hexonal !

hexonal:
Thanks @philtweir!

On the red CodSpeed check, in case it holds this up: I do not think it is measuring anything real here.

All four regressed benchmarks are `match_list[*]` in `crates/atuin-nucleo/matcher/benches/fuzzy_match.rs`. This PR touches one file, `crates/atuin-client/src/theme.rs`, and does not touch `atuin-nucleo` at all — the diff is a colour constant plus a test.

Two more signs it is noise rather than signal: an unrelated benchmark, `push_batch[10]` in `atuin-client`'s sqlite store, is reported as 14.74% *faster* on the same run, and CodSpeed itself flags the run with "Unknown Walltime execution environment detected — using the Walltime instrument on standard Hosted Runners will lead to inconsistent data."

Happy to acknowledge it on CodSpeed if that is the convention here, or to rebase onto current main for a fresh measurement — whichever you prefer.

hexonal:
Rebased onto current main — the branch was 35 commits behind, and the CodSpeed comparison above was against `c49892e`, which is now well out of date.

The production diff is unchanged (still the two `Annotation` colour lines plus the test); none of those 35 commits touch `crates/atuin-client/src/theme.rs`, so the rebase was a clean replay. Re-verified locally afterwards: `cargo fmt --check` clean and `cargo test -p atuin-client --lib theme_tests` gives 25/25, including `test_annotation_has_contrast_in_builtin_themes`.

On the red CodSpeed check I raised earlier — it resolved itself, and I'd rather flag why than let it look like I worked around it. #3812 (`ci: remove codspeed workflow`, 30 Jul) deleted `.github/workflows/codspeed.yml` outright. That commit wasn't in my old base, which is why the check kept reporting; it is now, so the check no longer runs. So this wasn't a real regression that got fixed — the benchmark job simply isn't part of CI anymore.

All 24 checks are green and the PR reports `mergeable` / `CLEAN`. Nothing outstanding from my side.

PR #3647, opened 2026-07-16
feat: add hickory-dns resolver for musl target to improve DNS resolution

sebdanielsson:
## Checks

Feel free to close this PR if you would prefer enterprise users to use the glibc build instead, but might be worth documenting somewhere in the docs.

## Environment

I work in an enterprise environment with an always on VPN, internal root certs, and HTTP proxy that does SSL incerception. Client is Windows 11 laptop + WSL2 running Ubuntu 26.04. I installed Atuin using mise, which pulls down the musl build by default.

## The problem

I tried to setup a personal sync server internally but couldn't login:

```sh
Error: error sending request for url (https://atuin.company.local/login)

Caused by:
   0: client error (Connect)
   1: dns error
   2: failed to lookup address information: Name does not resolve

Location:
    crates/atuin-client/src/api_client.rs:126:16
```

## Root cause

I let Claude do some digging in the background and it quickly found the issue:

> The musl release binaries fail to resolve hostnames when `resolv.conf` lists nameservers that don't serve the same zones. This is common in enterprise networks with split DNS, and especially on WSL2, which generates `resolv.conf` from all Windows network adapters. A typical result is two corporate nameservers plus a home router as the third.

> musl's resolver sends the query to all nameservers at once and takes the first answer it gets, including NXDOMAIN. A router that answers NXDOMAIN for an internal name in 2 ms always beats the corporate server that answers correctly in 14 ms, so internal hostnames fail with "Name does not resolve" every time. glibc tries nameservers in order, which is why the gnu binary works on the same machine. Pinning the name in `/etc/hosts` doesn't help either: musl matches hosts entries case sensitively, and reqwest lowercases the host, so the mixed-case entries WSL generates never match. Other static musl tools (uv for example) show the same failure.

This is actually documented on the musl wiki: https://wiki.musl-libc.org/functional-differences-from-glibc.html

## The fix

Enable reqwest's hickory-dns feature for musl targets only. Hickory reads `resolv.conf` and `/etc/hosts` like glibc does.

## Testing

When using the hickory-dns build I can now sign in without any issues.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1115,6 +1115,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "19d374276b40fb8bbdee95aef7c7fa6b5316ec764510eb64b8dd0e2ed0d7e7f5"
 
 [[package]]
+name = "critical-section"
+version = "1.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "790eea4361631c5e7d22598ecd5723ff611904e3344ce8720784c93e3d83d40b"
+
+[[package]]
 name = "crossbeam-channel"
 version = "0.5.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


ellie (approved):
thanks for the fix!

PR #3650, opened 2026-07-17
feat(bash): bundle bash-preexec with Atuin; automatically load if no other preexec backend is loaded

taylordotfish:
This PR bundles [bash-preexec](https://github.com/rcaloras/bash-preexec) with Atuin and updates `atuin init bash` to automatically load bash-preexec if no other preexec backend (ble.sh or an external copy of bash-preexec) has been loaded.

This behavior can be overwritten by setting `ATUIN_NO_BUILTIN_PREEXEC=1`; ideally this is passed to `atuin init` directly:

```bash
eval "$(ATUIN_NO_BUILTIN_PREEXEC=1 atuin init bash)"
```

since then we will avoid emitting an unused function containing the contents of bash-preexec, but we also check the variable dynamically in the init script, so if a user caches the result of `atuin init bash` and then later sets the override variable, it will still take effect.

One thing to note: if you initialize a preexec backend in your .bashrc *after* Atuin, it will not be detected, and Atuin will load the builtin bash-preexec. In the case of another copy of bash-preexec, this is not an issue, as it contains its own include guard. In the case of ble.sh, I tested this locally and did not notice any issues; however, I am not a regular ble.sh user.

### DIFF
--- a/vendor/README.md
+++ b/vendor/README.md
@@ -1,0 +1,16 @@
+This directory contains vendored repositories.
+
+Use [vendor.sh](vendor.sh) to manage vendored repositories:
+
+```
+Usage:
+  vendor.sh add <repo-url> <ref> [name]
+  vendor.sh list
+  vendor.sh update <name> <ref>
+
+<ref> specifies which branch/tag/commit to check out in the vendored
+repository.
+
+<name> is the name of the subdirectory in 'vendor/'. By default, it is
+inferred from the repository URL.
+```
--- a/.github/workflows/bats.yaml
+++ b/.github/workflows/bats.yaml
@@ -1,0 +1,19 @@
+name: Bats tests
+on: [push, pull_request]
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    strategy:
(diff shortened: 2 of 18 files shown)


markovejnovic:
add `CONTRIBUTING.md` info pls

vendor/README.md
```diff
@@ -0,0 +1,16 @@
...
+Usage:
+  vendor.sh add <repo-url> <ref> [name]
+  vendor.sh list
+  vendor.sh update <name> <ref>
+
+<ref> specifies which branch/tag/commit to check out in the vendored
+repository.
+
+<name> is the name of the subdirectory in 'vendor/'. By default, it is
+inferred from the repository URL.
```
markovejnovic, line 15: i would drop this and just stick it at the top of the script

markovejnovic (approved):
lgtm

vendor/README.md
taylordotfish: added a brief comment to the script, the rest is just the help message which is already in the script

PR #3651, opened 2026-07-17
docs: document built-in bash-preexec

taylordotfish:
This PR updates the docs with respect to the fact that bash-preexec is now automatically loaded by `atuin init bash`.


taylordotfish:
Now that #3688 has landed, we can merge the docs changes immediately. I removed the `install.sh` changes and will open another PR with those.

BinaryMuse (approved):

PR #3657, opened 2026-07-17
DisplayRichExt -- A mechanism for passing rich display options to paths

markovejnovic:
Tiny refactor I had Claude do while I was working on other things.

- Adds a new `DisplayRichExt` which allows you to call `Path::display_rich().trailing_slash(true)` which returns a type which `impl Display`. This new type will then gracefully print a trailing slash as you requested.
- There's also support for formatting a `~` character instead of `/home/me` and formatting relative paths too!
- Also removes a couple allocations in the atuin AI UI.
- Makes it so our config uses `PathBuf` rather than plain `String`.

Makes some of our path usage more intentional too -- so the code is slightly more self-documenting.


crates/atuin-daemon/src/components/search.rs
```diff
@@ -353,14 +376,14 @@ fn convert_filter_mode(
     match (mode, context) {
         (FilterMode::Global, _) => IndexFilterMode::Global,
         (FilterMode::Directory, Some(ctx)) => {
-            IndexFilterMode::Directory(with_trailing_slash(&ctx.cwd))
+            IndexFilterMode::Directory(ctx.cwd.display_rich().trailing_slash(true).to_string())
```
taylordotfish, line 379: I kind of wish this was `.trailing_slash()`. of course then we'd need another method that takes a bool. idk if it's worth it

crates/atuin-common/src/path/display_rich.rs
```diff
@@ -0,0 +1,297 @@
...
+        Self {
+            tilde: Some(Cow::Borrowed(home.as_ref())),
+            ..self
+        }
+    }
+
+    /// Abbreviate the path relative to the current user's home directory.
+    #[must_use]
+    pub fn tilde_me(self) -> Self {
+        match directories::BaseDirs::new() {
```
taylordotfish, line 80: I'm sure we were already doing stuff like this, but this is so sad. `BaseDirs::new` does so much: https://docs.rs/directories/6.0.0/src/directories/lin.rs.html#10

crates/atuin-common/src/path/display_rich.rs
```diff
@@ -0,0 +1,297 @@
...
+        }
+
+        if let Some(home) = self.tilde.as_deref()
+            && let Ok(relative) = self.path.strip_prefix(home)
+        {
+            return self.render(f, true, relative);
+        }
+
+        self.render(f, false, self.path)
+    }
```
taylordotfish: I think this could use some restructuring:
* `render` shouldn't be a trait method; it's a private implementation detail
* the `bool` parameter doesn't feel great, especially since it's kind of duplicating state that's already in `self`
* `fmt` is basically two nearly identical code paths that each dispatch to the same function differing only by a bool; surely there's a way to restructure this

ellie (approved):

ellie:
this is neat! i like

crates/atuin-common/src/path/display_rich.rs
taylordotfish: maybe something like

```rust
let relative = if let Some(base) = self.relative_to.as_deref() {
    self.path.strip_prefix(base)
} else if let Some(home) = self.tilde.as_deref() {
    // write tilde
    self.path.strip_prefix(home)
} else {
    self.path
};
```

or similar but returning a `(relative, tilde)` pair. either would be better I think

crates/atuin-common/src/path/display_rich.rs
markovejnovic: yeah you're right this is kind of nasty

markovejnovic: im glad you have the prudence where I don't... i kind of feel like we need the bun-style lazy_static! crap we built way back when, but i want to avoid bloating this PR

crates/atuin-common/src/path/display_rich.rs
markovejnovic: yeah i went on this back and forth and this is what i landed upon in terms of readability

crates/atuin-common/src/path/display_rich.rs
taylordotfish: `directories` internally uses https://docs.rs/dirs-sys/latest/dirs_sys/fn.home_dir.html

idk if that's for all platforms though

markovejnovic: added a TODO

PR #3660, opened 2026-07-17
fix(hook): drop agent commands containing NUL bytes via NonNulStr (#3589)

### LINKED ISSUE #3589: [Bug]: NUL bytes in history

markovejnovic:
Fixes #3589 by preventing agent hooks from writing NUL characters into history. My understanding of #3589 was that the failure arose from Claude invoking

```bash
echo '{
  "hook_event_name": "PreToolUse",
  "tool_name": "Bash",
  "tool_use_id": "beepboop",
  "tool_input": {
    "command": "hello\u0000world",
    "description": "measure the size of the globe"
  }
}' | atuin hook claude-code
```

When we deserialize `command`, we observe the byte array `hello\0world`, which we gladly commit to our database. Unfortunately, the command, as claude suggested, could never have theoretically been executed -- at some point, some shell will call `execve("hello", ["world"])` and since these **must** be C strings, there can be no `nul` character floating around. Bash, for example, strips NUL characters from its input.


crates/atuin-daemon/src/components/history.rs
```diff
@@ -153,9 +154,16 @@ impl HistorySvc for HistoryGrpcService {
                 )
             })?;
 
+        let command = NonNulStr::new(req.command).map_err(|err| {
+            Status::invalid_argument(format!(
+                "command contains a NUL byte at index {}",
+                err.index
+            ))
+        })?;
+
```
BinaryMuse: Been a while since I've been down in this code, but wondering if we shouldn't move this up into `atuin::command::client::history`, which covers both the daemon path and the non-daemon path?

markovejnovic: good call, checked with claude to better understand what's going on and i think this makes sense! will do it =)

BinaryMuse (approved):
Heck yeah ✨

PR #3659, opened 2026-07-17
Fix search functionality to respect character offsets and non-ASCII

Dilyxs:
### Reason for PR
Fixes #2898 
byte-by-byte traversal breaks when non-ascii characters are involved, 
```rust
source.chars().nth(i) 
```
when it is used with source.len(), goes out of bound and does an unwrap on a None causing error

### Solution
rewrote 4 functions to use chars_indices() to traverse while keeping word-boundary logic as it was before

added tests to verify behavior

## Checks
- [x] I am happy for maintainers to push small adjustments to this PR, to speed up the review cycle
- [x] I have checked that there are no existing pull requests for the same thing

PR #3662, opened 2026-07-18
chore: clean up `WordJumper` in atuin/src/command/client/search/cursor.rs

taylordotfish:
Followup to https://github.com/atuinsh/atuin/pull/3659

Clean up `WordJumper` in atuin/src/command/client/search/cursor.rs:

* Define `chars_within` helper for getting char indices within a range
* Use `tuple_windows` from `itertools` instead of peekable iterators
* Update tests to fit `rstest` pattern

### DIFF
--- a/crates/atuin/src/command/client/search/cursor.rs
+++ b/crates/atuin/src/command/client/search/cursor.rs
@@ -1,4 +1,25 @@
 use atuin_client::settings::WordJumpMode;
+use itertools::Itertools;
+use std::ops::Range;
+
+/// Like [`str::char_indices`], but only yields characters whose byte positions
+/// are in `start..end`.
+///
+/// # Panics
+///
+/// This function does not panic.
+fn chars_within(
+    string: &str,
+    range: Range<usize>,
+) -> impl DoubleEndedIterator<Item = (usize, char)> + use<'_> {
+    let start = string.ceil_char_boundary(range.start);
+    // `ceil_char_boundary` clamps to `string.len()` if out of bounds. Use
+    // `.max` to prevent `end` from being less than `start`.
+    let end = string.ceil_char_boundary(range.end).max(start);
+    string[start..end]
+        .char_indices()
+        .map(move |(i, c)| (start + i, c))
+}
 
 pub struct Cursor {
     source: String,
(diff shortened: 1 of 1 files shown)


crates/atuin/src/command/client/search/cursor.rs
```diff
@@ -1,4 +1,25 @@
 use atuin_client::settings::WordJumpMode;
+use itertools::Itertools;
+use std::ops::Range;
+
+/// Like [`str::char_indices`], but only yields characters whose byte positions
+/// are in `start..end`.
+///
+/// # Panics
+///
+/// This function does not panic.
```
markovejnovic, line 10:
```suggestion
```

imo

crates/atuin/src/command/client/search/cursor.rs
```diff
@@ -1,4 +1,25 @@
...
+    range: Range<usize>,
+) -> impl DoubleEndedIterator<Item = (usize, char)> + use<'_> {
+    let start = string.ceil_char_boundary(range.start);
+    // `ceil_char_boundary` clamps to `string.len()` if out of bounds. Use
+    // `.max` to prevent `end` from being less than `start`.
+    let end = string.ceil_char_boundary(range.end).max(start);
+    string[start..end]
+        .char_indices()
+        .map(move |(i, c)| (start + i, c))
+}
```
markovejnovic, line 22: I would stick this function in a `StringExt` trait in `atuin-common/` and I would call it

```rust
/// Iterator over the [`char`]s of a string yielding characters within the specified range.
///
/// The character indices are relative to the start of the original string.
///
/// ```
/// assert_eq(vec![('e', 1), ('l', 2)], chars_within("hello", [1, 3]).collect::<Vec<char>>())
fn char_indices_within<S: AsRef<Str>>(self: &S, range: Range<usize>) -> impl DoubleEndedIterator<Item = (usize, char)>;
```

obviously the templatizing for `S` belongs in the trait itself.

Thinking about this more, the offsetting logic feels like a generally applicable algorithm, so I would actually create a new

```rust
// iterator.rs

/// An offset of an element within an iterator, relative to a perfect subspan of that iterator.
///
/// ```text
/// Assuming an example `SubspanOffset` points to `2` in this range and a subspan [1, 3).
///
///        [-----span----)
/// 0      1      2      3      4      5      ...
/// ^-start^      ^
///        |-head |
///               |-(abs = 2, rel = 1)
/// ```
///
/// See [`IteratorExt<Item>::indices_within_subspan`] for an example usage.
impl SubspanOffset<IdxT> {
  /// The start of the original iterator.
  fn start(&self) -> IdxT;

  /// The index of the the subspan's first element within the iterator.
  fn head(&self) -> IdxT;

  /// Get the index relative to [`start`]
  fn abs(&self) -> IdxT;

  /// Get the index relative to 
  fn rel(&self) -> IdxT;
};

trait IteratorExt<Item> {
  /// New iterator yielding [`SubspanOffset<Idx>`] indices within the specified range.
  // not sure if `self` or `&self`, my guess is `self`.
  // also i feel like DoubleEndedIterator might be wrong here, semantically -- I feel like a plain Iterator might be correct.
  fn indices_within_subspan<Index>(self, range: Range<Index>) -> DoubleEndedIterator<SubspanOffset<Idx>>;

  /// Fetch the absolute offsets of elements within the specified range of the iterator.
  fn indices_within_subspan_abs<Index>(self, range: Range<Index>) -> DoubleEndedIterator<SubspanOffset<Idx>> {
    self.indices_within_subspan(range).map(SubspanOffset::abs)
  }

  fn indices_within_subspan_rel<Index>(self, range: Range<Index>) -> DoubleEndedIterator<SubspanOffset<Idx>> {
    self.indices_within_subspan(range).map(SubspanOffset::rel)
  }
}
```

Then in `string.rs`, you could implement a wrapper for your use-case. You might not even need it since I think `&str impl Iterator`, and our more general algo here is semantically equivalent (I think).

crates/atuin/src/command/client/search/cursor.rs
```diff
@@ -355,88 +332,66 @@ mod cursor_tests {
...
-            assert_eq!(EMACS_WORD_JUMPER.get_next_word_pos(&s, i_src), i_dest);
-        }
+    #[case::from_start_to_end_of_aaa(JUMPER_SUBJECT, 0, 6)]
+    #[case::from_within_aaa_to_end_of_aaa(JUMPER_SUBJECT, 3, 6)]
+    #[case::from_gap_skips_parens_to_end_of_bbb(JUMPER_SUBJECT, 7, 18)]
+    #[case::from_after_bbb_to_end_of_string(JUMPER_SUBJECT, 19, 30)]
+    #[case::non_ascii_from_start_to_test(EMACS_SUBJECT, 0, 10)]
+    #[case::non_ascii_from_emoji_to_test(EMACS_SUBJECT, 1, 10)]
+    #[case::trailing_multibyte(TRAILING_MULTIBYTE_SUBJECT, 3, 5)]
+    #[case::empty_string("", 0, 0)]
```
markovejnovic, line 347: 👍

markovejnovic (approved):
Your implementation is net positive, but I think we can generalize your algorithm. Worth asking claude whether there are other uses to see if we can re-use some of this logic.

Approving now, but I would **really** like to discuss my proposal.

crates/atuin/src/command/client/search/cursor.rs
taylordotfish: disagree, most functions that take a str and a range of byte indices in that str panic if the indices are out of range; it's worth noting that this function does not, in order to keep the semantics of the code it's replacing

if this were in the standard library I might agree with you since the standard library is diligent about documenting panics and you could reasonable infer no documented panics = doesn't panic, but our codebase is definitely not diligent about this kind of documentation

crates/atuin/src/command/client/search/cursor.rs
taylordotfish: your proposal will not work for this use case; `chars_within(s, n..n+1)` is O(1) whereas your proposal would be O(n)

PR #3663, opened 2026-07-18
chore(atuin-common): Add ellipsize_or_pad for unicode pad

markovejnovic:
This PR introduces a new `pad_ellipsize` utility to the `ellipsis.rs` set of utilities.

This fixes a tiny bug in the TUI on 2-col-width glyphs (such as 丑) which would get padded to three characters rather than two, since it was using byte padding historically.


crates/atuin-common/src/string/align.rs
```diff
@@ -0,0 +1,97 @@
...
+    Center,
+    /// Content flush to the end, padding on the start (right-aligned).
+    End,
+}
+
+pub trait AlignExt: AsRef<str> {
+    /// Pad `self` with spaces to fill `budget`, distributing the padding per `align`.
+    ///
+    /// Does not truncate.
+    fn pad_to<'a>(&'a self, budget: Measure, align: Alignment) -> Cow<'a, str> {
```
taylordotfish:
```suggestion
    fn pad_to(&self, budget: Measure, align: Alignment) -> Cow<'_, str> {
```

taylordotfish (approved):

crates/atuin-common/src/string/align.rs
markovejnovic: 👍

taylordotfish (approved):

PR #3665, opened 2026-07-18
refactor: commonize MessagePack (rmp) helpers

markovejnovic:
This PR commonizes some code regarding `rmp` `^(de)?serializers$`.

- `atuin-scripts` no longer panics on deserialize failures.
- there were 6 `error_report` copy-pastas, now there's one `DecodeError`.
- there were 7 checks for trailing bytes in the serialized rmp bytes. now there are some nice utilities to help against evil code
- there are a lot more tests now

Downside is:

- the caller needs to juggle `rmp::` and `atuin_common::rmp::`


crates/atuin-common/src/rmp/encode.rs
```diff
@@ -0,0 +1,62 @@
...
+    out: &mut W,
+    value: Option<T>,
+    write: impl FnOnce(&mut W, T) -> Result<(), EncodeError<W::Error>>,
+) -> Result<(), EncodeError<W::Error>> {
+    match value {
+        Some(v) => write(out, v),
+        None => rmp::encode::write_nil(out)
+            .map_err(|e| EncodeError::from(ValueWriteError::InvalidMarkerWrite(e))),
+    }
+}
```
taylordotfish, line 62: can you keep this function the way it was before?

crates/atuin-common/src/rmp/encode.rs
```diff
@@ -0,0 +1,62 @@
...
+pub fn write_bin<W: RmpWrite>(out: &mut W, data: &[u8]) -> Result<(), EncodeError<W::Error>> {
+    rmp::encode::write_bin(out, data)?;
+    Ok(())
+}
+
+pub fn write_array_len<W: RmpWrite>(out: &mut W, len: u32) -> Result<(), EncodeError<W::Error>> {
+    rmp::encode::write_array_len(out, len)?;
+    Ok(())
+}
+
```
taylordotfish, line 51: can we generate these with a macro instead?

crates/atuin-common/src/rmp/decode.rs
```diff
@@ -0,0 +1,174 @@
...
+        Err(e) => {
+            let e = e.into();
+            if let Some(Marker::Null) = e.type_mismatch() {
+                Ok(None)
+            } else {
+                Err(e)
+            }
+        }
+    }
+}
```
taylordotfish, line 115: can we keep these functions the way they were before?

crates/atuin-client/src/history.rs
```diff
@@ -214,30 +213,38 @@ impl History {
...
+        atu_rmp::encode::write_str(&mut output, &self.id.0)?;
+        atu_rmp::encode::write_u64(&mut output, self.timestamp.unix_timestamp_nanos() as u64)?;
+        atu_rmp::encode::write_sint(&mut output, self.duration)?;
+        atu_rmp::encode::write_sint(&mut output, self.exit)?;
+        atu_rmp::encode::write_str(&mut output, &self.command)?;
+        atu_rmp::encode::write_str(&mut output, &self.cwd)?;
+        atu_rmp::encode::write_str(&mut output, &self.session)?;
+        atu_rmp::encode::write_str(&mut output, &self.hostname)?;
+
+        atu_rmp::encode::write_optional(
```
taylordotfish, line 232: can we import encode and decode like before to minimize the diff?

taylordotfish (changes requested):
not sure why a bunch of comments were changed, imo most are worse now

also prefer type parameters and where clauses to `impl Trait`, not enough to block a PR but it's weird that they were changed

also not sure why a bunch of the functions take `Bytes` instead of a generic `RmpRead` or `RmpWrite`, I think they were better before

diff is also larger than necessary, imo this PR changes a bunch of things that didn't need changing

PR #3668, opened 2026-07-19
chore: Update us to sqlx 9

markovejnovic:
This PR bumps our sqlx to 0.9. Exciting (if it works)!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -90,8 +90,11 @@ features = ["json", "rustls-no-provider", "stream"]
 default-features = false
 
 [workspace.dependencies.sqlx]
-version = "0.8"
-features = ["runtime-tokio-rustls", "time", "postgres", "uuid"]
+version = "0.9"
+# Deliberately excludes `postgres` and TLS: the client binary only ever talks to
+# a local SQLite file. Crates that need the Postgres driver (and TLS to reach a
+# remote server) opt in individually -- see atuin-server-postgres.
+features = ["runtime-tokio", "time", "uuid"]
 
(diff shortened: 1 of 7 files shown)


markovejnovic:
Had claude filter the sqlx changelog for us:

---

- SQLite integer decode now errors on overflow (no silent truncation). One live i32 decode: `crates/atuin-client/src/database.rs:789` — let count: `i32 = row.get("count")` on a `COUNT(*)/group` aggregate. Pre-0.9 an out-of-range value truncated silently; now it returns a decode error. Only trips at >2³¹ rows in a single command/exit group — not realistic for shell history, so acceptable, but it's the one spot that changed semantics. (Widen to i64 if you want zero risk.)

Cargo.toml
```diff
@@ -90,8 +90,8 @@ features = ["json", "rustls-no-provider", "stream"]
 default-features = false
 
 [workspace.dependencies.sqlx]
-version = "0.8"
-features = ["runtime-tokio-rustls", "time", "postgres", "uuid"]
+version = "0.9"
+features = ["runtime-tokio", "time", "uuid"]
```
ellie, line 94: is rustls a default?

markovejnovic: It's not, but since `atuin-client` doesn't need TLS features, I've added a specific feature in `crates/atuin-server-postgres/Cargo.toml` for `atuin-server-postgres`.

ellie: ahh ok sounds good, thank you!

ellie (approved):

PR #3671, opened 2026-07-19
fix(search): reject non-interactive daemon-fuzzy

### LINKED ISSUE #3670: [Bug]: Non-interactive daemon-fuzzy is accepted but silently executes timestamp-ordered SQL

`daemon-fuzzy` appears to be an interactive/TUI search engine. If it is intentionally unsupported for non-interactive searches, I expect Atuin to reject it clearly:

```console
$ atuin search --search-mode daemon-fuzzy query
error: search mode 'daemon-fuzzy' is only supported with --interactive
help: use '--search-mode fuzzy' for non-interactive search
```

Alternatively, Atuin could implement and document an explicit fallback to ordinary `fuzzy`.

It should not accept `daemon-fuzzy` while silently executing a different search implementation.

### What happened?

Non-interactive search accepts `--search-mode daemon-fuzzy`, but does not contact or start the daemon. Instead, it passes `SearchMode::DaemonFuzzy` directly to the database search path.

That path:

- treats the query as an SQL character-subsequence match;
- orders candidates by timestamp;
- applies `LIMIT` before relevance ranking; and
- skips ordinary fuzzy reranking because `reorder_fuzzy` only handles `SearchMode::Fuzzy`.

Consequently, non-interactive `daemon-fuzzy`:

- does not use the daemon or Nucleo;
- ignores the daemon-fuzzy frequency, recency, and frecency settings;
- differs from interactive daemon-fuzzy matching and ranking; and
- is not an explicit fallback to ordinary non-interactive `fuzzy`.

The documentation says daemon-fuzzy uses an in-memory daemon index, without an interactive-only qualification:

https://docs.atuin.sh/cli/configuration/config/#search_mode

### Atuin doctor output

```yaml
Atuin Doctor
Checking for diagnostics


{
  "atuin": {
    "version": "18.17.0",
    "commit": "NO_GIT",
    "sync": null,
    "sqlite_version": "3.53.3",
    "daemon_enabled": true
  },
  "shell": {
    "name": "fish",
    "default": "bash",
    "plugins": [
      "atuin"
    ],
    "preexec": "built-in"
  },
  "system": {
    "os": "Arch Linux",
    "arch": "x86_64",
    "version": "rolling",
    "disks": [
      {
        "name": "/dev/nvme0n1p2",
        "filesystem": "btrfs"
      },
      {
        "name": "/dev/nvme0n1p2",
        "filesystem": "btrfs"
      },
      {
        "name": "/dev/nvme0n1p2",
        "filesystem": "btrfs"
      },
      {
        "name": "/dev/nvme0n1p2",
        "filesystem": "btrfs"
      },
      {
        "name": "/dev/nvme0n1p2",
        "filesystem": "btrfs"
      },
      {
        "name": "/dev/nvme0n1p2",
        "filesystem": "btrfs"
      },
      {
        "name": "/dev/nvme0n1p1",
        "filesystem": "vfat"
      }
    ]
  }
}
```

### Code of Conduct

zhangli091011:
## Summary

- Resolve and validate the effective search mode before database and record-store initialization.
- Reject `daemon-fuzzy` for non-interactive searches instead of silently using the database path.
- Apply CLI search-mode overrides before validation.
- Preserve interactive `daemon-fuzzy` and ordinary non-interactive `fuzzy`.

Fixes #3670

## Tests

- `cargo test -p atuin --bin atuin command::client::search::tests:: --locked`
- `cargo fmt --all -- --check`
- `cargo check -p atuin --all-targets --locked`
- `cargo clippy -p atuin --all-targets --locked -- -D warnings -D clippy::redundant_clone`

## Disclosure

This change was developed with AI assistance and reviewed and tested against the reported control flow. The final implementation was also exercised with invalid database paths to verify that mode validation happens before database setup.

## Checks

### DIFF
--- a/crates/atuin/src/command/client.rs
+++ b/crates/atuin/src/command/client.rs
@@ -163,6 +163,10 @@ impl Cmd {
     ) -> Result<()> {
         tracing::trace!(command = ?self, "client command");
 
+        if let Self::Search(search) = &self {
+            settings.search_mode = search.effective_search_mode(&settings)?;
+        }
+
         // Skip initializing any databases for history
         // This is a pretty hot path, as it runs before and after every single command the user
         // runs
--- a/crates/atuin/src/command/client/search.rs
+++ b/crates/atuin/src/command/client/search.rs
@@ -186,11 +186,6 @@ impl Cmd {
         store: SqliteStore,
         theme: &Theme,
     ) -> Result<()> {
-        if let Some(search_mode) = self.search_mode {
-            settings.search_mode = search_mode;
-        }
-        self.validate_search_mode(settings)?;
-
         let query = if self.query.is_empty() {
             std::env::var("ATUIN_QUERY").map_or_else(
                 |_| vec![],
--- a/crates/atuin/src/command/client/search.rs
+++ b/crates/atuin/src/command/client/search.rs
@@ -163,6 +163,18 @@ impl Cmd {
         self.interactive
     }
 
+    pub(super) fn effective_search_mode(&self, settings: &Settings) -> Result<SearchMode> {
+        let search_mode = self.search_mode.unwrap_or(settings.search_mode);
+
+        if !self.interactive && search_mode == SearchMode::DaemonFuzzy {
+            eyre::bail!(
+                "search mode 'daemon-fuzzy' is only supported with --interactive; use '--search-mode fuzzy' for non-interactive search"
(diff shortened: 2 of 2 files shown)


taylordotfish:
Closing this in favor of https://github.com/atuinsh/atuin/pull/3744 but thank you for your contribution!

PR #3678, opened 2026-07-20
fix(ui): calculate preview height based on command display width instead of byte count

lilydjwg:
## Checks


markovejnovic:
lgtm

PR #1885, opened 2024-03-15
feat(search): add better search scoring

ellie:
1. Prefer prefix matches over anything else
2. Prefer entire substring matches
3. Prefer more recent history

This will need tweaking, but should feel a little more natural to most people. I'll expose the values in config before release.

This applies to all search modes

## Checks
- [ ] I am happy for maintainers to push small adjustments to this PR, to speed up the review cycle
- [ ] I have checked that there are no existing pull requests for the same thing

PR #3682, opened 2026-07-20
docs(config): document smart_sort and tmux popup settings

### LINKED ISSUE #2167: docs: `smart_sort` should be documented
The `smart_sort` config option is missing some documentation. I had to refer to https://github.com/atuinsh/atuin/pull/1885 to understand what it actually does.

It should probably be documented in https://docs.atuin.sh/configuration/config/ as well as [`config.toml`](https://github.com/atuinsh/atuin/blob/main/crates/atuin-client/config.toml).

### LINKED ISSUE #3165: docs: documentation for tmux popup feature
Are there already any plans to add documentation for the tmux popup feature (PR #3058)? Additionally, it would be helpful to note that iTerm users who rely on tmux integration need to disable this feature, since iTerm does not support `tmux display-popup` (https://github.com/tmux/tmux/pull/4361)

If this isn’t already planned, I’d be happy to submit a PR.

Bartok9:
## Summary

- Document `smart_sort` in the config reference and example `config.toml`.
- Document the `[tmux]` popup settings (enabled/width/height), including the iTerm2 limitation.

## Motivation

Closes #2167.
Closes #3165.

`smart_sort` has been available for a while (default off) but was only discoverable via source/PR #1885. Users keep asking what it does. Ellie previously noted it was ready to document once proven.

The tmux popup shipped in 18.12 (`[tmux]` in the example config) but never landed on the docs site; #3165 also needs the iTerm2 native-tmux caveat.

## Verification

- Compared `smart_sort` text to `atuin-history::sort::sort` and the interactive-only call sites in `search/interactive.rs`.
- Compared `[tmux]` docs to `settings.rs` defaults, shell `ATUIN_TMUX_POPUP*` usage, and CHANGELOG 18.12.0 (#3058).
- Docs-only; no code/runtime change.

## Checklist

---

**Agent-Owner:** `sera` · **Platform:** hermes · **Claim-TTL:** 48h

### DIFF
--- a/crates/atuin-client/config.toml
+++ b/crates/atuin-client/config.toml
@@ -199,6 +199,11 @@
 ## Defaults to true. Configures whether to store commands that failed (those with non-zero exit status) or not.
 # store_failed = true
 
+## Defaults to false. When enabled, interactive search reorders results after the
+## chosen search mode runs: prefix matches first, then whole-substring matches,
+## then a light preference for more recent commands. Applies to all search modes.
+# smart_sort = false
+
 [stats]
 ## Set commands where we should consider the subcommand for statistics. Eg, kubectl get vs just kubectl
 # common_subcommands = [
--- a/docs/docs/configuration/config.md
+++ b/docs/docs/configuration/config.md
@@ -738,9 +738,11 @@ When `true` and Atuin is started inside a tmux session (tmux >= 3.2), interactiv
 search uses `tmux display-popup` instead of drawing inline in the current pane.
 Supported shell integrations: zsh, bash, and fish.
 
-`atuin init` exports `ATUIN_TMUX_POPUP` / `ATUIN_TMUX_POPUP_WIDTH` /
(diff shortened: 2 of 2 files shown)


Bartok9:
Thanks for the review! I double-checked the env-var claim against the `atuin init` source before writing it:

- `crates/atuin/src/command/client/init/zsh.rs` `print_tmux_config` (also `bash.rs` / `fish.rs`):
  - when `enabled = true` → exports `ATUIN_TMUX_POPUP_WIDTH` and `ATUIN_TMUX_POPUP_HEIGHT`
  - when `enabled = false` → exports `ATUIN_TMUX_POPUP=false`

So the docs deliberately say it exports *WIDTH/HEIGHT* (not `ATUIN_TMUX_POPUP=true`) on enable, and `ATUIN_TMUX_POPUP=false` on disable, which lines up with the code. `nu`/`powershell`/`xonsh` are still TODO for tmux popup, matching the 'zsh, bash, and fish' support note.

Happy to reword if there's a phrasing you'd prefer — just let me know the exact wording and I'll push it right away.

PR #3692, opened 2026-07-21
feat(search): filter by shell in interactive search

taylordotfish:
Followup to #3636 and #3658: add a new option, `search.shells`, which controls which shells' commands are shown in interactive search. By default, this is `"auto"`, which will show commands run by the current shell as well as commands with no shell recorded (from an older version of Atuin). It can also be `"all"` or an array of shell names.

See also: #608

### Testing

In addition to the unit tests, I tested locally in Bash and Zsh. I added some commands from Bash and Zsh and opened the interactive search UI in each shell. I examined the search results with an empty and non-empty query. I repeated the experiment for different values of `search.shells` in `config.toml` (`"auto"`, `"all"`, `["bash"]`, `["bash", ""]`, `[""]`, `["zsh"]`, `["zsh", "bash"]`).


crates/atuin/src/command/client/search/engines/skim.rs
```diff
@@ -62,6 +62,8 @@ async fn load_all_history(db: &dyn Database) -> Vec<(History, i32)> {
 
 #[allow(clippy::too_many_lines)]
 #[instrument(skip_all, level = Level::TRACE, name = "fuzzy_match", fields(history_count = all_history.len()))]
+// TODO: This function ignores `state.shell_filter`; the skim engine will be removed soon so it's
+// not worth adding.
```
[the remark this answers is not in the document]
taylordotfish: As explained by the comment, this is intended. The skim engine will be removed soon.

crates/atuin-daemon/src/components/search.rs
```diff
@@ -297,28 +329,24 @@ impl SearchSvc for SearchGrpcService {
...
-                                    .map(|s| s.display_rich().trailing_slash(true).to_string()),
-                                hostname: Some(ctx.hostname),
-                                session_id: Some(ctx.session_id),
-                            })
-                            .unwrap_or_default();
+                        let index = if let Some(new_index) =
+                            this.maybe_rebuild_index(search_req.shells).await
+                        {
+                            let mut guard = this.index.write().await;
+                            *guard = new_index;
```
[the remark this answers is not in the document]
taylordotfish: Wrong. When an index is installed, the lock is held until the search is completed. It is not possible for the index to change between installation and search. Also, the new indices are not built "from" an old index. They are completely fresh every time.

crates/atuin-client/src/database.rs
```diff
@@ -42,25 +42,27 @@ pub struct Context {
...
     pub reverse: bool,
     pub include_duplicates: bool,
     /// Author filter. Supports special values `$all-user` and `$all-agent`.
-    pub authors: Vec<String>,
-    pub shells: Vec<String>,
+    pub authors: &'a [String],
+    /// Shell filter. If empty, show commands from all shells. Empty string includes commands that
+    /// have no recorded shell.
+    pub shells: &'a [String],
 }
```
markovejnovic, line 66: 👍 

this should be very good for performance. i would love to see benchmarks before and after

crates/atuin-client/Cargo.toml
```diff
@@ -86,6 +86,7 @@ pretty_assertions = { workspace = true }
 rstest = { workspace = true }
 divan = "0.1.14"
 tempfile = "3"
+toml = "1.1"
```
markovejnovic, line 89: perhaps wiser to add this in the top-level `Cargo.toml` and then `workspace = true`. it's already used in the top-level tree through `config`.

crates/atuin-client/src/settings/shells.rs
```diff
@@ -0,0 +1,178 @@
...
+
+    #[default]
+    /// Include commands run from the current shell, or commands that have no recorded shell.
+    Auto,
+
+    /// Include commands run by any shell in a list. The empty string will include commands that
+    /// have no recorded shell.
+    ///
+    /// An empty list is treated the same as [`Self::All`], but [`Self::All`] should be preferred.
+    List(Vec<String>),
```
markovejnovic, line 17:
```suggestion
    List(Vec<ShellName>),
```

where `ShellName` is an enum. grep through the codebase, we already have an enum for this

crates/atuin-client/src/settings/shells.rs
```diff
@@ -0,0 +1,178 @@
...
+            {
+                let mut shells = Vec::new();
+                while let Some(shell) = seq.next_element()? {
+                    shells.push(shell);
+                }
+                Ok(Shells::List(shells))
+            }
+        }
+
+        deserializer.deserialize_any(Visitor)
```
markovejnovic:
```suggestion
        #[derive(Deserialize)]
        #[serde(rename_all = "lowercase")]
        enum Keyword {
            All,
            Auto,
        }
 
        /// Untagged unit variants only ever match `null`, so the keywords have to live one
        /// level down in [`Keyword`] rather than sitting alongside `List` here.
        #[derive(Deserialize)]
        #[serde(untagged, expecting = r#""all", "auto", or an array of strings"#)]
        enum Repr {
            Keyword(Keyword),
            List(Vec<String>),
        }
 
        Ok(match Repr::deserialize(deserializer)? {
            Repr::Keyword(Keyword::All) => Self::All,
            Repr::Keyword(Keyword::Auto) => Self::Auto,
            Repr::List(shells) => Self::List(shells),
        })
```

crates/atuin-client/src/settings/shells.rs
```diff
@@ -0,0 +1,178 @@
...
+}
+
+impl Shells {
+    /// Turn this setting into a concrete list of shells.
+    ///
+    /// The returned list is suitable for passing to [`Database::search`] via
+    /// [`OptFilters::shells`].
+    ///
+    /// [`Database::search`]: crate::database::Database::search
+    /// [`OptFilters::shells`]: crate::database::OptFilters::shells
```
markovejnovic, line 30: maybe explain the reasoning behind this and ShellList in here

crates/atuin-daemon/src/components/search.rs
```diff
@@ -28,6 +29,62 @@ const RESULTS_LIMIT: u32 = 200;
...
+        }
+    }
+}
+
+async fn build_frecency<F, R>(index: F, handle: &DaemonHandle)
+where
+    F: Fn() -> R,
+    R: Future<Output: Deref<Target = SearchIndex>>,
+{
+    // Build frecency map with current settings
```
markovejnovic:
```suggestion
```

crates/atuin/src/command/client/search/engines/skim.rs
markovejnovic, line 66: 👍

crates/atuin-client/Cargo.toml
taylordotfish: It's already being deduped; I'd feel more strongly if it were a regular dependency rather than a dev dependency

crates/atuin-daemon/src/search/index.rs
```diff
@@ -231,13 +223,104 @@ pub enum IndexFilterMode {
...
-/// Context for search queries.
-#[derive(Debug, Clone, Default)]
-pub struct QueryContext {
-    pub cwd: Option<String>,
-    pub git_root: Option<String>,
-    pub hostname: Option<String>,
-    pub session_id: Option<String>,
+#[derive(Clone, Debug, Default, Eq, PartialEq)]
+pub struct ShellFilter {
+    sorted: Vec<String>,
```
markovejnovic, line 231: doc pls

crates/atuin-daemon/src/components/search.rs
```diff
@@ -253,9 +244,43 @@ impl Component for SearchComponent {
...
+
 /// The gRPC service implementation.
+#[derive(Clone)]
 pub struct SearchGrpcService {
     index: Arc<RwLock<SearchIndex>>,
+    handle: DaemonHandle,
+}
+
+impl SearchGrpcService {
+    async fn maybe_rebuild_index(&self, shells: Vec<String>) -> Result<Option<SearchIndex>, ()> {
```
markovejnovic, line 296:
```suggestion
    #[instrumented]
    async fn maybe_rebuild_index(&self, shells: Vec<String>) -> Result<Option<SearchIndex>, ()> {
```

crates/atuin-daemon/src/components/search.rs
```diff
@@ -28,6 +29,62 @@ const RESULTS_LIMIT: u32 = 200;
...
+    F: Fn() -> R,
+    R: Future<Output: Deref<Target = SearchIndex>>,
+{
+    // Build frecency map with current settings
+    let settings = handle.settings().await;
+    index().await.rebuild_frecency(&settings.search).await;
+    info!("Frecency map built");
+}
+
+async fn build_index<F, R>(index: F, handle: &DaemonHandle) -> Result<(), ()>
```
markovejnovic, line 106:
```suggestion
/// todo pls docs
async fn build_index<F, R>(index: F, handle: &DaemonHandle) -> Result<(), ()>
```

crates/atuin-daemon/src/search/index.rs
```diff
@@ -231,13 +223,104 @@ pub enum IndexFilterMode {
...
+}
+
+impl ShellFilter {
```
markovejnovic, line 234: docs please

crates/atuin-common/src/utils.rs
```diff
@@ -195,12 +195,90 @@ where
...
+/// The sorted slice must not contain any duplicates.
+///
+/// All elements of `buffer` must be `false`. If `sorted.len() <= buffer.len()`, this function will
+/// not allocate any memory.
+pub fn iter_equals_sorted_deduped_slice<'a, I, T, S>(
+    iter: I,
+    sorted: &'a [S],
+    buffer: &mut [bool],
+) -> bool
+where
```
markovejnovic: `buffer` should be an allocator imo

markovejnovic (approved):

crates/atuin-daemon/src/components/search.rs
taylordotfish: comment was already there but sure

crates/atuin-client/src/settings/shells.rs
```diff
@@ -0,0 +1,173 @@
...
+    {
+        #[derive(Deserialize)]
+        #[serde(rename_all = "lowercase")]
+        enum Keyword {
+            All,
+            Auto,
+        }
+ 
+        /// Untagged unit variants only ever match `null`, so the keywords have to live one
+        /// level down in [`Keyword`] rather than sitting alongside `List` here.
```
taylordotfish:
```suggestion
```

crates/atuin-client/src/settings/shells.rs
taylordotfish: talked about this; string is more correct here since the `shell` field can be an arbitrary string

crates/atuin-client/src/settings/shells.rs
taylordotfish: done

crates/atuin-common/src/utils.rs
taylordotfish: agree, but allocators in rust aren't great/stable right now. I reworked this to avoid having to pass a buffer manually though

crates/atuin-daemon/src/components/search.rs
taylordotfish: done

crates/atuin-daemon/src/search/index.rs
taylordotfish: done

crates/atuin-daemon/src/components/search.rs
taylordotfish: I added it to the `build_index_only` and `build_frecency` functions instead

crates/atuin-daemon/src/search/index.rs
taylordotfish: done

markovejnovic: yeah even with the doc its painful... i really wish we had 

```
#[derive(Clone, Debug, Default, Eq, PartialEq)]
pub struct ShellFilter {
    /// The list of shells to include. If empty, all shells will be included.
    sorted: Vec<String>
      | |self| { self.sort() == self },
      && |self| { self.map(|x| self.count(x) == 1) }
}
```

ah a boy can dream

markovejnovic:
das good, i like `SortedDedupedSliceComparer`

markovejnovic (approved):

markovejnovic:
have you tried asking claude whether we could use `SortedDedupedSliceComparer` in some other spots?

PR #3700, opened 2026-07-21
fix: set correct umask for pty-proxy

### LINKED ISSUE #3695: [Bug]: `atuin pty-proxy init` changes user's umask

Shell umask not to change

### What happened?

When adding `eval "$(atuin pty-proxy init bash)"` to the top of my bashrc, my umask gets changed to `0077`, instead of the `0002` it is set to by PAM. This keeps catching me out and causing issues, e.g. taking down my Mastodon web interface, because after running an upgrade the www-data user had no access to read the precompiled asset files. I've tested this on both bash and zsh, across multiple Ubuntu machines, and a single Rocky Linux server; and get the same outcome. I've added a screen recording showing that adding or removing this line is what causes umask to change.

I don't *believe* this is intentional, but if it is, it should be documented somewhere.

https://github.com/user-attachments/assets/a7bb49b3-e218-4bd8-b346-c1534fbc36e3

### Atuin doctor output

```yaml
{
  "atuin": {
    "version": "18.17.1",
    "commit": "6245cc003a01380da6f62284e38cfcc5695e881d",
    "sync": null,
    "sqlite_version": "3.46.0",
    "daemon_enabled": true
  },
  "shell": {
    "name": "bash",
    "default": "bash",
    "plugins": [
      "atuin",
      "bash-preexec"
    ],
    "preexec": "bash-preexec"
  },
  "system": {
    "os": "Ubuntu",
    "arch": "x86_64",
    "version": "24.04",
    "disks": [
      {
        "name": "/dev/mapper/vg00-root",
        "filesystem": "ext4"
      },
      {
        "name": "/dev/nvme0n1p3",
        "filesystem": "ext4"
      },
      {
        "name": "/dev/nvme0n1p2",
        "filesystem": "vfat"
      }
    ]
  }
}
```

### Code of Conduct

ellie:
Atuin sets umask 0077 early on in the process start, which is great for most of what the cli does.

However, in the case of the cli proxy, this ends up being over-restrictive and can cause issues with newly created files.

Resolves #3695

## Checks

### DIFF
--- a/crates/atuin/src/command/mod.rs
+++ b/crates/atuin/src/command/mod.rs
@@ -41,13 +41,12 @@ pub enum AtuinCmd {
 
 impl AtuinCmd {
     pub fn run(self) -> Result<()> {
+        // set umask before we potentially open/create files
+        // or in other words, 077. Do not allow any access to any other user.
+        // Keep the previous umask so pty-proxy can restore it in the shell it
+        // spawns — the shell must not inherit ours (#3695).
         #[cfg(not(windows))]
-        {
-            // set umask before we potentially open/create files
-            // or in other words, 077. Do not allow any access to any other user
-            let mode = Mode::RWXG | Mode::RWXO;
-            umask(mode);
-        }
+        let prev_umask = umask(Mode::RWXG | Mode::RWXO);
 
         match self {
             // Client commands initialize their own logging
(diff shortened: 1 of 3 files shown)


crates/atuin/src/command/mod.rs
taylordotfish: Could we inline this function? The benefit of having a dummy non-Unix implementation was to avoid having to use `#[cfg]` at the call site, but now that the two versions of `run_pty_proxy` have different signatures, we have to do that anyway, so there's no benefit.

taylordotfish (commented):
lgtm overall, but the dummy non-unix `run_pty_proxy` should probably be removed

taylordotfish (approved):

PR #3703, opened 2026-07-22
ci: preview environments for docs pull requests

markovejnovic:
Adds Cloudflare Pages preview deployments for pull requests that touch `docs/**`, plus a strict build gate. Split out of the docs branding branch so it can be reviewed on its own.

Docs currently have **no PR check at all** — `docs-main.yml` and `docs-release.yml` only run `mike deploy` on push-to-main and on tags. Nothing builds the docs before a merge.

## Fork safety

This is deliberately two workflows rather than one.

| | runs | secrets | permissions |
|---|---|---|---|
| `docs-preview-build` | untrusted PR code | **none** | `contents: read` |
| `docs-preview-deploy` | default-branch code | Cloudflare + `GITHUB_TOKEN` | scoped writes |

Most docs PRs here come from forks, and on `pull_request` a fork gets no secrets — so it cannot deploy. The tempting fix is `pull_request_target`, and it would be genuinely dangerous: the PR owns `docs/mkdocs.yml`, and **mkdocs plugins execute arbitrary Python at build time**, so a fork could ship a plugin that exfiltrates the deploy token.

Instead the build half uploads an artifact with no credentials, and `workflow_run` — which executes the copy of the workflow on the default branch, not the one in the PR — downloads it and deploys. Both files carry a comment warning against collapsing them.

The artifact is still attacker-influenced HTML/JS. That is inherent to previews; it is contained by deploying to a dedicated `atuin-docs-preview` project on `*.pages.dev`, never to `docs.atuin.sh`.

## Alternatives considered

Worth recording, because ~200 lines of workflow looks like over-engineering without it.

- **Cloudflare Pages' native Git integration.** Would need no workflows at all, but it [does not create preview URLs for pull requests from forks](https://developers.cloudflare.com/pages/configuration/preview-deployments/) — which is where most docs contributions come from. Rejected on that alone.
- **`AdrianGonz97/refined-cf-pages-action`.** Purpose-built for exactly this, supports the fork-safe `workflow_run` pattern, and creates GitHub Deployments. Rejected on supply chain: last release May 2025, 44 stars, single maintainer — not something to hand a deploy token in a repo this size. Its useful part (the GitHub Deployment) is ~20 lines of `gh api` instead.
- **Netlify's Git integration.** Technically the best fit — it builds fork PRs, posts the comment and status natively, and needs no CI code. Rejected as a product/vendor decision.
- **gh-pages subdirectory previews.** No new secrets, but `mike` owns that branch and its `versions.json`, and it would publish untrusted PR content on `docs.atuin.sh` itself. Not worth saving a secret.

## Strict gate

The build runs `mkdocs build --strict`, so a broken internal link or a dead `#anchor` fails the check instead of shipping.

That is why the first commit is here: **`--strict` fails on `main` today**, with five long-standing link warnings. Verified — the workflow would break every docs PR if merged without it. One of those five was a real 404 (`faq.md` → `../configuration/key-binding.md` resolves to a path that does not exist in the built site); the other four were noisy but rendered correctly, and the fix is byte-identical output that mkdocs can now actually validate.

## Setup required before this works

1. Create a Cloudflare Pages project named `atuin-docs-preview` using **Direct Upload**, not the Git integration.
2. Create an API token with **Cloudflare Pages: Edit**.
3. Add repository secrets `CLOUDFLARE_API_TOKEN` and `CLOUDFLARE_ACCOUNT_ID`.

These are the **first third-party credentials in this repo** — every existing workflow uses only `GITHUB_TOKEN`. Worth a deliberate decision, not just a merge.

## Notes for review

- **The deploy half cannot run until this is on `main`.** `workflow_run` always uses the default-branch copy, so the build will run on this PR but no preview will appear. First real preview lands on the next docs PR.
- **Preview URLs are public.** Anything on `*.pages.dev` is world-readable; Cloudflare Access can gate them if that is not wanted.
- `cloudflare/pages-action` is deprecated in favour of `cloudflare/wrangler-action`, which this uses at `@v3` per Cloudflare's current docs. This repo runs newer majors elsewhere (`checkout@v7`, `download-artifact@v8`), so worth confirming v3 is still current.
- `--branch=pr-N` gives each PR a stable preview alias, so pushing again reuses the URL rather than minting a new one.
- The PR gets a **GitHub Deployment** (the native "View deployment" button, marked transient so superseded previews grey out) plus one upserted comment. Both via `gh api` rather than actions, to keep the credentialed workflow's dependency surface small.

### DIFF
--- a/.github/workflows/docs-preview-build.yml
+++ b/.github/workflows/docs-preview-build.yml
@@ -1,0 +1,73 @@
+name: Docs preview (build)
+
+# Half one of a two-workflow preview deployment.
+#
+# This half builds the docs from the pull request and uploads the result as an
+# artifact. It runs UNTRUSTED code -- the PR owns docs/mkdocs.yml, and mkdocs
+# plugins execute arbitrary Python at build time -- so it deliberately holds no
+# secrets and no write permissions. The deploy half
+# (docs-preview-deploy.yml) picks the artifact up.
+#
+# Do NOT be tempted to merge these into one workflow using
+# `pull_request_target` to get at the deploy credentials. That trigger runs
+# with secrets against the pull request's own code, so any fork could ship a
+# mkdocs plugin that exfiltrates the Cloudflare token.
+
+on:
+  pull_request:
+    paths:
+      - "docs/**"
+      - ".github/workflows/docs-preview-build.yml"
+      - ".github/workflows/docs-preview-deploy.yml"
+      - ".github/actions/docs-setup/**"
+
+permissions:
+  contents: read
+
+concurrency:
+  group: docs-preview-build-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+jobs:
+  build:
+    name: Build docs
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout pull request
+        uses: actions/checkout@v7
+        with:
+          # git-revision-date-localized reads the commit history to date each
+          # page; a shallow clone makes every page show the same date.
+          fetch-depth: 0
+
+      - uses: ./.github/actions/docs-setup
+
+      # --strict turns warnings into failures, which is what makes this a gate
+      # rather than decoration. It catches broken internal links and, thanks to
+      # `validation.anchors` in mkdocs.yml, deep-links to a #section that no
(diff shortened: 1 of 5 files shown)


markovejnovic:
ai-generated junk slop

PR #3706, opened 2026-07-22
fix(hook): install agent hooks with absolute atuin path

### LINKED ISSUE #3698: [Bug]: Agent Hook setup needs to use absolute paths

Note: `bash` is not my default shell, so it does not get the `atuin init bash` setup when it loads and does not have `atuin` in its PATH.

### What happened?

• PreToolUse hook (failed)
  error: hook exited with code 127

• PostToolUse hook (failed)
  error: hook exited with code 127

### Atuin doctor output

```yaml
{
  "atuin": {
    "version": "18.17.1",
    "commit": "6245cc003a01380da6f62284e38cfcc5695e881d",
    "sync": {
      "auth_state": "Self-hosted (authenticated)",
      "records": true,
      "auto_sync": true,
      "last_sync": "2026-07-21 19:15:32.445895814 +00:00:00"
    },
    "sqlite_version": "3.46.0",
    "daemon_enabled": true
  },
  "shell": {
    "name": "fish",
    "default": "bash",
    "plugins": [
      "atuin"
    ],
    "preexec": "built-in"
  },
  "system": {
    "os": "Fedora Linux",
    "arch": "x86_64",
    "version": "44",
    "disks": [
      {
        "name": "/dev/mapper/vgfedora-fedora",
        "filesystem": "ext4"
      },
      {
        "name": "/dev/nvme0n1p2",
        "filesystem": "ext4"
      },
      {
        "name": "/dev/sda1",
        "filesystem": "ext4"
      },
      {
        "name": "/dev/nvme0n1p1",
        "filesystem": "vfat"
      }
    ]
  }
}
```

### Code of Conduct

Bartok9:
## Summary

- `atuin hook install` now writes an absolute path to the Atuin binary for Claude Code / Codex hooks.
- Re-running install upgrades older bare `atuin hook …` entries without duplicating them.

## Motivation

Closes #3698.

Agent processes often start hooks with a minimal environment. When bash is the default shell (but never got `atuin init`), PATH lacks `atuin` and hooks fail with exit 127 (`command not found`). Pinning the hook command to `current_exe()` matches how the daemon is already spawned.

## Verification

- `cargo test -p atuin hook` — 35 passed
- `cargo clippy -p atuin --bin atuin -- -D warnings` — clean
- `cargo fmt -p atuin -- --check` — clean
- Unit coverage: absolute formatting, space quoting, bare→absolute upgrade, idempotent reinstall

## Checklist

---

**Agent-Owner:** `sera` · **Platform:** hermes · **Claim:** `sera` · **Claim-TTL:** 48h

### DIFF
--- a/crates/atuin/src/command/client/hook.rs
+++ b/crates/atuin/src/command/client/hook.rs
@@ -338,6 +338,11 @@ fn add_hook_entries(hooks: &mut Value, agent: &Agent, hook_command: &str) -> Res
 
         // Rewrite every Atuin-managed entry to the absolute command and drop any
         // duplicates so reinstalling never leaves stale bare `atuin hook ...` lines.
+        //
+        // `seen_managed` is intentionally shared across all wrapper entries for this
+        // event type: we want exactly one Atuin hook to fire per event, so if two
+        // different-matcher wrappers both contain a managed hook, only the first is
+        // kept (and an emptied wrapper is pruned below).
         let mut seen_managed = false;
         for entry in arr.iter_mut() {
             let Some(hooks_arr) = entry.get_mut("hooks").and_then(Value::as_array_mut) else {
--- a/crates/atuin/src/command/client/hook.rs
+++ b/crates/atuin/src/command/client/hook.rs
@@ -265,11 +265,30 @@ fn resolve_hook_command(agent: &Agent) -> Result<String> {
 }
 
 fn format_hook_command(exe: &Path, hook_agent: &str) -> String {
-    let exe = shlex::try_quote(exe.to_string_lossy().as_ref()).map_or_else(
-        |_| format!("\"{}\"", exe.display()),
+    format!("{} hook {hook_agent}", quote_exe(exe))
+}
+
+/// Quote an executable path for the platform shell that will run the hook.
+///
+/// Agent hook commands are executed by the host shell, so quoting must match it:
+/// POSIX (single-quote) rules on Unix and `cmd.exe` (double-quote) rules on Windows.
+#[cfg(windows)]
+fn quote_exe(exe: &Path) -> String {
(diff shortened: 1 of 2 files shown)


Bartok9:
Thanks @greptile-apps — all three addressed in 316ce1c:

1. **Windows quoting** — `format_hook_command` now quotes per-platform: POSIX single-quote on Unix and `cmd.exe`-style double-quote on Windows, so the emitted command is valid for whichever shell actually runs the hook.
2. **Duplicate migration** — the migration no longer stops at the first managed entry. It now rewrites *every* Atuin-managed hook to the absolute command, drops any duplicates, and prunes emptied wrapper entries, so reinstalling can't leave stale bare `atuin hook …` lines.
3. **Path handling** — kept `current_exe()` + platform-correct quoting; added tests for multi-managed dedup and unmanaged-command preservation.

`cargo test -p atuin hook` (37 passed), `cargo clippy -p atuin --bin atuin -- -D warnings` clean, `cargo fmt --check` clean.

Bartok9:
Thanks for the thorough second pass @greptile-apps — agreed all three are edge notes rather than blockers:

1. **`trim_end_matches(['\\','/'])`** — only used to isolate the binary basename for the `is_managed` check; a trailing-separator exe path isn't producible via `current_exe()`, so it's a non-issue in practice.
2. **Windows `&`/`^`/`%`** — a real concern only for exotic install paths; `current_exe()` on Windows resolves to a normal Program Files-style path, so space-only quoting covers the realistic cases. Noted for any future broader Windows hook support.
3. **Shared `seen_managed` dedup** — that's the intended semantics (one atuin hook per event, never two firing). I've added an inline comment making that explicit so the behavior isn't implicit.

No functional change from the review — just the clarifying comment. Ready for maintainer review whenever convenient.

PR #3713, opened 2026-07-22
AI WIP | Convert daemon locks to parking_lot

markovejnovic:
## Summary

Converts `std::sync`/`tokio::sync` locks to `parking_lot` wherever no guard needs to cross an `.await`, removing async lock machinery from the daemon's hot paths:

- **Search path** (runs per keystroke): `SearchIndex::search`/`rebuild_frecency`/`nucleo_item_count` are now fully synchronous; both the inner index locks and the outer `Arc<RwLock<SearchIndex>>` are parking_lot.
- **Settings**: `parking_lot::RwLock<Arc<Settings>>` snapshot pattern — `settings()` is now sync and returns a cheap `Arc<Settings>`; the config watcher no longer deep-clones `Settings` per reload.
- **History component** (per command completion): parking_lot with a clone-out pattern so no guard is held across the DB writes in `end_history`.
- **Semantic component** and **atuin-ai user-context cache**: direct Mutex conversions; std poison-recovery handling removed (parking_lot has no poisoning).

Intentionally untouched: tokio channels (`broadcast`/`mpsc`/`watch`), `DashMap`s, `std::sync::Once`/`OnceLock`, and the test-only mutex in `atuin-common` (rationale table in `docs/superpowers/plans/2026-07-22-parking-lot-conversion.md`).

## Safety notes

`atuin-nucleo` enables parking_lot's `send_guard` feature, and cargo feature-unification makes guards `Send` workspace-wide — so the compiler cannot catch a guard held across an `.await`. Every conversion was manually audited for this, and a final whole-branch review independently re-derived the guard lifetime at every lock site plus a lock-ordering analysis (no cycles; deepest nesting was flattened in the last commit).

## Behavior deltas (benign, for the record)

- Settings reloads no longer wait for an in-flight sync tick: previously the tokio read guard held across `do_sync_tick` blocked `apply_settings` until the tick finished. Now the reload applies immediately and the in-flight tick completes on its stale snapshot; the next tick sees the new settings.
- Concurrent interactive searches block an OS worker thread on the nucleo write lock instead of yielding to tokio. Millisecond-scale, and the old code already blocked a runtime thread on the same CPU-bound work while holding the tokio lock.

## Testing

- `cargo test` green on `atuin-daemon` (23), `atuin-ai` (270), `atuin-nucleo` (12), plus `atuin`/`atuin-client` suites; the only workspace failures are the pre-existing Postgres-dependent `tests/users.rs` (environmental, untouched by this branch).
- `cargo clippy --workspace` zero warnings; `cargo fmt --all -- --check` clean; `cargo tree` confirms a single `parking_lot v0.12.5`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -302,6 +302,7 @@ dependencies = [
  "futures",
  "glob-match",
  "imara-diff",
+ "parking_lot",
  "pretty_assertions",
  "pulldown-cmark",
  "ratatui",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -438,6 +438,7 @@ dependencies = [
  "hyper-util",
  "lasso",
  "listenfd",
+ "parking_lot",
  "prost",
  "prost-types",
  "protox",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,6 +56,7 @@ tokio = { version = "1", features = ["full"] }
 uuid = { version = "1.9", features = ["v4", "v7", "serde"] }
 whoami = "2.1.0"
 typed-builder = "0.23.2"
+parking_lot = "0.12.1"
 pretty_assertions = "1.3.0"
 proptest = "1.11.0"
 rstest = "0.26.1"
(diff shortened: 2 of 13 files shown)


markovejnovic:
junk, seems like we don't need `parking_lot`

PR #3728, opened 2026-07-23
feat(caps): client-side capability negotiation

markovejnovic:


markovejnovic:
Superseded by #3737, which now targets `main` and contains these commits plus the packfile bundling work. Splitting them across a stack made review harder, not easier, so they're combined into one PR.

PR #3730, opened 2026-07-23
perf: actually delete history rows, and index active history

ellie:
## What

Three related changes to history deletion and the cost of the `deleted_at` filter:

1. **`Sqlite::delete` now deletes the row** instead of scrambling the command and setting `deleted_at`. The tombstone dance existed so sync v1 could propagate deletions without leaking command text - v1 is gone (#3634), and sync v2 propagates deletions through the record store (`HistoryRecord::Delete`), which already hard-deletes via `delete_rows` when applied. The only remaining caller is the `store_failed = false` cleanup in `handle_end`, which runs before the entry is ever pushed to the record store, so no other host can know about it.

   `deleted_at` is still **read** everywhere: old databases contain tombstones, which query filters keep hiding, and `init_store` still converts them to deletion records when building a store from a legacy DB. The column and record wire format are unchanged.

2. **Partial index over active history**: `create index ... on history(timestamp) where deleted_at is null`. Counting live history was a full table scan - measured ~430ms cold on a 96MB / 167k-row database, ~975ms at 1M rows. With the index: ~8ms and ~30ms respectively. The planner also uses it for timestamp-ordered listings of live history. No measurable insert overhead (241µs vs 253µs per committed insert - the fsync dominates, and the entry is tiny next to the existing command-text indexes). Complements #3729, which moved the count off the TUI critical path.

3. **`build.rs` for atuin-client**: `sqlx::migrate!()` embeds migration directories at compile time, but without `rerun-if-changed` an incremental build does not recompile when a new migration file is added - producing a binary silently missing the migration. This bit me while testing (2).

## Checks

### DIFF
--- a/crates/atuin-client/build.rs
+++ b/crates/atuin-client/build.rs
@@ -1,0 +1,9 @@
+fn main() {
+    // sqlx::migrate!() embeds these directories at compile time, but cargo only
+    // reruns the macro when a tracked file changes - without this, adding a new
+    // migration file does not trigger a recompile and silently ships a binary
+    // missing the migration.
+    println!("cargo:rerun-if-changed=migrations");
+    println!("cargo:rerun-if-changed=meta-migrations");
+    println!("cargo:rerun-if-changed=record-migrations");
+}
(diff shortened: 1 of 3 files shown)


ellie:
I'd like to remove the deleted_at column in the future, but for now let's keep it in and focus solely on perf here

PR #3733, opened 2026-07-24
docs(search): explain how to search history from another host

### LINKED ISSUE #2442: Docs request: how to search by host
(Note, I suspect I'm misunderstanding something here, but ...)

I have three hosts: `void`, `prime`, and `bucket`; I'd like interactively search history on, say, `prime`, filtering for commands that were run on `bucket` and vice versa. 

Right now, I'm not even sure all of these hosts are seeing the same command list; `atuin sync` shows different numbers of entries on all of them, but at least I'd like to be able to search :D

Bartok9:
## Summary

- FAQ: how to search history that ran on another synced host.
- Advanced usage: clarify that context switch carries host/directory, so host filter mode can target that machine.

## Motivation

Closes #2442.

`host` filter mode only matches the current machine. Multi-host users need the global view + host column, or switch-context onto a remote entry then cycle to host mode. There is still no `--host` CLI flag; document a `--format` post-filter for scripts.

## Verification

- Matched `Context::from_history` (adopts hostname/session/cwd) and host-mode SQL (`lower(hostname)` vs context hostname).
- Matched default `ctrl-a` then `c` → `switch-context`, and `ctrl-r` → `cycle-filter-mode`.
- Docs-only.

## Checklist

---

**Agent-Owner:** `sera` · **Platform:** hermes · **Claim:** `sera` · **Claim-TTL:** 48h

### DIFF
--- a/docs/docs/faq.md
+++ b/docs/docs/faq.md
@@ -45,6 +45,33 @@ Add `--disable-ai`, e.g.:
 eval "$(atuin init zsh --disable-ai)"
 ```
 
+## How do I search history from another host?
+
+After [sync](guide/sync.md), the interactive **host** filter mode only includes history recorded on the machine you are currently using. Commands from other machines are still available under the default **global** filter.
+
+To label rows by machine, show the `host` column:
+
+```toml
+[ui]
+columns = ["duration", "time", "host", "command"]
+```
+
+To focus one other host inside the TUI:
+
+1. Stay in **global** filter mode and find any command that ran on that host (the host column helps).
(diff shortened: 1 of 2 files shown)


Bartok9:
Thanks for the review! Good catch on the scripted example — I've added a note clarifying that `grep` is line-oriented and truncates multiline commands in that output, pointing readers to a whole-record filter if they need the full command text. Pushed in b0d444e.

Bartok9:
Fixed Vale `write-good.ThereIs` on the FAQ host-search sentence and rebased onto current `main`.

Agent-Owner: sera

PR #3741, opened 2026-07-24
docs: clarify how to uninstall shell integration

### LINKED ISSUE #3565: Clarify better where to disable atuin upon uninstallation of a `setup.atuin.sh` installation
Please, provide a more detailed instructions on where to disable atuin upon uninstallation.

In the documentation, we can find "Remove the line referencing 'atuin init' from your shell config". However, since the shell-specific configuration is not implemented[^1], this is actually too vague in my opinion.

Instead, I would suggest to explicitly list which files to modify, which includes at least:

- `~/.zshrc`
-`~/.bashrc`
- `~/.config/fish/config.fish`
- `~/.profile`

I am not sure where exactly the last one is modified in the installation process, but proved especially problematic in my case. It locked me out of the interactive session of KDE Plasma on Kubuntu 26.04. Since I have previously deleted atuin binaries as instructed, I was immediately returned to the login screen after logging in, until I modified `.profile` via TTY.

This happened with atuin v18.16.1 installed using https://setup.atuin.sh/ .

[^1]: See Line 54:

https://github.com/atuinsh/atuin/blob/ba0ce4fbdec53217e9997d536a2ead1b5174642d/install.sh#L54

taylordotfish:
Fixes #3565


.vale.ini
```diff
@@ -13,6 +13,36 @@ Packages = https://github.com/errata-ai/Microsoft/releases/download/v0.14.2/Micr
...
+#     destroys the "and/or" token and silently drops the check that fires on it.
+#
+# A backtick, a quote or a ":" precedes the path in the first two cases, and a
+# space never does, so requiring whitespace excludes all three.
+#
+# The trailing [\w/-] stops before sentence punctuation, so masking a path
+# cannot merge two sentences and change what the readability rules see.
+#
+# Run "vale --minAlertLevel=suggestion docs/docs" before and after editing this
+# line: the alert count must only ever go down.
```
markovejnovic:
```suggestion
```

markovejnovic (approved):

PR #3744, opened 2026-07-25
fix(search): add `DbSearchMode`; correctly fall back to "fuzzy" in non-interactive searches

### LINKED ISSUE #3670: [Bug]: Non-interactive daemon-fuzzy is accepted but silently executes timestamp-ordered SQL

taylordotfish:
Add `DbSearchMode`; correctly fall back to "fuzzy" in non-interactive searches when the "daemon-fuzzy" or "skim" modes are specified. Fixes https://github.com/atuinsh/atuin/issues/3670.

### DIFF
--- a/crates/atuin-ai/src/tools/mod.rs
+++ b/crates/atuin-ai/src/tools/mod.rs
@@ -1123,8 +1123,7 @@ impl PermissibleToolCall for AtuinHistoryToolCall {
 
 impl AtuinHistoryToolCall {
     pub(crate) async fn execute(&self, db: &atuin_client::database::Sqlite) -> ToolOutcome {
-        use atuin_client::database::{self, Database as _, OptFilters};
-        use atuin_client::settings::SearchMode;
+        use atuin_client::database::{self, Database as _, DbSearchMode, OptFilters};
 
         // query_context rather than current_context: when running outside an
         // atuin-hooked shell (e.g. as an MCP server) there is no ATUIN_SESSION.
@@ -1161,7 +1160,7 @@ impl AtuinHistoryToolCall {
 
         let results = match db
             .search(
-                SearchMode::Fuzzy,
+                DbSearchMode::Fuzzy,
                 filter_mode,
                 &context,
                 &self.query,
--- a/crates/atuin-client/benches/ordering.rs
+++ b/crates/atuin-client/benches/ordering.rs
@@ -1,7 +1,7 @@
 use crate::_util::context::BenchCtx;
 use crate::history::BenchHistory;
+use atuin_client::database::DbSearchMode;
 use atuin_client::ordering::reorder_fuzzy;
-use atuin_client::settings::SearchMode;
 
 /// reorder_fuzzy is invoked on every keystroke during interactive fuzzy search, so keeping the
 /// performance in check is important.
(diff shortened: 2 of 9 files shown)


ellie:
I'm not certain that we want this, it would be strange to have different search results depending on if you run it interactively or via the CLI. Longer term, I would really like us to avoid the old fuzzy search entirely

ellie:
I DO think that the CLI should work though - even if that means starting the daemon just like the tui does

taylordotfish:
Talked about this in person -- eventually we want to truly use the daemon-fuzzy engine in non-interactive searches but this PR is an acceptable stopgap:

* Before this PR: non-interactive daemon-fuzzy search returns "fuzzy" results in wrong order
* After this PR: non-interactive daemon-fuzzy search returns "fuzzy" results in correct order
* Ideal (future work): non-interactive daemon-fuzzy search returns daemon-fuzzy results

markovejnovic (approved):

PR #3763, opened 2026-07-27
bench: daemon search benchmark (divan + CodSpeed)

ellie:
Benchmarks + CI only — no behavior changes. Merging this first records the nucleo-era baseline on CodSpeed, so the frizbee PR stacked on top (#3782) shows its improvement directly in its CodSpeed report.

## What

Adds `crates/atuin-search-bench`: a [divan](https://github.com/nvzqz/divan) benchmark measuring the daemon's `SearchIndex::search` **end-to-end** — index built from shell-history-shaped data, then queried the way the interactive UI does (realistic query set, limit 200, including the empty-query and directory-filtered paths).

- **Engine-agnostic benchmark names** (`daemon_search[<scale>/<query>]`): they measure whatever fuzzy matcher the daemon currently uses, so an engine swap shows up as a step change in the tracked series rather than as a new set of benchmarks.
- Deterministic synthetic history generator shaped after a real 168k-line history (zipf-ish repeated head, templated body with high-cardinality args so uniqueness lands near real history's ~30%, long tail incl. very long commands and some unicode). `BENCH_DATA` runs against a real exported history instead; `BENCH_SCALES`/`BENCH_SEED` env knobs, sampling tuned via divan's own `DIVAN_*` vars.
- Result counts are sanity-checked per scale so a broken query or filter shows up in the logs rather than as a suspiciously fast run.
- The `divan` dependency is CodSpeed's drop-in compat crate, so the same bench runs natively via `cargo bench` and instrumented under `cargo-codspeed`.
- CI: wired into the existing `codspeed.yml` workflow (adds `-p atuin-search-bench` to the build/run, with a reduced `BENCH_SCALES` since instrumented index builds at 1M+ lines would dominate the job).

## Baseline this records (M-series MacBook, native, 1M history lines / 188k unique commands)

| query | current daemon (nucleo) |
|---|---|
| empty | 9.1 ms |
| `git` | 18.5 ms |
| `git p` | 19.2 ms |
| `docker compose up` | 20.5 ms |
| `zzznomatchzzz` | 16.9 ms |
| `git` (directory filter) | 39.0 ms |

#3782 drops these 69–95%.


ellie:
Merged regardless of codspeed flag - microseconds are too small. will adjust that bench

PR #3766, opened 2026-07-28
fix(config): preserve comments in `atuin config set`

taylordotfish:
Currently, when `atuin config set` modifies an existing key, any comment above that key is deleted:

Before:

```toml
auto_sync = true

# some comment here

# another comment here
style = "auto"
```

After `atuin config set style full`:

```toml
auto_sync = true
style = "full"
```

This PR fixes the issue by transferring the old decor from the existing value.


BinaryMuse (approved):
Oof, good catch

PR #3768, opened 2026-07-28
chore: add `OrFilter` as a replacement for `Vec`s as filter lists

taylordotfish:
In several parts of the code, we use `Vec`s as filters, where the filter allows an item if it matches any element of the `Vec`, or if the `Vec` is empty, all items are allowed.

These semantics are not obvious from the type, particularly the empty `Vec` meaning "allow all". This PR adds a new type, `OrFilter`, with exactly these semantics. As a benefit, it also has O(log n) lookups instead of O(n).


crates/atuin-client/src/database.rs
```diff
@@ -58,11 +59,30 @@ pub struct OptFilters<'a> {
...
+            after: None,
+            limit: None,
+            offset: None,
+            reverse: false,
+            include_duplicates: false,
+            authors: DisjunctiveFilter::all(),
+            shells: DisjunctiveFilter::all(),
+        }
+    }
 }
```
markovejnovic: possible to use `#[default]`?

crates/atuin-common/src/filter.rs
```diff
@@ -0,0 +1,417 @@
...
+
+impl<T: Ord> FilterStorage for &'_ [T] {
+    const DEFAULT: Self = &[];
+    type Item = T;
+}
+
+impl<T: Ord> FilterStorage for &'_ mut [T] {
+    // Using a separate function is necessary here -- if `empty_mut_slice()` is replaced with
+    // `&mut []`, it won't compile.
+    const DEFAULT: Self = empty_mut_slice();
```
markovejnovic:
```suggestion
    const DEFAULT: Self = (|| { &mut [] })();
```

crates/atuin-common/src/filter.rs
```diff
@@ -0,0 +1,417 @@
...
+        let mut next_write: usize = 1;
+
+        while next_read < self.len() {
+            if self[next_read] != self[next_write - 1] {
+                self.swap(next_read, next_write);
+                next_write += 1;
+            }
+            next_read += 1;
+        }
+        &mut self[..next_write]
```
markovejnovic: `slice.rs` with a `SliceExt`

crates/atuin-common/src/filter.rs
```diff
@@ -0,0 +1,417 @@
...
+    }
+}
+
+/// Helper type for comparing a [`DisjunctiveFilter`] with an iterator.
+pub struct DisjunctiveFilterComparer<'a, T, I> {
+    /// [`None`] if no iterator was provided.
+    comparer: Option<SortedDedupedSliceComparer<'a, T, I>>,
+    /// Whether the filter is an ["all" filter](DisjunctiveFilter::is_all).
+    is_all: bool,
+}
```
markovejnovic: enumaroni

markovejnovic (commented):
- `OrFilter` better name
- `SortedDedupedSliceComparer` -> `compare.rs`
- `DEFAULT` -> `EMPTY_SENTINEL` (or something like that)
- Do not derive `Default`
- Seal `FilterStorage`
- Do not overload the meaning of `Option::None` to mean `Match::All`
- proptest on `filter.rs`.

crates/atuin-common/src/filter.rs
taylordotfish: Doesn't work :(

PR #3771, opened 2026-07-28
feat(config): add bind_up_arrow / bind_ctrl_r config option

borngraced:
`--disable-up-arrow` does works but only for the current session. I'd have to edit my shell init file to add the flag, which kinda defeats the point of having a config file.

 in `~/.config/atuin/config.toml`
```shell
[keys]
bind_up_arrow = true
bind_ctrl_r = true
```

CLI flags still work and override the config if you pass them.

## Checks


virtuallyunknown:
This would be great because right now if you're loading atuin via antidote, there is nowhere to pass the `--disable-up-arrow` flag to.

Current workaround I found is to set `bindkey '^r' atuin-search` and `export ATUIN_NOBIND="true"` in my zsh config files.

PR #3782, opened 2026-07-28
daemon: replace nucleo with frizbee, rank by match quality with frecency tiebreak

ellie:
Experimenting with frizbee, an alternative fuzzy matcher. It shows promising performance increases vs nucleo for our usage.

I've removed our dep on nucleo, and vendored the nucleo diatric normalisation table

Flagging that this is MPL licensed. This is copyleft, but is fine to include in the codebase as it is limited to just the one file. cc @taylordotfish who knows much more about licenses than I do!


crates/atuin-daemon/src/search/normalize.rs
```diff
@@ -0,0 +1,981 @@
...
+//! (helix-editor/nucleo, `src/chars/normalize.rs`) so dropping the matcher
+//! dependency doesn't lose accent-insensitive search.
+//!
+//! nucleo is licensed under the Mozilla Public License 2.0; this file remains
+//! covered by MPL-2.0 (<https://mozilla.org/MPL/2.0/>) independently of the
+//! crate's license.
+
+// Keep the vendored body byte-for-byte comparable with upstream.
+#![allow(clippy::manual_range_contains)]
+
```
taylordotfish: Instead of this we could also use an outer attribute on the `mod normalize` declaration in `search/mod.rs`:

```rust
#[allow(clippy::manual_range_contains)]
mod search;
```

That would keep this file closer to upstream.

crates/atuin-daemon/src/search/normalize.rs
```diff
@@ -0,0 +1,981 @@
+//! Latin-diacritic to ASCII normalization table.
+//!
+//! Vendored from the atuin-nucleo fork of the nucleo fuzzy matcher
+//! (helix-editor/nucleo, `src/chars/normalize.rs`) so dropping the matcher
+//! dependency doesn't lose accent-insensitive search.
```
taylordotfish: Just to be safe we should add an explicit link to https://github.com/atuinsh/atuin/blob/main/crates/atuin-nucleo/matcher/src/chars/normalize.rs; it might not be obvious to a reader how to get to the original file

crates/atuin-daemon/src/search/index.rs
```diff
@@ -224,13 +222,52 @@ pub enum IndexFilterMode {
...
+        Some((end, _)) => &query[..end],
+        None => query,
+    }
+}
+
+/// Normalize Latin diacritics to their ASCII equivalents (`é` → `e`) using
+/// the same table as nucleo's `Normalization::Smart`, so unaccented queries
+/// keep matching accented history entries — frizbee does no normalization of
+/// its own. Maps char to char, so char positions are preserved. Borrows the
+/// input when it is already normalized.
```
taylordotfish:
```suggestion
/// Normalize Latin diacritics to their ASCII equivalents (`é` → `e`) so
/// unaccented queries match accented history entries. Maps char to char so
/// char positions are preserved.
```

crates/atuin-daemon/src/search/index.rs
```diff
@@ -224,13 +222,52 @@ pub enum IndexFilterMode {
...
+    Cow::Owned(s.chars().map(normalize).collect())
+}
+
+/// One entry in the fuzzy matcher's haystack: the command text plus the text
+/// the matcher actually runs against (diacritics normalized to ASCII).
+/// `matchable` shares `command`'s allocation when normalization is a no-op.
+struct Haystack {
+    command: Arc<str>,
+    matchable: Arc<str>,
+}
```
taylordotfish: It's confusing that an “entry in a haystack” is itself called a `Haystack`. Perhaps:

* Rename `Haystack` to `HaystackEntry`
* Rename the `SearchIndex::haystacks` field to `haystack` (a haystack is a vector of haystack entries)

crates/atuin-daemon/src/search/index.rs
```diff
@@ -224,13 +222,52 @@ pub enum IndexFilterMode {
...
+    }
+    Cow::Owned(s.chars().map(normalize).collect())
+}
+
+/// One entry in the fuzzy matcher's haystack: the command text plus the text
+/// the matcher actually runs against (diacritics normalized to ASCII).
+/// `matchable` shares `command`'s allocation when normalization is a no-op.
+struct Haystack {
+    command: Arc<str>,
+    matchable: Arc<str>,
```
taylordotfish:
```suggestion
    /// The original text of the command.
    command: Arc<str>,
    /// The [normalized](normalize_diacritics) text of the command.
    normalized_command: Arc<str>,
```

or alternatively (more general naming):

```suggestion
    /// The original text of the entry.
    original: Arc<str>,
    /// The [normalized](normalize_diacritics) text of the entry.
    normalized: Arc<str>,
```

crates/atuin-daemon/src/search/index.rs
```diff
@@ -336,45 +384,143 @@ impl SearchIndex {
...
+        let query = normalize_diacritics(query);
+
+        let haystacks = self.haystacks.read().expect("haystacks lock poisoned");
+
+        // Filter target resolved once per search (intern/parse the string
+        // here) so the per-command checks below are integer-set lookups
+        enum Filter {
+            All,
+            Dir(Spur),
+            Workspace(String),
```
taylordotfish: Can this contain a `&str` instead of a `String` so we don't have to clone?

crates/atuin-daemon/src/search/index.rs
```diff
@@ -336,45 +384,143 @@ impl SearchIndex {
...
-        let snapshot = nucleo.snapshot();
-        let matched_count = snapshot.matched_item_count().min(limit);
+        let config = frizbee::Config::default()
+            .casing(frizbee::CaseMatching::Smart)
+            .sort(frizbee::SortStrategy::IndexAsc);
+        let mut matcher = frizbee::Matcher::from_query(&query, &config);
+
+        // An empty query matches every candidate with fuzzy score 0, so skip
+        // the matcher and rank purely by frecency
+        let mut scored: Vec<(u16, u32, u32)> = if matcher.patterns().is_empty() {
```
taylordotfish: What is this 3-tuple of integers? Needs documentation, or ideally turning into a named struct

crates/atuin-daemon/src/search/index.rs
```diff
@@ -241,10 +278,13 @@ pub struct SearchIndex {
...
-    nucleo: RwLock<Nucleo<String>>,
-    /// Injector for adding new commands to Nucleo.
-    injector: Injector<String>,
+    /// Unique commands in insertion order — the fuzzy matcher's haystack.
+    /// Shares the `Arc<str>` allocations with `commands`.
+    ///
+    /// Lock order: paths that need both this lock and a `commands` shard lock
+    /// must take this one first (see `add_history`, `search`,
+    /// `rebuild_frecency`); acquiring in the reverse order can deadlock.
+    haystacks: std::sync::RwLock<Vec<Haystack>>,
```
taylordotfish: Why are we using a std RwLock instead of a tokio one here? Not saying it should be tokio, just don't know what the rationale is

crates/atuin-daemon/src/search/index.rs
```diff
@@ -293,16 +327,35 @@ impl SearchIndex {
...
-            // New command - create Arc<str> once and share it
-            let Some(data) = CommandData::new(history, &self.interner) else {
-                return; // Invalid UUIDs, skip this entry
-            };
+            // New command; the entry API holds the shard lock so a concurrent
+            // insert of the same command can't double-push the haystack. The
+            // haystacks lock must be taken before the shard lock: search and
+            // rebuild_frecency read `commands` while holding the haystacks
+            // lock, so waiting on it here while entry() holds a shard lock
+            // would invert the order and deadlock.
```
taylordotfish: Hallucinated comment? It doesn't make sense to me. `self.commands` doesn't appear to have a lock on it?

crates/atuin-daemon/src/search/index.rs
```diff
@@ -224,13 +222,52 @@ pub enum IndexFilterMode {
 }
 
 /// Shareable frecency map: command -> frecency score.
-/// Wrapped in Arc for zero-copy sharing with scorer callbacks.
-type FrecencyMap = Arc<HashMap<Arc<str>, u32>>;
+/// Aligned with `SearchIndex::haystacks` by index, so scoring a match is an
+/// array access instead of hashing the command string. Commands added after
+/// the last rebuild sit past the end and score 0 until the next rebuild.
```
taylordotfish: This comment should be moved to the `SearchIndex::frecency_map` field

crates/atuin-daemon/src/search/index.rs
```diff
@@ -336,45 +384,143 @@ impl SearchIndex {
...
-                    // DashMap<Arc<str>, _>::get accepts &str via Borrow trait
+            // Rank by match quality first; frecency only breaks ties between
+            // equally good matches. Adding frecency to the fuzzy score instead
+            // lets a frequently-run scattered match outrank a contiguous one
+            // (see #3702), while dropping frecency entirely loses any
+            // most-recently-used ordering among equal matches - a tiebreaker
+            // gives both.
+            let rank = |a: &(u16, u32, u32), b: &(u16, u32, u32)| {
+                b.0.cmp(&a.0).then(b.1.cmp(&a.1)).then(a.2.cmp(&b.2))
+            };
```
taylordotfish: This is really hard to understand because I don't know what the tuple fields are, but also, this just seems like a reimplementation of normal tuple comparison? Like I'm pretty sure you would get the same result if you got rid of this and changed `select_nth_unstable_by` to `select_nth_unstable`

Edit: I see, the first two fields should actually be ordered in reverse

crates/atuin-daemon/src/search/index.rs
```diff
@@ -336,45 +384,143 @@ impl SearchIndex {
...
+            IndexFilterMode::Session(session) => {
+                parse_uuid_bytes(session).map_or(Filter::Nothing, Filter::Session)
+            }
+        };
+
+        // Filter pre-pass: collect the candidate commands for this filter
+        // mode, remembering each one's haystack index for frecency lookups.
+        // Filtered modes walk the command map directly (no string hashing);
+        // indices are sorted so ranking ties stay deterministic.
+        let candidates: Vec<(u32, &Haystack)> = tracing::span!(Level::TRACE, "index_search_filter")
```
taylordotfish: AFAICT we don't need to store both the index and haystack reference because they're always in sync. We'd save a lot of space by just storing the index, and then updating `candidates[i as usize].1.command.as_ref()` below to `haystacks[candidates[i as usize]].command.as_ref()`.

crates/atuin-daemon/src/search/index.rs
```diff
@@ -336,45 +384,143 @@ impl SearchIndex {
...
+                            Filter::Host(hostname) => entry.has_invocation_on_host(*hostname),
+                            Filter::Session(session) => entry.has_invocation_in_session(session),
+                        })
+                        .map(|entry| entry.haystack_index)
+                        .collect();
+                    indices.sort_unstable();
+                    indices
+                        .into_iter()
+                        .filter_map(|i| haystacks.get(i as usize).map(|hay| (i, hay)))
+                        .collect()
```
taylordotfish: Instead of:

* Iterator of indicates -> collect into intermediate `Vec`
* Sort intermediate `Vec`
* Iterate intermediate `Vec` -> filter by whether haystack entry exists -> collect into final `Vec`

We can avoid the intermediate `Vec` entirely:

* Iterator of indicates -> filter by whether haystack entry exists -> collect into final `Vec`
* Sort final `Vec`

crates/atuin-daemon/src/search/index.rs
```diff
@@ -336,45 +384,143 @@ impl SearchIndex {
...
+                .collect()
+        } else {
+            let commands: Vec<&Arc<str>> =
+                candidates.iter().map(|&(_, hay)| &hay.matchable).collect();
+            // Use every core, like nucleo did — mid-sized indexes regress
+            // badly on many-core machines if matching stays single-threaded.
+            // But below ~10k candidates, stay serial: frizbee hands out
+            // 2048-item chunks, so a small list gets barely any overlap and
+            // scoped thread spawn/join costs more than it saves (measured
+            // ~500us on CodSpeed's runners, walltime mode).
```
taylordotfish: Hallucinated comment? I thought CodSpeed didn't use their own runners…

crates/atuin-daemon/src/search/index.rs
taylordotfish: This method is very large now; there are several parts I think could be extracted into separate functions:

* The `Filter` enum and conversion from `filter_mode` to `Filter`
* Building the `candidates` vec
* The `frecency` closure and the building of the `scored` vec
* The final ranking code

crates/atuin-daemon/src/search/index.rs
```diff
@@ -293,16 +327,35 @@ impl SearchIndex {
...
-                cols[0] = cmd.clone().into();
-            });
+            let mut haystacks = self.haystacks.write().expect("haystacks lock poisoned");
+            match self.commands.entry(Arc::clone(&command_arc)) {
+                dashmap::Entry::Occupied(mut entry) => {
+                    entry.get_mut().add_invocation(history, &self.interner);
+                }
+                dashmap::Entry::Vacant(vacant) => {
+                    let Some(data) =
+                        CommandData::new(history, &self.interner, haystacks.len() as u32)
```
taylordotfish: It would be nice if we had better handling for the case where `haystacks.len()` is greater than 2^32, but it might be more trouble than it's worth. I'm not sure what the correct semantics would be.

crates/atuin-daemon/src/search/index.rs
```diff
@@ -224,13 +222,52 @@ pub enum IndexFilterMode {
...
+/// Longest query the fuzzy matcher will see. Frizbee's `u16` scores overflow
+/// (and panic) somewhere past ~2700 needle chars; no real query is anywhere
+/// near either limit, so longer input is truncated.
+const MAX_QUERY_LEN: usize = 512;
+
+/// Truncate a query to the longest length frizbee can score without
+/// panicking in `Matcher::from_query`. Anything that hands a query to
+/// frizbee (including client-side highlighting) must apply this.
+pub fn truncate_query(query: &str) -> &str {
+    match query.char_indices().nth(MAX_QUERY_LEN) {
```
taylordotfish:
```suggestion
pub fn truncate_query(query: &str) -> &str {
    // O(1) happy path -- query cannot exceed `MAX_QUERY_LEN` chars if it doesn't
    // even have that many bytes.
    if query.len() <= MAX_QUERY_LEN {
        return query;
    }
    match query.char_indices().nth(MAX_QUERY_LEN) {
```

crates/atuin-daemon/src/search/mod.rs
```diff
@@ -1,11 +1,12 @@
...
 
 mod index;
+mod normalize;
 
 // Include the generated proto code
 tonic::include_proto!("search");
 
 // Re-export the index and related types
-pub use index::{IndexFilterMode, SearchIndex};
+pub use index::{IndexFilterMode, SearchIndex, normalize_diacritics, truncate_query};
```
taylordotfish: Can we inline `normalize_diacritics` and `truncate_query` into this module? Since they're used outside of it it seems more natural to define them here than reexport them from a child.

crates/atuin-daemon/src/search/index.rs
```diff
@@ -336,45 +384,143 @@ impl SearchIndex {
...
+
+            scored
+                .iter()
+                .filter_map(|&(_, _, i)| {
                     self.commands
-                        .get(cmd.as_str())
+                        .get(candidates[i as usize].1.command.as_ref())
                         .map(|data| data.most_recent_id())
                 })
                 .collect()
```
taylordotfish: This PR isn't responsible for this, but I noticed we're doing this:

* Convert `[u8; 16]` UUID in each command to `String` (-> return `Vec<String>`)
* In the search component, parse each `String` UUID *back* into a `[u8; 16]`
* Convert each `[u8; 16]` into a `Vec<u8>` (-> send `Vec<Vec<u8>>`)

Could we return `Vec<[u8; 16]>` from this function instead to avoid the round-trip conversion? Ideally we wouldn't have to convert the arrays into `Vec<u8>`s either, but that's a harder change and would require modifying the protobuf message. So let's at least get rid of the `String` round-trip.

taylordotfish (commented):
Excited for this change! Looks good, but I think we should address these things first:

Performance:

* Reduce the size of the `candidates` vec by just storing the index -- the haystack entry reference is redundant
* Avoid collecting into an intermediate vec when building `candidates` by filtering first, then doing the sort on the final vec
* Add happy path to `truncate_query` when \# bytes <= max \# of chars
* `Filter` should contain a `&str` instead of `String` so we don't have to clone
* Not caused by this PR, but in `SearchIndex::search` we're converting `[u8; 16]` UUIDs to `String`s, only to immediately parse them back into `[u8; 16]`s in `components/search.rs`. Can `SearchIndex::search` return `Vec<[u8; 16]>` instead?

Naming:

* `Haystack` is actually a haystack *entry*; should be called `HaystackEntry` or similar
* `haystacks` should be called `haystack` (a haystack is a `Vec` of haystack entries)
* `matchable` field on `Haystack` is unclear; fields should be `command` and `normalized_command`, or `text` and `normalized`
* The `(u16, u32, u32)` tuple in `scored` should be a named struct (can derive `Ord` for same ordering as before)

Organization:

* `SearchIndex::search` is very long now -- split function body into 3 or 4 functions (see the diff comment I left for a suggestion; the steps are fairly discrete and self-contained and thus a good candidate for this)
* Manual implementation of `rank` for use with the `scored` vec seems completely unnecessary -- it does the same thing as Rust's normal ordering of tuples
* Move `truncate_query` and `normalize_diacritics` up one level rather than reexporting

Comments:

* Hallucination in comment about the “shard lock” and avoiding deadlock? It suggests `commands` has a lock on it but it doesn't.
* Hallucination about performance justification regarding multithreading? CodSpeed doesn't use their own runners I thought.
* Too much AI verbosity / irrelevant information in some of the comments (e.g., `normalize_diacritics`)
* Add comments to fields of `Haystack`

Licensing:

* Looks good, just one suggestion to add an explicit link to the original file (from our fork).

Questions:

* `tokio::sync::RwLock` vs. `std::sync::RwLock`? We're using both now -- is that correct?
* Handle case where `haystacks.len()` is greater than 2^32? Right now we handle this very incorrectly by truncating the index and colliding with older commands. I'm not sure it's worth adding to this PR though, and I'm not sure what the right behavior is.

PR #3785, opened 2026-07-28
test: adopt rstest across the workspace test suite

markovejnovic:


crates/atuin-client/src/history.rs
```diff
@@ -616,117 +642,45 @@ mod tests {
...
-                    );
-                }
-            }
+    #[rstest]
+    #[case::from_v0(Version::Zero, BYTES_V0.to_vec(), expected_v0())]
+    #[case::from_v1(Version::One, BYTES_V1.to_vec(), expected_v1())]
+    #[case::from_v2(Version::Two, expected_v2().serialize().unwrap().0, expected_v2())]
+    fn deserialize_across_versions(
+        #[case] source: Version,
+        #[case] bytes: Vec<u8>,
```
taylordotfish: doesn't really matter because it's a test but

```suggestion
    #[case::from_v0(Version::Zero, BYTES_V0, expected_v0())]
    #[case::from_v1(Version::One, BYTES_V1, expected_v1())]
    #[case::from_v2(Version::Two, &expected_v2().serialize().unwrap(), expected_v2())]
    fn deserialize_across_versions(
        #[case] source: Version,
        #[case] bytes: &[u8],
```

PR #3793, opened 2026-07-29
docs: adopt a subset of ASD-STE100 Simplified Technical English

markovejnovic:
Most people reading our docs are not reading them in their first language. This applies the mechanically checkable parts of [ASD-STE100 Simplified Technical English](https://asd-ste100.org) (issue 9) to `docs/docs` and to the strings the program prints.

## The linter was pulling the other way

`Microsoft.Contractions` is an **error-level** rule whose autofix rewrites `cannot` → `can't`. CI was actively pushing the docs toward the construction a second-language reader has the most trouble with — the elided word carries the negation, and the apostrophe form is not derivable from the words it stands for.

That rule is now off, and `.vale/styles/STE/` holds the inverse. This had to come first; without it the two rules fight.

## New `STE` Vale style

Only what a linter can actually check. Each rule file names the STE rule behind it.

| Rule file | STE rule | Checks |
|---|---|---|
| `Contractions`, `AmbiguousContractions` | 4.2 | `cannot`, not `can't` |
| `LatinAbbrev` | GR-6 | no `e.g.`, `i.e.`, `etc.`, `via` |
| `PhrasalVerbs` | 9.3 | `find`, not `figure out` |
| `SentenceLength` | 6.3 | 25 words max |
| `Spelling` | 1.14 | American spelling |

## Applied across the docs

- **202 contractions** expanded
- **32 Latin abbreviations** and bare `via` replaced with the English word that fits each sentence — not mechanically from a swap table
- **62 over-long sentences** split, or turned into vertical lists (rule 4.3)

Vale is clean at the level CI gates on.

## Bug fix: `TokenIgnores` was silently disabling the linter

Independent of the STE work, and worth a look on its own.

The pattern `(\s[~.]*/[\w./-]*[\w/-])` also matched bare `/word` — **including inside fenced code blocks**, where a match desynchronizes Vale's block parser and drops every alert in the rest of the file. Ten of 51 pages were unlinted, some from as early as line 32.

Requiring the leading `[~.]` fixes it, and immediately surfaced two error-level alerts that were **already red on `main`** (both fixed here).

## What is deliberately *not* adopted

Three STE rules would make these docs worse, and `CONTRIBUTING.md` says so:

- **Approved-word list (1.1)** — a 900-word aerospace vocabulary with no entry for `daemon`, `binary`, or `sync`.
- **No `-ing` forms (3.5)** — would rename every "Getting started" heading.
- **Strict active voice (3.6)** — would force us to invent agents for descriptions that genuinely have none ("history is stored in `~/.local/share/atuin`").

`Microsoft.Passive` stays at `suggestion` for that reason. The genuinely reversible ones still get fixed by hand.

## Rust strings

Semicolons removed from messages (rule 8.1), contractions expanded, and the format-error help rewritten as short imperatives:

```
- "If your formatting string contains literal curly braces, you need to escape them by doubling:"
+ "To put a literal curly brace in a formatting string, write it two times:"
```

Message strings only — no logic changes.

## Verification

- `vale --minAlertLevel=warning docs/docs` → **0 errors**, 2 warnings (both `alex.Ablist`, pre-existing on `main`)
- `mkdocs build --strict` → clean; spot-checked the rendered admonitions and new vertical lists
- `cargo fmt --check`, `cargo clippy --workspace --all-targets` → clean
- `cargo test -p atuin-client -p atuin-daemon -p atuin` → 319 passed

The 3 `tests/users.rs` failures are a missing local Postgres — verified identical on a pristine `main` worktree.

## Review notes

The doc rewrites are the part worth human eyes. Mechanical replacements were driven off Vale's own reported spans so code blocks stayed untouched, but every Latin-abbreviation and long-sentence change was a judgement call about the sense of that particular sentence.

One thing CI still cannot see: mkdocs admonition bodies are indented 4 spaces, so Vale treats them as code blocks and never lints them. I fixed the contractions in there by hand, but a regression there will not turn CI red. `CONTRIBUTING.md` already flagged this; it is still true.


markovejnovic:
overkill. not really necessary

PR #3797, opened 2026-07-30
feat: capability negotiation protocol (client + server)

markovejnovic:
Capability system used by atuin.

# Context

A node advertises capabilities about itself and, if it is a client, can read the server's.

Atuin's client and server versions are not necessarily always compatible. There are features
that clients may support, but outdated servers will not.

The capability system is designed to help us bridge the gap between the two.

# Design

- Each capability has a unique `CRI` (capability resource identifier), eg.
  `sh.atuin.server/capabilities`.
- Each capability has arbitrary associated data, for example `{ "version": 1 }`.

The client passes a header with each request it makes, `x-atuin-capabilities-known: <hash>`
which communicates to the server what capabilities the client is aware of. If the server's
capability hash does not match that of what the client believes, the server rejects the request
with a 412, after which the client polls `/api/v0/capabilities` to get the new capability list
as well as the new hash of the capability list.

The client then passes this updated hash back to the server and all is well.

The server capabilities are sent with every response as part of `x-atuin-capabilities-available`
in order to eagerly communicate to the client that the capability set needs to be updated
(hopefully to avoid unnecessary 412s).

# Implementation

The client side is implemented as reqwest middleware in [`client::CapClient`].
The server side is implemented as a plain struct that can be embedded in any server, in
[`client::CapServer`].


crates/atuin-domain/src/caps/mod.rs
```diff
@@ -0,0 +1,127 @@
...
+    pub version: String,
+
+    /// The list of capabilities this server supports, as a map of capability name to its value.
+    pub capabilities: HashMap<String, serde_json::Value>,
+}
+
+/// The capabilities a node advertises about itself.
+#[derive(Default)]
+struct OwnCaps {
+    caps: RwLock<HashMap<CapKey, Box<dyn Any + Send + Sync>>>,
```
taylordotfish:
```suggestion
    caps: RwLock<HashMap<CapKey, Box<dyn Capability>>>,
```

crates/atuin-domain/src/caps/mod.rs
```diff
@@ -0,0 +1,127 @@
...
+struct CapKey(String);
+
+impl Borrow<str> for CapKey {
+    fn borrow(&self) -> &str {
+        &self.0
+    }
+}
+
+/// A capability which two peers may negotiate.
+pub trait Capability: Serialize + DeserializeOwned + Send + Sync + 'static {
```
taylordotfish:
```suggestion
pub trait Capability: Any + Serialize + DeserializeOwned + Send + Sync + 'static {
```

crates/atuin-domain/src/caps/mod.rs
```diff
@@ -0,0 +1,127 @@
...
+            .insert(CapKey(C::NAME.to_string()), Box::new(cap));
+    }
+
+    fn support<C: Capability + Clone>(&self) -> Option<C> {
+        self.caps
+            .read()
+            .get(C::NAME)
+            .and_then(|cap| cap.downcast_ref::<C>())
+            .cloned()
+    }
```
taylordotfish: Docs please

crates/atuin-domain/src/caps/mod.rs
```diff
@@ -0,0 +1,127 @@
...
+    /// An opaque capability token issued by the server.
+    pub version: String,
+
+    /// The list of capabilities this server supports, as a map of capability name to its value.
+    pub capabilities: HashMap<String, serde_json::Value>,
+}
+
+/// The capabilities a node advertises about itself.
+#[derive(Default)]
+struct OwnCaps {
```
taylordotfish: Get rid of this type? It's never serialized

crates/atuin-domain/src/caps/client.rs
```diff
@@ -0,0 +1,219 @@
...
+
+    /// Register a capability this client advertises.
+    pub fn can<C: Capability>(&self, cap: C) {
+        self.inner.own.can(cap);
+    }
+
+    /// Check whether this client advertises the given capability.
+    pub fn support<C: Capability + Clone>(&self) -> Option<C> {
+        self.inner.own.support()
+    }
```
taylordotfish: These names are super confusing: I would never expect `can` to modify state, and I actually *would* expect `support` to do so.

Suggestion:

* `can` → `add`
* `support` → `supports` or `get`

crates/atuin-domain/src/caps/client.rs
```diff
@@ -0,0 +1,219 @@
...
+/// Client-side capability set: advertises its own capabilities and can read the server's.
+///
+/// The server's capabilities are populated by [`CapClient::refresh`], which the crate drives at
+/// negotiation time; [`CapClient::server_support`] is then a pure, offline read of that cache.
+///
+/// Cloning is cheap.
+#[derive(Debug, Clone)]
+pub struct CapClient {
+    inner: Arc<CapClientInner>,
+}
```
taylordotfish: Suggestion:

* `CapClientInner` → `CapClient`
* `CapClient` → `Arc<CapClient>`

Reasoning:

If I see `some_cap_client.clone()` where `some_cap_client: CapClient`, it's extremely non-obvious that this is performing a *shallow* clone where both clients are actually references to the same client. If they're `Arc`s, it's clearer.

If we had the [`Share` trait](https://rust-lang.github.io/rust-project-goals/2026/ergonomic-rc.html), this would be a non-issue; `CapClient` would simply implement `Share`. But as-is I'd prefer to have `Arc` be explicit.

crates/atuin-domain/src/caps/server.rs
```diff
@@ -0,0 +1,182 @@
...
+}
+
+#[derive(Debug)]
+struct Inner {
+    /// Opaque version token (xxh3 of the canonical capability set), computed once.
+    token: String,
+    /// Pre-serialized capabilities document (a `CapabilitiesResponse` as JSON).
+    body: String,
+    /// The advertised capabilities, kept for introspection (`advertises`).
+    caps: BTreeMap<String, Value>,
```
taylordotfish: It's somewhat confusing that we're storing this data in two places: here *and* in `OwnCaps`

taylordotfish: We should be dynamically building this from OwnCaps or something

crates/atuin-domain/src/caps/server.rs
```diff
@@ -0,0 +1,182 @@
...
+    Current,
+    /// The client presented a *differing* token -- its cached capabilities are stale.
+    Stale,
+}
+
+/// Immutable, cheaply-cloneable set of capabilities a server advertises.
+#[derive(Debug, Clone)]
+pub struct CapServer {
+    inner: Arc<Inner>,
+}
```
taylordotfish: Comment about `CapClient` and `Arc` applies here too

crates/atuin-domain/src/caps/mod.rs
```diff
@@ -0,0 +1,143 @@
...
+
+impl<C: Capability> DynCapability for C {
+    fn name(&self) -> &'static str {
+        C::NAME
+    }
+
+    fn json(&self) -> Result<serde_json::Value, serde_json::Error> {
+        serde_json::to_value(self)
+    }
+}
```
taylordotfish: I would prefer to unify these traits:

```rust
trait Capability {
    fn static_name() -> &'static str
    where
        Self: Sized;

    fn name(&self) -> &'static str;
}

struct SomeCap;

impl Capability for SomeCap {
    fn static_name() -> &'static str {
        "some_cap"
    }

    fn name(&self) -> &'static str {
        Self::static_name()
    }
}

fn must_be_dyn(_: &dyn Capability) {}
```

Alternatively, add `DynCapability` as a supertrait of `Capability`. But I prefer the first approach.

crates/atuin-domain/src/caps/server.rs
```diff
@@ -0,0 +1,176 @@
...
+        self
+    }
+
+    /// Finalize: compute the version token and pre-serialize the document. Cheap work done once.
+    pub fn build(self) -> Arc<CapServer> {
+        // A `BTreeMap` serializes its keys in sorted order, so the token is byte-identical on every
+        // node running the same capability set.
+        let wire = self.caps.to_wire();
+        let canonical = serde_json::to_vec(&wire).expect("capability map serializes");
+        let token = format!("{:016x}", xxhash_rust::xxh3::xxh3_64(&canonical));
```
taylordotfish: Replace with `add_many` as we talked about

crates/atuin-domain/src/caps/mod.rs
```diff
@@ -0,0 +1,140 @@
...
+            .get(C::static_name())
+            .and_then(|cap| {
+                let cap: &dyn Any = &**cap;
+                cap.downcast_ref::<C>()
+            })
+            .cloned()
+    }
+
+    /// Serialize every advertised capability into a name -> JSON map, in sorted key order.
+    fn to_wire(&self) -> BTreeMap<String, serde_json::Value> {
```
taylordotfish: Just return a serde json Value directly here

crates/atuin-domain/src/caps/mod.rs
```diff
@@ -0,0 +1,140 @@
...
+/// `dyn Capability`. The type-level name lives in [`static_name`](Capability::static_name), gated on
+/// `Self: Sized` so it stays off the vtable; [`name`](Capability::name) is its object-safe
+/// counterpart, for when only a `dyn Capability` is in hand.
+pub trait Capability: Any + Send + Sync {
+    /// The name this capability is indexed by on the wire, eg `sh.atuin.server/records.batch`.
+    fn static_name() -> &'static str
+    where
+        Self: Sized;
+
+    /// The name of this capability. Implementors return [`Self::static_name`].
```
taylordotfish: we need to fix this lol

crates/atuin-domain/src/caps/mod.rs
```diff
@@ -0,0 +1,145 @@
...
+            .insert(CapKey(cap.name().to_string()), cap);
+    }
+
+    /// Check whether this node advertises the given capability.
+    pub fn get<C: Capability + Clone>(&self) -> Option<C> {
+        self.caps
+            .read()
+            .get(C::static_name())
+            .and_then(|cap| {
+                let cap: &dyn Any = &**cap;
```
taylordotfish: can we make this clearer somehow

crates/atuin-domain/src/caps/mod.rs
```diff
@@ -0,0 +1,145 @@
...
+    fn name(&self) -> &'static str;
+
+    /// Convert this capability's associated data into a JSON value.
+    fn json(&self) -> Result<serde_json::Value, serde_json::Error>;
+}
+
+/// The capabilities a node advertises about itself.
+#[derive(Default)]
+pub struct CapsBundle {
+    caps: RwLock<BTreeMap<CapKey, Box<dyn Capability>>>,
```
taylordotfish: Make this `BTreeSet<Box<dyn Capability>>` where they're ordered by the name or type id. You may need to wrap `Box<dyn Capability>` in a newtype.

crates/atuin-domain/src/caps/server.rs
```diff
@@ -0,0 +1,272 @@
...
+    pub fn token(&self) -> &str {
+        // `bake` writes ASCII hex, so this is always valid UTF-8.
+        self.token.to_str().expect("token is valid UTF-8")
+    }
+
+    /// The pre-serialized capabilities document, served verbatim by the capabilities endpoint.
+    /// Deserializes into a [`crate::api::CapabilitiesResponse`].
+    pub fn body(&self) -> &str {
+        // `bake` writes JSON from `serde_json`, so this is always valid UTF-8.
+        self.body.to_str().expect("body is valid UTF-8")
```
taylordotfish: why not store `body` as a `String` rather than a `BString` so we don't have to do this?

crates/atuin-domain/src/caps/server.rs
```diff
@@ -0,0 +1,272 @@
...
+    /// The capabilities this server advertises, for typed introspection.
+    pub fn caps(&self) -> &CapsBundle {
+        &self.caps
+    }
+
+    /// Decide whether a request whose client echoed `known` is current.
+    ///
+    /// Absent (`None`) or matching tokens are [`Negotiation::Current`]; only a present *differing*
+    /// token is [`Negotiation::Stale`]. A client that sends no token is therefore never rejected.
+    pub fn negotiate(&self, known: Option<&str>) -> Negotiation {
```
taylordotfish, line 111: The name "negotiate" is a little confusing to me -- it sounds like it should have a side effect but it's just a pure check. I don't really have a good idea for a better name though

crates/atuin-domain/src/caps/axum.rs
```diff
@@ -0,0 +1,206 @@
...
+        .into_response()
+}
+
+/// Capability-negotiation middleware for the sync API.
+///
+/// Rejects with `412` + `X-Atuin-Capabilities-Available` **only** when the request carries
+/// `X-Atuin-Capabilities-Known` with a token differing from ours. Absent or matching tokens pass
+/// straight through, so pre-capabilities clients are never affected. A non-UTF-8 known header is
+/// treated as absent.
+pub async fn negotiate(
```
taylordotfish: Does this have to / should this be public?

crates/atuin-domain/src/caps/middleware.rs
```diff
@@ -0,0 +1,321 @@
...
+        // Refresh (coalesced across concurrent callers) and retry exactly once with the new token.
+        self.caps.refresh(&self.http).await?;
+        let fresh = self.caps.known_token();
+        if let Some(value) = fresh.as_deref().and_then(|t| HeaderValue::from_str(t).ok()) {
+            retry_req
+                .headers_mut()
+                .insert(HeaderName::from_static(KNOWN_HEADER), value);
+        }
+
+        next.run(retry_req, ext).await
```
taylordotfish: In the case where `refresh == true`, we retry the request but we don't actually modify the request in any way based on the new capabilities we received (you need `refresh == false` for that).

What about modifying the `refresh == true` case so it avoids actually resending the request? Instead, it sends "it's ok if my caps don't match yours, just do the thing I requested but send me your new caps as part of your response" to the server.

crates/atuin-domain/src/caps/middleware.rs
```diff
@@ -0,0 +1,321 @@
...
+    }
+}
+
+/// Install capability negotiation onto a [`reqwest::Client`].
+pub trait CapabilitiesExt {
+    /// Wrap this client so it negotiates capabilities.
+    fn with_capabilities(self, caps: Arc<CapClient>, refresh: bool) -> ClientWithMiddleware;
```
taylordotfish: also if we could take the opportunity to make this an enum instead of bool that would be nice

crates/atuin-domain/src/caps/mod.rs
markovejnovic: Pulled the upcast + downcast into a small documented helper, so `get` reads as "look up by name, downcast to the requested type":

```rust
/// Recover the concrete type behind a stored capability.
///
/// `Capability: Any`, so a `&dyn Capability` upcasts to `&dyn Any`, which then downcasts to the
/// requested type -- yielding `None` if the stored capability is a different type.
fn downcast<C: Capability>(cap: &dyn Capability) -> Option<&C> {
    let cap: &dyn Any = cap;
    cap.downcast_ref::<C>()
}
```

`get` is then:

```rust
self.caps
    .read()
    .get(C::static_name())
    .and_then(|cap| downcast::<C>(cap.as_ref()))
    .cloned()
```

The `&**cap` / bare `&dyn Any` mechanics now live in one named place. Went with a free helper rather than an `as_any` trait method to keep every capability impl free of that boilerplate.

crates/atuin-domain/src/caps/mod.rs
markovejnovic: Done -- `CapsBundle` now holds a `BTreeSet<CapEntry>`, where `CapEntry(Box<dyn Capability>)` orders and `Borrow<str>`s by `name()`. The name is the wire identifier, so it's the set's real key and there's no separate `CapKey` on this side anymore.

Two details worth calling out:

- Ordered by **name** rather than `TypeId`, so `BTreeSet::get(&str)` via `Borrow<str>` still works for lookups (a `TypeId` ordering would be inconsistent with the borrowed `str`).
- `add_dyn` uses `.replace()` rather than `.insert()` -- `BTreeSet::insert` keeps the *existing* equal element, but we want the latest registration with a given name to win.

crates/atuin-domain/src/caps/server.rs
markovejnovic: Agreed -- reverted `token` and `body` to `String` and dropped the `bstr` dependency. The token is always ASCII hex and the body always UTF-8 JSON, so there was no non-UTF-8 case that `BString` bought us. The accessors go back to plain `&self.token` / `&self.body`, no `to_str()`.

crates/atuin-domain/src/caps/axum.rs
markovejnovic: No -- made it private (`async fn negotiate`). Nothing outside this module referenced it; the router installs it through `negotiate_capabilities`. Also dropped the `[negotiate]` intra-doc link from that method's docs so a public item no longer links to a private one.

crates/atuin-domain/src/caps/mod.rs
```diff
@@ -95,10 +95,50 @@ fn downcast<C: Capability>(cap: &dyn Capability) -> Option<&C> {
...
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for CapEntry {
+    fn cmp(&self, other: &Self) -> Ordering {
+        self.name().cmp(other.name())
+    }
+}
+
```
taylordotfish, line 137: derive_more?

crates/atuin-domain/src/caps/middleware.rs
```diff
@@ -49,50 +63,64 @@ impl Middleware for CapMiddleware {
...
+                .map(str::to_owned);
+            if let Some(available) = advertised {
+                // `refresh_if_stale` is coalesced and idempotent, so a burst of stale responses
+                // drives exactly one fetch. Best-effort: a refresh failure must not fail the
+                // request the server already served.
+                let caps = self.caps.clone();
+                let http = self.http.clone();
+                tokio::spawn(async move {
+                    let _ = caps.refresh_if_stale(&http, &available).await;
+                });
```
taylordotfish, line 93: add another check if stale to avoid spawn in most cases

crates/atuin-common/src/sync/single_flight.rs
```diff
@@ -0,0 +1,319 @@
...
+        }
+
+        // We are the leader: run the fetch with no data lock held, so reads stay concurrent.
+        let fresh = Arc::new(fetch().await?);
+
+        // Brief critical section, no `.await` inside -- never hold a parking_lot guard across one.
+        *self.value.write() = Some(fresh.clone());
+        self.generation.fetch_add(1, Ordering::Release);
+
+        Ok(fresh)
```
taylordotfish: implement in terms of `refresh_if` with `|| true` as `needed` or remove?

crates/atuin-domain/src/caps/axum.rs
```diff
@@ -0,0 +1,254 @@
...
+        Negotiation::Stale => {
+            // The client did not ask us to enforce, so serve the request and tell it our token.
+            let mut response = next.run(request).await;
+            if let Some(value) = available {
+                response
+                    .headers_mut()
+                    .insert(HeaderName::from_static(AVAILABLE_HEADER), value);
+            }
+            response
+        }
```
taylordotfish:
```suggestion
        Negotiation::Stale => {
            let mut response = if enforce {
                (StatusCode::PRECONDITION_FAILED, "capabilities out of date").into_response()
            } else {
                // The client did not ask us to enforce, so serve the request and tell it our token.
                next.run(request).await
            };
            if let Some(value) = available {
                response
                    .headers_mut()
                    .insert(HeaderName::from_static(AVAILABLE_HEADER), value);
            }
            response
        }
```

crates/atuin-domain/src/caps/client.rs
```diff
@@ -0,0 +1,236 @@
...
+    pub async fn refresh_if_stale(
+        &self,
+        client: &reqwest::Client,
+        available: &str,
+    ) -> reqwest::Result<()> {
+        self.server
+            .refresh_if(
+                || self.known_token().as_deref() != Some(available),
+                || self.fetch_server_caps(client),
+            )
```
taylordotfish: turns out this is just a mutex, get rid of singleflight

taylordotfish: check if stale -> acquire mutex -> check if stale again -> fetch

taylordotfish (approved):

PR #3800, opened 2026-07-30
refactor(server): use Arc<dyn Database> instead of a DB type parameter

markovejnovic:
## What

`atuin-server` was generic over the `Database` trait — `AppState<DB>`, `router<DB>`, the `UserAuth` extractor, and every handler carried a `<DB: Database>` parameter, monomorphized twice. But the trait has **two live implementations** (`Postgres`, `Sqlite`) selected at runtime by the `db_uri` scheme, so the generic was never resolved to a single type in practice.

This collapses the compile-time generic into a single `Arc<dyn Database>` behind dynamic dispatch. Both backends are preserved.

## Changes

- **`atuin-server-database`**: make `Database` object-safe — drop the `Sized + Clone` supertraits and the `new()` method.
- **`atuin-server-postgres` / `atuin-server-sqlite`**: move `new()` from the trait impl into an inherent `impl`.
- **`atuin-server`**: `AppState`, `router`, `UserAuth`, all handlers, and `launch`/`launch_with_tcp_listener`/`make_router` lose `<DB>`. `AppState` holds `Arc<dyn Database>` (clones as a refcount bump, matching the shared-pool semantics). Backend selection (`postgres://` vs `sqlite:`) moves out of `main.rs` into a single `connect()` helper.
- Update the two test harnesses; drop the now-unused `atuin-server-postgres` dev-dependency from `atuin`.

No behaviour change: same backends, same runtime selection.

## Verification

- `cargo check --workspace --all-targets` — clean
- `cargo clippy` on affected crates + all targets — clean
- `cargo test -p tests-database --test db_story` — passes via the `Arc<dyn Database>` path on the sqlite backend

Not run locally (needs a live Postgres): the Postgres integration test and the `atuin` sync tests. They compile.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -231,7 +231,6 @@ dependencies = [
  "atuin-scripts",
  "atuin-server",
  "atuin-server-database",
- "atuin-server-postgres",
  "clap",
  "clap_complete",
  "clap_complete_nushell",
--- a/crates/atuin-server-database/src/lib.rs
+++ b/crates/atuin-server-database/src/lib.rs
@@ -84,9 +84,7 @@ impl Debug for DbSettings {
 }
 
 #[async_trait]
-pub trait Database: Sized + Clone + Send + Sync + 'static {
-    async fn new(settings: &DbSettings) -> DbResult<Self>;
-
+pub trait Database: Send + Sync + 'static {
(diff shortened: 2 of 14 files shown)


markovejnovic:
no net improvement

PR #3802, opened 2026-07-30
chore(deps): bump frizbee to 0.12.0

saghen:
Hey folks, this PR bumps frizbee from [v0.11.0 -> v0.12.0](https://github.com/saghen/frizbee/compare/v0.11.0...v0.12.0) which includes fixes for unicode, u8/backend selection, and some remaining arithmetic panics. I've also added a `config.scoring.max_needle_len()` to avoid the scoring panic, but haven't wired that up here since you already have a sane 512 byte limit.

Would you be interested in a follow-up PR adding opt-in support for typo resistance? In blink.cmp, we do this via `max_typos = Some(needle.len() / 4)`

Thanks for adopting the project and for the really generous sponsorship! It's pretty surreal seeing my fuzzy matcher end up in projects I use daily :)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1889,9 +1889,9 @@ dependencies = [
 
 [[package]]
 name = "frizbee"
-version = "0.11.0"
+version = "0.12.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a151c9ed9ab8cefc76a76b7a15292fc83575f70cf2076d238e0a3251f69738ba"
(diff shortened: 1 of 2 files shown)


markovejnovic:
Thank you @saghen!

markovejnovic:
As for 

> Would you be interested in a follow-up PR adding opt-in support for typo resistance? In blink.cmp, we do this via max_typos = Some(needle.len() / 4)

Gladly!

ellie:
> Thanks for adopting the project and for the really generous sponsorship!

You are very welcome! Thank you for frizbee, it's been a very nice change!

If you think there's any way we could use it more effectively, please lmk! Very open to feedback

PR #3806, opened 2026-07-30
refactor: adopt enum_dispatch for enum trait delegation

markovejnovic:
Use `enum_dispatch` for one less pointer indirection

### DIFF
--- a/crates/atuin-client/src/auth.rs
+++ b/crates/atuin-client/src/auth.rs
@@ -1,6 +1,6 @@
 use std::collections::HashMap;
 
-use ambassador::{delegatable_trait, Delegate};
+use enum_dispatch::enum_dispatch;
 use eyre::{Context, Result, bail};
 use reqwest::{StatusCode, Url, header::USER_AGENT};
 use serde::Deserialize;
@@ -43,7 +43,7 @@ pub enum MutateResponse {
 ///
 /// CLI commands use this trait so they don't need to know which backend is
 /// active — they just prompt for input and call these methods.
-#[delegatable_trait]
+#[enum_dispatch]
 #[allow(async_fn_in_trait)]
 pub trait IsAuthClient: Send + Sync {
     /// Log in with username + password, optionally providing a TOTP code.
@@ -74,8 +74,7 @@ pub trait IsAuthClient: Send + Sync {
 }
 
 /// Static-dispatch enum over the two auth backends.
-#[derive(Delegate)]
-#[delegate(IsAuthClient)]
+#[enum_dispatch(IsAuthClient)]
 pub enum AuthClient {
     Legacy(LegacyAuthClient),
     Hub(HubAuthClient),
--- a/crates/atuin-client/src/auth.rs
+++ b/crates/atuin-client/src/auth.rs
@@ -1,6 +1,6 @@
 use std::collections::HashMap;
 
-use async_trait::async_trait;
+use ambassador::{delegatable_trait, Delegate};
 use eyre::{Context, Result, bail};
 use reqwest::{StatusCode, Url, header::USER_AGENT};
 use serde::Deserialize;
@@ -43,7 +43,8 @@ pub enum MutateResponse {
 ///
 /// CLI commands use this trait so they don't need to know which backend is
 /// active — they just prompt for input and call these methods.
-#[async_trait]
+#[delegatable_trait]
+#[allow(async_fn_in_trait)]
 pub trait AuthClient: Send + Sync {
     /// Log in with username + password, optionally providing a TOTP code.
     async fn login(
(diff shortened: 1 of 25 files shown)


crates/atuin-client/src/auth.rs
```diff
@@ -43,8 +43,9 @@ pub enum MutateResponse {
 ///
 /// CLI commands use this trait so they don't need to know which backend is
 /// active — they just prompt for input and call these methods.
-#[async_trait]
-pub trait AuthClient: Send + Sync {
+#[delegatable_trait]
+#[allow(async_fn_in_trait)]
```
taylordotfish: ?

crates/atuin-daemon/src/daemon.rs
```diff
@@ -237,6 +239,16 @@ pub trait Component: Send + Sync {
     async fn stop(&mut self) -> Result<()>;
 }
 
+/// Static-dispatch enum over the daemon components.
+#[derive(Delegate, derive_more::From)]
+#[delegate(IsComponent)]
+pub enum Component {
```
taylordotfish: Perhaps?

* `IsComponent` → `Component`
* `Component` → `AnyComponent`

crates/atuin-client/src/auth.rs
```diff
@@ -44,8 +44,11 @@ pub enum MutateResponse {
 /// CLI commands use this trait so they don't need to know which backend is
 /// active — they just prompt for input and call these methods.
 #[enum_dispatch]
+// enum_dispatch can only delegate a bare `async fn` (not the `-> impl Future +
+// Send` form), and this trait is public and called across crates by the CLI, so
+// the async_fn_in_trait lint cannot be resolved by reducing visibility here.
```
taylordotfish: This comment isn't really addressing the lint.

Here's what `async_fn_in_trait` says:

```
warning: use of `async fn` in public traits is discouraged as auto trait bounds cannot be specified
  = note: you can suppress this lint if you plan to use the trait only in your own code, or do not care about auto traits like `Send` on the `Future`
  = note: `#[warn(async_fn_in_trait)]` on by default
help: you can alternatively desugar to a normal `fn` that returns `impl Future` and add any desired bounds such as `Send`, but these cannot be relaxed without a breaking API change
  |
2 -     async fn f();
2 +     fn f() -> impl std::future::Future<Output = ()> + Send;
  |
```

Let's update the comment to address why it's ok for us to have a public trait with functions that return a non-`Send` future.

crates/atuin-daemon/Cargo.toml
```diff
@@ -14,12 +14,12 @@ readme.workspace = true
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-atuin-client = { path = "../atuin-client", version = "18.19.0-beta.1" }
-atuin-common = { path = "../atuin-common", version = "18.19.0-beta.1" }
+atuin-client = { path = "../atuin-client", version = "18.19.0-beta.2" }
+atuin-common = { path = "../atuin-common", version = "18.19.0-beta.2" }
```
taylordotfish, line 18: ?

markovejnovic: image

crates/atuin/src/command/client/search/engines.rs
```diff
@@ -14,19 +14,19 @@ pub mod db;
...
             // Fall back to fuzzy mode if daemon feature is not enabled
-            SearchEngine::Db(db::Search(DbSearchMode::Fuzzy))
+            AnySearchEngine::Db(db::Search(DbSearchMode::Fuzzy))
         }
-        SearchMode::Prefix => SearchEngine::Db(db::Search(DbSearchMode::Prefix)),
-        SearchMode::FullText => SearchEngine::Db(db::Search(DbSearchMode::FullText)),
-        SearchMode::Fuzzy => SearchEngine::Db(db::Search(DbSearchMode::Fuzzy)),
+        SearchMode::Prefix => AnySearchEngine::Db(db::Search(DbSearchMode::Prefix)),
+        SearchMode::FullText => AnySearchEngine::Db(db::Search(DbSearchMode::FullText)),
+        SearchMode::Fuzzy => AnySearchEngine::Db(db::Search(DbSearchMode::Fuzzy)),
```
taylordotfish: I think we can change these to `.into()` calls; `enum_dispatch` derives `From` for all the inner types.

crates/atuin-daemon/Cargo.toml
markovejnovic: not sure what you mean by this, the version looks correct

crates/atuin-daemon/src/daemon.rs
```diff
@@ -416,8 +435,8 @@ impl DaemonBuilder {
     /// Register a component.
     ///
     /// Components are started in registration order and stopped in reverse order.
-    pub fn component(mut self, component: impl Component + 'static) -> Self {
-        self.components.push(Box::new(component));
+    pub fn component(mut self, component: impl Into<AnyComponent>) -> Self {
+        self.components.push(component.into());
         self
```
taylordotfish: Dunno if this would work because it's kind of recursive, but can we add `Into<AnyComponent>` as a supertrait of `Component` and then make this `impl Component`?

markovejnovic: this pr is becoming too much manual back and forth. it's clear to me that the amount of work both on your and my side for something of marginal value is not worth it

crates/atuin-client/src/auth.rs
```diff
@@ -43,7 +43,16 @@ pub enum MutateResponse {
...
+#[enum_dispatch]
+// rustc's `async_fn_in_trait` lint: `async fn` in a publicly-reachable trait
+// doesn't promise the returned futures are `Send`. The lint's own guidance is to
+// allow it when the trait is "used only in your own code" / you "do not care
+// about auto traits like `Send` on the `Future`"; every call goes through the
+// concrete `AnyAuthClient` enum (never a generic `T: AuthClient` or `dyn`), so
+// the missing `Send` bound is never relied on. Its suggested fix — desugaring to
+// `-> impl Future + Send` — isn't available: enum_dispatch can only delegate a
+// bare `async fn`.
+#[allow(async_fn_in_trait)]
```
taylordotfish:
```suggestion
#[allow(async_fn_in_trait, reason = "only used within our code and we don't need it to be Send")]
```

taylordotfish (approved):

PR #3825, opened 2026-07-31
feat(search): remove the "skim" engine

taylordotfish:
Remove the "skim" engine in favor of "daemon-fuzzy". If somemone has `search_mode = "skim"`, this PR will fall back to "fuzzy" but show a warning in the interactive search UI encouraging the user to switch to daemon-fuzzy.

This PR also changes the text of `atuin search --help` very slightly: previously, we accepted both `fulltext` and `full-text` as search modes, but listed `full-text` as the canonical spelling in the help message. This PR changes the canonical spelling to `fulltext` while still accepting both variants. The reasoning for this is that `fulltext` is the only valid spelling in config.toml.

skim-warning


markovejnovic (approved):

ltrzesniewski:
Argh, I use skim on Windows, which works better than the default fuzz. I can't use the daemon... 😞

Is there a good reason to remove it?

markovejnovic:
Pinging @taylordotfish for 

> Is there a good reason to remove it?

I know that skim has some pretty severe performance problems -- I profiled it yesterday and noticed more than 300ms of stutter using it, compared to either of the two other options.

As for:

> I can't use the daemon

Why is that? The daemon is definitely the best way to use Atuin (in terms of performance and search accuracy). If you have any issues with it, please ping me -- they're top priority!

ltrzesniewski:
@markovejnovic AFAIK the daemon is not implemented for Windows. Did I miss some huge news?

ltrzesniewski:
@ellie could you please reconsider this? Atuin just got *much* worse for me, favoring 3 month old commands over stuff from 2 minutes ago... 😢 

image

xidiot:
I'm also trying to figure out how to make my results always chronologically ordered. I've read [here](https://forum.atuin.sh/t/search-output-order-is-not-what-i-expected/108) that `skim` might help, but now I see it was removed :\

PR #3832, opened 2026-08-01
Swap vt100 for rio-vt

raphamorim:
This swaps the terminal emulation backend from vt100 to [rio-vt](https://crates.io/crates/rio-vt), the terminal core we extracted from [Rio](https://github.com/raphamorim/rio).

Three crates used vt100 (atuin-common's `ansi` module, atuin-ai's live command preview, atuin-pty-proxy's screen mirror). Since the latter two already depend on atuin-common's `ansi` feature, the emulator now lives in one place — a new `atuin_common::vt` module exposing the small parser/screen surface those callers consume — and atuin-ai/atuin-pty-proxy drop their emulator dependency entirely.

Notes on the wrapper:

- `Screen::contents()` reproduces vt100's soft-wrap semantics (wrapped rows join without a newline) via rio-vt's per-row wrap flag, so `ansi::to_plain_text` behaves identically — the existing rstest/proptest suite passes unchanged.
- `rows_formatted()` for the pty-proxy wire format is re-implemented as a small SGR serializer over rio-vt's style tables (bold/dim/italic/underline/inverse/hidden/strike, 16/256/truecolor). Each row starts from default attributes and resets at the end, so the client-side behavior of writing rows straight to stdout is preserved. The module has its own tests.
- One unrelated one-liner: rio-vt ships a `PartialOrd<Column> for usize` impl, which made a bare `.into()` in `interactive.rs` ambiguous — now `usize::from(...)`.

A nice side effect worth mentioning: atuin-common's `never_panics_and_strips_controls` proptest immediately caught a real panic in rio-vt (`step_by(0)` in graphics insertion when cell dimensions are unset, i.e. exactly the headless configuration atuin runs). That's fixed and released in rio-vt 0.5.4, with the found seed committed to `proptest-regressions/` here — so this PR's fuzzing hardened the engine before the swap landed.

Why rio-vt: actively maintained, safe Rust, correct unicode width + grapheme clustering, and it drops the unmaintained `net2` transitive dependency (RUSTSEC-2020-0016) that vt100 pulls in. Benchmarks against vt100 and alacritty_terminal: https://github.com/raphamorim/rio-vt-benchmark.

Verified: full workspace builds; atuin-common (329 tests incl. all proptests), atuin-ai (341), and atuin-pty-proxy (74) all pass; clippy clean on the touched crates.


markovejnovic:
Thank you for your PR but we have decided to proceed with our [`vt100-rust`](https://github.com/atuinsh/vt100-rust) fork.

PR #3833, opened 2026-08-01
fix: ignore empty or relative XDG base directory variables

M1lan:
## Problem

`config_dir()`, `data_dir()` and `runtime_dir()` in `atuin-common` use the raw value of `XDG_CONFIG_HOME` / `XDG_DATA_HOME` / `XDG_RUNTIME_DIR` without validating it. The [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/latest/) requires that empty or non-absolute values be ignored.

With `XDG_CONFIG_HOME` set but empty, `config_dir()` returns the relative path `atuin`, and `Settings::get_config_path()` then `create_dir_all()`s it — leaving a stray empty `./atuin` directory in whatever cwd the process ran from.

Reproduce:

```sh
cd /tmp && XDG_CONFIG_HOME= atuin info >/dev/null; ls -d atuin
```

## Fix

Shared `xdg_dir()` helper: use the env var only when it parses to an absolute path, otherwise fall back to the spec default. Applied to `config_dir()`, `data_dir()` and `runtime_dir()`.

Unit tests added for the empty and relative cases (wired into the existing sequential `test_dirs` runner).

### DIFF
--- a/crates/atuin-common/src/utils.rs
+++ b/crates/atuin-common/src/utils.rs
@@ -106,21 +106,31 @@ pub fn home_dir() -> PathBuf {
         .expect("could not determine home directory")
 }
 
+/// Resolve an XDG base directory environment variable.
+///
+/// Per the XDG Base Directory Specification, values that are empty or not
+/// absolute paths are invalid and must be ignored, falling back to the
+/// default. Honoring a relative value would make the directory depend on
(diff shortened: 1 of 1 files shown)


taylordotfish:
Hi @M1lan, thanks for the PR. The non-compliance has now been fixed in https://github.com/atuinsh/atuin/pull/3932 and https://github.com/atuinsh/atuin/pull/3910.

PR #3852, opened 2026-08-04
fix(nix): add an RPATH for OpenSSL so the built binary runs

phinze:
Since #3807 moved TLS from rustls to native-tls, the Nix build produces an `atuin` that can't start:

```
$ nix build github:atuinsh/atuin/v18.19.0
$ ./result/bin/atuin --version
./result/bin/atuin: error while loading shared libraries: libssl.so.3: cannot open shared object file: No such file or directory
```

The binary needs libssl.so.3 and libcrypto.so.3, but it has an empty RPATH and OpenSSL isn't in its runtime closure, so nothing puts it on the loader's path.

What took a while to see is that the build is perfectly happy. The `preBuild` LD_LIBRARY_PATH added in #3807 is still exported when `postInstall` runs `$out/bin/atuin gen-completions`, so completions generate fine and `nix build` goes green. It only breaks once you run the installed binary, which CI never does.

This adds the RPATH in postFixup, guarded to Linux since Darwin doesn't use one. Afterwards OpenSSL shows up in the runtime closure and the binary runs in an empty environment.

I also added a `./result/bin/atuin --version` step to the Nix workflow. Happy to drop it if you'd rather keep this to the fix, but as it stands `nix build` can't catch this class of bug.

Reproduced with `nix build github:atuinsh/atuin/v18.19.0`; main carries the same atuin.nix. Fix built and tested against main (18.20.0-beta.1) on x86_64-linux/NixOS. I haven't tested Darwin.

## Checks

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -33,6 +33,9 @@ jobs:
       - name: Run nix build
         run: nix build --print-build-logs
 
+      - name: Check the built binary runs
+        run: ./result/bin/atuin --version
+
   nixfmt:
     runs-on: depot-ubuntu-24.04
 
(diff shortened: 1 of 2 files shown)


ellie (approved):
thank you!

PR #3870, opened 2026-08-07
feat: Add turn time and tips to Atuin AI

BinaryMuse:
This PR adds a tip system to Atuin AI, allowing the app to call out lesser known features.

`crates/atuin-ai/src/tui/tips.rs` defines `TIPS`, which is a `const &[Tip]`. Each tip has an ID, text, and a relevance callback (if it turns false, the tip is skipped).

This PR has a few default tips, but we'll certainly want to think through these.

Additionally, Atuin AI now ends a turn with the total time; this is where the tip hangs when a turn is complete.

image

Closes ATU-550

### DIFF
--- a/crates/atuin-ai/src/tui/tips.rs
+++ b/crates/atuin-ai/src/tui/tips.rs
@@ -1,0 +1,179 @@
+//! Feature tips surfaced under the turn spinner and the "Responded in"
+//! line. One tip is pulled per turn and rides through both.
+
+use atuin_client::settings::Settings;
+
+/// Facts a tip's relevance predicate may consult. Assembled fresh at each
+/// pull so predicates see current session state, not startup state.
+pub(crate) struct TipContext<'a> {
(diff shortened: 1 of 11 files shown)


crates/atuin-ai/src/tui/tips.rs
```diff
@@ -0,0 +1,219 @@
...
+    /// Stable key — the hook for cross-session "already seen" persistence.
+    #[cfg_attr(not(test), allow(dead_code))]
+    pub id: &'static str,
+    /// Rendered after a "Tip: " prefix; keep it to one short sentence.
+    pub text: &'static str,
+    /// `None` = always relevant.
+    pub relevant: Option<fn(&TipContext) -> bool>,
+}
+
+pub(crate) const TIPS: &[Tip] = &[
```
BinaryMuse, line 27: Tips start here

PR #3874, opened 2026-08-08
chore(tests): fix flaky pty-proxy test

taylordotfish:
Fix flakiness in `a_stalled_client_does_not_wedge_the_socket_server`. The root cause is the unreliability of terminal echo; under load, partial lines can be written, causing the pty screen not to contain as many characters as expected. This PR turns off terminal echo on the pty used in tests.


BinaryMuse (approved):
✨

PR #3876, opened 2026-08-08
fix(common): survive wide characters on a one-column screen

BinaryMuse:
`never_panics_and_strips_controls` proptest with a double-width character (e.g. `⺀`) rendered onto a one-column screen panics inside vt100 with `attempt to subtract with overflow` (`vt100-0.16.2/src/screen.rs:730` computes `size.cols - width` unchecked). Proptest biases toward edge values, so `cols = 1` comes up often enough to flake regularly — it failed a run on #3859 today with minimal input `[226, 186, 128]`, `cols = 1`.

The fix clamps the emulated screen to two columns in `to_plain_text`, which only affects the degenerate one-column case. The regression seed is checked in so the proptest replays it deterministically, and an explicit unit test documents the vt100 quirk.


taylordotfish:
Upstream issue in vt100: https://github.com/doy/vt100-rust/issues/37

BinaryMuse:
Is there any merit to merging this, or are we just waiting for upstream?

taylordotfish:
I think it's worth merging, let's just add a link to the upstream issue in the code though with a TODO to remove our hack once upstream is fixed

PR #3879, opened 2026-08-09
feat: enable syntax highlighting on illumos

papertigers:
illumos platforms support tree-sitter so enable syntax highlighting in atuin command history.

## Checks

### DIFF
--- a/crates/atuin/Cargo.toml
+++ b/crates/atuin/Cargo.toml
@@ -136,7 +136,7 @@ daemonize = "0.5.0"
 # Enable tree-sitter shell parsing on platforms where tree-sitter's bundled C
 # compiles cleanly. tree-sitter 0.26's portable/endian.h fails on illumos,
 # Windows cross-compiles, and potentially other exotic targets.
-[target.'cfg(any(target_os = "linux", target_os = "macos", target_os = "windows"))'.dependencies]
+[target.'cfg(any(target_os = "linux", target_os = "macos", target_os = "windows", target_os = "illumos"))'.dependencies]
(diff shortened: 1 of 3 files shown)


crates/atuin/Cargo.toml
```diff
@@ -136,7 +136,7 @@ daemonize = "0.5.0"
 # Enable tree-sitter shell parsing on platforms where tree-sitter's bundled C
 # compiles cleanly. tree-sitter 0.26's portable/endian.h fails on illumos,
 # Windows cross-compiles, and potentially other exotic targets.
-[target.'cfg(any(target_os = "linux", target_os = "macos"))'.dependencies]
+[target.'cfg(any(target_os = "linux", target_os = "macos", target_os = "illumos"))'.dependencies]
```
papertigers: The comment above needs to be adjusted still

papertigers:
This change also reworks the illumos test in GitHub actions. Do we want to pull it out into its own PR? 

In an effort to keep CI time down i have chosen to only built the client and the server.

ellie (approved):
nice one! ty for keeping illumos in good shape

papertigers:
@ellie thanks for driving this over the line, I had some other stuff come up I needed to get to first.

ellie:
do you know anywhere we would be able to get fast illumos runners btw? would love to speed up CI there and unsure of best practices

papertigers:
@ellie unfortuantely not that I know of 😕. I don't want to burden the project with long CI times, we could switch the job to just a cargo check rather than a build. The vmactions thing is using nested virt which is why it's a little slow. 

Another option is setting up buildomat which is what we use at work for CI. I can see what spare hardware I have lying around and maybe set that up. 

Of course open to other suggestions!

PR #3883, opened 2026-08-09
feat: add windows/powershell highlighting

ltrzesniewski:
This adds highlighting on Windows along with the PowerShell syntax:

image

I don't know why there were comments saying that tree-sitter doesn't build on Windows, or why there's no PowerShell grammar. The build went smoothly both on Windows and Linux, and I simply referenced the `tree-sitter-powershell` crate. ¯\\\_(ツ)\_/¯

I'm not entirely satisfied with this yet, since `Meaning::SyntaxFlag` isn't supported. I'll handle that later.

## Checks

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -306,6 +306,7 @@ dependencies = [
  "tree-sitter",
  "tree-sitter-bash",
  "tree-sitter-fish",
+ "tree-sitter-powershell",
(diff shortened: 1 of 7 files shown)


ellie (approved):
looks good, thank you for doing this! I actually didn't think tree sitter worked on windows :O

PR #3892, opened 2026-08-10
fix(ai): execute commands from enter in the bash integration

BinaryMuse:
In the bash AI integration, pressing enter in the TUI inserted the chosen command instead of executing it, just like tab.

This adopts the two-step macro dispatch the main bash integration already uses for `enter_accept`: `?` expands to two intermediate sequences, the first runs the handler via `bind -x`, and the handler rebinds the second to `accept-line` to execute, or a no-op otherwise. The handler resets the chain binding on every invocation so a stale `accept-line` from a previous execute can't fire on a later `?` press. The sequences use `\C-x\C-_B<n>\a`, mirroring the main integration's `\C-x\C-_A<n>\a` without colliding with it. Because the command now goes through readline's normal accept path, it lands in bash history and fires bash-preexec, so atuin history records it too.

Bash <= 4.2 cannot `bind -x` key sequences longer than two bytes, so it keeps the direct binding; enter degrades to insert there, now documented instead of silently broken. Also removes the unused `__atuin_ai_accept_line` / `_atuin_ai_question_mark_wrapper` scaffolding.

Verified in an interactive bash 5.3 under a pty with a stubbed `atuin`: the execute marker runs the command (and only then), a mid-line `?` after an execute inserts literally, and the insert marker leaves the line unexecuted.

Fixes ATU-516

### DIFF
--- a/crates/atuin-ai/src/commands/init.rs
+++ b/crates/atuin-ai/src/commands/init.rs
@@ -100,6 +100,12 @@ pub fn generate_bash_integration() -> &'static str {
     r#"
 # Question mark at start of line - natural language mode
 _atuin_ai_question_mark() {
+    # Reset the chain sequence to a no-op; the execute branch below rebinds
+    # it to accept-line so readline runs the command after this handler
+    # returns.  Without the reset, a stale accept-line binding from a
+    # previous execute would fire on the next '?' press.
+    bind '"\C-x\C-_B0\a": ""'
+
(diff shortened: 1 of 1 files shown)


taylordotfish (approved):
Confirmed fixed on Bash 5.2.37 on Debian trixie

BinaryMuse:
> Confirmed fixed on Bash 5.2.37 on Debian trixie

Thank you!

PR #3894, opened 2026-08-10
fix(search): display daemon errors without exiting

### LINKED ISSUE #3891: [Bug]: daemon-fuzzy exits on first query when daemon is unavailable

When `search_mode = "daemon-fuzzy"` is configured but the daemon is unavailable and autostart is disabled, Atuin should give an actionable error before or inside the search UI. The message could suggest starting the daemon, enabling `daemon.autostart`, or switching to `search_mode = "fuzzy"`.

Falling back to local fuzzy search with a warning would also be reasonable. The TUI should not disappear without explaining the configuration problem.

### What happened?

The interactive search UI opens normally. Typing the first character closes it with exit status 1. Through the Nushell keybinding this looks like the history UI crashed, with no useful error shown.

I reproduced this with the real TUI in a PTY:

```text
mode=daemon-fuzzy alive_before_key=true alive_after_key=false exit_status=1
mode=fuzzy        alive_before_key=true alive_after_key=true
```

The initial empty query succeeds because it uses the database path. The first non-empty query reaches the daemon search engine, fails to connect, and propagates the error out of the render loop.

### Steps to reproduce

1. Configure Atuin without a running or autostarted daemon:

   ```toml
   search_mode = "daemon-fuzzy"

   [daemon]
   enabled = false
   autostart = false
   ```

2. Run `atuin search --interactive`.
3. Type one character.

The process exits with status 1. Changing only `search_mode` to `"fuzzy"` keeps the TUI open.

This is related to #3274 and #3561, but those cover an incompatible daemon and a daemon killed after autostart. This report covers the configuration where no daemon is available and Atuin provides no actionable feedback.

### Atuin doctor output

```yaml
{
  "atuin": {
    "version": "18.19.0",
    "commit": "NO_GIT",
    "daemon_enabled": false
  },
  "shell": {
    "name": "nu",
    "default": "nu"
  },
  "system": {
    "os": "Arch Linux",
    "arch": "x86_64"
  }
}
```

### Code of Conduct

tianrking:
Fixes #3891.

When `daemon-fuzzy` cannot connect, the query error escapes the render loop and closes the interactive search UI. This keeps the UI open, clears stale results, and shows the error at the bottom. Retryable daemon failures also get an actionable message and reset the cached client so a later query can reconnect.

Tested with:
- `cargo +1.97.1 test -p atuin unavailable_daemon_keeps_search_open_with_error -- --nocapture`
- `cargo +1.97.1 test -p atuin command::client::search::interactive::tests -- --nocapture` (47 passed)
- `cargo +1.97.1 test -p atuin command::client::search::engines::daemon::tests -- --nocapture` (4 passed)
- `cargo +1.97.1 fmt --all -- --check`
- `git diff --check`

## Checks

### DIFF
--- a/crates/atuin/src/command/client/search/engines/daemon.rs
+++ b/crates/atuin/src/command/client/search/engines/daemon.rs
@@ -195,7 +195,7 @@ impl SearchEngine for Search {
         let span =
             span!(Level::TRACE, "daemon_search.req_resp", query = %query, query_id = query_id);
 
-        let mut stream = self
+        let stream = self
             .client
             .try_with_autostart(&self.params, async |client| {
                 let search_params = SearchParams {
@@ -207,7 +207,18 @@ impl SearchEngine for Search {
                 };
                 client.search(search_params).await
             })
-            .await?;
+            .await;
+
+        let mut stream = match stream {
+            Ok(stream) => stream,
+            Err(err) if LazyClient::should_retry(&err) => {
+                self.client.0 = None;
+                return Err(err.wrap_err(
+                    "daemon unavailable; start it, enable daemon.autostart, or use search_mode = \"fuzzy\"",
+                ));
+            }
+            Err(err) => return Err(err),
+        };
 
         let mut ids = Vec::with_capacity(200);
         span!(Level::TRACE, "daemon_search.resp")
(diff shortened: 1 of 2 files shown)


ellie:
It would be much better to display an error! I don't think silently falling back to something worse is a good UX.

 Please confirm if you read and followed our CONTRIBUTING.md? This PR desc feels pretty agent-y.

tianrking:
Yes—I read and followed `CONTRIBUTING.md` before opening this. You're right that the fallback changes search semantics without making that visible. I've updated the PR so the UI stays open and displays an actionable daemon error instead; the fallback is removed.

taylordotfish:
Closing now that #3895 is merged.

PR #3896, opened 2026-08-10
refactor(client): own sync_addr via Arc<Url>, drop Client lifetime

markovejnovic:
There are some nasty `Arc`s inside of `reqwest::Client`, so the value of holding a reference to `Url` is so marginal over just adding an `Arc`

### DIFF
--- a/crates/atuin-client/src/api_client.rs
+++ b/crates/atuin-client/src/api_client.rs
@@ -1,5 +1,6 @@
 use std::collections::HashMap;
 use std::env;
+use std::sync::Arc;
 use std::time::Duration;
 
 use eyre::{Result, bail};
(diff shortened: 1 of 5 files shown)


taylordotfish (approved):
looks good to merge after rustfmt issue fixed

crates/atuin-client/src/api_client.rs
```diff
@@ -240,9 +241,9 @@ async fn handle_resp_error(resp: Response) -> Result<Response> {
     Ok(resp)
 }
 
-impl<'a> Client<'a> {
+impl Client {
     pub fn new(
-        sync_addr: &'a Url,
+        sync_addr: &Url,
```
taylordotfish: we could have this be `impl Into<Arc<Url>>` so callers can avoid a clone?

PR #3916, opened 2026-08-13
fix: 'atuin login' prints 401 failures until authorized

BinaryMuse:
`atuin_client::hub::handle_resp_error` was previously hard-coded to `error!` any time an HTTP request had a non-success response, in addition to returning a `Err(eyre!(String))`.

This PR changes `handle_resp_error` to return a properly typed `HubError` with three variants:

* `Status { status: StatusCode, reason: Option<String> }` for regular HTTP errors
* `Request(reqwest::Error)` for errors from Reqwest
* `Url(atuin_common::url::UrlAppendError)` for URL construction errors

Verified that all call sites handle the error appropriately, and updated the login loop to check for 401s and treat them as pending.

Closes ATU-585

### DIFF
--- a/crates/atuin-client/src/hub.rs
+++ b/crates/atuin-client/src/hub.rs
@@ -65,14 +65,19 @@ pub enum HubError {
 }
 
 fn status_message(status: StatusCode, reason: Option<&str>) -> String {
-    if status == StatusCode::SERVICE_UNAVAILABLE {
-        "Service unavailable: check https://status.atuin.sh".to_string()
-    } else if status == StatusCode::TOO_MANY_REQUESTS {
-        "Rate limited; please wait before trying again".to_string()
-    } else if let Some(reason) = reason {
-        format!("Hub error: {status} - {reason}")
-    } else {
-        format!("Hub request failed with status: {status}")
+    match status {
+        StatusCode::SERVICE_UNAVAILABLE => {
+            "Service unavailable: check https://status.atuin.sh".to_string()
+        }
+        StatusCode::TOO_MANY_REQUESTS => {
+            "Rate limited; please wait before trying again".to_string()
+        }
+        status if let Some(reason) = reason => {
+            format!("Hub error: {status} - {reason}")
+        }
+        status => {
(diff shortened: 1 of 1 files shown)


crates/atuin-client/src/hub.rs
```diff
@@ -44,6 +45,37 @@ pub enum HubAuthStatus {
...
+    if status == StatusCode::SERVICE_UNAVAILABLE {
+        "Service unavailable: check https://status.atuin.sh".to_string()
+    } else if status == StatusCode::TOO_MANY_REQUESTS {
+        "Rate limited; please wait before trying again".to_string()
+    } else if let Some(reason) = reason {
+        format!("Hub error: {status} - {reason}")
+    } else {
+        format!("Hub request failed with status: {status}")
+    }
+}
```
ellie: I'd probably rewrite this as a match

crates/atuin-client/src/hub.rs
```diff
@@ -44,6 +45,42 @@ pub enum HubAuthStatus {
...
+            "Rate limited; please wait before trying again".to_string()
+        }
+        status if let Some(reason) = reason => {
+            format!("Hub error: {status} - {reason}")
+        }
+        status => {
+            format!("Hub request failed with status: {status}")
+        }
+    }
+}
```
taylordotfish: we can avoid some allocations using `std::fmt::from_fn` (not a blocking issue though)

```suggestion
fn status_message(status: StatusCode, reason: Option<&str>) -> impl std::fmt::Display {
    std::fmt::from_fn(move |f| match status {
        StatusCode::SERVICE_UNAVAILABLE => {
            write!(f, "Service unavailable: check https://status.atuin.sh")
        }
        StatusCode::TOO_MANY_REQUESTS => {
            write!(f, "Rate limited; please wait before trying again")
        }
        status if let Some(reason) = reason => {
            write!(f, "Hub error: {status} - {reason}")
        }
        status => {
            write!(f, "Hub request failed with status: {status}")
        }
    })
}
```

PR #3921, opened 2026-08-14
Store history.db UUIDs as 16-byte blobs

ellie:
## Checks

## Summary

`history.db` stored the `id` and `session` UUIDs as 32-character hex strings. This PR stores them as raw 16-byte blobs instead, halving their footprint in the table and in the two indexes that contain them (the `id` primary key and `idx_history_session_timestamp`).

Following the plan:

1. **Read paths handle either form.** The row mapper checks the value's storage class and hex-decodes blobs back to the usual simple-format string, so `History` is unchanged in memory (and on the sync wire). Lookups (`load`, `load_active`, `update`, `delete_rows`) and session filters (`list`, `search`, `stats`) match the value against both its binary and legacy text forms, which stays an index-range lookup.
2. **The write path always writes bytes** — with one deliberate exception: only strings in the one format atuin generates (32 lowercase hex chars, `Uuid::as_simple`) are converted, so decoding always round-trips exactly. An arbitrary user-set `ATUIN_SESSION` is stored as text, as before.
3. **`atuin store rebuild history` converts the rest.** Since the rebuild inserts-or-ignores, pre-existing rows would stay text forever; a new `Database::convert_uuids_to_binary` rewrites them in place (`unhex()`, guarded to exactly the strings the write path would convert) and vacuums so the space is actually returned to the filesystem.

No schema migration is needed: SQLite's TEXT column affinity stores blob values unmodified.

One subtlety: `Paged` used a single descending `id` cursor, but in SQLite's ordering every blob sorts after every text value, and the decoded cursor no longer says which form the row was stored in — a naive cursor would skip all text rows after passing the smallest binary id. It now pages in two phases (binary ids, then text ids), split on the sargable `id > x''` / `id < x''` boundary.

## Testing

- New unit tests: binary round-trip, non-UUID sessions staying text, legacy-text-row load/update/delete/search/stats interop, conversion (including its no-op second run), and mixed-form pagination.
- Full workspace suite passes (`cargo nextest run --workspace`, 2089 tests incl. the Postgres-backed server tests), plus clippy with CI flags, `cargo fmt --check`, and the rustdoc check.
- End-to-end with the built binary: recorded history lands as blobs, hand-inserted legacy text rows show up in `history list`/`search` alongside them, and `atuin store rebuild history` converts them in place with ids preserved. On a 50k-row database the file shrank from 23.3 MB to 20.0 MB (~14%; short test commands, so table savings are dominated by the fixed ~64 bytes/row across table + indexes).

<div><a href="https://cursor.com/agents/bc-307f84bd-baba-40a2-a0b8-fea8f6c6b89d?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-307f84bd-baba-40a2-a0b8-fea8f6c6b89d&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/crates/atuin-client/src/database.rs
+++ b/crates/atuin-client/src/database.rs
@@ -14,7 +14,7 @@ use fs_err as fs;
 use itertools::Itertools;
 use sql_builder::{SqlBuilder, SqlName, bind::Bind, esc, quote};
 use sqlx::{
-    Result, Row,
+    Result, Row, TypeInfo, ValueRef,
     sqlite::{
         SqliteConnectOptions, SqliteJournalMode, SqlitePool, SqlitePoolOptions, SqliteRow,
         SqliteSynchronous,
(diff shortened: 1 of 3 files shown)


crates/atuin-client/src/database.rs
```diff
@@ -178,6 +178,78 @@ fn apply_shell_filter(sql: &mut SqlBuilder, shells: OrFilter<&[String]>) {
...
+    if value.len() != 32
+        || !value
+            .bytes()
+            .all(|b| matches!(b, b'0'..=b'9' | b'a'..=b'f'))
+    {
+        return None;
+    }
+
+    Uuid::try_parse(value).ok().map(Uuid::into_bytes)
+}
```
markovejnovic, line 199: can we use a rich type for this? that's my general feedback for this. I'd like to see some sort of `UuidV7` type that has a custom `Serialize/Deserialize`

PR #3925, opened 2026-08-15
fix(wrapped): panic when every command is in stats.ignored_commands

VXNCXNX:
`atuin wrapped` panics if every command in the year is in `stats.ignored_commands`.

```
$ cat ~/.config/atuin/config.toml
[stats]
ignored_commands = ["cd", "ls"]

$ atuin wrapped 2026
thread 'main' panicked at crates/atuin/src/command/client/wrapped.rs:348:52:
Failed to compute stats
```

After:

```
Every command in your 2026 history is in stats.ignored_commands, so there is nothing to wrap up!
```

`compute` returns `None` in two cases: no history, and history where everything was filtered out. `wrapped` guards the first with an `is_empty()` check and then `.expect()`s the second, which is reachable whenever the ignore list covers the whole range. `atuin stats` handles the same `None` with `if let Some(stats)` and exits cleanly, so the two commands disagree on identical data.

## Checks

## Testing

Reproduced with the config above and a single `cd /tmp` entry. With one non-ignored command added, `wrapped` prints normally before and after, and empty history still gives the existing "history is empty" message.

`all_commands_ignored` in `crates/atuin-history/src/stats.rs` sits next to the existing `ignored_commands` test, which covers one ignored plus one kept command and so stops just short of this case. It asserts `compute` returns `None` for non-empty, wholly-ignored history, which is the precondition `wrapped` was assuming away.

Gates: `cargo test -p atuin-history` 38 passed, `cargo fmt --check` clean, `cargo clippy --tests -- -D warnings -D clippy::redundant_clone` clean. Built with `--features atuin-client/vendored-tls`, since this machine has no system OpenSSL.

*AI disclosure: I used Claude Code on this. I found the panic by reading the two call sites of `compute`, reproduced it against a real database, and ran every command above myself.*


markovejnovic:
This looks correct! Thank you for your PR! Will slightly adjust and merge!

PR #3926, opened 2026-08-15
fix(bash): localize "READLINE_{LINE,POINT}" in Bash <= 3.2

### LINKED ISSUE #3924: [Bug]: Bash 3.2 Ctrl-R leaves READLINE_* set, causing bash-preexec to stop recording history

Using `Ctrl-R` to invoke Atuin search should not interfere with recording subsequent commands in Atuin history.

In particular, after invoking `Ctrl-R` and then either:

* cancelling with `Esc`,
* selecting a command with `Tab` for editing, or
* selecting/executing a command with `Enter`,

subsequent commands typed normally at the Bash prompt should continue to be recorded by Atuin.

Environment:
~~~
macOS 26.5.2, Apple Silicon
Apple Bash 3.2.57
Atuin 18.18.1
bash-preexec backend
~~~

Atuin is initialized with:

~~~
eval "$(atuin init bash --disable-up-arrow --disable-ai)"
~~~

I am deliberately using Apple Bash 3.2 rather than a newer Bash, and am not using `ble.sh`.

### What happened?

After any use of `Ctrl-R`, subsequent commands stop being recorded by Atuin.

I was able to reduce this to a deterministic interaction between Atuin's Bash-3.2 integration and `bash-preexec`.

In a fresh shell, before using `Ctrl-R`:
~~~
$ declare -p READLINE_LINE READLINE_POINT 2>/dev/null
$ echo ATUIN_BASELINE
ATUIN_BASELINE
$ atuin history list --format '{command}' | grep -Fx 'echo ATUIN_BASELINE'
echo ATUIN_BASELINE
~~~

`READLINE_LINE` and `READLINE_POINT` are initially unset, and Atuin records commands normally.

Then I press `Ctrl-R`, open Atuin search, and cancel with `Esc`.

Afterward:
~~~
$ declare -p READLINE_LINE READLINE_POINT 2>/dev/null
declare -- READLINE_LINE=""
declare -- READLINE_POINT="0"

$ echo ATUIN_AFTER_CTRLR
ATUIN_AFTER_CTRLR
$ atuin history list --format '{command}' | grep -Fx 'echo ATUIN_AFTER_CTRLR'
$
~~~
The command is present in normal Bash history but is not recorded by Atuin.

The same thing occurs after `Ctrl-R` → Tab and `Ctrl-R` → Enter. For example, after selecting `pwd` with `Tab`:
~~~
$ declare -p READLINE_LINE READLINE_POINT 2>/dev/null
declare -- READLINE_LINE="pwd"
declare -- READLINE_POINT="3"
~~~

and subsequent commands again cease to be recorded.

Manually unsetting these variables immediately restores Atuin recording:
~~~
$ unset READLINE_LINE READLINE_POINT

$ echo ATUIN_AFTER_UNSET
ATUIN_AFTER_UNSET
$ atuin history list --format '{command}' | grep -Fx 'echo ATUIN_AFTER_UNSET'
echo ATUIN_AFTER_UNSET
~~~

## Apparent cause

`bash-preexec` currently contains this guard in its DEBUG/preexec machinery:
~~~
if [[ -n "${COMP_POINT:-}" || -n "${READLINE_POINT:-}" ]]; then
    # We're in the middle of a completer or a keybinding set up by "bind -x".
    return
fi
~~~
This guard was added to avoid treating `bind -x` callbacks as ordinary interactive commands. In this case, however, `READLINE_POINT` remains nonempty after Atuin's keybinding has completed, so `bash-preexec` subsequently suppresses every real interactive preexec hook.

Looking at Atuin 18.18.1's Bash integration, this seems particularly relevant to the Bash <4 code path. `__atuin_history` explicitly does:
~~~
[[ ${BLE_ATTACHED-} ]] || ((BASH_VERSINFO[0] >= 4)) ||
    READLINE_LINE="" READLINE_POINT=0
~~~
before entering the search. If the search is cancelled, the function returns without unsetting these variables. When a command is selected, the Bash-3.2 implementation needs `READLINE_LINE` to transfer the selected command into the real Readline editing buffer via macros, but the variables likewise remain set afterward.

## Tested workaround

I tested a proof-of-concept workaround which cleans up the variables after Atuin's Bash-3.2 Readline macros have consumed them, and then re-arms `bash-preexec`:
~~~
__atuin_bash32_cleanup()
{
    unset READLINE_LINE READLINE_POINT
    __bp_preexec_interactive_mode="on"
}

bind -m emacs -x '"\C-xT": __atuin_bash32_cleanup'

__atuin_macro_accept_line='"\C-x\C-_A10\a\C-xT\C-x\C-_A5\a"'
__atuin_macro_insert_line='"\C-x\C-_A10\a\C-x\C-_A6\a\C-xT"'

__atuin_widget_run()
{
    local data=${__atuin_widget[$1]}
    local keymap=${data%%:*} widget=${data#*:}
    local __atuin_macro_chain_keymap=$keymap

    bind -m "$keymap" "\"$__atuin_macro_chain\": \"\\C-xT\""

    builtin eval -- "$widget"
}
~~~

With this workaround installed, I tested all of:

* `Ctrl-R` → `Esc`
* `Ctrl-R` → select → `Tab` → edit/execute or cancel
* `Ctrl-R` → select → `Enter`

multiple successive `Ctrl-R` invocations using different exit paths

After each, `READLINE_LINE` and `READLINE_POINT` were unset and subsequent commands were recorded correctly by Atuin.

The need to set:
~~~
__bp_preexec_interactive_mode="on"
~~~
is important: simply invoking another cleanup `bind -x` which unsets the variables itself consumes/disturbs bash-preexec's interactive state, and the following command is still skipped. Re-arming that state makes the proof-of-concept work reliably.

This appears related to #1003, especially the report there that `Ctrl-R` → `Esc` causes the next command not to be added to history, and to the `bash-preexec` `bind -x` handling discussed in Atuin #2059.

The relevant `bash-preexec` `READLINE_POINT` test was itself introduced to address Atuin/#1003, so simply removing that guard does not look like the right solution. The issue seems instead to be that Atuin's Bash-3.2 path leaves the special Readline variables populated after the keybinding has finished.

### Atuin doctor output

```yaml
Atuin Doctor
Checking for diagnostics


{
  "atuin": {
    "version": "18.18.1",
    "commit": "NO_GIT",
    "sync": null,
    "sqlite_version": "3.46.0",
    "daemon_enabled": false
  },
  "shell": {
    "name": "bash",
    "default": "bash",
    "plugins": [
      "atuin",
      "bash-preexec"
    ],
    "preexec": "bash-preexec"
  },
  "system": {
    "os": "Darwin",
    "arch": "arm64",
    "version": "26.5.2",
    "disks": [
      {
        "name": "Macintosh HD",
        "filesystem": "apfs"
      },
      {
        "name": "Macintosh HD",
        "filesystem": "apfs"
      }
    ]
  }
}
```

### Code of Conduct

akinomyoga:
Fixes #3924

I've originally localized those variables in commit a7bed6146 (#1533), but after that, I decided to leave the change to global variables in commit a6abc717b (#2953) for macro chaining in Bash 3.2 to expand the value later. However, I missed that it conflicted with the change in bash-preexec https://github.com/rcaloras/bash-preexec/pull/152.

The first and second commits refactor the code, and the third commit is the fix for the issue.

I checked that the problem reported in #3924 happens with the `main` branch in Bash 3.2, and then I checked that the changes in this PR fix the problem in Bash 3.2.

## Checks

### DIFF
--- a/crates/atuin-ai/src/commands/init.rs
+++ b/crates/atuin-ai/src/commands/init.rs
@@ -148,24 +148,30 @@ _atuin_ai_question_mark() {
     fi
 }
 
-# A "bind -x" handler cannot invoke readline bindable functions such as
-# accept-line, so binding '?' directly to the handler could only ever edit
-# READLINE_LINE, never execute it.  We use the same two-step macro dispatch
-# as atuin's main bash integration: '?' expands to two intermediate
-# sequences, the first runs the handler via "bind -x", and the handler
-# decides what the second does by rebinding it — accept-line to execute, a
-# no-op otherwise.  The sequences \C-x\C-_B<n>\a mirror the main
-# integration's \C-x\C-_A<n>\a chain, with B instead of A to avoid
-# colliding with it.
-if ((BASH_VERSINFO[0] >= 5 || BASH_VERSINFO[0] == 4 && BASH_VERSINFO[1] >= 3)); then
-    bind '"\C-x\C-_B0\a": ""'
-    bind -x '"\C-x\C-_B1\a": _atuin_ai_question_mark'
-    bind '"?": "\C-x\C-_B1\a\C-x\C-_B0\a"'
-else
-    # Bash <= 4.2 cannot "bind -x" a key sequence longer than two bytes,
-    # so bind the handler directly.  Enter in the TUI then inserts the
-    # command instead of executing it.
-    bind -x '"?": _atuin_ai_question_mark'
+# We only set up keybinding in Bash >= 4.0.  Bash < 4.0 does not provide
+# READLINE_LINE, so if bound, it would always start an AI session regardless of
+# the contents of the line buffer.  This means that it would make impossible
+# to input "?" in Bash 3.2.
+if ((BASH_VERSINFO[0] >= 4)) || [[ ${BLE_VERSION-} ]]; then
+    # A "bind -x" handler cannot invoke readline bindable functions such as
+    # accept-line, so binding '?' directly to the handler could only ever edit
+    # READLINE_LINE, never execute it.  We use the same two-step macro dispatch
+    # as atuin's main bash integration: '?' expands to two intermediate
+    # sequences, the first runs the handler via "bind -x", and the handler
+    # decides what the second does by rebinding it — accept-line to execute, a
+    # no-op otherwise.  The sequences \C-x\C-_B<n>\a mirror the main
+    # integration's \C-x\C-_A<n>\a chain, with B instead of A to avoid
+    # colliding with it.
+    if ((BASH_VERSINFO[0] >= 5 || BASH_VERSINFO[0] == 4 && BASH_VERSINFO[1] >= 3)); then
+        bind '"\C-x\C-_B0\a": ""'
+        bind -x '"\C-x\C-_B1\a": _atuin_ai_question_mark'
+        bind '"?": "\C-x\C-_B1\a\C-x\C-_B0\a"'
+    else
+        # Bash <= 4.2 cannot "bind -x" a key sequence longer than two bytes,
+        # so bind the handler directly.  Enter in the TUI then inserts the
+        # command instead of executing it.
+        bind -x '"?": _atuin_ai_question_mark'
+    fi
 fi
 "#
     .trim()
--- a/crates/atuin/src/shell/atuin.bash
+++ b/crates/atuin/src/shell/atuin.bash
@@ -193,6 +193,20 @@ __atuin_clear_prompt() {
     printf '%s' "${__atuin_clear_prompt_cache[offset]}"
 }
 
+# Insert $1 into the line buffer of the line editor.  Special handling is
+# needed in Bash <= 3.2 because the old Bash does not support READLINE_LINE and
+# READLINE_POINT.  When this function is called from inside a keybinding set up
+# by `atuin-bind`, the contents is inserted into the command line by macro
+# chaining.
+__atuin_insert_line() {
+    local __atuin_command=$1
+    READLINE_LINE=$__atuin_command
+    READLINE_POINT=${#READLINE_LINE}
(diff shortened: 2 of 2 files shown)


akinomyoga:
I added a fix for Atuin AI integration (5afd0bca94388e1469aefa8a62251717d358a100). The AI integration also had the same issue, but in the first place, if the keybinding for the AI integration is enabled, it would make it impossible to input `?` in Bash 3.2. That's because `_atuin_ai_question_mark` always triggers the AI mode regardless of the existing contents of the line buffer as the shell variable `READLINE_LINE` is unavailable in Bash 3.2. The commit disables the Atuin AI integration in Bash <= 3.2.

akinomyoga:
@dubiousjim If you have time, you can check the behavior of this branch.

dubiousjim:
> @dubiousjim If you have time, you can check the behavior of this branch.

Tested commit `f601e210a77a4ab46ceec9d96c18332d187461b7` on macOS 26.5.2 arm64 with Apple Bash 3.2.57 and `bash-preexec`.

I tested the same cases I used to reproduce #3924, with my local workaround completely disabled:

* normal command recording baseline
* `Ctrl-R` → `Esc`
* `Ctrl-R` → select → `Tab` → execute/edit
* `Ctrl-R` → select → `Enter`
* several successive `Ctrl-R` invocations using different exit paths

All passed. After the `Ctrl-R` interactions,
~~~
declare -p READLINE_LINE READLINE_POINT 2>/dev/null
~~~
produced no output, and subsequent commands continued to be recorded by Atuin in every case.

I tested using the locally built PR binary throughout (`atuin 18.20.0-beta.3`, commit `f601e210...`) with isolated Atuin databases and without my workaround from #3924.

So this fixes the problem for me.

dubiousjim:
Thanks for your attention to this issue.

taylordotfish (approved):
Thank you!

akinomyoga:
Thank you!

PR #3927, opened 2026-08-15
refactor(bash): remove redundant check of existing preexec-backend

akinomyoga:
The check is not just redundant, but it's not working as expected...

The present detection using `__atuin_update_preexec_backend` does not work as expected.  First of all, the function
`__atuin_update_preexec_backend` intends to detect the active status of the preexec backend, so it is not suitable for checking the existence of a backend at the session initialization stage.  In particular, the function `__atuin_update_preexec_backend` intentionally does not detect ble.sh even if ble.sh is loaded, because ble.sh is not activated at the initialization stage.

To check whether ble.sh is loaded, one needs to explicitly check the shell variable `$BLE_VERSION`.  However, all the complication is actually unnecessary because bash-preexec.sh already has the necessary duplicate detection for an early return.  See the added code comment for the details of the mechanism.

## Checks

### DIFF
--- a/crates/atuin/src/shell/atuin.bash
+++ b/crates/atuin/src/shell/atuin.bash
@@ -711,10 +711,20 @@ fi
 
 if command -v __atuin_load_builtin_preexec > /dev/null; then
     if [[ -z ${ATUIN_NO_BUILTIN_PREEXEC-} ]]; then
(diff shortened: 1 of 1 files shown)


taylordotfish:
Thank you @akinomyoga!

PR #3933, opened 2026-08-17
fix: fix panics from `vt100::Parser::new`

### LINKED ISSUE #3930: [Bug]: pty-proxy clamps the vt100 grid to 1x1, but 2x2 is the panic-free floor

Opening a dropdown ("quake style") terminal with `atuin pty-proxy` in the shell init should show a prompt and nothing else, and the screen mirror should serve a snapshot at the real terminal size.

### What happened?

A window manager creates that kind of terminal unmapped, so `terminal::size()` in `runtime.rs` returns `0, 0`. `ParserState::new` clamps that to `1x1`:

```rust
// crates/atuin-pty-proxy/src/screen.rs:259 (main)
let (rows, cols) = (rows.max(1), cols.max(1));
```

But `1x1` is itself a panicking geometry: the first wrapping write lands in `col_wrap` — the exact path `vt100_guarded`'s own doc comment names. So in this configuration the guard isn't absorbing a rare surprise, it's on the normal startup path of every new session, and the panic hook's report lands in the user's first screenful, as the comment at `screen.rs:281-283` anticipates ("The panic hook's one-line report still reaches stderr; noisy").

I measured vt100 0.16.2's real floor by driving `atuin pty-proxy` in a PTY pinned at each size and writing output that wraps (500 ASCII columns, then 200 double-width glyphs):

|            | cols=1            | cols=2 | cols=3 | cols=80         |
| ---------- | ----------------- | ------ | ------ | --------------- |
| **rows=1** | `grid.rs:689`     | `grid.rs:689` | `grid.rs:689` | `grid.rs:689` |
| **rows=2** | `screen.rs:896`   | ok     | ok     | ok              |
| **rows=3** | `screen.rs:827`   | ok     | ok     | ok              |
| **rows=24**| `screen.rs:827`   | ok     | ok     | ok              |

`grid.rs:689` is the `.unwrap()` inside `pub fn col_wrap`. Two things fall out of the matrix:

- **`rows=1` panics at every width**, including 80, so flooring rows at 1 buys nothing.
- **2x2 is clean in every direction.**

So flooring at `(rows.max(2), cols.max(2))` in `ParserState::new` and in the `Msg::Resize` arm would take the whole degenerate-size class out of the guard's hands and let `vt100_guarded` go back to covering only genuine surprises. It would also line the floor up with #3876, which already picked two columns in `atuin_common::ansi::to_plain_text` for this same vt100 quirk.

### Reproduction

```python
import os, pty, fcntl, termios, struct, select, time

pid, fd = pty.fork()
if pid == 0:
    os.execvp("atuin", ["atuin", "pty-proxy", "--shell", "/bin/zsh"])
fcntl.ioctl(fd, termios.TIOCSWINSZ, struct.pack("HHHH", 1, 1, 0, 0))  # rows, cols
time.sleep(1)
os.write(fd, b"printf 'A%.0s' {1..500}; printf '\\n'\n")

out, t = b"", time.time()
while time.time() - t < 2:
    if select.select([fd], [], [], 0.1)[0]:
        out += os.read(fd, 65536)
print([l for l in out.decode(errors="replace").splitlines() if "panicked" in l])
```

Set the winsize to `0, 0` instead and it reproduces the original path, since the clamp turns it into the same `1x1` grid.

### Notes on versions

The numbers above are from 18.18.1 (the current nixpkgs build), which predates `vt100_guarded`. There the parser thread simply dies, so the snapshot socket serves nothing for the rest of the shell's life — the search popup silently loses its restore-over-scrollback behaviour:

```
start 0x0     panic=True   screen mirror: got 0B
start 120x30  panic=False  screen mirror: got 365B (screen 120x30)
```

On `main` I'd expect the guard to turn that into repeated hook output plus a `1x1` mirror until the first real resize arrives. I read `main` rather than building it, so please treat that half as inference — the clamp/floor point stands either way.

Real-world setup where I hit this: GNOME with the quake-terminal extension launching WezTerm, `pty-proxy` initialised from `.zshrc`. Anything that spawns a terminal hidden should behave the same.

### Atuin doctor output

```yaml
{
  "atuin": {
    "version": "18.18.1",
    "commit": "NO_GIT",
    "sync": {
      "auto_sync": true,
      "last_sync": "2026-08-17 10:27:36.474115475 +00:00:00"
    },
    "sqlite_version": "3.46.0",
    "daemon_enabled": true
  },
  "shell": {
    "name": "bash",
    "default": "zsh",
    "plugins": [
      "atuin"
    ],
    "preexec": null
  },
  "system": {
    "os": "NixOS",
    "arch": "x86_64",
    "version": "26.11",
    "disks": [
      {
        "name": "/dev/nvme0n1p2",
        "filesystem": "ext4"
      },
      {
        "name": "/dev/nvme0n1p1",
        "filesystem": "vfat"
      }
    ]
  }
}
```

(`shell.name` reads `bash` because `atuin doctor` was run from a non-interactive shell; the affected sessions are zsh, per `shell.default`.)

### Code of Conduct

taylordotfish:
`vt100` has an [upstream issue](https://github.com/doy/vt100-rust/issues/37) where panics can occur if `Parser::new` is called with a `rows` or `cols` parameter less than 2. This PR adds a wrapper function that clamps the values so they are never less than 2.

Fixes #3930.

### DIFF
--- a/clippy.toml
+++ b/clippy.toml
@@ -1,8 +1,31 @@
-disallowed-methods = [
-    { path = "std::time::Duration::new", reason = "panics on overflow; use atuin_common::time::DurationExt::try_new, or Duration::from_secs when nanos are 0" },
-    { path = "time::Duration::new", reason = "panics on overflow; use atuin_common::time::DurationExt::try_new" },
-    { path = "std::time::Duration::from_secs_f64", reason = "panics on NaN/negative/overflow; use Duration::try_from_secs_f64" },
-    # https://github.com/time-rs/time/issues/802
-    { path = "time::OffsetDateTime::from_unix_timestamp_nanos", reason = "wraps out-of-range input to a valid-looking wrong date (time-rs/time#802); use atuin_common::time::OffsetDateTimeExt::from_unix_nanos" },
-    { path = "time::UtcDateTime::from_unix_timestamp_nanos", reason = "same truncation bug as OffsetDateTime (time-rs/time#802); range-check the full i128 before converting" },
-]
+[[disallowed-methods]]
+path = "std::time::Duration::new"
+reason = """panics on overflow; use atuin_common::time::DurationExt::try_new, or \
+    Duration::from_secs when nanos are 0"""
+
+[[disallowed-methods]]
+path = "time::Duration::new"
+reason = "panics on overflow; use atuin_common::time::DurationExt::try_new"
+
+[[disallowed-methods]]
+path = "std::time::Duration::from_secs_f64"
+reason = "panics on NaN/negative/overflow; use Duration::try_from_secs_f64"
+
+# https://github.com/time-rs/time/issues/802
+[[disallowed-methods]]
+path = "time::OffsetDateTime::from_unix_timestamp_nanos"
+reason = """wraps out-of-range input to a valid-looking wrong date (time-rs/time#802); use \
+    atuin_common::time::OffsetDateTimeExt::from_unix_nanos"""
+
+[[disallowed-methods]]
+path = "time::UtcDateTime::from_unix_timestamp_nanos"
+reason = """same truncation bug as OffsetDateTime (time-rs/time#802); range-check the full i128 \
+    before converting"""
+
+[[disallowed-methods]]
+path = "vt100::Parser::new"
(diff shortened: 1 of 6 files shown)


markovejnovic (approved):
lgtm

PR #3936, opened 2026-08-18
chore(ai): remove obsolete `atuin ai init` command

taylordotfish:
As per discussion with @ellie, the `atuin ai init` command is no longer needed as the initialization for Atuin AI is now included in `atuin init`. This PR removes the obsolete command; its existence was complicating other fixes such as #3928.

As part of this change, the shell init snippets are now exposed as simple constant strings rather than functions.


markovejnovic (approved):
diff itself looks good, haven't confirmed existing logic for this exists, but since you say it does, i trust you

PR #3943, opened 2026-08-18
refactor: Add a `CmdOrigin` Type

markovejnovic:
Add a new `CmdOrigin` type which is a type-safe idea around the `<host>:<user>` pair.


crates/atuin-daemon/src/client.rs
```diff
@@ -299,7 +299,7 @@ impl From<Context> for RpcSearchContext {
         RpcSearchContext {
             session_id: context.session,
             cwd: context.cwd,
-            hostname: context.hostname,
+            hostname: context.cmd_origin,
```
markovejnovic: have to keep the `hostname` field here unfortunately since `RpcSearchContext` goes over the wire

crates/atuin-domain/src/lib.rs
```diff
@@ -67,3 +67,176 @@ macro_rules! new_uuid {
...
+    fn as_ref(&self) -> &str {
+        self.0.as_ref()
+    }
+}
+
+impl<T: AsRef<str>> CmdHost<T> {
+    pub fn owned(&self) -> CmdHost<String> {
+        CmdHost(self.0.as_ref().to_string())
+    }
+}
```
taylordotfish: Unless it would cause issues in generic code I would make this `T: Into<String>`

crates/atuin-domain/src/lib.rs
```diff
@@ -67,3 +67,176 @@ macro_rules! new_uuid {
...
+
+impl CmdHost<String> {
+    pub fn probe() -> Self {
```
taylordotfish: Still not a huge fan of this name; I would have to read the code to figure out what it does

crates/atuin-domain/src/lib.rs
```diff
@@ -67,3 +67,176 @@ macro_rules! new_uuid {
...
+            .map(Self)
+            .unwrap_or_default()
+    }
+}
+
+impl<'a> CmdHost<&'a str> {
+    pub fn as_str(&self) -> &'a str {
+        self.0
+    }
+}
```
taylordotfish: Why not implemented for all `CmdHost<T: AsRef<str>>`?

crates/atuin-domain/src/lib.rs
```diff
@@ -67,3 +67,176 @@ macro_rules! new_uuid {
 pub mod api;
 pub mod caps;
 pub mod record;
+
+/// A hostname, generic over its backing storage so it can be an owned
+/// `CmdHost<String>` or a zero-copy `CmdHost<&str>` view into a [`CmdOrigin`].
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, derive_more::Display, derive_more::From)]
```
taylordotfish: derive Copy

crates/atuin-domain/src/lib.rs
```diff
@@ -67,3 +67,176 @@ macro_rules! new_uuid {
...
+pub struct CmdHost<T: AsRef<str> = String>(T);
+
+impl<T: AsRef<str>> AsRef<str> for CmdHost<T> {
+    fn as_ref(&self) -> &str {
+        self.0.as_ref()
+    }
+}
+
+impl<T: AsRef<str>> CmdHost<T> {
+    pub fn owned(&self) -> CmdHost<String> {
```
taylordotfish: make this `into_owned`

crates/atuin-domain/src/record/cmd_origin.rs
```diff
@@ -0,0 +1,169 @@
...
+    pub fn probe() -> Self {
+        Self::new(CmdHost::probe(), CmdUser::probe())
+    }
+
+    /// The host portion, as a zero-copy view.
+    pub fn host(&self) -> CmdHost<&str> {
+        CmdHost(&self.raw[..self.sep])
+    }
+
+    /// The user portion, as a zero-copy view (empty when there is no `:`).
```
taylordotfish: Comment needs clarification; there is always a `:` in `raw`

crates/atuin-client/src/history/builder.rs
```diff
@@ -119,7 +123,7 @@ impl From<HistoryFromDb> for History {
             cwd: from_db.cwd,
             duration: from_db.duration,
             session: from_db.session,
-            hostname: from_db.hostname,
+            cmd_origin: CmdOrigin::parse_fuzzy(from_db.hostname),
```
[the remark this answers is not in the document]
markovejnovic: I have fixed the filter

crates/atuin-client/src/import/nu_histdb.rs
```diff
@@ -45,7 +46,10 @@ impl From<HistDbEntry> for History {
             .exit(histdb_item.exit_status)
             .duration(histdb_item.duration_ms)
             .session(format!("{:x}", histdb_item.session_id))
-            .hostname(String::from_utf8_lossy(&histdb_item.hostname).into_owned());
+            .cmd_origin(CmdOrigin::new(
+                CmdHost::from(String::from_utf8_lossy(&histdb_item.hostname).into_owned()),
+                CmdUser::default(),
+            ));
```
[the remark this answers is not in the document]
markovejnovic: Historically it was wrong -- we used to use the `hostname` as the `CmdOrigin`, which would result in a string `<host>` rather than a string `<host>:<user>` -- even if a user was empty, a safer option would have been `<host>:`. We now use `<host>:unknown-user` which is most definitely the correct option here.

crates/atuin-client/src/import/resh.rs
```diff
@@ -143,7 +144,7 @@ impl Importer for Resh {
                 .duration(duration)
                 .exit(entry.exit_code)
                 .cwd(entry.pwd)
-                .hostname(entry.host)
+                .cmd_origin(CmdOrigin::new(CmdHost::from(entry.host), CmdUser::default()))
```
markovejnovic, line 147: See my reply to greptile for the functional change

crates/atuin-common/Cargo.toml
```diff
@@ -23,9 +23,10 @@ unicode = ["dep:unicode-width", "dep:unicode-segmentation"]
 # via a `vt100` terminal emulator. Compiled only when enabled.
 ansi = ["dep:vt100"]
 # Enables OS-specific utilities.
-os = ["dep:rustix"]
+os = ["dep:rustix", "dep:whoami"]
 
 [dependencies]
+whoami = { workspace = true, optional = true }
```
taylordotfish: Why?

crates/atuin-domain/src/record/cmd_origin.rs
```diff
@@ -0,0 +1,175 @@
+/// A hostname, generic over its backing storage so it can be an owned
+/// `CmdHost<String>` or a zero-copy `CmdHost<&str>` view into a [`CmdOrigin`].
+#[derive(
+    Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, derive_more::Display, derive_more::From,
+)]
+pub struct CmdHost<T: AsRef<str> = String>(T);
```
taylordotfish: Remove the bound, it's not necessary

crates/atuin-domain/src/record/cmd_origin.rs
```diff
@@ -0,0 +1,175 @@
...
+
+impl Default for CmdHost<String> {
+    fn default() -> Self {
+        Self(String::from("unknown-host"))
+    }
+}
+
+/// A username, generic over its backing storage (owned `String` or `&str` view).
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, derive_more::Display, derive_more::From)]
+pub struct CmdUser<T: AsRef<str> = String>(T);
```
taylordotfish: Remove the bound, it's not necessary

crates/atuin-domain/src/record/cmd_origin.rs
```diff
@@ -0,0 +1,175 @@
...
+            .map(Self)
+            .unwrap_or_default()
+    }
+}
+
+impl<'a> CmdHost<&'a str> {
+    pub fn as_str(&self) -> &'a str {
+        self.0
+    }
+}
```
taylordotfish: Idea: instead of this:

```rust
impl<T> CmdHost<T> {
    pub fn into_inner(self) -> T {
        self.0
    }
}
```

(this depends on my "remove the bound" comment; otherwise you need to add `T: AsRef<str>`)

Then if you replace all the `.as_str()` calls with `.into_inner()` it'll still work, because `CmdHost<&str>` is `Copy`, so it doesn't actually "consume" the value

this is more general so I'd prefer it

crates/atuin-domain/src/record/cmd_origin.rs
```diff
@@ -0,0 +1,175 @@
...
+            .map(Self)
+            .unwrap_or_default()
+    }
+}
+
+impl<'a> CmdUser<&'a str> {
+    pub fn as_str(&self) -> &'a str {
+        self.0
+    }
+}
```
taylordotfish: same idea about `into_inner` instead

taylordotfish (approved):

PR #3950, opened 2026-08-19
chore: remove atuin-nucleo

taylordotfish:
After #3782 we no longer use `atuin-nucleo`, so we can remove it from the tree. The source code is not lost; we can always recover it if necessary.

One file, `crates/atuin-daemon/src/search/normalize.rs`, was vendored from `atuin-nucleo` as part of #3782. There is a comment linking to the original source; [the link](https://github.com/atuinsh/atuin/blob/v18.18.1/crates/atuin-nucleo/matcher/src/chars/normalize.rs) goes to a specific tag rather than `main`, so it will continue to work even after the removal.


markovejnovic (approved):
lgtm

PR #3951, opened 2026-08-19
fix(history): don't misclassify users named after an agent as agents

ellie:
- a user named pi (or any known agent name) is no longer auto-classified as an agent. this has been caught by a number of raspberry pi users
- new, optional AuthorKind field. this flags if the history is from a user, or from an agent. it is nullable, so as to be backwards compatible
- in order to be compatible with existing agent hooks, if the --author flag is explicitly provided, and is a known agent name, then the record is marked as from an agent. There is no reason for a human to be passing --author
- this is all overridden by --author-kind, as the primary source of truth
- finally, for a record where the author and the username in `hostname` are the same, we assume that it is NOT an agent. eg, for the case where both `hostname` and `author` say that the username is `pi`, we assume not an agent
- extensions (pi, opencode) and hook paths now stamp the kind explicitly, but need updating in order to do so

### DIFF
--- a/crates/atuin-ai/src/history_format.rs
+++ b/crates/atuin-ai/src/history_format.rs
@@ -1,4 +1,4 @@
-use atuin_client::history::{History, is_known_agent};
+use atuin_client::history::History;
 use atuin_common::time::{DurationExt, OffsetDateTimeExt};
 use time::UtcOffset;
 
@@ -37,11 +37,11 @@ fn format_history_metadata(history: &History, local_offset: UtcOffset) -> String
 }
 
 /// Attribution for agent-run commands: which agent, and its stated intent.
-/// A stated intent marks a command as agent-run even when the agent is not in
-/// KNOWN_AGENTS, so its author is still named. User-run commands (no intent,
-/// author not a known agent) get nothing.
+/// A stated intent marks a command as agent-run even when the entry is not
+/// recognised as one, so its author is still named. User-run commands (no
+/// intent, not an agent) get nothing.
 fn format_attribution(history: &History) -> String {
-    match (is_known_agent(&history.author), &history.intent) {
+    match (history.is_agent(), &history.intent) {
         (true, Some(intent)) => format!(" — {}: {intent}", history.author),
         (true, None) => format!(" — {}", history.author),
         (false, Some(intent)) if !history.author.is_empty() => {
@@ -87,6 +87,7 @@ mod tests {
             intent: None,
             deleted_at: None,
             shell: Some("zsh".into()),
+            author_kind: None,
         }
     }
 
--- a/crates/atuin-client/migrations/20260818000000_history_author_kind.sql
+++ b/crates/atuin-client/migrations/20260818000000_history_author_kind.sql
@@ -1,0 +1,4 @@
+-- Whether a human or an agent ran the command, when the integration that
+-- captured it said so. Null means it did not, and the author name is all we
+-- have to go on. See `atuin_client::history::AuthorKind`.
+alter table history add column author_kind integer;
--- a/crates/atuin-client/migrations/20260818000000_history_author_kind.sql
+++ b/crates/atuin-client/migrations/20260818000000_history_author_kind.sql
@@ -1,4 +1,3 @@
--- Whether a human or an agent ran the command, when the integration that
--- captured it said so. Null means it did not, and the author name is all we
--- have to go on. See `atuin_client::history::AuthorKind`.
+-- Whether a human or an agent ran the command. If null, we have to go off
+-- the author name.
 alter table history add column author_kind integer;
--- a/crates/atuin-client/src/database.rs
+++ b/crates/atuin-client/src/database.rs
@@ -2071,6 +2071,50 @@ mod test {
         }
     }
 
+    /// A legacy row whose colonless hostname IS a known agent name (e.g. a machine hostnamed
+    /// `pi`, imported before hostnames were `host:user`): the author defaulted to the whole
+    /// hostname, so it tells us nothing — human on both the SQL and [`History::is_agent`] sides.
+    #[tokio::test(flavor = "multi_thread")]
+    #[rstest]
+    async fn author_filter_treats_a_colonless_agent_hostname_as_human() {
+        let db = Sqlite::new("sqlite::memory:", test_local_timeout()).await.unwrap();
+
+        let history: History = History::import()
+            .timestamp(OffsetDateTime::now_utc())
+            .command("echo hello")
+            .cwd("/tmp")
+            .author("pi")
+            .build()
+            .into();
+        db.save(&history).await.unwrap();
+        // `CmdOrigin` cannot represent a colonless hostname (parse_lenient rewrites it), so plant
+        // the legacy shape directly.
+        sqlx::query("update history set hostname = 'pi'").execute(&db.pool).await.unwrap();
+
+        let loaded = db.load(history.id.0.as_str()).await.unwrap().unwrap();
+        assert!(!loaded.is_agent());
+
+        let context = Context {
+            cmd_origin: CmdOrigin::try_from("pi:unknown-user".to_owned()).unwrap(),
+            session: "session".into(),
+            cwd: "/tmp".into(),
+            host_id: "host".into(),
+            git_root: None,
+        };
+        for (pattern, expected) in [(AuthorPattern::AllUser, 1), (AuthorPattern::AllAgent, 0)] {
+            let authors = OrFilter::from_list(vec![pattern]).unwrap();
+            let filters = OptFilters {
+                authors: authors.as_slice_filter(),
+                ..Default::default()
+            };
+            let results = db
+                .search(DbSearchMode::FullText, FilterMode::Global, &context, "echo", filters)
+                .await
+                .unwrap();
+            assert_eq!(results.len(), expected, "{authors:?}");
+        }
+    }
+
     /// A user called `pi` shares a name with the `pi` agent, so on their machine the author name
     /// alone cannot say who ran a command: only an entry that states its kind is an agent's.
     #[tokio::test(flavor = "multi_thread")]
--- a/crates/atuin-client/src/database.rs
+++ b/crates/atuin-client/src/database.rs
@@ -121,7 +121,7 @@ fn apply_author_filter(sql: &mut SqlBuilder, authors: OrFilter<&[AuthorPattern]>
     // Mirrors [`History::is_agent`]: a recorded kind wins, and without one a known agent name means
     // an agent, unless the author is only the username it defaulted to. A kind we don't recognise
     // (written by a newer version) falls through to the name heuristic, exactly like
-    // [`AuthorKind::from_u8`] mapping it to `None` — and so does a NULL kind, because
+    // [`AuthorKind::from_repr`] mapping it to `None` — and so does a NULL kind, because
     // `NULL IN (...)` is not true.
     let is_agent = || {
         format!(
@@ -404,7 +404,7 @@ impl Sqlite {
         let shell: Option<String> = row.try_get("shell").ok().flatten();
         let author_kind: Option<i64> = row.try_get("author_kind").ok().flatten();
         let author_kind =
-            author_kind.and_then(|kind| u8::try_from(kind).ok()).and_then(AuthorKind::from_u8);
+            author_kind.and_then(|kind| u8::try_from(kind).ok()).and_then(AuthorKind::from_repr);
 
         History::from_db()
             .id(row.get("id"))
@@ -2023,7 +2023,7 @@ mod test {
     }
 
     /// An author_kind value this version doesn't recognise (written by a newer one) must fall
-    /// through to the name heuristic in SQL, exactly like [`AuthorKind::from_u8`] returning `None`
+    /// through to the name heuristic in SQL, exactly like [`AuthorKind::from_repr`] returning `None`
     /// does in [`History::is_agent`] — otherwise the two classifiers disagree on the same row.
     #[tokio::test(flavor = "multi_thread")]
     #[rstest]
--- a/crates/atuin-client/src/database.rs
+++ b/crates/atuin-client/src/database.rs
@@ -127,7 +127,7 @@ fn apply_author_filter(sql: &mut SqlBuilder, authors: OrFilter<&[AuthorPattern]>
         format!(
             "CASE WHEN author_kind IN ({kinds}) THEN author_kind = {agent} ELSE {author_expr} IN \
              ({names}) AND {author_expr} <> {user_expr} END",
-            kinds = AuthorKind::VARIANTS.iter().map(|kind| kind.as_u8().to_string()).join(", "),
+            kinds = AuthorKind::VARIANTS.iter().map(|kind| kind.as_u8()).join(", "),
             agent = AuthorKind::Agent.as_u8(),
             names = KNOWN_AGENTS.iter().map(quote).join(", "),
         )
--- a/crates/atuin-client/src/database.rs
+++ b/crates/atuin-client/src/database.rs
@@ -114,8 +114,9 @@ fn apply_author_filter(sql: &mut SqlBuilder, authors: OrFilter<&[AuthorPattern]>
     let user_expr = "CASE WHEN instr(hostname, ':') > 0 THEN substr(hostname, instr(hostname, \
                      ':') + 1) ELSE hostname END";
 
-    let author_expr =
-        format!("CASE WHEN author IS NULL OR trim(author) = '' THEN {user_expr} ELSE author END");
+    let author_expr = std::fmt::from_fn(|f| {
+        write!(f, "CASE WHEN author IS NULL OR trim(author) = '' THEN {user_expr} ELSE author END")
+    });
 
     // Mirrors [`History::is_agent`]: a recorded kind wins, and without one a known agent name means
     // an agent, unless the author is only the username it defaulted to. A kind we don't recognise
--- a/crates/atuin-client/src/database.rs
+++ b/crates/atuin-client/src/database.rs
@@ -118,15 +118,20 @@ fn apply_author_filter(sql: &mut SqlBuilder, authors: OrFilter<&[AuthorPattern]>
         write!(f, "CASE WHEN author IS NULL OR trim(author) = '' THEN {user_expr} ELSE author END")
     });
 
+    let defaulted_expr = "CASE WHEN instr(hostname, ':') = 0 THEN hostname WHEN substr(hostname, \
+                          instr(hostname, ':') + 1) = 'unknown-user' THEN substr(hostname, 1, \
+                          instr(hostname, ':') - 1) ELSE substr(hostname, instr(hostname, ':') + \
+                          1) END";
+
     // Mirrors [`History::is_agent`]: a recorded kind wins, and without one a known agent name means
-    // an agent, unless the author is only the username it defaulted to. A kind we don't recognise
+    // an agent, unless the author is only the name it defaulted to. A kind we don't recognise
     // (written by a newer version) falls through to the name heuristic, exactly like
     // [`AuthorKind::from_repr`] mapping it to `None` — and so does a NULL kind, because
     // `NULL IN (...)` is not true.
     let is_agent = || {
         format!(
             "CASE WHEN author_kind IN ({kinds}) THEN author_kind = {agent} ELSE {author_expr} IN \
-             ({names}) AND {author_expr} <> {user_expr} END",
+             ({names}) AND {author_expr} <> {defaulted_expr} END",
             kinds = AuthorKind::VARIANTS.iter().map(|kind| kind.as_u8()).join(", "),
             agent = AuthorKind::Agent.as_u8(),
             names = KNOWN_AGENTS.iter().map(quote).join(", "),
@@ -2090,6 +2095,45 @@ mod test {
         // `CmdOrigin` cannot represent a colonless hostname (parse_lenient rewrites it), so plant
         // the legacy shape directly.
         sqlx::query("update history set hostname = 'pi'").execute(&db.pool).await.unwrap();
+
+        let loaded = db.load(history.id.0.as_str()).await.unwrap().unwrap();
+        assert!(!loaded.is_agent());
+
+        let context = Context {
+            cmd_origin: CmdOrigin::try_from("pi:unknown-user".to_owned()).unwrap(),
+            session: "session".into(),
+            cwd: "/tmp".into(),
+            host_id: "host".into(),
+            git_root: None,
+        };
+        for (pattern, expected) in [(AuthorPattern::AllUser, 1), (AuthorPattern::AllAgent, 0)] {
+            let authors = OrFilter::from_list(vec![pattern]).unwrap();
+            let filters = OptFilters {
+                authors: authors.as_slice_filter(),
+                ..Default::default()
+            };
+            let results = db
+                .search(DbSearchMode::FullText, FilterMode::Global, &context, "echo", filters)
+                .await
+                .unwrap();
+            assert_eq!(results.len(), expected, "{authors:?}");
+        }
+    }
+
+    #[tokio::test(flavor = "multi_thread")]
+    #[rstest]
+    async fn author_filter_treats_a_placeholder_user_agent_hostname_as_human() {
+        let db = Sqlite::new("sqlite::memory:", test_local_timeout()).await.unwrap();
+
+        let history: History = History::import()
+            .timestamp(OffsetDateTime::now_utc())
+            .command("echo hello")
+            .cwd("/tmp")
+            .cmd_origin(CmdOrigin::try_from("pi:unknown-user".to_owned()).unwrap())
+            .author("pi")
+            .build()
+            .into();
+        db.save(&history).await.unwrap();
 
         let loaded = db.load(history.id.0.as_str()).await.unwrap().unwrap();
         assert!(!loaded.is_agent());
--- a/crates/atuin-client/src/database.rs
+++ b/crates/atuin-client/src/database.rs
@@ -6,7 +6,7 @@ use std::time::Duration;
 use atuin_common::filter::{self, OrFilter};
 use atuin_common::time::OffsetDateTimeExt;
 use atuin_common::utils;
-use atuin_domain::record::CmdOrigin;
+use atuin_domain::record::{CmdOrigin, UNKNOWN_USER};
 use fs_err as fs;
 use itertools::Itertools;
 use sql_builder::bind::Bind;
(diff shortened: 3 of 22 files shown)


crates/atuin-client/src/history.rs
```diff
@@ -30,15 +33,51 @@ pub fn is_known_agent(author: &str) -> bool {
...
+            1 => Some(Self::User),
+            2 => Some(Self::Agent),
+            _ => None,
+        }
+    }
+
+    pub const fn as_u8(self) -> u8 {
+        self as u8
+    }
+}
```
markovejnovic, line 64: Hmm, I would suggest something else. On `origin/main`, inside `atuin-domain/src/record/cmd_origin.rs`, there is a type `CmdUser`.

I would **strongly** suggest you edit that file as I describe here.

There is a struct, right now:

```
/// A username, generic over its backing storage (owned `String` or `&str` view).
#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, derive_more::Display, derive_more::From)]
pub struct CmdUser<T = String>(T);
```

I would suggest modifying this, relatively aggressively:

1. Rename `CmdUser` to be `CmdUsername`, probably can keep the rest as-is.
2. Create a new `CmdUser` like so

```rust
#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
pub enum CmdUser<S> {
  Agent(CmdUsername<S>),
  Human(CmdUsername<S>),
}

impl Display for CmdUser<S> {
  fn display(&self) {
      match self {
        CmdUser::Agent(v) => write!("a${}", v.as_str()),
        CmdUser::Human(v) => write!("h${}", v.as_str()),
      }
    )
  }
}
```

This means that the calling semantics of `CmdOrigin::user` will change -- it will now return a `CmdUser<&str>` which is virtually free, however, enforces careful handling of the different type of users.

You will have to make some other changes, but this is much more type-rich and plays nicely with the new `CmdOrigin` utility I've made.

Since I strongly suspect you'll need to deal with the `Kind`, strum has some utilities for you.

I would suggest looking at this usage (also on `origin/main`):

```rust
#[derive(Debug, Eq, PartialEq, Clone, strum_macros::EnumDiscriminants)]
#[strum_discriminants(name(HistoryRecordKind))]
#[allow(
    clippy::large_enum_variant,
    reason = "`Create` records are much more common than `Delete` records; wrapping in a `Box`
        would use more memory overall"
)]
pub enum HistoryRecord {
    Create(History),   // Create a history record
    Delete(HistoryId), // Delete a history record, identified by ID
}
```

As this can help you **significantly** avoid hand-rolling the `Kind`, which, in-turn, improves the quality of our application.

markovejnovic (changes requested):

crates/atuin-client/src/history.rs
ellie: sorry I still had a lot of stuff to push

crates/atuin-client/src/history.rs
```diff
@@ -210,11 +270,15 @@ impl History {
...
         shell: Option<String>,
+        author_kind: Option<AuthorKind>,
     ) -> Self {
         let session = session
             .or_else(|| env::var("ATUIN_SESSION").ok())
             .unwrap_or_else(|| uuid_v7().as_simple().to_string());
         let cmd_origin = cmd_origin.unwrap_or_else(CmdOrigin::probe_current);
+        // `From<HistoryCaptured>` pre-resolves the same flag -> env chain to infer `author_kind`;
+        // a precedence change here must be mirrored there, or the stored kind can disagree with
+        // the stored author.
```
taylordotfish: I don't understand this comment

crates/atuin-client/src/history.rs
```diff
@@ -245,6 +324,10 @@ impl History {
     /// * `intent` is always written; if `None`, nil is written to the output.
     /// * Added new field `shell`.
     ///
+    /// Fields appended after V2 shipped (see [`V2_SERIALIZED_FIELDS`]) are read back only if the
+    /// encoded array is long enough to contain them, so older clients ignore them and newer clients
+    /// can still read older entries.
+    ///
```
taylordotfish:
```suggestion
```

crates/atuin-client/src/history.rs
```diff
@@ -235,6 +299,21 @@ impl History {
...
+        }
+    }
+
+    /// Whether an AI agent, rather than a human, ran this command.
+    ///
+    /// [`Self::author_kind`] is authoritative when the integration that captured the entry set it.
+    /// Otherwise we guess: a known agent name identifies an agent, unless it is also the username
+    /// recorded in the entry's origin, in which case the author is just the default it fell back
+    /// to and tells us nothing. That exception is what stops a user called `pi` from looking like
+    /// the `pi` agent.
```
taylordotfish, line 304:
```suggestion
```

crates/atuin-client/src/history.rs
```diff
@@ -128,6 +166,24 @@ impl Version {
...
+/// Number of fields [`History::serialize`] writes.
+///
+/// This is deliberately not [`Version::min_fields`]: fields appended to V2 grow this count, while
+/// `min_fields` stays at the 12 fields V2 launched with so that entries written before the new
+/// fields existed still decode.
+///
+/// This count must never gate reading a field back. Each appended field gets its own frozen
+/// threshold constant (its position in the array, like [`V2_FIELDS_THROUGH_AUTHOR_KIND`]) —
+/// gating on this growing total would make records written before the *next* field silently lose
+/// this one.
```
taylordotfish: I don't understand what “gate reading a field back” and “gating on this growing total” mean

crates/atuin-client/src/history.rs
```diff
@@ -30,15 +31,51 @@ pub fn is_known_agent(author: &str) -> bool {
...
+    /// Decode the representation written by [`Self::as_u8`].
+    ///
+    /// Unknown values are treated as "not stated" rather than an error, so that entries written by
+    /// a future version with more kinds still decode.
+    pub const fn from_u8(value: u8) -> Option<Self> {
+        match value {
+            1 => Some(Self::User),
+            2 => Some(Self::Agent),
+            _ => None,
+        }
```
taylordotfish:
```suggestion
        Self::VARIANTS.get(usize::from(u8))
```

crates/atuin-client/src/history.rs
```diff
@@ -30,15 +31,51 @@ pub fn is_known_agent(author: &str) -> bool {
...
+
+impl AuthorKind {
+    /// Every recognised kind. The SQL author filter derives its recognised-value list from this,
+    /// so it stays in lockstep with [`Self::from_u8`] (a test pins the two together).
+    pub const VARIANTS: [Self; 2] = [Self::User, Self::Agent];
+
+    /// Decode the representation written by [`Self::as_u8`].
+    ///
+    /// Unknown values are treated as "not stated" rather than an error, so that entries written by
+    /// a future version with more kinds still decode.
```
taylordotfish:
```suggestion
```

crates/atuin-client/src/history.rs
```diff
@@ -30,15 +31,51 @@ pub fn is_known_agent(author: &str) -> bool {
...
+pub enum AuthorKind {
+    /// A human ran this command.
+    User = 1,
+    /// An AI agent ran this command.
+    Agent = 2,
+}
+
+impl AuthorKind {
+    /// Every recognised kind. The SQL author filter derives its recognised-value list from this,
+    /// so it stays in lockstep with [`Self::from_u8`] (a test pins the two together).
```
taylordotfish:
```suggestion
    /// This enum's variants in order.
```

crates/atuin-client/src/history.rs
```diff
@@ -30,15 +31,51 @@ pub fn is_known_agent(author: &str) -> bool {
     KNOWN_AGENTS.contains(&author)
 }
 
+/// Who wrote a history entry, as declared by whatever captured it.
+///
+/// This is stored as a small integer, both on the wire and in the database, and is optional: an
+/// entry captured before this field existed, or by an integration that does not set it, has no
+/// kind, and [`History::is_agent`] falls back to inspecting the author name.
```
taylordotfish:
```suggestion
/// This is stored as a small integer, both on the wire and in the database.
```

crates/atuin-client/src/history/builder.rs
```diff
@@ -71,10 +75,30 @@ pub struct HistoryCaptured {
...
+        let author_kind = captured
+            .author_kind
+            .or_else(|| {
+                // Parsed like the `--author-kind` flag, so the two channels can't drift —
+                // case-insensitively, since a bad env value is silently ignored rather than
+                // rejected loudly the way a bad flag is.
+                let value = std::env::var(HISTORY_AUTHOR_KIND_ENV).ok()?;
+                clap::ValueEnum::from_str(&value, true).ok()
+            })
+            .or_else(|| author.as_deref().is_some_and(is_known_agent).then_some(AuthorKind::Agent));
```
taylordotfish: Is this the right place for this logic? Why not in `History::new`? Now we have two places that read from `HISTORY_AUTHOR_ENV`.

crates/atuin-client/src/history/builder.rs
```diff
@@ -71,10 +75,30 @@ pub struct HistoryCaptured {
...
+        // from integrations that predate `--author-kind`. An explicit kind, from the builder or
+        // `ATUIN_HISTORY_AUTHOR_KIND` (`user`/`agent`), still wins.
+        let author = normalize_optional_string(captured.author)
+            .or_else(|| normalize_optional_string(std::env::var(HISTORY_AUTHOR_ENV).ok()));
+        let author_kind = captured
+            .author_kind
+            .or_else(|| {
+                // Parsed like the `--author-kind` flag, so the two channels can't drift —
+                // case-insensitively, since a bad env value is silently ignored rather than
+                // rejected loudly the way a bad flag is.
```
taylordotfish: I don't understand what case sensitivity has to do with environment variable vs. flag differences.

crates/atuin-client/src/history/builder.rs
```diff
@@ -71,10 +75,30 @@ pub struct HistoryCaptured {
...
+    author_kind: Option<AuthorKind>,
 }
 
 impl From<HistoryCaptured> for History {
     fn from(captured: HistoryCaptured) -> Self {
+        // An author stated by the caller (`--author` or `ATUIN_HISTORY_AUTHOR`), unlike the
+        // hostname-derived default, is an authorship claim: a known agent name there identifies
+        // the agent even on a machine whose username matches it. This is what classifies entries
+        // from integrations that predate `--author-kind`. An explicit kind, from the builder or
+        // `ATUIN_HISTORY_AUTHOR_KIND` (`user`/`agent`), still wins.
```
taylordotfish: This comment was very hard for me to understand; I think I got there eventually but I'm not sure. Might benefit from rewriting it

crates/atuin-client/src/history/builder.rs
taylordotfish: If this really is the only correct place for this logic, can we stop reading `HISTORY_AUTHOR_ENV` in `History::new`? It just seems like a footgun to me to have to places where we read `HISTORY_AUTHOR_ENV`, but only one of them reads `HISTORY_AUTHOR_KIND_ENV` too. Duplication lets things drift out of sync.

crates/atuin-client/src/database.rs
```diff
@@ -99,28 +99,39 @@ impl Context {
...
+    // Mirrors [`History::is_agent`]: a recorded kind wins, and without one a known agent name means
+    // an agent, unless the author is only the username it defaulted to. A kind we don't recognise
+    // (written by a newer version) falls through to the name heuristic, exactly like
+    // [`AuthorKind::from_u8`] mapping it to `None` — and so does a NULL kind, because
+    // `NULL IN (...)` is not true.
+    let is_agent = || {
+        format!(
+            "CASE WHEN author_kind IN ({kinds}) THEN author_kind = {agent} ELSE {author_expr} IN \
+             ({names}) AND {author_expr} <> {user_expr} END",
+            kinds = AuthorKind::VARIANTS.iter().map(|kind| kind.as_u8().to_string()).join(", "),
```
taylordotfish: I believe the `.to_string()` is unnecessary here; `.join()` works on any `T: Display`

```suggestion
            kinds = AuthorKind::VARIANTS.iter().map(|kind| kind.as_u8()).join(", "),
```

crates/atuin-client/src/database.rs
```diff
@@ -99,28 +99,39 @@ impl Context {
...
-                       ELSE hostname END ELSE author END";
-
-    let mut agent_list: Option<String> = None;
-    let get_agent_list = || KNOWN_AGENTS.iter().map(quote).join(", ");
+    // The username half of `hostname`, which is what `author` falls back to when nothing set it.
+    let user_expr = "CASE WHEN instr(hostname, ':') > 0 THEN substr(hostname, instr(hostname, \
+                     ':') + 1) ELSE hostname END";
+
+    let author_expr =
+        format!("CASE WHEN author IS NULL OR trim(author) = '' THEN {user_expr} ELSE author END");
```
taylordotfish:
```suggestion
    let author_expr = std::fmt::from_fn(|f| {
        write!(f, "CASE WHEN author IS NULL OR trim(author) = '' THEN {user_expr} ELSE author END")
    });
```

avoids creating an intermediate heap-allocated string

crates/atuin-client/migrations/20260818000000_history_author_kind.sql
```diff
@@ -0,0 +1,4 @@
+-- Whether a human or an agent ran the command, when the integration that
+-- captured it said so. Null means it did not, and the author name is all we
+-- have to go on. See `atuin_client::history::AuthorKind`.
```
taylordotfish:
```suggestion
-- Whether a human or an agent ran the command. If null, we have to go off
-- the author name.
```

crates/atuin-daemon/proto/history.proto
```diff
@@ -1,6 +1,16 @@
 syntax = "proto3";
 package history;
 
+// Whether a human or an agent ran a command. Unspecified means the caller did
+// not say, and the author name is all the receiver has to go on. Unspecified is
+// also what keeps this backwards compatible: it is proto3's zero value, so a
+// peer from before this field existed reads and writes it implicitly.
```
taylordotfish, line 7:
```suggestion
// Whether a human or an agent ran a command. `UNSPECIFIED` means the caller did
// not say; it is 0 to make it the default value for backwards compatibility.
```

crates/atuin-client/src/history.rs
```diff
@@ -128,6 +166,24 @@ impl Version {
...
+///
+/// This is deliberately not [`Version::min_fields`]: fields appended to V2 grow this count, while
+/// `min_fields` stays at the 12 fields V2 launched with so that entries written before the new
+/// fields existed still decode.
+///
+/// This count must never gate reading a field back. Each appended field gets its own frozen
+/// threshold constant (its position in the array, like [`V2_FIELDS_THROUGH_AUTHOR_KIND`]) —
+/// gating on this growing total would make records written before the *next* field silently lose
+/// this one.
+const V2_SERIALIZED_FIELDS: u32 = 13;
```
taylordotfish: nit: I would prefer this be called `LATEST_SERIALIZED_FIELDS` or `CURRENT_SERIALIZED_FIELDS` or similar. It matches the `Version::LATEST.min_fields()` call it's replacing more closely, and we only ever care about the # of serialized fields for the latest version. If/when we bump to V3, we won't care about serializing V2 (only deserializing), just like we don't care about serializing V1 now.

crates/atuin-client/src/history.rs
```diff
@@ -128,6 +166,24 @@ impl Version {
...
+
+/// A V2 record contains `author_kind` iff it has at least this many fields.
+///
+/// Frozen forever at the position `author_kind` was appended at; do not grow it alongside
+/// [`V2_SERIALIZED_FIELDS`].
+const V2_FIELDS_THROUGH_AUTHOR_KIND: u32 = 13;
```
taylordotfish: nit: I would call this `V2_AUTHOR_KIND_FIELD_NUMBER` or similar to make it clearer it's a frozen index

taylordotfish (commented):
The serialization looks correct, just with a couple nits. I'm somewhat unsure about the logic added to `From<HistoryCaptured> for History`, both in terms of location and its similarity to existing logic in `History::new`. I also left some suggestions for a handful of easy efficiency/readability wins.

The verbose LLM-written comments are overall quite confusing and, at least for me, make this PR harder to understand than it actually is. My eyes gravitate towards reading comments before code, but in many cases, ignoring the comment and just reading the code would have been clearer. I would prefer seeing most of them removed or tightened up.

crates/atuin-daemon/proto/history.proto
ellie: hm that comment assumes more understanding of proto3 than many have, I would prefer to keep it as is.

crates/atuin-client/src/history/builder.rs
ellie: It actually makes more sense to have it in History::capture than anywhere else, because all other ways of constructing history are not assumed to be part of the user environment. 

The existing `capture` function looks like this

```rust
    /// Builder for a history entry that is captured via hook.
    ///
    /// This builder is used only at the `start` step of the hook,
    /// so it doesn't have any fields which are known only after
    /// the command is finished, such as `exit` or `duration`.
    ///
    /// ## Examples
    /// ```rust
    /// use atuin_client::history::History;
    ///
    /// let history: History = History::capture()
    ///     .timestamp(time::OffsetDateTime::now_utc())
    ///     .command("ls -la")
    ///     .cwd("/home/user")
    ///     .build()
    ///     .into();
    /// ```
    ///
    /// Command without any required info cannot be captured, which is forced at compile time:
    ///
    /// ```compile_fail
    /// use atuin_client::history::History;
    ///
    /// // this will not compile because `cwd` is missing
    /// let history: History = History::capture()
    ///     .timestamp(time::OffsetDateTime::now_utc())
    ///     .command("ls -la")
    ///     .build()
    ///     .into();
    /// ```
    pub fn capture() -> builder::HistoryCapturedBuilder {
        builder::HistoryCaptured::builder()
    }
```

so putting logic in there was kinda breaking the existing pattern, where these functions just called the builder. I can change it though, I don't really feel strongly about this as they are ~equivalent

crates/atuin-client/src/history.rs
ellie: I can see the argument for

```
// This is stored as a small integer, both on the wire and in the database, and is optional: an
/// entry captured before this field existed, or by an integration that does not set it
```

much more than cutting out the actual reason why it exists

crates/atuin-client/src/history.rs
markovejnovic: I think there are utilities for this in `strum`.

crates/atuin-daemon/proto/history.proto
```diff
@@ -1,6 +1,16 @@
...
+enum AuthorKind {
+  AUTHOR_KIND_UNSPECIFIED = 0;
+  AUTHOR_KIND_USER = 1;
+  AUTHOR_KIND_AGENT = 2;
+}
```
markovejnovic, line 12: Hmm, I recognize we can't change this in this PR, but could you create a ticket and add a `TODO(ATU-XYZ)` to track merging `AuthorKind` and the `Author` field.

I would much rather prefer we had a typed message going forwards:

```protobuf
message Author {
  AuthorKind kind = 1;
  string name = 2;
}
```

crates/atuin-daemon/src/history/mod.rs
```diff
@@ -5,6 +5,28 @@
 // Include the generated proto code
 tonic::include_proto!("history");
 
+// Translate between the client's `Option<AuthorKind>` (where "not stated" is `None`) and the
+// proto enum (where proto3 forces a zero variant, `Unspecified`, to mean the same thing).
```
markovejnovic:
```suggestion
```

crates/atuin/src/command/client/history.rs
```diff
@@ -49,6 +49,14 @@ pub enum Cmd {
         #[arg(long)]
         author: Option<String>,
 
+        /// Whether a human or an AI agent ran this command
+        ///
+        /// When omitted, an explicitly passed author that is a known agent name counts as an
+        /// agent; anything else is left unstated and later classified by the author name, which
+        /// cannot tell an agent apart from a user who shares its name.
```
markovejnovic:
```suggestion
        ///
        /// [`Option::None`] will cause us to perform a best-guess effort.
```

crates/atuin-client/src/history.rs
```diff
@@ -516,22 +612,34 @@ mod tests {
...
-    fn author_matches_filters(author: &str, filters: OrFilter<&[AuthorPattern]>) -> bool {
+    fn author_matches_filters(history: &History, filters: OrFilter<&[AuthorPattern]>) -> bool {
         // `contains` is true for an "all" filter, so that case needs no separate check.
-        filters.contains(&AuthorPattern::Name(author.to_owned()))
-            || (filters.contains(&AuthorPattern::AllUser) && !is_known_agent(author))
-            || (filters.contains(&AuthorPattern::AllAgent) && is_known_agent(author))
+        filters.contains(&AuthorPattern::Name(history.author.clone()))
+            || (filters.contains(&AuthorPattern::AllUser) && !history.is_agent())
+            || (filters.contains(&AuthorPattern::AllAgent) && history.is_agent())
+    }
```
markovejnovic, line 631: Heeding my advice will result in this function being more type-safe and granular (as well as semantically correct):


```suggestion
    fn author_matches_filters(user: &CmdUser, filters: OrFilter<&[AuthorPattern]>) -> bool {
// `contains` is true for an "all" filter, so that case needs no separate check.
        filters.contains(&AuthorPattern::Name(history.author.clone()))
            || (filters.contains(&AuthorPattern::AllUser) && matches!(user, CmdUser::Human))
            || (filters.contains(&AuthorPattern::AllAgent) && matches!(user, CmdUser::Agent))
    }
```

crates/atuin-client/src/history.rs
```diff
@@ -174,6 +230,10 @@ pub struct History {
     pub deleted_at: Option<OffsetDateTime>,
     /// The shell used to run the command.
     pub shell: Option<String>,
+    /// Whether a human or an agent wrote this command, if whatever captured it said so.
+    ///
+    /// When this is `None`, [`History::is_agent`] guesses from the author name.
+    pub author_kind: Option<AuthorKind>,
```
markovejnovic, line 231: Hmmm... There are some problems here:

- Currently `author` is typed as a `String`.
- There is a `cmd_origin` field that is remapped with `sqlx` on L224.
- There is an additional `author_kind` field, now.

We discussed in DMs, but my opinion is that these three fields should not exist. There should be one field `CmdOrigin`, and there should be an `impl FromRow for History` that gracefully deserializes it.

This will result in simpler code.

One thing I don't completely grasp on _my_ PR is why I didn't catch before that `cmd_origin` and `author` are disjoint. I mentioned in DMs (though I haven't verified yet) that certain importers actually stored `<host>:<user>` in the `hostname` column, which is why I did what I did.

The result of what I'm saying is that there is homework in figuring out whether my claim is correct or not, and if not, commonizing the fields.

If yes, there is still homework in finding good workarounds in order to merge `author, cmd_origin, author_kind` into one field.

crates/atuin-client/src/database.rs
```diff
@@ -99,28 +99,39 @@ impl Context {
...
 
-    let author_expr = "CASE WHEN author IS NULL OR trim(author) = '' THEN CASE WHEN \
-                       instr(hostname, ':') > 0 THEN substr(hostname, instr(hostname, ':') + 1) \
-                       ELSE hostname END ELSE author END";
-
-    let mut agent_list: Option<String> = None;
-    let get_agent_list = || KNOWN_AGENTS.iter().map(quote).join(", ");
+    // The username half of `hostname`, which is what `author` falls back to when nothing set it.
+    let user_expr = "CASE WHEN instr(hostname, ':') > 0 THEN substr(hostname, instr(hostname, \
+                     ':') + 1) ELSE hostname END";
```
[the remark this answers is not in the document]
markovejnovic: this seems correct

crates/atuin-ai/src/history_format.rs
```diff
@@ -37,11 +37,11 @@ fn format_history_metadata(history: &History, local_offset: UtcOffset) -> String
...
 /// Attribution for agent-run commands: which agent, and its stated intent.
-/// A stated intent marks a command as agent-run even when the agent is not in
-/// KNOWN_AGENTS, so its author is still named. User-run commands (no intent,
-/// author not a known agent) get nothing.
+/// A stated intent marks a command as agent-run even when the entry is not
+/// recognised as one, so its author is still named. User-run commands (no
+/// intent, not an agent) get nothing.
 fn format_attribution(history: &History) -> String {
-    match (is_known_agent(&history.author), &history.intent) {
+    match (history.is_agent(), &history.intent) {
```
markovejnovic, line 44:
```suggestion
    match (matches!(history.cmd_origin.user(), CmdUser::Agent), &history.intent) {
```

feels much more logical

markovejnovic (commented):
Added some more comments, mostly revolving around type safety

crates/atuin-client/src/history.rs
ellie: I think it's much clearer to explain the precedence chain that this function follows in a comment, rather than in the code.  I am happy to delete it if you disagree, but think that new human or agent readers will find this clearer.

crates/atuin-ai/src/history_format.rs
ellie: it seems much more difficult to read, why is it more logical?

crates/atuin-client/src/history.rs
taylordotfish: I don't feel that strongly; leaving it is fine

crates/atuin-client/src/history.rs
taylordotfish: My rationale is that the fact that the field is optional is not really the responsibility of the `AuthorKind` type itself, but rather a type that *contains* an `AuthorKind` (i.e., `History`). Put another way, I see `AuthorKind` as a “dumb” or “plain” type that shouldn't care whether it's being used in a required context or optional.

That said, I don't feel that strongly, happy to leave as-is or go with your other suggestion.

crates/atuin-ai/src/history_format.rs
ellie: I think it's also important not to confuse the concept of "author" and "user", especially in field naming

author: something that has written history. this could be an agent, or it could be a human
user: the system user

crates/atuin-client/src/history/builder.rs
taylordotfish: `From<HistoryCaptured> for History` or `History::capture` both sound like fine places to me then; my priority would be to make sure the logic is consolidated in one place. So if we can put all of the fields-from-environment logic in one of those two places, and then delete the partial fields-from-environment logic in `History::new` (reading from `HISTORY_AUTHOR_ENV` again), that would be great.

crates/atuin-client/src/history.rs
ellie: That's fair. I was thinking of keeping it because it's only use is optional, and if you were looking for an AuthorKind and wanted to know what it was, it would be a single docstring to read vs hunting around for context too

crates/atuin-client/src/history/builder.rs
ellie: yeah agreed. there's already precedence for things to be pretty mixed up, and I don't think that should be cleaned up here. I will follow up making sure that all capture-specific logic lives where it makes sense, vs split up

crates/atuin-client/src/history/builder.rs
ellie: IMPoRTANT_THING foo

vs 

foo --improtant-thing

the former won't error, the latter will

not really a necessary comment anyway

crates/atuin-client/src/history.rs
ellie: there are! `FromRepr`

crates/atuin-daemon/proto/history.proto
ellie: followup in stacked pr

PR #3952, opened 2026-08-19
perf(tracing): OTel export + instrumentation coverage

markovejnovic:
Adds support for Jaeger. See `CONTRIBUTING.md` for more details


crates/atuin/src/command/client.rs
```diff
@@ -197,7 +199,11 @@ impl Cmd {
         // doing anything else. History commands are performance-sensitive and run before and after
         // every shell command, so we want to skip any unnecessary initialization for them.
         let settings = Settings::new().wrap_err("could not load client settings")?;
-        self.init_logging(&settings);
+        let _logging = self
+            .log_config(&settings)
+            .map(|c| LogCtx::try_enable("atuin", &c))
+            .transpose()
+            .wrap_err("failed to enable logging")?;
```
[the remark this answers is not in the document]
markovejnovic: that's intended -- if the user explicitly requests `ATUIN_OTEL`, we better crash if we can't set it up.

crates/atuin/src/logs/mod.rs
```diff
@@ -1,22 +1,215 @@
...
+        );
+        if show_time {
+            base.with(
+                config.stderr.as_ref().map(|c| stderr_layer::<_, fmt::time::SystemTime>(c, filter)),
+            )
+            .try_init()?;
+        } else {
+            base.with(config.stderr.as_ref().map(|c| stderr_layer::<_, ()>(c, filter)))
+                .try_init()?;
+        }
```
taylordotfish, line 101: :(

crates/atuin/Cargo.toml
```diff
@@ -107,9 +107,13 @@ futures-util = "0.3"
...
 open = "5"
 ratatui = { workspace = true }
-tracing = "0.1"
+tracing = { workspace = true }
 tracing-subscriber = { workspace = true }
 tracing-appender = "0.2"
+tracing-opentelemetry = "0.33"
+opentelemetry = "0.32"
+opentelemetry_sdk = "0.32"
+opentelemetry-otlp = "0.32"
```
ellie: I would prefer that we pull these in under a build flag. There is no real need to ship production builds with the OTEL sdk included, I don't imagine users needing that

crates/atuin/src/logs/otel.rs
```diff
@@ -0,0 +1,149 @@
+//! OpenTelemetry span export, gated behind the `profiling-traced` feature.
+#[cfg(feature = "profiling-traced")]
+mod inner {
```
taylordotfish: The two versions of `inner` might be nicer as separate files so you lose a level of indentation (look, now *I'm* advocating for more files...)

crates/atuin/src/logs/otel.rs
```diff
@@ -0,0 +1,149 @@
...
+                .with_resource(
+                    opentelemetry_sdk::Resource::builder().with_service_name(service_name).build(),
+                )
+                .build();
+
+            Ok(Some(Self { provider }))
+        }
+
+        /// The [`tracing_opentelemetry`] layer that exports spans through this context.
+        pub fn layer<S>(&self, service_name: &'static str) -> Box<dyn Layer<S> + Send + Sync>
```
taylordotfish: Why `Box` instead of returning `impl Layer<S> + Send + Sync`?

ellie (approved):

PR #3954, opened 2026-08-20
refactor: build out an object-mapping for sqlite

markovejnovic:


markovejnovic:
It looks like this is getting dropped on the floor, unfortunately.

PR #3955, opened 2026-08-20
perf(sync): parallel record downloads

markovejnovic:
# Downloads are now 339% faster.

Before

image

After

image

## Current State

Currently records are downloaded in the following algorithm:

```mermaid
flowchart TD
  download-page
  commit-page-to-db
  
  download-page --> commit-page-to-db --> download-page
```

Each page download is against a cursor: "Download page of records in range idx=[0, 100]". Since we know how many records we need to download total, we take our total record range `total=[0, 1K]`, and we loop over with some gait (in my example 100).

## This PR

This PR makes it so that we download pages in a stream, concurrently. Since we know the gait, we take our `total=[0, 1K]` range and split it into tiles: `total=[[0,100], [101, 200], [201, 300]...]`.

We then use tokio's `buffered` utility to run those requests in parallel and create an ordered insertion stream, so the chart ends up looking something like

```mermaid
flowchart TD
  download-page-1 --> insert-page-1
  insert-page-1 --> insert-page-2
  download-page-2 --> insert-page-2
  insert-page-2 --> insert-page-3
  download-page-3 --> insert-page-3
```

Inserting pages is an order-of-magnitude cheaper (in terms of latency) operation.

The implementation is slightly more complicated than this naive implementation as we cannot trust that a server will gracefully return fully loaded-up pages (note we assume as part of the parallelization flow that the server will return `gait`-many entries in a page), so, consequently, there is a fallback branch which downloads pages serially in case we find one page which has a gait samller than expected. Practically, this happens if the `SQL` query page size is smaller than what the client requested, which, in our case, is never, but more fault tolerance, more good.

## Other Changes

- Created a new type `Tiled`, intended to be used as `(0..100).tiled(5)` to return `[(0,5), (6,10), ...]`. It's got all the `Iterator` impls you'd expect.
- Tweaked `api_client.rs` a little bit to have a calling convention `.records().stream()` to get a stream of records and `.records().one()` if you only want one record.
- Renamed `RecordSeriesKey::host -> host_id`.
- Trimmed some tests.
- `sync/packfile.rs` got removed in favor of being inlined inside `sync/mod.rs`. I realized it made no sense to keep this code completely separate.

## Review Order

- Check `tiled.rs`. It's mostly self-contained, simple, but quite useful.
- Check `api_client.rs` -- there's a new calling convention `client.records(series).stream()`.
- Check everything else.

## Benchmarks

Ran some real benchmarks on my personal ~52K records. Before 1m35. After 28s to download them. 339% improvement.


ellie (approved):
I focused on a higher level functionality review here, seeing as we want to ship asap.

Generally looks good to me, my main concern was wrt order preservation but it seems like that it well handled. I am doing a thorough e2e test of this separately to ensure that correctness is maintained in the store, but going by the code I think it will be fine!

markovejnovic:
going to wait on @taylordotfish's review and we'll be good to merge

crates/atuin-common/src/range/tiled.rs
```diff
@@ -0,0 +1,160 @@
...
+mod sealed {
+    pub trait Sealed {}
+    impl Sealed for u8 {}
+    impl Sealed for u16 {}
+    impl Sealed for u32 {}
+    impl Sealed for u64 {}
+    impl Sealed for usize {}
+}
+
+#[doc(hidden)]
```
taylordotfish:
```suggestion
```

If you really need to hide the members, put them in `Sealed`

crates/atuin-common/src/range/tiled.rs
```diff
@@ -0,0 +1,160 @@
...
+pub trait TileIdx: sealed::Sealed + Copy + Ord {
+    type NonZero: Copy + std::fmt::Debug;
+
+    /// `self` advanced by a non-zero tile width, saturating at the type's maximum.
+    fn saturating_advance(self, by: Self::NonZero) -> Self;
+
+    /// `size` as a non-zero tile width, falling back to `1` when it is `0`.
+    fn width_or_one(size: Self) -> Self::NonZero;
```
taylordotfish: Confusing name, should be called something like `into_nonzero_or_one` or `into_nonzero_saturating`. The actual conversion has nothing to do with "width" or "size"

crates/atuin-common/src/range/tiled.rs
```diff
@@ -0,0 +1,160 @@
...
+    pub trait Sealed {}
+    impl Sealed for u8 {}
+    impl Sealed for u16 {}
+    impl Sealed for u32 {}
+    impl Sealed for u64 {}
+    impl Sealed for usize {}
+}
+
+#[doc(hidden)]
+pub trait TileIdx: sealed::Sealed + Copy + Ord {
```
taylordotfish: Would prefer to call this `TileInt` or `TileItem` as it's not specific to indexes

crates/atuin-common/src/range/tiled.rs
taylordotfish, line 1: The term "tile" is confusing. When I first saw "tile" I thought the iterator was going to be two-dimensional somehow. Can we call it "chunked" instead or similar?

crates/atuin-common/src/range/tiled.rs
```diff
@@ -0,0 +1,160 @@
...
+use std::num::{NonZeroU8, NonZeroU16, NonZeroU32, NonZeroU64, NonZeroUsize};
+use std::ops::Range;
+
+mod sealed {
+    pub trait Sealed {}
+    impl Sealed for u8 {}
+    impl Sealed for u16 {}
+    impl Sealed for u32 {}
+    impl Sealed for u64 {}
+    impl Sealed for usize {}
```
taylordotfish: `impl_tile_idx!` should emit an impl of `Sealed` instead of having to do it manually here. The impl doesn't need to exist in the `sealed` module, you can just add it to the macro

crates/atuin-common/src/range/tiled.rs
```diff
@@ -0,0 +1,160 @@
...
+        let tile = self.cursor..stop;
+        self.cursor = stop;
+        Some(tile)
+    }
+}
+
+impl<T: TileIdx> FusedIterator for Tiled<T> {}
+
+/// Tile a [`Range`] into fixed-size sub-ranges. See [`Tiled`].
+pub trait RangeTiledExt<T: TileIdx> {
```
taylordotfish: Just call it `RangeExt`; `atuin_common::range::tiled::RangeExt` is already clear, and it gets imported `as _` anyway. If you really want to re-export in the parent module, then re-export `tiled::RangeExt` as `RangeTiledExt`, but the non-reexported version is only two characters longer for the `::` anyway

crates/atuin-common/src/range/tiled.rs
```diff
@@ -0,0 +1,160 @@
...
+        }
+    )+};
+}
+
+impl_tile_idx!(
+    u8 => NonZeroU8,
+    u16 => NonZeroU16,
+    u32 => NonZeroU32,
+    u64 => NonZeroU64,
+    usize => NonZeroUsize,
```
taylordotfish: `NonZero<T>` is now stable, so this could be made simpler by not having to explicitly name every nonzero equivalent

crates/atuin-client/src/api_client.rs
```diff
@@ -268,6 +278,119 @@ pub fn caps_client(
...
+            }
+        }
+    }
+
+    /// Fetch one page of `series`' records: `page.end - page.start` records starting at
+    /// `page.start`.
+    #[instrument(level = "trace", skip(self), err)]
+    async fn page(&self, page: Range<RecordIdx>) -> Result<(u64, Vec<Record<EncryptedData>>)> {
+        let base_url = self.client.sync_addr.append_path("api/v0/record/next")?;
+        let width = page.end - page.start;
```
taylordotfish: It doesn't really make a lot of sense that this method unconditionally returns `page.end - page.start`. The caller can do that if they really need. This should just return the `Vec`

crates/atuin-client/src/api_client.rs
```diff
@@ -268,6 +278,119 @@ pub fn caps_client(
...
+    /// The plan lets us prefetch several pages in parallel at predictable offsets; if the server
+    /// returns a short page mid-stream we fall back to a serial finish from real progress so no
+    /// records are skipped.
+    pub fn stream(
+        self,
+        chunks: Tiled<RecordIdx>,
+    ) -> impl Stream<Item = Result<Vec<Record<EncryptedData>>>> + 'static {
+        try_stream! {
+            // Download the pages in parallel.
+            let mut fetches = stream::iter(chunks).map(|p| self.page(p))
```
taylordotfish: Related to my previous comment, I would do this here instead of returning it from `page`:

```suggestion
            let mut fetches = stream::iter(chunks).map(|p| (p.end - p.start, self.page(p)))
```

crates/atuin-client/src/api_client.rs
```diff
@@ -268,6 +278,119 @@ pub fn caps_client(
...
+            .query(&[
+                ("host", self.series.host_id.to_string()),
+                ("tag", self.series.tag.as_str().to_owned()),
+                ("start", page.start.to_string()),
+                ("count", width.to_string()),
+            ])
+            .send()
+            .await?;
+        let resp = handle_resp_error(resp).await?;
+        let records = resp.json::<Vec<Record<EncryptedData>>>().await?;
```
taylordotfish, line 396: fwiw, turbofish here isn't necessary, the type is unambiguous from the function signature

crates/atuin-client/src/api_client.rs
```diff
@@ -268,6 +278,119 @@ pub fn caps_client(
...
+    }
+
+    /// Stream the record pages the `chunks` plan covers for this series.
+    ///
+    /// The plan lets us prefetch several pages in parallel at predictable offsets; if the server
+    /// returns a short page mid-stream we fall back to a serial finish from real progress so no
+    /// records are skipped.
+    pub fn stream(
+        self,
+        chunks: Tiled<RecordIdx>,
```
taylordotfish: Question: instead of accepting a `Tiled`, can this accept any `impl IntoIterator<Item = Range<T>>`? If it would be very hard to generalize this function as such, then skip it, just wondering if it wouldn't be too much work to add

crates/atuin-client/src/api_client.rs
```diff
@@ -347,9 +470,29 @@ impl Client {
         Ok(())
     }
 
-    /// Upload the given packfile.
+    /// Upload a stream of packfile blobs, transferring several concurrently.
+    #[instrument(level = "trace", skip_all, err)]
+    pub async fn upload_packfiles(
+        &self,
+        packfiles: impl Stream<Item = Result<(RecordId, Vec<RecordId>, Vec<u8>)>>,
```
taylordotfish: The 3-tuple is confusing here; can we make it a named struct?

crates/atuin-client/src/api_client.rs
```diff
@@ -347,9 +470,29 @@ impl Client {
...
+        packfiles
+            .map(move |packfile| {
+                let client = client.clone();
+                async move {
+                    let (manifest_id, record_ids, blob) = packfile?;
+                    client.upload_packfile(manifest_id, &record_ids, blob).await
+                }
+            })
+            .buffered(MAX_CONCURRENT_PACKFILE_UPLOADS)
+            .try_for_each(|()| async { Ok(()) })
```
taylordotfish: imo this reads a bit better, and could be slightly faster due to not having to check a `Result`. works because `()` implements `Extend<()>`

```suggestion
            .try_collect::<()>()
```

taylordotfish (commented):
Partial review on everything except `sync/mod.rs`

crates/atuin-client/src/record/sync/mod.rs
```diff
@@ -401,42 +463,78 @@ impl Keyed<'_> {
...
+        let tiles = (first_missing_local..first_missing_local + expected).tiled(page_size);
+        let ret = self.download_pages(series, tiles, &pb).await?;
+
+        pb.finish_with_message("Downloaded records");
+
+        Ok(ret)
+    }
+
+    #[instrument(level = "trace", skip_all, fields(id = ?manifest.id), err)]
+    pub(super) async fn download_packed(
```
taylordotfish: Does this need to be `pub(super)`? Seems like it's only used in this module.

crates/atuin-client/src/record/sync/mod.rs
```diff
@@ -401,42 +463,78 @@ impl Keyed<'_> {
...
+
+        let pages = self.engine.client.records(series).stream(tiles);
+        futures::pin_mut!(pages);
+        while let Some(page) = pages.next().await {
+            let page = page.map_err(|e| SyncError::RemoteRequestError { msg: e.to_string() })?;
 
-            // We commit the packfile's history into the local store before we persist the manifests, so
-            // a manifest we recorded always has its associated history.
+            // A packfile manifest's history must land in the store *before* the manifest itself, so
+            // a stored manifest always has its records. Loose pages have no such step.
```
taylordotfish, line 542: I don't understand this comment. What is a "loose page"?

taylordotfish (commented):
Latter part of review: `sync/mod.rs`

crates/atuin-common/src/range/tiled.rs
markovejnovic: The iterator is two dimensional though. `.tiled()` raises the dimensionality of the iterator. We go from

```
[1, 2, 3, 4, 5]
```

to

```
[
  [1,2],
  [3,4],
  [5,]
]
```


That being said, I do recognize the name is perhaps less than ideal. I went through some options and I'll list them out:

- `chunked` isn't ideal because this _isn't_ chunking. It's taking a range, and creating an iterator of ranges. There already is a `chunk` operation on `Range`, as `Range` implements `IntoIterator`, meaning that you can call `(x..y).iter().chunked()` and guess what that does -- returns an `Iterator<Iterator<usize>>` (or similar)
- `partition` would mean taking the range and partitioning it on some ordering key. Less than ideal.

crates/atuin-client/src/api_client.rs
markovejnovic: I have no clue how this snuck in

crates/atuin-client/src/api_client.rs
markovejnovic: Sure

crates/atuin-client/src/record/sync/mod.rs
markovejnovic: non packfile

crates/atuin-client/src/api_client.rs
markovejnovic: Unfortunately no since the short page recovery relies on `Tiled`'s methods.

crates/atuin-common/src/range/tiled.rs
taylordotfish: I see your point. In concept, though, I think it is the same as `chunks`. `Range`s *are* iterators, so `tiled` does give you an `Iterator<Iterator<usize>>`, the difference is just that the chunks have a concrete type, `Range`, instead of an opaque struct that implements `Iterator`, enabling you to do more operations on them, like `.start` and `.end`.

So maybe we can call it `.range_chunks`? Or is there a name that communicates “chunks, but where the chunks have the same type as the parent iterator”? `self_chunks`? (don't really like that)

crates/atuin-client/src/record/sync/mod.rs
taylordotfish: I see. Maybe just get rid of the last sentence? I think it's clear without it. Non-packfile records obviously don't have any child history to expand; they're leaf nodes.

If you'd prefer to keep it in, maybe reword to “pages of non-packfile records” to make it clearer for a future reader?

Issue #3561, opened 2026-07-05
[Bug]: `history start` fails after the daemon is killed on terminal exit


nahoj:
This is an issue that I investigated with Claude and the writeup is mostly by it.

With `daemon.enabled = true` and `autostart = true`, the daemon is spawned by whichever client first needs it. Under systemd that means it lands in that terminal's scope — with Konsole, even a per-tab scope:

    /user.slice/user-1000.slice/user@1000.service/app.slice/app-org.kde.konsole-612472.scope/tab(612486).scope

Close that tab and systemd kills the daemon with the scope:

    systemd[5503]: app-org.kde.konsole-16181.scope: Killing process 17354 (atuin) with signal SIGKILL.

Any `history start` racing that window fails once (the next client re-autostarts the daemon). This causes ATUIN_HISTORY_ID to be empty, which crashes `atuin history end` (similar to https://github.com/atuinsh/atuin/pull/3537, except I use Xonsh).

Shipping a systemd service with atuin, as listed in #2008 ("Support automatically installing some service or startup script") would fix the bug. Also, the documentation could mention that `autostart = true` is what ties the daemon's lifetime to an arbitrary terminal.

I tried these 2 ways, to start the daemon either at login or upon first use via the socket, both work:

**Variant A — simple user service, started at login:**

```ini
# ~/.config/systemd/user/atuin-daemon.service
[Unit]
Description=Atuin daemon
# bound retries so a crash-looping daemon (e.g. version rollback with a
# newer DB schema, #3382) fails fast instead of looping forever
StartLimitIntervalSec=60
StartLimitBurst=5
[Service]
# the daemon dies with "Address in use" on a stale socket file left by an
# unclean death, which would otherwise break every auto-restart, #3516
ExecStartPre=/bin/rm -f %t/atuin.sock
ExecStart=/path/to/atuin daemon start
Restart=on-failure
RestartSec=1
[Install]
WantedBy=default.target
```

**Variant B — socket activation** (uses `systemd_socket = true`): add a socket unit, drop the `ExecStartPre` (systemd owns the socket), and add `Requires=atuin.socket` to the service:

```ini
# ~/.config/systemd/user/atuin.socket
[Unit]
Description=Atuin daemon socket
[Socket]
ListenStream=%t/atuin.sock
SocketMode=0600
Service=atuin-daemon.service   # required: unit names differ
[Install]
WantedBy=sockets.target
```

### Atuin doctor output

```yaml
{
  "atuin": {
[...]

Issue #2289, opened 2024-07-17
Error running atuin daemon


raffaelewylde:
### What did you expect to happen?

expected atuin daemon to startup

### What happened?

I know the daemon isn't completely implemented yet. Figured I'd let you guys know about this issue. I've tried running the daemon every so often fo a couple weeks and always get the same error.
```
 atuin daemon                     
Error: No such file or directory (os error 2)

Location:
    /home/wylde/.cargo/registry/src/index.crates.io-6f17d22bba15001f/atuin-daemon-0.2.0/src/server.rs:206:10
```
and yet that file exists...
```
 ls  /home/wylde/.cargo/registry/src/index.crates.io-6f17d22bba15001f/atuin-daemon-0.2.0/src/ 
server/  client.rs  history.rs  lib.rs  server.rs*
```
The contents of server.rs are 

```
use eyre::WrapErr;

use atuin_client::encryption;
use atuin_client::history::store::HistoryStore;
use atuin_client::record::sqlite_store::SqliteStore;
use atuin_client::settings::Settings;
use std::path::PathBuf;
use std::sync::Arc;
use time::OffsetDateTime;
use tracing::{instrument, Level};

use atuin_client::database::{Database, Sqlite as HistoryDatabase};
use atuin_client::history::{History, HistoryId};
use dashmap::DashMap;
use eyre::Result;
use tonic::{transport::Server, Request, Response, Status};

use crate::history::history_server::{History as HistorySvc, HistoryServer};

use crate::history::{EndHistoryReply, EndHistoryRequest, StartHistoryReply, StartHistoryRequest};

mod sync;

#[derive(Debug)]
pub struct HistoryService {
    // A store for WIP history
    // This is history that has not yet been completed, aka a command that's current running.
    running: Arc<DashMap<HistoryId, History>>,
    store: HistoryStore,
    history_db: HistoryDatabase,
}

impl HistoryService {
    pub fn new(store: HistoryStore, history_db: HistoryDatabase) -> Self {
        Self {
            running: Arc::new(DashMap::new()),
            store,
            history_db,
        }
    }
}

#[tonic::async_trait()]
impl HistorySvc for HistoryService {
    #[instrument(skip_all, level = Level::INFO)]
    async fn start_history(
        &self,
        request: Request<StartHistoryRequest>,
    ) -> Result<Response<StartHistoryReply>, Status> {
        let running = self.running.clone();
        let req = request.into_inner();

        let timestamp =
[...]

PR #3957, opened 2026-08-20
fix(daemon): recover from a stale socket after an unclean shutdown

### LINKED ISSUE #3516: [Bug]: atuin daemon will not recover after an abrupt termination or system reboot

If the atuin daemon is stopped abruptly, .or the machine is rebooted, the atuin daemon should be able to start on next boot without needing manual human intervention.

Some people also reported this on https://github.com/atuinsh/atuin/issues/2289 - but the actual original issue there seems to be a separate issue.

### What happened?

On reboot, the atuin daemon will fail to start.

It will continually try to restart every 10 seconds - and spam the logfile (e.g. `~/.atuin/logs/daemon.log.2026-06-05`) with the following, ad infinitium:
```
2026-06-05T16:25:19.685229+1000  INFO atuin_daemon::components::sync: sync component started
2026-06-05T16:25:19.686023+1000  INFO atuin_client::settings::watcher: starting config file watcher: "/Users/victorhooi/.config/atuin/config.toml"
2026-06-05T16:25:19.688368+1000  INFO atuin_client::settings::watcher: config file watcher initialized for: "/Users/victorhooi/.config/atuin"
2026-06-05T16:25:19.688424+1000  INFO atuin_daemon::server: listening on unix socket "/Users/victorhooi/.local/share/atuin/daemon.sock"
2026-06-05T16:25:29.730920+1000  INFO atuin_daemon::daemon: starting component component="history"
2026-06-05T16:25:29.731967+1000  INFO atuin_daemon::components::history: history component started
2026-06-05T16:25:29.731998+1000  INFO atuin_daemon::daemon: starting component component="search"
2026-06-05T16:25:29.732014+1000  INFO atuin_daemon::components::search: search component started
2026-06-05T16:25:29.732024+1000  INFO atuin_daemon::daemon: starting component component="sync"
2026-06-05T16:25:29.732037+1000  INFO atuin_daemon::components::sync: sync component started
2026-06-05T16:25:29.732533+1000  INFO atuin_client::settings::watcher: starting config file watcher: "/Users/victorhooi/.config/atuin/config.toml"
2026-06-05T16:25:29.735702+1000  INFO atuin_client::settings::watcher: config file watcher initialized for: "/Users/victorhooi/.config/atuin"
2026-06-05T16:25:29.735807+1000  INFO atuin_daemon::server: listening on unix socket "/Users/victorhooi/.local/share/atuin/daemon.sock"
2026-06-05T16:25:39.796947+1000  INFO atuin_daemon::daemon: starting component component="history"
2026-06-05T16:25:39.799261+1000  INFO atuin_daemon::components::history: history component started
2026-06-05T16:25:39.799298+1000  INFO atuin_daemon::daemon: starting component component="search"
2026-06-05T16:25:39.799319+1000  INFO atuin_daemon::components::search: search component started
2026-06-05T16:25:39.799334+1000  INFO atuin_daemon::daemon: starting component component="sync"
2026-06-05T16:25:39.799350+1000  INFO atuin_daemon::components::sync: sync component started
2026-06-05T16:25:39.800130+1000  INFO atuin_client::settings::watcher: starting config file watcher: "/Users/victorhooi/.config/atuin/config.toml"
2026-06-05T16:25:39.804194+1000  INFO atuin_client::settings::watcher: config file watcher initialized for: "/Users/victorhooi/.config/atuin"
2026-06-05T16:25:39.804281+1000  INFO atuin_daemon::server: listening on unix socket "/Users/victorhooi/.local/share/atuin/daemon.sock"
2026-06-05T16:25:49.862432+1000  INFO atuin_daemon::daemon: starting component component="history"
2026-06-05T16:25:49.863792+1000  INFO atuin_daemon::components::history: history component started
2026-06-05T16:25:49.863823+1000  INFO atuin_daemon::daemon: starting component component="search"
2026-06-05T16:25:49.863847+1000  INFO atuin_daemon::components::search: search component started
2026-06-05T16:25:49.863865+1000  INFO atuin_daemon::daemon: starting component component="sync"
2026-06-05T16:25:49.863885+1000  INFO atuin_daemon::components::sync: sync component started
```

To fix it - you have to manually delete the lockfile using rm.

### Atuin doctor output

```yaml
Ah drat - sorry - I ran this *after* I'd manually remove the lockfile:


✦ ❯ atuin doctor
Atuin Doctor
Checking for diagnostics


{
  "atuin": {
    "version": "18.16.1",
    "commit": "NO_GIT",
    "sync": {
      "records": true,
      "auto_sync": true,
      "last_sync": "2026-06-05 9:55:18.092128 +00:00:00"
    },
    "sqlite_version": "3.46.0"
  },
  "shell": {
    "name": "nu",
    "default": "nu",
    "plugins": [
      "atuin"
    ],
    "preexec": "built-in"
  },
  "system": {
    "os": "Darwin",
    "arch": "arm64",
    "version": "26.5.1",
    "disks": [
      {
        "name": "Macintosh HD",
        "filesystem": "apfs"
      },
      {
        "name": "Macintosh HD",
        "filesystem": "apfs"
      }
    ]
  }
}
```

### Code of Conduct

retif:
Fixes #3516.

## The problem

If the daemon dies without shutting down cleanly, it never starts again. Its socket file outlives it, `bind` then fails with `EADDRINUSE` forever, and the only way out is deleting the file by hand.

This is *not* a missing-cleanup bug — the daemon does remove its socket. But it does so from inside the shutdown-signal future in `run_grpc_server`, gated on receiving `DaemonEvent::ShutdownRequested`. `SIGKILL` delivers no event and runs no destructor, so that path simply never executes. What is missing is **recovery on the next start**.

## How I hit it

atuin 18.16.1 on NixOS, under a systemd user unit with `Restart=on-failure` / `RestartSec=5s`.

`systemd-oomd` killed the daemon under memory pressure:

```
systemd-oomd[712]: Marked .../atuin-daemon.service for killing due to memory pressure
  for .../user@1000.service being 92.91% > 90.00% for > 1min
atuin-daemon.service: Main process exited, code=killed, status=9/KILL
```

Every restart after that failed:

```
atuin: Error: Address already in use (os error 98)
Location: crates/atuin-daemon/src/server.rs:69:10
```

which the restart policy turned into an unbounded loop — the restart counter reached **11205**. The socket file was there but dead: `connect()` to `/run/user/1000/atuin.sock` returned `ECONNREFUSED`, and `ss -lx` showed no listener for it.

The part that makes this worth fixing upstream is that **nothing surfaces to the user**. Shell history silently stopped being recorded for about 20 hours. No error at the prompt; up-arrow and `ctrl-r` just quietly stopped showing new commands.

It is reproducible without the OOM killer — `kill -9` the daemon and start it again. #3561 hits the same thing when a terminal scope is torn down, and #2289 has reports of it too.

## The fix, and why it is not just `remove_file`

The obvious patch is to unlink the path before binding. That would be worse than the bug it fixes: two daemons racing at startup means the second one silently steals the socket of the first, and the first keeps running while serving nobody.

So the socket is probed before it is reclaimed:

- try `bind`;
- only on `ErrorKind::AddrInUse`, `connect` to the same path;
- `ECONNREFUSED` means the file exists with nothing behind it — it is stale, so remove it and `bind` once more;
- a successful `connect` means a live daemon owns the socket — return an error, do not unlink;
- any other `connect` error is not evidence either way, so return the original bind error and leave the file alone.

That is the same probe I used by hand to diagnose the incident above.

The systemd socket-activation branch is deliberately untouched: it receives a pre-bound listener from `listenfd` and returns `None` for the cleanup path, so it never binds the path itself. The existing `.context("reading socket: …")` wrapping is preserved.

The only other change is moving `use tokio::net::UnixListener` into the `cfg(target_os = "linux")` block, since after this change that is the only place in `run_grpc_server` still naming it, and leaving it at the top would be an unused import on other unix targets.

## Tests

Three unit tests in `crates/atuin-daemon/src/server.rs`, covering the three outcomes that matter:

- binding a path that does not exist yet;
- reclaiming a socket left behind by a dead daemon (bind, close the listener without unlinking — which is what a `SIGKILL`ed process leaves — then assert the next bind succeeds and accepts connections);
- refusing to steal a live daemon's socket (assert it fails with `AddrInUse`, the file is still there, and the original daemon is still reachable).

## What I verified, and what I did not

I diagnosed and reproduced the failure on the affected machine, and confirmed against `main` (not just the packaged 18.16.1) that both the bind site and the shutdown-gated cleanup are unchanged. I also verified the workaround that this change makes unnecessary — an `ExecStartPre=-rm -f %t/atuin.sock` in the unit file — which fixed the loop on that machine, but every packager has to rediscover it independently.

I have to be straight about one thing: my laptop cannot build this workspace (it OOMs), so I have **not** run `cargo build`/`test`/`clippy` against this change locally. I checked formatting with a nightly `rustfmt` and wrote the code against the current lint config, but the compile, the tests and clippy are on CI here. If CI turns something up I will fix it; if you would rather have a maintainer's build behind it before merging, that is completely fair.

## Checks


retif:
The Greptile summary above is correct and I've fixed it in d5d98988 — its review is pinned to the first commit, so that comment is now stale.

I'd read `CONTRIBUTING.md` but not `AGENTS.md`, which asks for `#[rstest]` on every test rather than a bare `#[test]` (`#[rstest]` + `#[tokio::test]` for async), and for shared setup to come from a `#[fixture]` returning an RAII guard where teardown is needed. The three tests now take a `socket` fixture that holds the `TempDir` alongside the path, instead of each building its own. The assertions are unchanged, and all three still run:

```
PASS atuin-daemon server::unix_tests::binds_a_path_that_does_not_exist_yet
PASS atuin-daemon server::unix_tests::reclaims_a_socket_left_behind_by_a_dead_daemon
PASS atuin-daemon server::unix_tests::refuses_to_steal_a_live_daemons_socket
```

`server.rs` had no other tests, so there were no plain `#[test]`s to migrate alongside them.

PR #3961, opened 2026-08-20
refactor(record): name the (host, tag) pair as RecordSeriesKey

markovejnovic:
Introduces a `RecordSeriesKey`, which is a struct which uniquely identifies the `(tag, host)` pair we historically had.

This new type semantically conveys what the old loose types did.

Benefits:
- Functions that needed `2` args now only need one.
- Semantic meaning is verbosely conserved.
- Type safety is always good.

### DIFF
--- a/crates/atuin-domain/src/record/mod.rs
+++ b/crates/atuin-domain/src/record/mod.rs
@@ -84,12 +84,6 @@ impl RecordSeriesKey {
     }
 }
 
-impl From<(HostId, RecordTag)> for RecordSeriesKey {
-    fn from((host, tag): (HostId, RecordTag)) -> Self {
-        Self { host, tag }
-    }
-}
-
 /// A single record stored inside of our local database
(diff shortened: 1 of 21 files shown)


crates/atuin-domain/src/record/mod.rs
```diff
@@ -71,6 +70,34 @@ new_uuid!(HostId);
 
 pub type RecordIdx = u64;
 
+/// The composite key identifying a single append-only record stream.
+///
+/// Every [`Record`] belongs to exactly one stream, addressed by its owning [`HostId`] and its
+/// [`RecordTag`]; a record's [`RecordIdx`] is only unique *within* one such stream. This pair
+/// travels together through nearly every record-store and sync API, so we give it a name rather
+/// than threading `(HostId, RecordTag)` as two separate arguments everywhere.
```
markovejnovic:
```suggestion
/// The composite key identifying record WAL journal uniqueness.
```

ellie (approved):

PR #3987, opened 2026-08-25
docs: document settings missing from config.toml

harshil-mistry:
Seven settings have a `set_default` in `settings.rs` but appear in neither `config.toml` nor the docs site, so the only way to find them is to read the source:

| Setting | Default |
| --- | --- |
| `keymap_mode_shell` | `"auto"` |
| `no_mouse` | `false` |
| `smart_sort` | `false` |
| `keys.prefix` | `"a"` |
| `search.recency_score_multiplier` | `1.0` |
| `search.frequency_score_multiplier` | `1.0` |
| `search.frecency_score_multiplier` | `1.0` |

Added them to `config.toml`, commented out with their real defaults, following the file's existing convention. Descriptions come from the code rather than being invented — the three multipliers use the wording from the `rebuild_frecency` doc comment in `atuin-daemon/src/search/index.rs` ("Weight for recency component", "Weight for frequency component", "Overall multiplier for final score").

I deliberately didn't add the rest of the undocumented `set_default` keys: the `*.db_path` ones are derived from `data_dir` rather than things you'd normally set, and `theme.*`, `logs.*` and `shell_up_key_binding` are already covered on the docs site.

While checking this I noticed something unrelated that I've left alone: `Search::default()` orders filters as global, host, session, **session-preload**, workspace, directory, while `set_default("search.filters", ...)` orders them global, host, session, workspace, directory, **session-preload**. Happy to open a separate issue if that's not intentional.

---

AI disclosure: I used Claude (Opus 5) to diff `set_default` keys against `config.toml` and the docs. I checked each setting's usage in the code myself before writing its description.

### DIFF
--- a/crates/atuin-client/config.toml
+++ b/crates/atuin-client/config.toml
@@ -185,11 +185,6 @@ enter_accept = true
 ## mode in the Atuin search is forced to be the specified one.
 # keymap_mode = "emacs"
 
-## the keymap mode used when Atuin is invoked from a shell up-key binding.
-## "auto" follows the shell's own keymap; otherwise the accepted values are
-## identical to those of "keymap_mode"
-# keymap_mode_shell = "auto"
-
 ## disable mouse capture in the interactive search TUI, so the terminal keeps
 ## its native selection and scrolling behaviour
 # no_mouse = false
--- a/crates/atuin-client/config.toml
+++ b/crates/atuin-client/config.toml
@@ -185,6 +185,18 @@ enter_accept = true
 ## mode in the Atuin search is forced to be the specified one.
 # keymap_mode = "emacs"
 
+## the keymap mode used when Atuin is invoked from a shell up-key binding.
+## "auto" follows the shell's own keymap; otherwise the accepted values are
+## identical to those of "keymap_mode"
+# keymap_mode_shell = "auto"
+
+## disable mouse capture in the interactive search TUI, so the terminal keeps
+## its native selection and scrolling behaviour
+# no_mouse = false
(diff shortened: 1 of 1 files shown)


harshil-mistry:
Thanks @greptile-apps — both points addressed.

**CRLF:** already fixed. The bot reviewed `b7f66a4`; the branch was force-updated to LF-only before that comment landed. Current diff is `+20/-0` on a single hunk set, no line-ending churn.

**`keymap_mode_shell` (P1):** correct, and I have dropped it from this PR. Two reasons, both confirmed in source:

1. `crates/atuin/src/command/client/search.rs:236` assigns it unconditionally:
   ```rust
   settings.keymap_mode = match settings.keymap_mode {
       KeymapMode::Auto => self.keymap_mode,
       value => value,
   };
   settings.keymap_mode_shell = self.keymap_mode;
   ```
   `keymap_mode` deliberately lets config.toml win over the `--keymap-mode` keybinding flag; `keymap_mode_shell` has no such guard, so a value set in config.toml is always clobbered.
2. Its only reader is `Self::finalize_keymap_cursor` (`interactive.rs:295`), which restores the *cursor shape* on exit — it does not participate in keymap selection at all (`mode_keymap` reads `self.keymap_mode`).

So documenting it as a user-facing setting would have been wrong on both counts. The remaining six settings (`no_mouse`, `smart_sort`, `prefix`, and the three score multipliers) are unaffected.

Happy to open a separate issue about the asymmetry in (1) if that is worth tracking on its own.

PR #3988, opened 2026-08-25
chore: revert changes to `atuin-pty-proxy` from `lab share` PR / feat: store pty-proxy sockets in `/tmp/atuin-$UID`

taylordotfish:
### Summary

The `lab share` PR (https://github.com/atuinsh/atuin/pull/3795) was intended to be self-contained given the “experimental” nature of the code; however, it contained many large changes to `pty-proxy`. These ~2400 lines of code have not passed rigorous review and should not be present in `pty-proxy`, which powers core features, not just `lab` experiments.

This PR reverts the changes in `atuin-pty-proxy` while making the necessary changes to `atuin-lab-share` to preserve as much functionality as possible.

### Backward compatibility

This PR should continue to work with old pty-proxy servers. An explicit write-side socket shutdown is performed to prevent old pty-proxy servers from waiting 100ms for a magic byte that we no longer send.

### PTY proxy socket path change

The `lab share` PR caused pty-proxy sockets to be stored in a subdirectory of `/tmp`. Previously, they were directly stored at `/tmp/atuin-pty-proxy-{pid}.sock`. This is not ideal, as socket permissions are not standard across systems, and the best way to prevent unauthorized access is to put the socket in a directory readable only by the owner. To avoid regressing this case, this PR keeps (but rewrites) the change to store pty-proxy sockets in subdirectories. This time, however, **they're stored in `/tmp/atuin-$UID/`**, the same directory used for the daemon socket. This should not affect existing sessions, as the socket path is passed in an environment variable.

If we fail to create the socket or socket directory, we print an error but continue running, so that the user still gets a subshell. This matches the existing behavior (even before the `lab share` PR) when we fail to bind to the socket.

### Blocking `send`

This PR changes the PTY-proxy to use blocking `send` instead of `try_send`. If the buffer is full and a `try_send` fails, the terminal snapshot can diverge with potentially disastrous results. This is especially true of “resize” messages. The performance impact of changing to `send` is minor:

`seq 1 200000` end-to-end:

| method     | run 1 | run 2 |
| ---------- | ----- | ----- |
| `try_send` | 2.08s | 2.13s |
| `send`     | 2.17s | 2.22s |

Only 4% slower on this pathological case.

### Review order

I suggest reviewing commit-by-commit. The first two commits are very large and mostly mechanical – this is the churn for the revert. I suggest mostly giving them a quick overview to ensure nothing was unintentionally touched. The rest of the commits should be reviewed in more detail.


crates/atuin-common/src/os/unix.rs
```diff
@@ -29,6 +29,95 @@ pub fn touch_file(path: &Path) -> std::io::Result<()> {
     Ok(())
 }
 
+/// Get the global temporary directory.
+pub fn tmp_dir() -> PathBuf {
+    // TODO: We should perhaps use `std::env::temp_dir()` instead, but that would be a breaking
+    // change and could cause clients to fail to connect to an older running daemon.
+    crate::utils::env_nonempty("TMPDIR").map_or_else(|| "/tmp".into(), Into::into)
+}
```
markovejnovic, line 37: I dislike this utility because I think it's a footgun, for multiple sets of reasons:

1. I do not think this needs to unanimously return a `PathBuf`. If it lives on the `AppCtx` (as I proposed it with you), we get to return a `&Path`. Less allocations, more good.

I'll leave my other comments in `create_named_temp_dir`

crates/atuin-common/src/os/unix.rs
```diff
@@ -29,6 +29,95 @@ pub fn touch_file(path: &Path) -> std::io::Result<()> {
...
+///
+/// Every component of the path except the last must exist: this function does not create ancestors.
+///
+/// This function ensures the directory is owned by the current user with appropriate permissions to
+/// prevent other users from accessing its contents. This is especially important for sockets --
+/// some systems ignore permissions on sockets themselves and allow any user who can access the
+/// socket file to connect to it.
+///
+/// On success, returns `path`. This may allow resources to be reused if `P` is an owned type.
+pub fn create_named_temp_dir<P>(path: P) -> Result<P, NamedTempDirError>
```
markovejnovic: Feedback, continued:

1. The utility does not give me the ability to change the mode. Do we **really** want to use `0o700`, for every tmp dir we create? Probably not.
2. The utility does not enforce management of the tempdir. Nothing is stopping me from forgetting to call `P.rmdir()` as a user of this utility.
3. The utility is overly-specific. All this mode logic, all this chown logic, it's all very specific to some sort of _secure_-ish tmpdir. That can be the default, but I'd like to see more flexibility here.

I would much more prefer if we wrote this:

```rust
struct TmpDirBuilder<P>;

impl<P> TmpDirBuilder<P> {
  pub fn name(self, name: &str) -> Self;
  pub fn perms(self, octal: u8) -> TmpDirPermsBuilder<P>;
  pub fn open(self) -> TmpDir<P>;
}

struct TmpDirPermsBuilder<P> {
  builder: &mut TmpDirBuilder<P>,
}

impl<P> TmpDirPermsBuilder<P> {
  /// Sets `mode` to `0o700`, bans symlinks, and sets `chown` to the given user.
  pub fn ultrastrict(self, user: u32) -> Self;
  pub fn mode(self, chmod: u32);
  pub fn user(self, user: u32);
  pub fn allow_symlink(self, allow: bool);
}

struct TmpDir<P>;

impl<P> TmpDir<P> {
  pub fn builder() -> TmpDirBuilder<P>;
  pub fn leak(self) -> std::fs::File;
}

impl<P> Drop for TmpDir<P> {
  fn drop(&mut self) {
    rmdir(self.fd);
  }
}
```

Can do later, won't block this PR on it, obviously.

markovejnovic (approved):

crates/atuin-common/src/os/unix.rs
taylordotfish: Happy to make that change once `AppCtx` is merged.

crates/atuin-common/src/os/unix.rs
taylordotfish: Renamed to `create_secure_temp_dir`. If we need flexible permissions later, we can refactor and make it more general.

Discussed in person: we do not want to remove the temp subdirectory itself, only the contents in it. An RAII wrapper for the sockets we put *in* the subdirectory would be good and I'd support that in a followup PR. As for the subdirectory itself, keeping it around is beneficial to reduce the chance that a different user creates `/tmp/atuin-{our uid}`.

PR #3990, opened 2026-08-25
fix(client): checkpoint WAL files before they grow unbounded

### LINKED ISSUE #3989: [Bug]: history.db-wal grows unbounded on machines running many concurrent shells

`history.db-wal` should stay roughly bounded — close to SQLite's own ~4 MB compile-time auto-checkpoint threshold — and shell hooks (`history start`/`history end`, which run on every single prompt) should stay fast regardless of how many terminal sessions are open at once.

### What happened?

On a machine that regularly runs dozens of concurrent shell/terminal sessions, `history.db-wal` grew to **23 MB** next to a **1.8 MB** `history.db` (3,640 history rows) — 13x the size of the database itself, and never checkpointed on its own.

Traced with `fs_usage` while triggering a single shell-hook invocation (one `cd`): **5,639 `pread` syscalls** against the WAL file for that one event. After manually running `PRAGMA wal_checkpoint(TRUNCATE)` (which dropped the WAL to 0 bytes), the identical event cost **1,071** syscalls — an 82% reduction, for what should be a near-instant operation on a 3,640-row table.

Day to day this showed up as `mc` (Midnight Commander) being slow to change directory — `mc`'s own panel navigation never touches a shell, but typing `cd` at its built-in command line runs through a real background subshell with the same `precmd`/`chpwd` hooks as any interactive terminal, and that subshell's atuin hook was paying the WAL-scan cost on every single `cd`.

**Root cause**, confirmed with a clean sequential repro (zero concurrency, no reader contention involved): SQLite's automatic checkpoint runs in `PASSIVE` mode, which never shrinks the WAL file on disk even when it fully succeeds — only `TRUNCATE` does. Atuin never issues one. 30 sequential `history start`/`history end` pairs (clean process-exit gap after each) grew the WAL unboundedly (~70 KB/pair, 2.39 MB after 30 pairs) with no reduction at any point. Since every shell hook invocation opens a fresh short-lived connection pool against the same database file, and there's rarely a moment with zero concurrent readers on a machine running many shells at once, the default auto-checkpoint threshold is never enough to keep the file bounded — reader-overlap starvation of `TRUNCATE`-level checkpoints makes it worse still, though it isn't required to reproduce the underlying growth.

I have a fix ready (explicit threshold-based `wal_checkpoint(TRUNCATE)`, both on pool-open for the CLI and periodically for the daemon) and will open a PR referencing this issue.

### Atuin doctor output

Original environment where this was observed:

```yaml
atuin: 18.18.1 (Homebrew via Nix/home-manager)
shell: zsh, `atuin init zsh --disable-up-arrow`, `auto_sync = false`
os: macOS 26.6.2
```

Re-confirmed against a current build while preparing the fix:

```json
{
  "atuin": {
    "version": "18.20.0-beta.3",
    "sqlite_version": "3.46.0",
    "daemon_enabled": false
  },
  "shell": {
    "name": "zsh",
    "default": "zsh",
    "plugins": ["atuin"],
    "preexec": "built-in"
  },
  "system": {
    "os": "Darwin",
    "arch": "arm64",
    "version": "26.6.2"
  }
}
```

Vany:
Fixes #3989

## Summary

`history.db-wal` (and the other client-side WAL-mode stores: record store, KV, scripts, AI sessions) can grow unbounded on machines that run many concurrent shells. Root cause: SQLite's automatic checkpoint runs in `PASSIVE` mode, which never shrinks the WAL file on disk even when it fully succeeds — only `TRUNCATE` does, and Atuin never issued one. See #3989 for the full repro and `fs_usage` numbers (23 MB WAL next to a 1.8 MB db; 5,639 `pread` syscalls for one `cd` hook vs 1,071 after a manual `wal_checkpoint(TRUNCATE)`).

- `atuin-common::sqlite::checkpoint_wal_if_needed`: `stat()`s the `-wal` sidecar, and if it's over 16 MiB, runs `PRAGMA wal_checkpoint(TRUNCATE)` on a dedicated short-lived connection (own `busy_timeout`, never ties up a shared pool connection). Wired into every client-side WAL store, on every pool-open.
- The CLI opens a fresh pool per shell-hook invocation, so the pool-open check alone is enough there.
- The daemon opens its pools once and holds them for its whole lifetime, so it additionally gets a periodic (5 min) background re-check — otherwise the checkpoint would only ever fire once, at daemon boot.

## Verification

- Unit tests in `atuin-common` (temp-file-backed WAL db, threshold passed directly): missing WAL is a no-op, under-threshold WAL is left alone, over-threshold WAL gets truncated.
- End-to-end against the real CLI binary, at the actual production threshold (not just synthetic unit-test thresholds): 500 sequential `history start`/`history end` pairs, isolated data dir, zero concurrency. WAL saw-tooths between 0 and ~16.8 MB, truncating every ~230 pairs (matches the original report's own estimate), averaging ~8 MB steady-state.
- Re-verified again after the final connection-handling change (dedicated connection instead of the shared pool): 300 pairs, truncates at pair 231, same behavior.

## Test plan

- [x] `cargo build --workspace --all-features`
- [x] `cargo clippy --workspace --all-features -- -D warnings`
- [x] `cargo +nightly fmt --check`
- [x] `cargo test -p atuin-common --lib --features sqlite`
- [x] Manual end-to-end repro against the built CLI binary (see above)


markovejnovic (changes requested):
This PR seems exclusively vibe-coded with little human review. As per our AI policy, I am closing this:

> While we are very happy for you to use AI to build your contribution, it is essential that any pull requests you open are reviewed, understood, and tested by you. Pull requests with giant, AI-written descriptions imply the opposite. Please take the time to write a PR body that outlines your intent, understanding and summary of the change.

Thank you for your PR, but as I find the quality of this PR to be less than the standard we are setting, I am going to close it and open a PR myself with this, in a couple minutes. Should be linked to your issue.

PR #3995, opened 2026-08-26
feat: switch terminal emulation to `atuin-vt100`

taylordotfish:
Switch terminal emulation from [`vt100`](https://github.com/doy/vt100) to [`atuin-vt100`](https://github.com/atuinsh/vt100-rust), Atuin's fork of `vt100`.

This is necessary to enable us to capture the output of commands. It also fixes panics that required us to clamp screen size to a minimum of 2x2, and improves the performance of handling certain kinds of terminal data (multi-line and multi-char escape sequences, in particular).

### Why a fork?

To support some upcoming features in Atuin, like capturing the output of commands, we need more flexible terminal emulation. In particular, we need to make sure output capture is fast and low on memory use, but the design of `vt100` makes this difficult. `atuin-vt100` adds a mechanism for streaming capture of output without having to create a large scrollback buffer. Additionally, we want to preserve formatting like colors and bold, but the “formatted” output provided by `vt100` contains escape sequences that make processing difficult, like sequences that move the cursor to an absolute position. `atuin-vt100` adds a “basic formatted” style that restricts escape sequences to SGR escapes, and control characters to `'\n'`.

We can also use this opportunity to fix some bugs and improve performance. `atuin-vt100` fixes panics on small screen sizes, improves handling of IL and DL sequences to be spec-compliant, and improves the performance of many operations, especially escape sequences that insert/delete multiple rows/characters. See the [changelog](https://github.com/atuinsh/vt100-rust/blob/main/CHANGELOG.md#0170---2026-08-25) for more details.

### Can the changes be upstreamed?

The fix for panics on small screen sizes has an open upstream PR [here](https://github.com/doy/vt100-rust/pull/41). However, upstream seems unmaintained or inconsistently maintained, so we don't want to be blocked on upstream acceptance. Additionally, in my opinion, many of the changes in `atuin-vt100` are opinionated to suit our needs and wouldn't necessarily be good candidates for upstreaming—for example, the new `on_scroll` callback only supports the “basic formatted” style for row contents rather than plain text and fully formatted too. If upstreamed, it would likely need to have support for these other styles, and the basic formatted style itself, which is also new in our fork, could be the subject of debate as well. In the interest of maintaining high development speed on our new features that require more flexible terminal emulation, at this time, a fork seems like the best approach. Of course, iike `vt100`, `atuin-vt100` is fully open-source, released under the permissive MIT license, so everyone is free to copy and adapt our code.

### What changed in the fork?

See the [changelog](https://github.com/atuinsh/vt100-rust/blob/main/CHANGELOG.md#0170---2026-08-25) for an overview. For more details, I recommend looking at the [pull requests](https://github.com/atuinsh/vt100-rust/pulls?q=is%3Apr+is%3Amerged+sort%3Aupdated-asc) that have been merged.

### Who will maintain the fork?

I, @taylordotfish, will maintain the fork, although contributions are welcome from all. I am not expecting the fork to require much maintenance except when we need modifications to support a new feature in Atuin.

### Behavior changes

IL and DL now behave according to the spec (no-ops when the cursor is outside the scroll region). This is an improvement, but technically a change. Also, previously we would clamp terminal dimensions to 2 to avoid panics; now that the panics have been fixed, we clamp them to 1. This means that if you write a wide character to a 1x1 terminal, the character will be discarded (you can't render a wide character on such a terminal); previously it would “work” but only because we were emulating the wrong screen size.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -345,6 +345,7 @@ dependencies = [
  "atuin-common",
  "atuin-daemon",
  "atuin-domain",
+ "atuin-vt100",
  "chrono",
  "chrono-humanize",
  "clap",
@@ -390,7 +391,6 @@ dependencies = [
  "typed-builder",
  "unicode-width 0.2.2",
  "uuid",
- "vt100",
  "xxhash-rust",
  "yaml-rust2",
 ]
@@ -464,6 +464,7 @@ dependencies = [
 name = "atuin-common"
 version = "18.20.0"
 dependencies = [
+ "atuin-vt100",
  "base64",
  "crypto_secretbox",
  "derive_more",
@@ -499,7 +500,6 @@ dependencies = [
  "unicode-width 0.2.2",
  "url",
  "uuid",
- "vt100",
  "zeroize",
 ]
 
@@ -655,13 +655,13 @@ name = "atuin-pty-proxy"
 version = "18.20.0"
 dependencies = [
  "atuin-common",
+ "atuin-vt100",
  "clap",
  "crossterm",
  "eyre",
  "portable-pty",
  "rstest",
  "signal-hook",
- "vt100",
 ]
 
 [[package]]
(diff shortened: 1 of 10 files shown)


markovejnovic:
Let's talk about this in-person. The efficiency commits can be upstreamed, I suppose, and I'd like to understand more about the `String` commit.

markovejnovic (approved):

PR #4001, opened 2026-08-26
feat(sqlite): rename compactor to maintenance_task + incremental vacuum

markovejnovic:
## What

Renames the SQLite `compactor` module in `atuin-common` to `maintenance_task`, and extends its background task so that — in addition to compacting the WAL — it now reclaims free space via SQLite **incremental auto-vacuum**.

## Why

The WAL compactor (#3994) only bounds the WAL. Nothing reclaimed free pages left behind by deletes/retention, so the main DB file only ever grew. This generalises that one background task into a maintenance task that also vacuums, on a low-frequency cadence, without disrupting the frequent-write workload.

## Approach

The strategy was chosen after deep research into SQLite vacuuming (sources: sqlite.org VACUUM/PRAGMA/WAL docs, Jim Nelson's auto_vacuum writeup, PhotoStructure's WAL-VACUUM notes, sqlx discussion #3284):

- **`auto_vacuum = INCREMENTAL`** rather than `FULL` (which taxes every commit) or a periodic full `VACUUM` (2× disk, whole-DB write lock, WAL spike). Incremental gives bounded, low-lock, on-demand reclaim that coexists with the daemon and the existing checkpoint compactor.
- **New WAL databases** are created in `auto_vacuum=INCREMENTAL` at creation (the pragma runs before any table exists).
- **Existing field databases** (default `auto_vacuum=NONE`) are converted **once** via `PRAGMA auto_vacuum=INCREMENTAL; VACUUM; wal_checkpoint(TRUNCATE)` the first time the vacuum check runs.
- **Reclaim** is threshold-gated (`free_ratio ≥ 20% && reclaimable ≥ 8 MiB`), then done in bounded `incremental_vacuum(256)` chunks with a no-progress guard, ending in `wal_checkpoint(TRUNCATE)` so freed space actually leaves the WAL and shrinks the file on disk.
- The vacuum check runs on a **6 h cadence with a 5-minute initial delay** (never at startup); WAL compaction keeps its 1-minute cadence. Both share the maintenance task's single dedicated connection via `tokio::select!` (one arm at a time — no aliasing, and VACUUM/checkpoint self-serialise rather than contend).

## Commits

1. `refactor(sqlite)`: rename `compactor` module → `maintenance_task` (pure rename, `git mv`, no behaviour change)
2. `feat(sqlite)`: create new WAL databases with `auto_vacuum=INCREMENTAL`
3. `feat(sqlite)`: add incremental vacuum + legacy conversion helpers (with unit + real-DB tests)
4. `feat(sqlite)`: run periodic vacuum check in the maintenance task loop

## Tests

`cargo test -p atuin-common --features sqlite` — 343 passed. New tests build a genuine ~2 MiB freelist and assert it drains to 0, and assert a legacy DB actually flips `auto_vacuum` 0→2 (idempotent). `cargo clippy --all-targets -D warnings` clean; `cargo build --workspace` green.

## Notes for reviewers

- ⚠️ **One-time conversion cost on large legacy DBs:** the first successful conversion runs a full `VACUUM` on every existing default-WAL database (client `history.db`, kv, scripts, ai). On a multi-GB `history.db` this transiently ~doubles disk use and grows the WAL until the trailing `wal_checkpoint(TRUNCATE)` runs (`journal_size_limit` does not bound the WAL mid-VACUUM). It happens once per DB, during an idle window, and fails-safe under the 500 ms busy timeout (retrying in 6 h). Worth a release note / telemetry watch.
- A few module doc comments still say "Compactor"/"compacts" and now under-describe the task — left untouched deliberately (I don't edit comment/doc prose without a separate pass); happy to update them if you'd prefer.
- Non-goals (intentional): no full-`VACUUM` "big red button" for defragmentation; incremental only.

## Checks

### DIFF
--- a/crates/atuin-common/src/sqlite/builder.rs
+++ b/crates/atuin-common/src/sqlite/builder.rs
@@ -4,7 +4,7 @@ use std::time::Duration;
 
 use sqlx::sqlite::{SqliteConnectOptions, SqliteJournalMode, SqliteSynchronous};
 
-use super::compactor::Compactor;
+use super::maintenance_task::MaintenanceTask;
 use super::{Sqlite, SqliteOpenOrCreateError};
 use crate::path::PathExt;
 
@@ -17,7 +17,7 @@ pub enum Journaling {
         /// The maximum size of the journal before sqlite is configured to automatically sweep it.
         ///
         /// Do note that this is a suggestion for Sqlite and under heavy concurrent reads will not
-        /// be respected. See [`Compactor`] for a strict maximum size.
+        /// be respected. See [`MaintenanceTask`] for a strict maximum size.
         #[allow(rustdoc::private_intra_doc_links)]
         max_size_hint: u64,
     },
@@ -136,7 +136,8 @@ impl<P: AsRef<Path>> SqliteBuilder<P> {
         let mut sqlite = Sqlite::connect(opts.clone(), self.timeout).await?;
 
         if matches!(self.journal, Some(Journaling::Wal { .. })) {
-            sqlite.compactor = Compactor::spawn_active(opts, sqlite.info.clone()).await;
+            sqlite.maintenance_task =
+                MaintenanceTask::spawn_active(opts, sqlite.info.clone()).await;
         }
 
         #[cfg(unix)]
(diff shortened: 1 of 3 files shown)


markovejnovic:
This is way too large and risky to merge. Would need to revisit later.

PR #4003, opened 2026-08-27
feat: pty-proxy formalized ipc

markovejnovic:
This PR adds an IPC mechanism to `atuin-pty-proxy`.

## Checks


crates/atuin-pty-proxy/src/client/ipc.rs
```diff
@@ -0,0 +1,170 @@
...
+            stream,
+            request_timeout: self.request_timeout,
+            scratch: Vec::new(),
+        };
+
+        let rep = conn
+            .call(HelloReq {
+                version: PROTOCOL_VERSION,
+            })
+            .await?;
```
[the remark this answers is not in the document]
markovejnovic: fixed with an environ variable

taylordotfish (commented):
can IpcServer and IpcServerWorker be merged into one type? they basically seem like the same thing.

crates/atuin-pty-proxy/src/ipc/controller.rs
```diff
@@ -0,0 +1,46 @@
...
+
+    pub fn dump_screen(&self, _req: DumpScreenReq) -> DumpScreenRep {
+        DumpScreenRep {
+            screen: self.get_screen(),
+        }
+    }
+
+    pub fn goodbye(_req: GoodbyeReq) -> GoodbyeRep {
+        GoodbyeRep {}
+    }
```
taylordotfish: I would make `hello` and `goodbye` methods for consistency, even though they don't need `self`.

crates/atuin-pty-proxy/src/ipc/controller.rs
```diff
@@ -0,0 +1,46 @@
...
+//! This controller is injected into [`crate::ipc::IpcServer::spawn`] and is responsible for
+//! servicing client requests.
+use std::sync::mpsc::{self, SyncSender};
+
+use crate::ipc::domain::*;
+use crate::screen::{Msg, ScreenSnapshot};
+
+#[derive(Debug, Clone)]
+pub struct IpcController {
+    /// TODO(markovejnovic): World's biggest debt.
```
taylordotfish: please explain

crates/atuin-pty-proxy/src/server/unix/ipc/controller.rs
```diff
@@ -0,0 +1,46 @@
+//! The main controller responsible for servicing the [`crate::ipc::IpcServer`].
+//!
+//! This controller is injected into [`crate::ipc::IpcServer::spawn`] and is responsible for
+//! servicing client requests.
```
taylordotfish: this could use a clearer doc comment. I couldn't figure out whether the controller ran on the client or server side until I read the code. also this should probably be a comment on the struct itself show it shows up in lsp etc

crates/atuin-pty-proxy/src/ipc/domain.rs
```diff
@@ -0,0 +1,62 @@
+use derive_more::{From, TryInto};
+use serde::{Deserialize, Serialize};
+
+use crate::screen::ScreenSnapshot;
+
+pub const PROTOCOL_VERSION: u32 = 1;
+
+/// Associates a request type with the reply type the server sends back.
+#[cfg(feature = "client")]
+pub trait IsRequest: Into<Req> {
```
taylordotfish: I think this should be called `Request`; traits do not usually begin with `Is`. When I saw this at first I thought it had something to do with downcasting.

taylordotfish (commented):
I'm finding `Req` and `Rep` a bit hard to read. I keep parsing `Req` in particular as `Requirement`. I would prefer the full names, at least in some of the cases, like the associated types

crates/atuin-pty-proxy/src/server/unix/ipc/controller.rs
markovejnovic: This comes from backend land -- controllers usually refer to modules which service http requests (so server). https://education.launchcode.org/java-web-development/chapters/spring-controllers-and-routes/controllers-simple.html

markovejnovic:
> can IpcServer and IpcServerWorker be merged into one type? they basically seem like the same thing.

Not really, `IpcServer` is a facade over a thread in `IpcServerWorker`. really it should be holding a join handle but I wanted to keep it as before and make that sweeping change later

crates/atuin-pty-proxy/src/server/screen.rs
```diff
@@ -98,103 +96,37 @@ pub fn spawn_parser_thread(
...
-            if screen_tx.send(Msg::ScreenRequest(reply_tx)).is_err() {
-                break;
-            }
-            if let Ok(data) = reply_rx.recv() {
-                let _ = stream.write_all(&data);
-                let _ = stream.flush();
-            }
-        }
-    });
+trait SnapshotExt {
```
taylordotfish: Usually `SnapshotExt` would be extension methods on a type called `Snapshot`

crates/atuin-pty-proxy/src/server/screen.rs
```diff
@@ -98,103 +96,37 @@ pub fn spawn_parser_thread(
...
-    buf.extend_from_slice(&cols.to_be_bytes());
-    buf.extend_from_slice(&cursor_row.to_be_bytes());
-    buf.extend_from_slice(&cursor_col.to_be_bytes());
-
-    for row_data in screen.rows_formatted(0, cols) {
-        let len = row_data.len() as u32;
-        buf.extend_from_slice(&len.to_be_bytes());
-        buf.extend_from_slice(row_data.as_bytes());
+impl SnapshotExt for vt100::Parser {
+    fn snapshot(&self) -> ScreenSnapshot {
```
taylordotfish: I think this should just be an associated funciton on `ScreenSnapshot`. I don't see the value in an extension trait.

crates/atuin-pty-proxy/src/domain/ipc/wire.rs
```diff
@@ -0,0 +1,159 @@
...
+    };
+    header.to_slice(&mut buf[..Header::SERIALIZED_LEN]);
+
+    Ok(buf)
+}
+
+/// Decode a postcard-encoded body into a `T`.
+pub fn decode_body<'a, T: Deserialize<'a>>(buf: &'a [u8]) -> Result<T, postcard::Error> {
+    postcard::from_bytes(buf)
+}
```
taylordotfish, line 208: It feels weird that the whole encoding logic lives in `wire.rs` but we inline most of the decoding logic at the call site. Can we standardize it?

crates/atuin-pty-proxy/src/domain/ipc/mod.rs
```diff
@@ -0,0 +1,45 @@
...
+    DumpScreen(DumpScreenReq),
+    Goodbye(GoodbyeReq),
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, From, TryInto)]
+pub enum AnyResponse {
+    Hello(HelloRep),
+    DumpScreenRep(DumpScreenRep),
+    Goodbye(GoodbyeRep),
+}
```
taylordotfish, line 45: You liked my comment about `Req` and `Rep` from before, do you feel differently now?

markovejnovic: oof sorry, -- i renamed the trait to be Request but missed these

crates/atuin-pty-proxy/src/client/ipc.rs
```diff
@@ -0,0 +1,283 @@
...
+
+        let mut header_bytes = [0u8; Header::SERIALIZED_LEN];
+        self.stream.read_exact(&mut header_bytes).await.map_err(IpcError::Io)?;
+
+        let header = Header::parse(header_bytes).map_err(IpcError::Header)?;
+        let body_len = (header.message_width as usize).saturating_sub(Header::SERIALIZED_LEN);
+        let mut buf = vec![0u8; body_len];
+        self.stream.read_exact(&mut buf).await.map_err(IpcError::Io)?;
+
+        wire::decode_body::<AnyResponse>(&buf).map_err(IpcError::Decode)
```
taylordotfish: We should stream deserialize

crates/atuin-pty-proxy/src/server/unix/ipc/server.rs
```diff
@@ -0,0 +1,197 @@
...
+
+            if let Err(err) = self.service_stream(&mut stream) {
+                warn!(%err, "dropping ipc client connection");
+            }
+        }
+
+        Ok(())
+    }
+
+    fn service_stream(&self, stream: &mut UnixStream) -> Result<(), StreamServiceError> {
```
taylordotfish: Docs/clearer name please. Is “service” a verb here?

crates/atuin-pty-proxy/src/server/unix/ipc/server.rs
```diff
@@ -0,0 +1,197 @@
...
+    Decode(postcard::Error),
+
+    #[error("failed to encode reply: {0}")]
+    Encode(postcard::Error),
+
+    #[error("unexpected io error: {0}")]
+    Io(std::io::Error),
+}
+
+pub struct IpcServer;
```
taylordotfish, line 52: I'm not sure I see the utility of this type. If I call `spawn` only for it to return a ZST with no methods, is this not just a free function with extra steps?

crates/atuin-pty-proxy/src/server/unix/ipc/server.rs
```diff
@@ -0,0 +1,197 @@
...
+    fn from(err: std::io::Error) -> Self {
+        match err.kind() {
+            ErrorKind::ConnectionReset => Self::ConnectionReset(err),
+            ErrorKind::WouldBlock | ErrorKind::TimedOut => Self::Timeout(err),
+            _ => Self::Io(err),
+        }
+    }
+}
+
+impl From<wire::DecodeError> for StreamServiceError {
```
taylordotfish: We `use wire::EncodeError` but not `DecodeError`, standardizing would be clearer. At first I assumed this meant `EncodeError` wasn't from `wire` but it is

crates/atuin-pty-proxy/src/domain/screen.rs
```diff
@@ -0,0 +1,45 @@
...
+
+    #[must_use]
+    pub fn cursor_col(&self) -> u16 {
+        self.cursor_pos.1
+    }
+
+    #[must_use]
+    pub fn formatted_rows(&self) -> &[String] {
+        &self.rows
+    }
```
taylordotfish: This doesn't seem to do anything since the field is `pub`. A bit confusing to have two different names for the same thing too.

crates/atuin-pty-proxy/src/server/unix/ipc/server.rs
markovejnovic: I mentioned in another one of my messages, but in the future when we clean up all the thread situations, I'd like this to hold a `JoinHandle`.

taylordotfish (commented):
Did you try building with all different combinations of the new `pty-proxy` features? Should we add this to CI?

PR #4011, opened 2026-08-28
chore: some more clippy

markovejnovic:


crates/atuin-client/src/settings.rs
```diff
@@ -2042,7 +2042,7 @@ mod tests {
         assert_eq!(config.emacs.len(), 1);
         match &config.emacs["ctrl-c"] {
             super::KeyBindingConfig::Simple(s) => assert_eq!(s, "exit"),
-            _ => panic!("expected Simple variant"),
+            super::KeyBindingConfig::Rules(_) => panic!("expected Simple variant"),
         }
```
taylordotfish, line 2045: imo this is strictly worse. what this code was communicating before was “must be the `Simple` variant”. as a reader, I don't care about how many other variants there are, I just know that it has to be `Simple`. explicitly naming the other variant is not only visual noise, but it will result in compile errors if another variant is added, even though the semantics of this code are independent of how many non-`Simple` variants there are

crates/atuin-domain/src/record/mod.rs
```diff
@@ -740,7 +740,7 @@ mod tests {
             id: RecordId(Uuid::from_bytes([
                 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16,
             ])),
-            idx: 12345678910111213141_u64,
+            idx: 12_345_678_910_111_213_141_u64,
```
taylordotfish: I also think this is strictly worse in this case; the test is testing this struct against its JSON equivalent, but the JSON will *not* add underscores, so now it's harder to compare

this lint may still be worth it though

taylordotfish (commented):
mostly positive but there are a couple of code readability regressions I would like to address

taylordotfish (approved):
L.G.T.M.

PR #4012, opened 2026-08-28
fix(sqlite): remove user records during account deletion

cuishuang:
## Why?

Unlike the PostgreSQL and MySQL backends, SQLite's `delete_user()` does not delete rows from the record store. Since `store.user_id` has no foreign key or cascade, deleting an account permanently leaves its encrypted records orphaned.

## Checks


cuishuang:
@ellie Would you mind taking a look at this SQLite account-deletion fix?

The SQLite delete_user() path removes the user but leaves their record-store rows behind because store.user_id has no foreign-key cascade. This brings SQLite in line with the other database backends and adds focused coverage.

Thanks!

markovejnovic:
Hi @cuishuang. Thanks for your PR! I'm sorry that I've caused so many conflicts in https://github.com/atuinsh/atuin/pull/4015. The issue you point out should be resolved. The current `delete_user` is:

```rust
    #[instrument(skip_all)]
    async fn delete_user(&self, u: &User) -> DbResult<()> {
        db::query(Self::Dialect::DELETE_SESSIONS_BY_USER).bind(u.id).execute(self.pool()).await?;
        db::query(Self::Dialect::DELETE_HISTORY_BY_USER).bind(u.id).execute(self.pool()).await?;
        db::query(Self::Dialect::DELETE_STORE_BY_USER).bind(u.id).execute(self.pool()).await?;
        db::query(Self::Dialect::DELETE_USER_BY_ID).bind(u.id).execute(self.pool()).await?;
        Ok(())
    }
```

where

```rust
    const DELETE_STORE_BY_USER: &'static str = "DELETE FROM store WHERE user_id = $1";
```

Your PR _does_ add a transaction on top of this, which I suspect is good to add, so I'll do that in a separate transaction! The issue you outline should be resolved regardless.

Thank you!

PR #4013, opened 2026-08-28
refactor(history): make HistoryId a Copy Uuid

markovejnovic:
`HistoryId` wrapped a `String`, even though every id is already a `uuid_v7` minted in the simple (hyphen-less) 32-char form. This wraps `uuid::Uuid` instead so the type is **`Copy`**.

Unfortunately, couldn't use the existing `new_uuid!` macro because the wire representation is slightly different between the two


taylordotfish (commented):
it seems error-prone that we have to call `.0.as_simple().to_string()` everywhere; what if we mess up and call `.as_hyphenated()` somewhere? can we have a single method on `HistoryId` that turns it into the string format we've been using this whole time?

crates/atuin-client/migrations/20260828000000_drop_non_uuid_history_ids.sql
```diff
@@ -0,0 +1,9 @@
+-- `HistoryId` is now a UUID: rows are read via `try_get::<String>().parse()`, so
+-- an id that is not a valid UUID fails to decode and breaks every history query,
+-- not just its own row. Atuin has only ever minted ids as the 32-char simple
+-- (hyphen-less) UUID form, so drop anything that isn't exactly that. On a healthy
+-- database this matches every row and deletes nothing.
+--
+-- The character class allows upper-case hex too: `Uuid::parse_str` accepts it and
+-- the next write normalises it back to lower-case, so such rows still load fine.
+delete from history where length(id) <> 32 or id glob '*[^0-9a-fA-F]*';
```
taylordotfish: "this actually solves nothing"

crates/atuin-daemon/src/components/sync.rs
```diff
@@ -279,8 +279,7 @@ async fn do_sync_tick(
                 match batch {
                     Ok(histories) if !histories.is_empty() => {
                         // Only the IDs go on the bus; the rows themselves are already in sqlite.
-                        let ids: Arc<[HistoryId]> =
-                            histories.iter().map(|h| h.id.clone()).collect();
+                        let ids: Arc<[HistoryId]> = histories.iter().map(|h| h.id).collect();
```
markovejnovic, line 282: This change is quite good. Let's assume someone has 200K entries.

- Users will likely see a decrease in RSS usage on the daemon -- String is 24b overhead and then uuidv7 is 16 bytes. Used to be a ~12MB (8MB string value and the rest is glibc allocator header) spike due to the broadcast. Now it's 3.2MB.
- The other great thing is this optimization allows us to avoid 200K allocations and replace them with one allocation -- god knows what the overhead of that was.

200K allocations is actually an insane amount -- even with how optimized allocators are.

taylordotfish (commented):
I think most instances of `&HistoryId` should be `HistoryId`. What do you think?

crates/atuin-client/src/history.rs
```diff
@@ -189,9 +191,94 @@ const LATEST_SERIALIZED_FIELDS: u32 = 13;
...
+        serializer.collect_str(self)
+    }
+}
+
+impl<'de> Deserialize<'de> for HistoryId {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let s = <std::borrow::Cow<'de, str>>::deserialize(deserializer)?;
```
taylordotfish, line 225:
```suggestion
        let s = std::borrow::Cow::<'de, str>::deserialize(deserializer)?;
```

taylordotfish (approved):

PR #4016, opened 2026-08-29
refactor: Clean up the history service and fix minor bugs

markovejnovic:
This PR:

  - Removes the `HistoryComponent` we historically had, in favor of a split:
    - `HistoryJournal` is responsible for the logic of tracking which commands are in-flight. It is a simple, async component.
    - `grpc::history::Service` is a shim for the GRPC service which wraps `HistoryJournal` (and also the `DaemonHandle`).
  - Makes the protobuf slightly more strictly typed:
    - `common.Uuid` is a new UUID type which is sent over the wire.
    - `EndHistoryRequest.duration` is now a nullable `google.protobuf.Duration`. We used to use a `u64` in-place of `Option<u64>` where `0` meant `None`. We no longer do.
    - A new `common.RecordId` defines a record ID. It is semantically equivalent to the `RecordId` rust type.
    - A new `history.Id` defines a `HistoryId`. It is semantically equivalent to the `HistoryId` rust type.
    - `history.TailHistoryReply` has been improved through the use of a discriminated union, rather than a variant key and flat payload.
    - `DaemonEvent::(HistoryStarted|HistoryEnded)` have been removed.
    - The `atuin history tail` command no longer crashes on backpressure, rather prints that there was lag via `eprintln!`.
    - Fixed a bug where negative timestamps in the shell commands would result in ~1-2ms durations. Now, negative durations are clamped to `0ms`. I think negative durations can occur when a command which runs runs extremely briefly.

The motivations were:

  - Decouple the gRPC layer from the packing layer.
  - Slowly inch away towards moving away from the event bus architecture. If we ended up needing signals/slots, I'd rather we do the [Observer pattern](https://refactoring.guru/design-patterns/observer) as it trades boilerplate for looser coupling (which is a favorable tradeoff in my opinion).
  - Slightly generalize the History service to make room for implementing command storage.

## Followups

- Migrate more of the services into `grpc/`. I'm going to do this as a follow-up since I want to unblock the command capture effort (which would benefit from this PR), and the migration can be mostly done by claude.
- Add more richer types into the grpc layer.

## Checks


markovejnovic:
- [ ] What happens if an old gRPC packet in flight receives an updated client?

crates/atuin-daemon/proto/history.proto
```diff
@@ -25,25 +33,27 @@ message StartHistoryRequest {
 }
 
 message EndHistoryRequest {
-  string id = 1;
+  HistoryId id = 1;
```
ellie, line 36: are we sure that changing the type here is safe?

crates/atuin-daemon/proto/history.proto
```diff
@@ -25,25 +33,27 @@ message StartHistoryRequest {
...
   int64 exit = 2;
-  uint64 duration = 3;
+  // Measured command runtime. Absent means the client didn't measure it; the daemon then derives
+  // it from the command's start timestamp.
+  google.protobuf.Duration duration = 3;
```
ellie, line 40: verified that this is nanosecond resolution, in case anyone else is meaning to check

ellie (approved):

crates/atuin-daemon/src/history/mod.rs
```diff
@@ -2,7 +2,11 @@
 //!
 //! This module contains the proto-generated types for the history gRPC service.
 
-// Include the generated proto code
+pub mod common {
+    #![allow(clippy::must_use_candidate, reason = "prost-generated code")]
+
+    tonic::include_proto!("common");
+}
```
taylordotfish: This should probably be in its own module instead of in `history`?

crates/atuin-daemon/src/components/search.rs
```diff
@@ -219,16 +225,6 @@ impl Component for SearchComponent {
...
-            DaemonEvent::HistoryStarted(history) => {
-                debug!(id = %history.id, command = %history.command, "History started (no index action)");
-            }
-            DaemonEvent::HistoryEnded(history) => {
-                span!(Level::TRACE, "inject_history_ended")
-                    .in_scope(async || {
-                        self.index.read().await.add_history(history);
-                    })
-                    .await;
-            }
```
taylordotfish, line 231: We lost this logging. Should we add it back? Just a question; fine if the answer is no.

crates/atuin-daemon/src/grpc/history/model.rs
```diff
@@ -0,0 +1,443 @@
...
+    #[error("invalid id field: {0}")]
+    InvalidId(#[from] IdParseError),
+    #[error("invalid duration: {0}")]
+    InvalidDuration(#[from] prost_types::DurationError),
+}
+
+grpc_invalid_argument!(EndHistoryRequestParseError);
+
+/// Errors thrown parsing the [`EndHistoryRequest`].
+impl TryFrom<EndHistoryRequest> for (HistoryId, i64, Option<Duration>) {
```
taylordotfish: Could we make this a struct with named fields please?

crates/atuin-daemon/src/grpc/history/pb.rs
taylordotfish, line 1: This entire file appears to be implementations of `From` and `TryFrom` that convert between:

* Protobuf types from `crate::history`, and
* Our domain model types from `atuin_client::history`

But it's confusing to me that it lives in *neither* of those places.

I think it should live in the same place where the protobuf types are defined, i.e. in `src/history/mod.rs`. It confused me that we have a module called `model` that is neither a parent nor child of the module where any of these types are defined, that we never actually need to import because it only contains trait impls.

I'm open to other suggestions, but this was genuinely confusing for me so I would ideally like to see some kind of change or clarification.

**Edit:** As I realized in my comment below, this also makes it hard to find this code even if you're using an LSP. To find conversion functions I often find myself doing go-to-definition on the source or destination types because otherwise I tend to end up at the literal generic definition of `Into::into` in the standard library. But this relies on the assumption that the conversion functions will live next to one of those two types. I think that's generally a good assumption, so I think we should try to make our code abide by it.

crates/atuin-daemon/src/grpc/history/pb.rs
```diff
@@ -0,0 +1,443 @@
...
+            exit: history.exit,
+            duration: history.duration,
+            shell: history.shell.unwrap_or_default(),
+            author_kind: AuthorKind::from(history.author_kind) as i32,
+        }
+    }
+}
+
+/// Map a single journal event to its tail-stream reply.
+impl From<Result<CmdEvent, BroadcastStreamRecvError>> for TailHistoryReply {
```
taylordotfish: Some of these conversions would be a lot clearer as inherent methods.

When our critical application logic lives nowhere except a `From` impl, it's really hard to find where we're actually calling it. I think even LSP might have a hard time (I don't think my LSP would get here if I did go-to-definition on the `.into()`—if I came across this code I'd probably try go-to-definition on the source or result types and be confused that neither of them led me to the conversion code, as I mentioned in my earlier comment), and in any case, I think our code should be clear without LSP, if only to make reviewing PRs on GitHub easier.

`From<Result<_, _>>` is also *really* uncommon and unintuitive in Rust. In fact, there are **no** implementations like this in the Rust standard library. I would strongly prefer this be written in a more idiomatic and explicit/searchable way.

I recommend reading what [the standard library has to say about `From`](https://doc.rust-lang.org/stable/std/convert/trait.From.html#when-to-implement-from). In particular, `From` is meant to be used for conversions that are **infallible, lossless, value-preserving, and obvious**. In this case, the conversion is infallible, but it is **not** lossless (we discard the payload of `Cancelled`), value-preserving (somewhat arguable, but I think a `None` event means something similar to but distinct from a cancelled history), nor obvious (because of the non-idiomatic nature of `From<Result<_, _>>`, and the fact that `Err` doesn't map to `None` as I might expect at first).

taylordotfish (changes requested):
For the sake of clarity and maintainability, I would like to see changes in:

* Organization: in particular, I don't think `model.rs` is in the right location. I also had some confusion over the placement of `common`.
* Naming/explicitness: I think the anonymous 3-tuple in this PR should have named fields, and too much critical application logic is living in non-idiomatic `From` impls like `From<Result<_, _>>`, which is hard to search for and surprising to me as a reader.

From the correctness side, there are some outstanding questions about gRPC compatibility. We had some in-person discussions about this but I don't believe we reached a conclusion—are the backward-incompatible protobuf changes ok? Are we sure a daemon–client version mismatch will never cause a problem? I would like there to be a concrete answer to this so we're sure we're not introducing a regression.

taylordotfish:
> * [ ]  What happens if an old gRPC packet in flight receives an updated client?

I think we also need to answer this question before merging. We are changing our protobuf messages in a **backward-incompatible** way. I want to be **really, really** sure that this will not cause issues.

Ellie had a related question about a type change that I would also ideally like to have an answer to.

crates/atuin-daemon/src/grpc/history/model.rs
```diff
@@ -0,0 +1,443 @@
...
+                value: value.0.into_bytes().to_vec(),
+            }),
+        }
+    }
+}
+
+impl From<History> for HistoryEntry {
+    fn from(history: History) -> Self {
+        Self {
+            timestamp: history.timestamp.unix_timestamp_nanos() as u64,
```
taylordotfish: I think these will be rejected by the new Clippy lints that have been added—could you merge main?

crates/atuin-daemon/proto/history.proto
markovejnovic: Yeah `protobuf.Duration` encodes `{ secs: i64, nsecs: i64 }`, so it can represent a negative duration. `std::time::Duration` can't represent a negative duration, so we do some casting shenanigans.

crates/atuin-daemon/src/grpc/history/model.rs
markovejnovic: I fixed it in a commit I forgot to push. Should be resolved

crates/atuin-daemon/src/history/mod.rs
markovejnovic: Yeah, indeed, but at this stage nothing else is using it so I decided to keep it here. As I start cleaning up more and more of the protobuf, this will move into `grpc/common/proto.rs`, I'm guessing.

Would rather do this in a follow-up.

crates/atuin-daemon/src/components/search.rs
markovejnovic: - The rwlock `write/read` calls themselves probably don't need to be instrumented, I'm guessing -- the `write` path is, effectively, an `ArcSwap`. So the `read` is almost always instant.
- Indeed, we lost some logging on the `add_history` method, but it seems to be trivial:

```
    pub fn add_history(&self, history: &History) {
        if history.is_agent() {
            return;
        }
        if !self.shells.contains(history.shell.as_deref().unwrap_or_default()) {
            return;
        }

        let command = history.command.as_str();

        // DashMap with Arc<str> keys can be looked up with &str via Borrow trait
        if let Some(mut entry) = self.commands.get_mut(command) {
            // Existing command - just update invocations
            entry.add_invocation(history, &self.interner);
        } else {
            let mut haystack = self.haystack.write();
            match self.commands.entry(Arc::from(command)) {
                dashmap::Entry::Occupied(mut entry) => {
                    entry.get_mut().add_invocation(history, &self.interner);
                }
                dashmap::Entry::Vacant(vacant) => {
                    let Some(data) = CommandData::new(history, haystack.len(), &self.interner)
                    else {
                        return; // Skip invalid commands
                    };
                    haystack.push(HaystackEntry::new(vacant.insert(data).key().clone()));
                }
            }
        }
        // Note: frecency_map is rebuilt by background task, not invalidated here
    }
```

So the value of the span was marginal anyways

- However, what I do think is valuable is to create spans for the history element lifecycle -- when it gets started, we can open a span, when it gets finished/cancelled, we can close a span. I'll do that

crates/atuin-daemon/src/grpc/history/pb.rs
markovejnovic: Hmm, okay. I don't really agree with the standard on this one:

```cpp
struct MyThing {
  operator AnotherThing() const;
}
```

is very common and I kind of like that, but two things:

- Consistency > Style
- Even google's C++ style guide does not like implicit conversion operators.

Perhaps I am indeed trying to fit a square peg in a round hole, so I'll do as you say.

crates/atuin-daemon/src/history/mod.rs
markovejnovic: Talked in person, decided to do it now. It's done!

crates/atuin-daemon/src/grpc/history/model.rs
markovejnovic: Should be done

crates/atuin-daemon/src/grpc/history/pb.rs
markovejnovic: To confirm, are we ok with this new `pb.rs` structure?

crates/atuin-daemon/src/grpc/history/pb.rs
markovejnovic: I think this all should be cleaned up. Marking conversation as resolved, feel free to re-open if i missed something

crates/atuin-daemon/proto/history.proto
markovejnovic: Yeah, it should be. I'm writing analysis+docs in `crates/atuin-daemon/src/grpc/history/mod.rs`.

taylordotfish (approved):

PR #4017, opened 2026-08-29
feat: refactor/improve output capturing mechanism

taylordotfish:
This PR updates Atuin's command output capturing code. Much of this is to lay the groundwork for future changes, but there are also some immediate benefits:

* OSC 133 parsing has been improved. The parser is now **fully allocation-free**, and bugs relating to escape sequences that spanned chunk boundaries have been fixed.
* Output captures **match reality much more closely**. Previously, we did not correctly handle output that included certain kinds of escape sequences, like those that move the cursor to an absolute position. Now we use a more robust capture mechanism.
* SGR escape sequences (responsible for colors and formatting) are now included in captured output. **This enables Atuin AI to answer questions like “what does the red text mean in the command I just ran?”.**

Atuin AI can now answer questions like this:

atuin-ai-colors

### DIFF
--- a/crates/atuin-common/src/string/buffer.rs
+++ b/crates/atuin-common/src/string/buffer.rs
@@ -90,6 +90,7 @@ impl fmt::Write for BoundedBuffer {
 mod tests {
     use std::fmt::Write as _;
 
+    use proptest::prelude::*;
     use rstest::{fixture, rstest};
 
     use super::BoundedBuffer;
@@ -207,5 +208,114 @@ mod tests {
     fn into_data_returns_what_was_kept(mut buffer: BoundedBuffer) {
         let _ = buffer.write_str("abcdefghij");
         assert_eq!(buffer.into_data(), "abcdefgh");
+    }
+
+    /// Feed `writes` into a fresh buffer of `limit` bytes, ignoring the write results.
+    fn filled(limit: usize, writes: &[String]) -> BoundedBuffer {
+        let mut buffer = BoundedBuffer::new(limit);
+        for write in writes {
+            let _ = buffer.write_str(write);
+        }
+        buffer
+    }
+
+    prop_compose! {
+        /// A limit small enough that the writes below regularly overflow it.
+        fn limit_and_writes()(
+            limit in 0usize..24,
+            writes in prop::collection::vec("(?s).{0,8}", 0..8),
+        ) -> (usize, Vec<String>) {
+            (limit, writes)
+        }
+    }
+
+    proptest! {
+        /// The whole point of the type: whatever it is fed, it never grows past its limit.
+        #[test]
+        fn never_exceeds_the_limit((limit, writes) in limit_and_writes()) {
+            let buffer = filled(limit, &writes);
+            prop_assert!(buffer.data().len() <= limit);
+            prop_assert_eq!(buffer.limit(), limit);
+        }
+
+        /// What is kept is exactly the longest prefix of everything written that fits, cut on a
+        /// character boundary. `data()` being a `&str` at all also proves the cut kept it UTF-8.
+        #[test]
+        fn keeps_the_longest_prefix_that_fits((limit, writes) in limit_and_writes()) {
+            let all = writes.concat();
+            let buffer = filled(limit, &writes);
+            prop_assert_eq!(buffer.data(), &all[..all.floor_char_boundary(limit)]);
+        }
+
+        /// How the data is split across writes cannot change the result.
+        #[test]
+        fn chunking_does_not_matter((limit, writes) in limit_and_writes()) {
+            let chunked = filled(limit, &writes);
+            let at_once = filled(limit, &[writes.concat()]);
+
+            prop_assert_eq!(chunked.data(), at_once.data());
+            prop_assert_eq!(chunked.is_truncated(), at_once.is_truncated());
+        }
+
+        /// Truncation is reported if and only if something was actually dropped.
+        #[test]
+        fn reports_truncation_exactly_when_data_was_dropped((limit, writes) in limit_and_writes()) {
+            let all = writes.concat();
+            let buffer = filled(limit, &writes);
+            prop_assert_eq!(buffer.is_truncated(), all.len() > limit);
+        }
+
+        /// A write fails only once the buffer has given up, and from then on every write fails.
+        #[test]
+        fn writes_fail_only_after_truncation((limit, writes) in limit_and_writes()) {
+            let mut buffer = BoundedBuffer::new(limit);
+            for write in &writes {
+                let was_truncated = buffer.is_truncated();
+                prop_assert_eq!(buffer.write_str(write).is_err(), was_truncated);
+            }
+        }
+
+        /// `take` hands the whole state over and leaves a buffer indistinguishable from a new one.
+        #[test]
+        fn take_moves_the_state_out((limit, writes) in limit_and_writes()) {
+            let mut buffer = filled(limit, &writes);
+            let before = buffer.data().to_string();
+            let was_truncated = buffer.is_truncated();
+
+            let taken = buffer.take();
+            prop_assert_eq!(taken.data(), before);
+            prop_assert_eq!(taken.is_truncated(), was_truncated);
+            prop_assert_eq!(taken.limit(), limit);
+
+            prop_assert_eq!(buffer.data(), "");
+            prop_assert!(!buffer.is_truncated());
+            prop_assert_eq!(buffer.limit(), limit);
+        }
+
+        /// So does `clear`, minus the handing over.
+        #[test]
+        fn clear_is_as_good_as_a_new_buffer((limit, writes) in limit_and_writes()) {
+            let mut buffer = filled(limit, &writes);
+            buffer.clear();
+
+            prop_assert_eq!(buffer.data(), "");
+            prop_assert!(!buffer.is_truncated());
+            prop_assert_eq!(buffer.limit(), limit);
+
+            // And it takes writes again just like a fresh one would.
+            for write in &writes {
+                let _ = buffer.write_str(write);
+            }
+            let fresh = filled(limit, &writes);
+            prop_assert_eq!(buffer.data(), fresh.data());
+        }
+
+        /// Consuming the buffer returns what inspecting it showed.
+        #[test]
+        fn into_data_matches_data((limit, writes) in limit_and_writes()) {
+            let buffer = filled(limit, &writes);
+            let seen = buffer.data().to_string();
+            prop_assert_eq!(buffer.into_data(), seen);
+        }
     }
 }
--- a/crates/atuin-common/src/string/buffer.rs
+++ b/crates/atuin-common/src/string/buffer.rs
@@ -1,0 +1,211 @@
+use std::fmt;
+
+/// A string buffer limited to a certain length.
+#[derive(Clone)]
+pub struct BoundedBuffer {
+    inner: String,
+    limit: usize,
+    truncated: bool,
+}
+
+impl BoundedBuffer {
+    /// Create a new bounded buffer.
+    ///
+    /// `limit` is the maximum length, in bytes, of the buffer.
+    #[must_use]
+    pub fn new(limit: usize) -> Self {
+        Self {
+            inner: String::new(),
+            limit,
+            truncated: false,
+        }
+    }
+
+    /// Get the buffer's maximum length.
+    #[must_use]
+    pub fn limit(&self) -> usize {
+        self.limit
+    }
+
+    /// Inspect the data in the buffer so far.
+    #[must_use]
+    pub fn data(&self) -> &str {
+        &self.inner
+    }
+
+    /// Consume the [`BoundedBuffer`] and return the underlying [`String`].
+    #[must_use]
+    pub fn into_data(self) -> String {
+        self.inner
+    }
+
+    /// Return whether the buffer has been truncated.
+    #[must_use]
+    pub fn is_truncated(&self) -> bool {
+        self.truncated
+    }
+
+    /// Replace this buffer with an empty one, and return the old full buffer.
+    ///
+    /// The buffer's limit is maintained.
+    #[must_use]
+    pub fn take(&mut self) -> Self {
+        Self {
+            inner: std::mem::take(&mut self.inner),
+            truncated: std::mem::take(&mut self.truncated),
+            limit: self.limit,
+        }
+    }
+
+    /// Clear the buffer.
+    pub fn clear(&mut self) {
+        self.inner.clear();
+        self.truncated = false;
+    }
+}
+
+/// Implementation of [`fmt::Write`] for [`BoundedBuffer`].
+///
+/// [`BoundedBuffer`] will return [`fmt::Error`] only when the buffer is full and cannot accept any
+/// more data.
+impl fmt::Write for BoundedBuffer {
+    fn write_str(&mut self, s: &str) -> fmt::Result {
+        if self.truncated {
+            return Err(fmt::Error);
+        }
+
+        let available = self.limit - self.inner.len();
+        if available >= s.len() {
+            self.inner.push_str(s);
+            return Ok(());
+        }
+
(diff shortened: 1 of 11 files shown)


crates/atuin-pty-proxy/src/runtime.rs
```diff
@@ -72,53 +70,44 @@ fn run(options: RuntimeOptions) -> eyre::Result<()> {
...
-                        (capture_tracker.as_mut(), options.command_capture_sink.as_ref())
-                    {
-                        tracker.push(&buf[..n], sink);
-                    }
-
-                    if let Some(highlighter) = highlighter.as_mut() {
-                        let rendered = highlighter.render(&buf[..n]);
-                        let _ = msg_tx.send(Msg::Data(rendered.clone()));
+                    let raw_data = &buf[..n];
+                    let _ = msg_tx.send(Msg::Data(raw_data.to_vec()));
```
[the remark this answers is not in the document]
taylordotfish: Yes, the capture and snapshot emulators run serially, which could under high load block the stdout thread, but at the same time, the status quo before this PR is that the capture emulator runs directly on the stdout thread itself, which means it *always* blocks the stdout thread, which is arguably worse. Under this PR, bursts can be absorbed without delaying the stdout thread at all; before this PR, *any* burst would unconditionally block stdout forwarding. Running the emulators on a separate thread is the cleaner design; if we find that running the two emulators serially is too slow and causes the stdout thread to block under reasonable workloads, we can explore increasing the MPSC queue buffer size and/or running the emulators on separate threads. We should not make those decisions preeemptively, as they will increase memory, and we need to be conscious of memory use in the PTY proxy as well.

crates/atuin-common/src/string/trim.rs
```diff
@@ -0,0 +1,172 @@
+mod sealed {
+    pub trait Sealed {}
+}
+
+/// A pattern that, unlike `std::str::pattern::Pattern`, does not consume the pattern when used.
```
markovejnovic:
```suggestion
/// A pattern that, unlike [`std::str::pattern::Pattern`], does not consume the pattern when used.
```

crates/atuin-common/src/string/buffer.rs
```diff
@@ -0,0 +1,211 @@
+use std::fmt;
+
+/// A string buffer limited to a certain length.
+#[derive(Clone)]
+pub struct BoundedBuffer {
```
markovejnovic, line 5: Seems difficult to generalize this. I tried, but ultimately failed. Not sure it's applicable here, but are you interested in making this accept `atuin_common::string::Measure`?

crates/atuin-common/src/string/buffer.rs
```diff
@@ -0,0 +1,211 @@
...
+        }
+
+        self.inner.push_str(&s[..s.floor_char_boundary(available)]);
+        self.truncated = true;
+        Ok(())
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
markovejnovic, line 90: Some vibe-coded proptests could be good here to catch edge cases

crates/atuin-common/src/string/trim.rs
```diff
@@ -0,0 +1,172 @@
...
+            }
+
+            fn trim_end_matches<'a>(&mut self, s: &'a str) -> &'a str {
+                s.trim_end_matches(($to_pattern)(self))
+            }
+        }
+    };
+}
+
+// Implement `PatternRef` for all of the types that implement `Pattern`.
```
markovejnovic, line 32:
```suggestion
```

I'm guessing blanket impls didn't work?

crates/atuin-daemon/proto/semantic.proto
```diff
@@ -7,8 +7,13 @@ service Semantic {
 }
 
 message CommandCapture {
+  // The shell prompt, as rendered by the terminal. Contains SGR escape
+  // sequences.
   string prompt = 1;
+  // The command typed by the user, as rendered by the terminal. Contains SGR
+  // escape sequences.
```
markovejnovic:
```suggestion
  // The command typed by the user, as rendered by the terminal. May contain SGR
  // escape sequences.
```

crates/atuin-daemon/proto/semantic.proto
```diff
@@ -7,8 +7,13 @@ service Semantic {
 }
 
 message CommandCapture {
+  // The shell prompt, as rendered by the terminal. Contains SGR escape
+  // sequences.
```
markovejnovic:
```suggestion
  // The shell prompt, as rendered by the terminal. May contain SGR escape
  // sequences.
```

crates/atuin-pty-proxy/src/screen.rs
```diff
@@ -22,23 +25,65 @@ pub fn socket_path() -> Result<PathBuf, SecureTempDirError> {
...
+pub fn spawn_parser_thread(rows: u16, cols: u16, screen_rx: Receiver<Msg>, options: ParserOptions) {
+    std::thread::spawn(move || {
+        let rows = NonZeroU16::new(rows).unwrap_or(NonZeroU16::MIN);
+        let cols = NonZeroU16::new(cols).unwrap_or(NonZeroU16::MIN);
+        let mut parser = Parser::new(rows, cols, options);
+        for msg in screen_rx {
+            parser.handle_msg(msg);
+        }
     });
 }
```
markovejnovic: I generally prefer renaming `Parser` to be `ParserWorker`, while making a public-facing `Parser`.

```rust
struct Parser {
  thread: JoinHandle<()>,
}

impl Parser {
  pub fn spawn((rows, cols), screen_rx, options) -> Self;
}
```

That way the thread isn't detached -- that is its own can of worms. If you want, you can do this in this PR, or I can do it in a follow-up when I sweep through all of the pty proxy.

crates/atuin-pty-proxy/src/screen.rs
```diff
@@ -22,23 +25,65 @@ pub fn socket_path() -> Result<PathBuf, SecureTempDirError> {
...
+pub struct ParserOptions {
+    pub sink: Option<CommandCaptureSink>,
+    pub debug_osc133: bool,
+}
 
-        loop {
-            let Ok(first) = msg_rx.recv() else {
-                break;
-            };
+struct Parser {
```
markovejnovic, line 34: This name is very generic. While you're at it, I would strongly suggest you rename this to something like `ScreenParser` (even that kind of sucks). Would suggest coming up with a good name.

crates/atuin-pty-proxy/src/screen.rs
```diff
@@ -22,23 +25,65 @@ pub fn socket_path() -> Result<PathBuf, SecureTempDirError> {
...
+    emulator: vt100::Parser,
+    tracker: Option<CommandCaptureTracker>,
+    highlighter: Option<Osc133DebugHighlighter>,
```
markovejnovic, line 37:
```suggestion
    osc133_highlighter: Option<Osc133DebugHighlighter>,
```

crates/atuin-pty-proxy/src/screen.rs
```diff
@@ -22,23 +25,65 @@ pub fn socket_path() -> Result<PathBuf, SecureTempDirError> {
...
+                    highlighted = highlighter.render(&raw_data);
+                    &highlighted
+                } else {
+                    &raw_data
+                };
+                self.emulator.process(data);
+            }
+            Msg::Resize { rows, cols } => {
+                let rows = NonZeroU16::new(rows).unwrap_or(NonZeroU16::MIN);
+                let cols = NonZeroU16::new(cols).unwrap_or(NonZeroU16::MIN);
```
markovejnovic, line 70: Hmmm. Is it possible to have a terminal with zero columns and zero rows? `crossterm::terminal::size()` returns `(u16, u16)`.

- If that is not possible, I'd recommend moving this logic into the message (so the message carries `NonZeroU16`. You can `debug_assert` that this invariant is held in `atuin-pty-proxy/src/runtime.rs:160`.
- If that is possible, then you can leave the things here as-is, though I would make a note mentioning _why_ the Vt100 parser cannot accept col=0, row=0.

crates/atuin-pty-proxy/src/runtime.rs
```diff
@@ -160,7 +149,6 @@ fn run(options: RuntimeOptions) -> eyre::Result<()> {
 fn spawn_resize_handler(
     master: Box<dyn portable_pty::MasterPty + Send>,
     resize_tx: mpsc::SyncSender<Msg>,
-    current_cols: Arc<AtomicU16>,
```
markovejnovic, line 163: 👍

crates/atuin-pty-proxy/src/runtime.rs
```diff
@@ -72,53 +70,44 @@ fn run(options: RuntimeOptions) -> eyre::Result<()> {
...
                     } else {
-                        let _ = msg_tx.send(Msg::Data(buf[..n].to_vec()));
+                        raw_data
+                    };
 
-                        if stdout.write_all(&buf[..n]).is_err() {
-                            break;
-                        }
+                    if stdout.write_all(data).is_err() {
+                        break;
```
markovejnovic, line 110: I know this was here before you, but this `break` terminates the thread. Does this not leave pty-proxy in an inconsistent state, wherein the debug highlighter just stops working?

Won't block your PR on this but interested in your thoughts.

crates/atuin-pty-proxy/src/debug.rs
```diff
@@ -15,28 +15,22 @@ impl Osc133DebugHighlighter {
...
-            rendered.extend_from_slice(event_label(located.event));
+        for chunk in chunks {
+            rendered.extend_from_slice(chunk.data);
+            rendered.extend_from_slice(event_label(chunk.event));
             rendered.extend_from_slice(RESET);
-            start = offset;
         }
 
-        rendered.extend_from_slice(&data[start..]);
+        rendered.extend_from_slice(chunk_iter.trailing_data());
```
markovejnovic, line 33: 👍

crates/atuin-pty-proxy/src/osc133.rs
```diff
@@ -168,212 +292,140 @@ impl Default for Parser {
 
 impl Parser {
```
markovejnovic, line 289: Have you considered using [`winnow`](https://docs.rs/winnow/latest/winnow/) for this? This code reads very much like a hand-rolled combinatorial parser and I can't help but wonder if using a combinatorial parser library would have been helpful here.

markovejnovic (approved):
Mostly nits. I haven't had a chance to review `osc133.rs` -- it's a parser and that's not easy to review.

Fortunately, I don't really care for it -- it's a debugging utility users don't even hit. Arguably, I think we should tear it out behind a `debug`-only build.

crates/atuin-pty-proxy/src/runtime.rs
markovejnovic: Also I feel the ordering may have been wrong here -- probably needed `acq/rel`, so I think this change is good.

crates/atuin-common/src/string/trim.rs
taylordotfish: I'll test if this works. I was unsure since it's an unstable item, but maybe you can link to it still.

crates/atuin-common/src/string/buffer.rs
taylordotfish: Yes, I think that could be a good idea. Probably will push that to a followup PR if that's ok.

crates/atuin-common/src/string/trim.rs
taylordotfish: Yep, works

crates/atuin-common/src/string/trim.rs
taylordotfish: Yes. We can't name `std::str::pattern::Pattern` in our code because it's unstable.

crates/atuin-pty-proxy/src/screen.rs
taylordotfish: I think it's a matter of style. The fact that `Parser` is in the `screen` module clarifies it enough for me (`screen::Parser`), but I don't feel that strongly. If you do feel strongly about it I'm fine making the change.

crates/atuin-pty-proxy/src/screen.rs
taylordotfish: I mildly disagree with this; I think the type makes it clear, and it was already called `highlighter` in our existing code in runtime.rs. But if you feel strongly about it I don't mind changing it.

crates/atuin-pty-proxy/src/screen.rs
taylordotfish: I think it is possible. I'm happy to add a comment; however `atuin_vt100::Parser::new` takes `NonZeroU16`s, so the fact that we can't pass 0 is already enforced by the type system. If we were using upstream `vt100` I'd agree with you that we need a comment to explain that we can't pass 0, but now the compiler enforces it.

That said, I don't have an objection to adding a comment explicitly stating this.

crates/atuin-pty-proxy/src/runtime.rs
taylordotfish: Yes, I believe there was a race condition. I'm not even sure `acq/rel` would have fixed it honestly.

crates/atuin-pty-proxy/src/screen.rs
taylordotfish: I'll make a similar change. I think `Parser` is a fine name because it *is* the actual parser. In my mind, the parser thread owns the parser. I also prefer having the parser thread be named something that communicates it's a thread.

I'll make a `ParserThread` type that owns the `JoinHandle`. I'll also have it be responsible for creating and holding the `SyncSender`; there's no reason why the caller should have to do that manually, and then `ParserThread` can have a simple method called `send` that sends a message.

If you're not satisfied with that design, we can talk about it and address in a followup PR like you said.

taylordotfish: Never mind, that won't work because we need to clone the `SyncSender` and use it in multiple places.

taylordotfish: I'm going to leave this change for later so we can discuss, but for now, I'll make `spawn_parser_thread` return a `JoinHandle`. I won't do anything with it in this PR but I'll bind it to an underscore-prefixed variable name in `runtime.rs`.

crates/atuin-pty-proxy/src/runtime.rs
taylordotfish: If we fail writing to the parent terminal, it's almost certainly because the FD got closed, which means *everything* is broken. Not just the debug highlighter—we can't forward data from the PTY either…

crates/atuin-pty-proxy/src/osc133.rs
taylordotfish: I haven't. I'm open to it in theory but have questions:
* How does the performance of `winnow` compare? Is it allocation-free?
* The grammar of our parser is technically `*`. Nothing is actually a parse error. Is `winnow` a good choice for this?
* How does `winnow` handle streaming? We push arbitrary-sized chunks to our parser, and OSC 133 escape sequences can cross chunk boundaries.
* Our parser is quite simple, really. OSC 133 sequences aren't hard to parse. Is the added complexity of `winnow` worth it?

taylordotfish:
> Mostly nits. I haven't had a chance to review `osc133.rs` -- it's a parser and that's not easy to review.
> 
> Fortunately, I don't really care for it -- it's a debugging utility users don't even hit. Arguably, I think we should tear it out behind a `debug`-only build.

@markovejnovic: To clarify, `osc133.rs` actually isn't debug-only. It's used by `capture.rs` to figure out where command output starts and ends.

crates/atuin-pty-proxy/src/screen.rs
taylordotfish: The more I think about it, I feel a little more strongly that we shouldn't make this change. It's a private type, and it's in `screen.rs` so I think it's clear it's a “screen parser”. I generally dislike redundancy in names, and it's a pattern that's discouraged in Rust.

PR #4018, opened 2026-08-29
fix(nix): Export OpenSSL lib for LD_LIBRARY_PATH in nix devshell

Adda0:
From `atuin.nix`:

```nix
  # native-tls pulls OpenSSL into the sqlx-macros proc-macro, which rustc dlopens at
  # build time. With OPENSSL_NO_VENDOR it links libssl.so.3 dynamically, so the loader
  # needs it on LD_LIBRARY_PATH while the proc-macro is loaded.
  preBuild = ''
    export LD_LIBRARY_PATH="${lib.makeLibraryPath [ openssl ]}''${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH}"
  '';
```

The same fix is needed for the Nix dev shell so that you can run, e.g., `cargo run`, yet it was not present there. This PR copies the fix into the nix dev shell shell hook.

## Checks

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -55,6 +55,11 @@
             RUST_SRC_PATH = "${pkgs.rustPlatform.rustLibSrc}";
 
             shellHook = ''
+              # See atuin.nix's preBuild for explanation.
+              export LD_LIBRARY_PATH="${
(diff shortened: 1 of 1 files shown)


ellie:
Thank you!

PR #4019, opened 2026-08-29
feat(nu): Support shell integration for Nushell Helix edit modes

Adda0:
Adds support for Helix edit modes of Nushell (new in [Nushell 0.115](https://www.nushell.sh/blog/2026-08-15-nushell_v0_115_0.html#added-helix-edit-mode)).

Supports `helix_normal` and `helix_insert`.
Does not register for `helix_select`.

Note that this is a new Nushell edit mode, and Nushell will complain if there are invalid modes in the keybindings' mode config. Thus, the version guard is necessary to not break Nushell integration for Nushell versions < 0.115.

## Checks

### DIFF
--- a/crates/atuin/src/command/client/init/nu.rs
+++ b/crates/atuin/src/command/client/init/nu.rs
@@ -7,7 +7,7 @@ const BIND_CTRL_R: &str = r"$env.config = (
             name: atuin
             modifier: control
             keycode: char_r
-            mode: [emacs, vi_normal, vi_insert]
+            mode: ([emacs, vi_normal, vi_insert] | append (_atuin_helix_edit_modes_if_supported [helix_normal, helix_insert, helix_select]))
             event: { send: executehostcommand cmd: (_atuin_search_cmd) }
         }
     )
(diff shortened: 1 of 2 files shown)


kronberger-droid:
@Adda0
Nice, you were faster than me!

One thing worth a thought is adding `helix_select` to the `Ctrl-R` list.
Unlike `vi_visual`, helix select does not just reuse the normal keybindings table, but has its own, which is similar to the normal one but not the same.
It is seeded from the normal table though, so it already carries Reedline's default `Ctrl-R` -> `SearchHistory`, and that built-in reverse search opens there instead of Atuin unless the binding names the mode explicitly.

The up binding arguably makes less sense in this mode, since Reedline deliberately keeps Up and Down there on selection-extending line motions rather than history.

`_atuin_helix_edit_modes_if_supported` already takes the list as a parameter, so only the `Ctrl-R` call site would change.

Soon `vi_visual` will be a thing too, since it will help us fix some bugs, and the same decision has to be made there when the time comes.

Adda0:
Hey. I was unsure whether to include `helix_select` for the `Ctrl+r` binding, but I missed that `Ctrl+r` already works in the `helix_select` mode anyway, only with the default history search. Thanks for pointing that out. I will add `helix_select` for `Ctrl+r`.

PR #4020, opened 2026-08-30
fix(search): apply CLI filters to interactive search

### LINKED ISSUE #3869: [Bug]: `search --author` (and `--shell`) is ignored in interactive (-i) mode

`atuin search --author '$all-agent' -i` (and any other `--author` value) should filter the **interactive** TUI to matching authors, exactly as the non-interactive `atuin search --author '$all-agent'` already does.

### What happened?

`--author` is silently ignored whenever `-i`/`--interactive` is passed. The interactive TUI shows **all** history regardless of the `--author` value, while the non-interactive search of the same query filters correctly.

#### Reproduction

```console
$ # Seed one user-authored and one agent-authored command:
$ ID=$(atuin history start --author user  -- "echo USER_ONLY_MARKER");  atuin history end --exit 0 "$ID"
$ ID=$(atuin history start --author codex -- "echo AGENT_ONLY_MARKER"); atuin history end --exit 0 "$ID"

$ # Non-interactive: filter works — only the agent command is returned:
$ atuin search --author '$all-agent' --cmd-only
echo AGENT_ONLY_MARKER

$ atuin search --author codex --cmd-only
echo AGENT_ONLY_MARKER

$ # Interactive: filter is ignored — USER_ONLY_MARKER is still listed,
$ # and the header shows the full, unfiltered history count.
$ atuin search --author '$all-agent' -i
# TUI lists echo USER_ONLY_MARKER as well; "history count" == total, not the agent subset
```

Concretely, `atuin search --author '$all-agent' --cmd-only | wc -l` returns the agent subset, but `atuin search --author '$all-agent' -i` shows the header `history count: <total>` and lists user-authored commands that the non-interactive filter correctly excludes.

#### Root cause

In `crates/atuin/src/command/client/search.rs`, the `--author` list is only consumed on the **non-interactive** path:

```rust
// interactive branch — self.author is never passed through:
if self.interactive {
    let item = interactive::history(&query, settings, db, &history_store, theme).await?;   // ~L247–248
    ...
} else {
    let authors = OrFilter::from_list(self.author).unwrap_or_default();   // ~L265
    let shells  = OrFilter::from_list(self.shell).unwrap_or_default();    // ~L266
    ...
    authors: authors.as_slice_filter(),                                   // ~L280
    let mut entries = run_non_interactive(settings, opt_filter, &query, &db).await?;  // ~L284
}
```

`OrFilter::from_list(self.author)` / `self.shell` are built only in the `else` branch. The interactive entrypoint's signature has no author/shell parameter, so the values cannot reach the TUI query:

```rust
// crates/atuin/src/command/client/search/interactive.rs  (~L1663)
pub async fn history(
    query: &[String],
    settings: &Settings,
    mut db: impl Database,
    history_store: &HistoryStore,
    theme: &Theme,
) -> Result<String> { ... }
```

So `--author` (and, by the same construction, `--shell`) has no effect in interactive mode.

#### Suggested fix

Thread the parsed `authors`/`shells` `OrFilter`s into `interactive::history(...)` and apply them to the interactive query the same way `run_non_interactive` does — or, minimally, document that `--author`/`--shell` are non-interactive-only and error/warn when combined with `-i`.

### Atuin doctor output

```json
{
  "atuin": {
    "version": "18.20.0-beta.3",
    "commit": "c2f4718807109aec2246df1694cc5e9b4cb567c1",
    "sync": null,
    "sqlite_version": "3.46.0",
    "daemon_enabled": false
  },
  "shell": {
    "name": "bash",
    "default": "bash",
    "plugins": ["atuin", "bash-preexec"],
    "preexec": null
  },
  "system": {
    "os": "Arch Linux",
    "arch": "x86_64",
    "version": "unknown"
  }
}
```

taljeon:
## Summary

- apply `--author` and `--shell` filters to interactive search
- preserve the existing user-only and configured-shell defaults when those flags are omitted
- fall back to the database when an author filter cannot be represented by the daemon index
- merge current `main@dc493140` without force, preserving its newer search/daemon work while keeping this PR's five-file filter diff

## Tests

- focused `interactive_search_applies_author_and_shell_filters` cases: 2 passed
- `cargo +nightly fmt --all -- --check`: passed
- `cargo clippy -p atuin --all-targets --all-features -- -D warnings`: passed; current `clippy.toml` emits its existing unreachable disallowed-method warning outside this PR diff
- `cargo test -p atuin --bin atuin`: 552 unit tests passed
- host-level package tests: 107 integration tests passed; two bash-preexec exact-output waits timed out after the other 83/85 PTY cases passed
- the same two targeted bash-preexec cases timed out on pristine current `main`, while the other 10 cases in that group passed or skipped identically, confirming a current-base E2E/environment baseline rather than a PR regression
- `git diff --check` and the five-file publication leak scan: passed; the repository scanner reports only pre-existing tracked fixtures/examples outside the diff

Fixes #3869

## AI assistance

OpenAI `gpt-5.6-sol` through Codex assisted with the current-base merge, overlap review, verification, and baseline comparison. I reviewed the final five-file diff and the test evidence above.

## Checks


taljeon:
Merged current `main` and resolved the single `interactive.rs` import conflict by retaining `AuthorPattern`, `History`, and `HistoryId` while dropping the now-removed `HistoryStats` import. The final diff remains the original five-file author/shell-filter change.

Focused filter tests pass 2/2, formatting and all-target/all-feature Clippy pass, and all 473 unit tests pass. The full package run hit two bash-preexec PTY exact-output timeouts; both reproduce identically on pristine current `main`, so they are documented as a current-base E2E/environment baseline. Diff and publication-leak checks are clean. A re-review would be appreciated when convenient.

taljeon:
Merged current `main@dc493140` without force. The newer base touched `search.rs` and `engines/daemon.rs`; the merge was clean, and the final comparison against current main remains the intended five-file author/shell-filter diff.

Post-merge verification:

- focused author/shell regression: 2/2 passed;
- nightly Rustfmt and all-target/all-feature Clippy: passed (with the existing `clippy.toml` unreachable-method warning);
- 552 unit tests passed;
- 107 integration tests passed; the remaining two bash-preexec exact-output cases timed out identically on pristine current main (the other 10 targeted cases passed there), confirming the existing host timing baseline;
- diff, noreply metadata, and publication leak gates passed.

The PR body is updated for this exact head. A re-review would be appreciated when convenient.

Issue #3868, opened 2026-08-07
smart_sort does not always sort chronologically


xidiot:
I've found out about `smart_sort` from https://github.com/atuinsh/atuin/pull/1885, which I thought was working just fine until now (sorting the results chronologically). As you can see in the below screenshot the 1st match is older then the 2nd match (actually it is older than the 1-9 results): 

<img width="1896" height="99" alt="Image" src="https://github.com/user-attachments/assets/04b9a39b-13cc-4920-a21c-1faadbcaafe2" />

I know that `smart_sort` is NOT supposed to only prefer more recent history, but in that case is there something I can use to show the results always chronologically sorted?

ps: I use `search_mode = "fulltext"`

PR #4021, opened 2026-08-30
fix(history): don't let smart_sort's prefix boost defeat fulltext mode

### LINKED ISSUE #3923: [Bug]: search_mode=fulltext but results are similar to search_mode=prefix

If `search_mode=fulltext` I expect to match the text/search string anywhere in the command. 

### What happened?

Latest command:

Image

After i search for `grep`:
Image

As seen the `ps axu|grep -i git` command is not even in the first twenty results. It's only shown after ALL commands which starts with `grep`, no matter the `search_mode` this is always the case. Even if `ps axu|grep -i git` is not the latest command the same happens.

As per the docs:
 >fulltext mode searches for "\*query\*".

Is this expected?

### Atuin doctor output

```yaml
{
  "atuin": {
    "version": "18.19.0",
    "commit": "fb33978d9439ba7e97123c1adc932c6a74e3c8ae",
    "sync": null,
    "sqlite_version": "3.46.0",
    "daemon_enabled": true
  },
  "shell": {
    "name": "bash",
    "default": "bash",
    "plugins": [
      "atuin",
      "bash-preexec"
    ],
    "preexec": "bash-preexec"
  },
  "system": {
    "os": "Darwin",
    "arch": "arm64",
    "version": "26.5.2",
    "disks": [
      {
        "name": "Macintosh HD",
        "filesystem": "apfs"
      },
      {
        "name": "Macintosh HD",
        "filesystem": "apfs"
      }
    ]
  }
}
```

### Code of Conduct

nityanand123gupta:
Fixes #3923.

I used a coding agent (Claude) to help investigate and implement this, under my direction and review — disclosing per CONTRIBUTING.md.

## What I found

The reporter's symptom (`search_mode = fulltext`, but a match like `ps axu|grep -i git` doesn't show up until after every command that literally starts with `grep`) isn't in the SQL query building in `atuin-client/src/database.rs` — that already builds a genuine `%query%` fulltext condition. The DB results are correct.

The re-ranking happens afterward, in `atuin-history/src/sort.rs`'s `sort()`, which runs whenever `smart_sort` is enabled (the default), regardless of search mode. It always scores a command that *starts with* the query at 2.0 and one that merely *contains* it at 1.75, then multiplies by a small recency factor. In fulltext mode this defeats the mode's own definition (docs: `fulltext mode searches for "*query*"` — matches anywhere) by keeping every prefix match ranked above every non-prefix match no matter how old, instead of letting recency decide as the mode's docs imply.

(I also looked at #3868, "smart_sort does not always sort chronologically" — that one is about `smart_sort`'s recency/prefix blend behaving as originally designed per #1885, not a bug, so I left it alone.)

## Fix

Threaded the active `SearchMode` into `sort()`. Only in `FullText` mode, a prefix match and any other substring match now score identically, so the existing recency multiplier is what decides ranking among them. `Prefix` and `Fuzzy` modes are untouched.

## Testing

- Added two unit tests in `sort.rs`: one reproduces the reported scenario (an old prefix match should no longer outrank a much more recent substring-only match under `FullText`), and one confirms `Prefix` mode's existing behavior is unchanged.
- Verified both by reverting only the scoring change and confirming the fulltext test fails without it, then restoring it and confirming both pass.
- `cargo test -p atuin-history` (40 tests) and `cargo test -p atuin --bin atuin search::` (336 tests) both pass.
- `cargo +nightly fmt --check` and `cargo clippy` are clean on the changed files (no new warnings).

## Checks

### DIFF
--- a/crates/atuin-history/benches/smart_sort.rs
+++ b/crates/atuin-history/benches/smart_sort.rs
@@ -1,4 +1,5 @@
 use atuin_client::history::History;
+use atuin_client::settings::SearchMode;
 use atuin_history::sort::sort;
 use rand::rngs::StdRng;
 use rand::{Rng, SeedableRng};
@@ -18,7 +19,9 @@ const SEED_NOW: OffsetDateTime = datetime!(2026-01-01 12:59:59 -5);
 #[divan::bench(args=[100, 200, 400, 800, 1600, 10000])]
 fn smart_sort(bencher: divan::Bencher, lines: usize) {
     // Generating the history is not part of what we measure, so it happens as an input.
-    bencher.with_inputs(|| history(lines)).bench_values(|commands| sort("curl", commands));
+    bencher
+        .with_inputs(|| history(lines))
+        .bench_values(|commands| sort("curl", SearchMode::Fuzzy, commands));
 }
 
 /// Generate a few different sizes of "history". This will use a whole bunch of memory, sorry.
--- a/crates/atuin-history/src/sort.rs
+++ b/crates/atuin-history/src/sort.rs
@@ -23,7 +23,16 @@ pub fn sort(query: &str, search_mode: SearchMode, input: Vec<History>) -> Vec<Hi
             // buried mid-command would always rank below every prefix match no matter how old,
             // instead of being ranked primarily by recency as the docs promise.
             let score = if search_mode == SearchMode::FullText {
-                if h.command.contains(query) {
+                // Mirror the smart-case behavior of the DB-side LIKE/GLOB query
+                // (QueryToken::has_uppercase in database.rs): an all-lowercase query
+                // matches case-insensitively, so a lowercase-only query must not
+                // demote differently-cased commands that the DB already matched.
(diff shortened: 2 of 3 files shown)


nityanand123gupta:
Good catch — fixed in 6c627dc. The DB-side fulltext condition actually applies smart-case itself: an all-lowercase query token uses SQL \`LIKE\` (case-insensitive), while a query containing uppercase uses \`GLOB\` (case-sensitive) — see \`QueryToken::has_uppercase\` in \`database.rs\`. My scoring branch was plain case-sensitive regardless, so a differently-cased command the DB legitimately matched could get demoted to a non-match score.

Mirrored the same smart-case rule in \`sort()\`: case-insensitive contains when the query has no uppercase, case-sensitive otherwise. Added two regression tests (\`fulltext_lowercase_query_matches_case_insensitively\`, \`fulltext_uppercase_query_matches_case_sensitively\`) and verified the lowercase one fails without the fix and passes with it.

PR #4031, opened 2026-08-31
fix(daemon): reject invalid pidfile pids before signalling

markovejnovic:
`force_cleanup` (in `crates/atuin/src/command/client/daemon.rs`) read the daemon pidfile and parsed it as a `u32`, then handed it to `atuin_common::os::process::force_terminate`, which turns it into a rustix `Pid` via `Pid::from_raw(pid.cast_signed())`.

A corrupt or hand-edited pidfile value in `[2^31, 2^32)` overflows to a **negative** `i32`.

### DIFF
--- a/crates/atuin/src/command/client/daemon.rs
+++ b/crates/atuin/src/command/client/daemon.rs
@@ -707,7 +707,8 @@ async fn force_cleanup(settings: &Settings) {
     if pidfile_path.exists() {
         if let Ok(contents) = fs::read_to_string(pidfile_path)
             && let Some(pid_str) = contents.lines().next()
-            && let Some(pid) = parse_pid(pid_str)
+            && let Some(pid) = pid_str.trim().parse::<i32>().ok()
+            && pid > 0
             && let Err(e) = atuin_common::os::process::force_terminate(
                 pid.unsigned_abs(),
                 Duration::from_secs(2),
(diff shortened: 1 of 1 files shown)


ellie (approved):

crates/atuin/src/command/client/daemon.rs
```diff
@@ -696,13 +696,23 @@ async fn run(
...
+    }
+
     // Read and kill the existing process if pidfile exists
     if pidfile_path.exists() {
         if let Ok(contents) = fs::read_to_string(pidfile_path)
             && let Some(pid_str) = contents.lines().next()
-            && let Ok(pid) = pid_str.parse::<u32>()
-            && let Err(e) =
-                atuin_common::os::process::force_terminate(pid, Duration::from_secs(2)).await
+            && let Some(pid) = parse_pid(pid_str)
```
taylordotfish:
```suggestion
            && let Some(pid) = pid_str.parse::<i32>().ok().filter(|p| *p > 0)
```

& get rid of the separate function

crates/atuin/src/command/client/daemon.rs
```diff
@@ -696,13 +696,24 @@ async fn run(
 async fn force_cleanup(settings: &Settings) {
     let pidfile_path = Path::new(&settings.daemon.pidfile_path);
 
+    fn parse_pid(s: &str) -> Option<i32> {
+        match s.parse::<i32>() {
+            Ok(pid) if pid > 0 => Some(pid),
+            _ => None,
+        }
+    }
+
```
taylordotfish:
```suggestion
```

taylordotfish (approved):

PR #4033, opened 2026-08-31
fix(server): purge orphaned store rows on Postgres

markovejnovic:
## Defect

The historical `delete_user` bug -- where a deleted user's `store` rows were left behind -- lived in the **shared** blanket-impl `delete_user` path (`crates/atuin-server/src/db/mod.rs`) that serves **both** SQLite and Postgres. When it was fixed, a cleanup migration was shipped only for SQLite (`20260828000000_purge-orphaned-store.sql`). Postgres never got an equivalent, so Postgres deployments that ran the old buggy code still carry **orphaned `store` rows** (`user_id not in users`) that are never cleaned up. atuin.sh's primary production backend is Postgres, so this is a real correctness gap. (MySQL is unreleased and out of scope.)

## The fix

Add the mirroring Postgres migration `20260829000000_purge-orphaned-store.sql` (the `20260828000000` slot is already taken by `drop-history-count-trigger`, so a new strictly-higher version is used):

```sql
delete from store where user_id not in (select id from users);
```

The `NOT IN` NULL caveat is safe here: `users.id` is `bigserial` (never NULL) and `store.user_id` is `NOT NULL`. The shared `delete_user` Rust code is already converged and is not touched.

## TDD

Added `crates/atuin-server/tests/pg_purge_orphaned_store.rs`. It brings a real Postgres instance up to head via the embedded Postgres migration chain, seeds two owned `store` rows plus two orphaned rows, then applies the purge migration from the chain and asserts the orphans are gone while owned rows remain.

- **RED** (before the migration existed): the purge migration is absent from the embedded chain, so no purge runs and the orphans survive -- `assertion left == right failed ... left: 2, right: 0`.
- **GREEN** (after adding the migration): `test result: ok. 1 passed`.

Verified with `postgres:16` in Docker; the full chain also applies cleanly via the existing `db_story` harness against Postgres.

### DIFF
--- a/crates/atuin-server/src/db/postgres/migrations/20260829000000_purge-orphaned-store.sql
+++ b/crates/atuin-server/src/db/postgres/migrations/20260829000000_purge-orphaned-store.sql
@@ -1,0 +1,3 @@
+-- Postgres deployments hit the same delete_user bug (shared code path) that left
+-- orphaned store rows behind. Clean them up. See the sqlite counterpart.
+delete from store where user_id not in (select id from users);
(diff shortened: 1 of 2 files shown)


markovejnovic:
hallucination and slop.

PR #4034, opened 2026-08-31
fix(daemon): compare process identity before signalling on force-cleanup

markovejnovic:
## The defect

`atuin daemon start --force` could send `SIGTERM` (and then `SIGKILL`) to an innocent, unrelated process.

Two gaps combined:

1. **Unconditional signal before any identity check.** In `crates/atuin-common/src/os/unix/process.rs`, `force_terminate` snapshotted `process_start_time(pid)` — the start_time of *whatever* occupies the pid *now* — and then sent `SIGTERM` unconditionally. The start_time comparison only gated the later `SIGKILL` escalation, never the initial signal.
2. **Identity was never persisted.** The daemon pidfile (`crates/atuin/src/command/client/daemon.rs`) stored only `pid` + `version`. The original daemon's start_time was never written, so there was nothing to compare the live process against.

Net effect: when a daemon exits and the OS later reuses its pid for an unrelated process, a subsequent `--force` cleanup reads the stale pid from the pidfile and signals that innocent process. Verified empirically.

## The fix (scoped to this defect only)

- **Persist identity.** The daemon's start_time is now written as an optional third line in the pidfile. Old-format pidfiles without it are still handled — cleanup falls back to current best-effort liveness, never a hard error.
- **Compare before signalling.** `force_terminate` now takes the expected start_time and refuses to send *any* signal when the live process's start_time differs from it (pid reuse → different process → leave it alone, return cleanly). Only a matching identity — or an unknown/old-format one — proceeds to `SIGTERM` → `SIGKILL`.

No pid-range validation is added here; that is a separate concern.

## Tests (strict TDD)

Added unit tests in `crates/atuin-common/src/os/unix/process.rs` that spawn a real victim process the test owns (a `sleep`, killed on drop):

- `wrong_identity_does_not_terminate_victim` — records the victim's actual start_time, calls the termination path with a deliberately-wrong expected start_time, and asserts the victim survives. This test fails against the pre-fix code (which SIGTERMs the innocent victim) and passes with the guard.
- `matching_identity_terminates_victim` — matching identity terminates the victim.
- `unknown_identity_terminates_victim` — `None` (old-format pidfile) falls back to best-effort liveness and terminates.

The pidfile-guard test was updated to tolerate and verify the new start_time line.

### DIFF
--- a/crates/atuin-common/src/os/process.rs
+++ b/crates/atuin-common/src/os/process.rs
@@ -18,19 +18,46 @@ pub(crate) const EXIT_BACKOFF: Backoff = Backoff::Exponential {
 };
 
 /// Gracefully (and then forcefully) terminate the process.
-pub async fn force_terminate(pid: u32, timeout: Duration) -> Result<(), std::io::Error> {
+///
+/// `expected_start_time` is the start_time the caller expects the process at `pid` to have (e.g. the
+/// value persisted alongside the pid). It guards against pid reuse: if the live process's identity
+/// does not match, it is left untouched. Pass `None` when the expected identity is unknown (for
+/// example an old-format pidfile), in which case liveness is best-effort.
+pub async fn force_terminate(
+    pid: u32,
+    expected_start_time: Option<u64>,
+    timeout: Duration,
+) -> Result<(), std::io::Error> {
     #[cfg(unix)]
     {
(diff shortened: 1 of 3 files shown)


markovejnovic:
the current solution is good enough.

PR #4039, opened 2026-08-31
feat(search): show delete shortcut in help

### LINKED ISSUE #3978: Can we add some sort of indication on how to delete an entry in the CTRL-R screen?
Every couple weeks, I need to delete an entry from Atuin. And every single time, I have to search on how to do it (ctrl+a and then d). It would be nice if one had some sort of indication on how to do it in the ctrl+r screen. Just a "Press CTRL+a and then d to delete" somewhere on the screen, say, top left corner.

Or maybe just a "Press ? for keymaps", and then show the keymaps. Something, because it is annoying to have to search for it every single time.

nfvelten:
Fixes #3978

The interactive search help row now shows the existing delete shortcut (`Ctrl+A`, then `d`) alongside the other search actions. The displayed prefix follows the configured `keys.prefix` value, so custom prefixes remain accurate.

Added a focused test covering the default and custom prefix cases.

Verification:
- `git diff --check` passed
- `cargo test` was attempted, but the local build could not complete because the machine hit its disk quota while compiling `zstd-sys`.

AI note: This PR was prepared with help from OpenAI Codex.

### DIFF
--- a/crates/atuin/src/command/client/search/interactive.rs
+++ b/crates/atuin/src/command/client/search/interactive.rs
@@ -878,6 +878,11 @@ impl State {
         );
 
         let show_help = settings.show_help && (compactness == Compactness::Full || area.height > 1);
+        let help_height = if show_help && self.tab_index == 0 {
+            2
+        } else {
+            1
+        };
         let warnings = self.build_warnings(settings, theme);
         let warning_height = u16::try_from(warnings.height()).unwrap_or(u16::MAX);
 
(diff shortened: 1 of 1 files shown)


nfvelten:
Addressed in `952e2a4`: the Search help now uses two rows and splits the shortcuts so the delete hint is visible at common terminal widths. The Greptile review was against the earlier `aa697194` commit; the formatter and completed build/test checks pass on the updated commit.

PR #4042, opened 2026-09-01
chore: lift dependencies into top-level Cargo.toml

markovejnovic:
This PR:

  - Moves `Cargo.toml` dependencies out of our subpackages into the top-level workspace.

Motivation:

  - Easier maintenance
  - Avoiding accidental duplicate dependencies (we had two `tracing-appender` versions).


taylordotfish (approved):
would be nice if we could sort the dependencies in cargo.toml but it's not blocking

PR #4048, opened 2026-09-02
feat: Implement command capture and remove semantic.rs

markovejnovic:


crates/atuin-daemon/proto/history.proto
```diff
@@ -141,4 +233,13 @@ service History {
...
   rpc Shutdown(ShutdownRequest) returns (ShutdownReply);
+  // Store a completed command's captured output. Errors if output already exists for this id.
+  rpc RegisterCommandOutput(RegisterCommandOutputRequest) returns (RegisterCommandOutputResponse);
+  // Retrieve a command's captured output, optionally sliced by line ranges.
+  //
+  // See the docs for `GetCommandOutputRequest` for the calling convention.
+  //
+  // Returns:
+  //  - `tonic::Status::not_found` if the given output is not found.
+  rpc GetCommandOutput(GetCommandOutputRequest) returns (GetCommandOutputReply);
```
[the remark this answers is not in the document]
markovejnovic: Again, we haven't released protocol version 2 yet so this doesn't matter.

crates/atuin-daemon/src/output_capture.rs
```diff
@@ -0,0 +1,182 @@
...
+//! Output is keyed by `history_id` (16 UUID bytes) in a fjall keyspace; the
+//! value is the encoded `history::CommandCapture`. Each write runs in an
+//! optimistic transaction so the check-then-insert is atomic against concurrent
+//! writers, and all blocking fjall I/O runs on tokio's blocking pool via
+//! `spawn_blocking`.
+//!
+//! TODO(retention): the store grows unbounded; no eviction yet. See the design doc.
+
+use atuin_client::history::HistoryId;
+use fjall::{OptimisticTxDatabase, OptimisticTxKeyspace, Readable};
```
ellie, line 13:
```suggestion
use fjall::config::CompressionPolicy;
use fjall::{CompressionType, OptimisticTxDatabase, OptimisticTxKeyspace, Readable}
```

crates/atuin-daemon/src/output_capture.rs
```diff
@@ -0,0 +1,182 @@
...
+    /// If, in the future, you need to share the database for other uses, you should definitely
+    /// delete this function and inject the database via [`OutputCapture::new`].
+    pub fn open(path: impl AsRef<std::path::Path>) -> fjall::Result<Self> {
+        let db = OptimisticTxDatabase::builder(path.as_ref()).open()?;
+        Self::new(db)
+    }
+
+    /// Create a new [`OutputCapture`] system.
+    pub fn new(db: OptimisticTxDatabase) -> fjall::Result<Self> {
+        let keyspace = db.keyspace(KEYSPACE_NAME, fjall::KeyspaceCreateOptions::default)?;
```
ellie:
```suggestion
        let keyspace = db.keyspace(KEYSPACE_NAME, fjall::KeyspaceCreateOptions::default)?;
        let keyspace = db.keyspace(KEYSPACE_NAME, || {
            fjall::KeyspaceCreateOptions::default()
                .data_block_compression_policy(CompressionPolicy::all(CompressionType::Lz4))
        })?;
```

this seems to use about half as much disk space for build heavy outputs, more depending. been messing about with data and the default config is very conservative

ellie: it looks like fjall won't migrate keyspace settings, so whatever gets set we are stuck with (unless we do some process of our own)

probably fine if we enable this now, if better algorithms come in the future we may need to figure smth out

markovejnovic: Yeah, additionally, I noticed we should perhaps consider using kv separation since our key-value pairs have such high size disparities. I tried it and got about 20% improvement in compation speed -- good enough for me to justify it.

I also tried FIFO but that just seems like it's going to be a footgun in the future, although it does make things slightly faster.

taylordotfish (approved):
looks good, would just like to see some smoke testing of atuin ai

PR #4049, opened 2026-09-02
chore: enable Clippy cast warnings

taylordotfish:
Enable warnings about `as` casts in Clippy, now that we've changed most of them to `easy-cast`.

Clippy also warns about `<float> as <int>` casts even though these saturate, which, imo, is not as much of a problem as truncation/wrapping. It's not possible to warn only about `<int> as <int>` casts, so this PR updates the `<float> as <int>` casts as well:

* Ones where we were already clamping within the destination type's range have been changed to `easy_cast::ConvTo`.
* There were several where the saturating behavior of the cast was sufficient but not necessary for our purposes. All it did was make the integer out of range for a second conversion, where we would then fall back to a default value. These have been changed to use `try_cast_to` with the same fallback as before.
* There were some where we could avoid the cast by using a different library method.
* For the remaining ones where saturating was intended, an `#[expect]` attribute has been added to them.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -219,12 +219,11 @@ borrow_as_ptr = "warn"
 branches_sharing_code = "warn"
 case_sensitive_file_extension_comparisons = "warn"
 cast_lossless = "warn"
-# TODO(markovejnovic): These could be enabled in a separate PR, but we should
-# first try using easy_cast.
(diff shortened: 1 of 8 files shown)


ellie (approved):
easy-cast is NICE! love this

PR #4051, opened 2026-09-02
fix: unblock atuin sync when a large command is included

### LINKED ISSUE #2887: [Bug]: Sync: Content Too Large (HTTP 413)

`atuin sync` with large entities should succeed (or at least not error out).

### What happened?

I had some history entries in my database that exceeded 4kB (even 8kB), causing the sync to error out with a `HTTP 413 Content Too Large`.

```
~ % atuin sync

Uploading 18979 records to <redacted>/history
2025-08-25T07:11:15.599671Z ERROR atuin_client::record::sync: failed to post records: There was an error with the atuin sync service: Status 413.
If the problem persists, contact the host

Location:
    /private/tmp/atuin-20250804-8800-rmph0g/atuin-18.8.0/crates/atuin-client/src/api_client.rs:186:9
Error: a request to the sync server failed: "There was an error with the atuin sync service: Status 413.\nIf the problem persists, contact the host"

Location:
    crates/atuin/src/command/client/store/push.rs:90:29
~ %                                                      
```

Fix was to first prune these from my `history.db` (`delete from history where length(command) > ...`) and (re)move my `records.db`.

I suggest multiple things:

The server should reject large entities in a controlled manner. The `413` was followed by a `503 Service Unavailable` so I suppose the server crashed? (I haven't verified this). Still, a better error message would be helpful.

The client should handle this sync error more gracefully - maybe mark this record as `not accepted by the server` and move on?

In addition, a configuration option to filter out history entries by size would be nice. Nobody (or, at least, I) needs base64 encoded JPEGs in their history.

### Atuin doctor output

```yaml
Atuin Doctor
Checking for diagnostics


{
  "atuin": {
    "version": "18.8.0",
    "sync": {
      "cloud": true,
      "records": true,
      "auto_sync": true,
      "last_sync": "2025-08-25 7:25:16.3285 +00:00:00"
    },
    "sqlite_version": "3.46.0"
  },
  "shell": {
    "name": "zsh",
    "default": "zsh",
    "plugins": [
      "atuin"
    ],
    "preexec": "built-in"
  },
  "system": {
    "os": "Darwin",
    "arch": "arm64",
    "version": "15.6.1",
    "disks": [
      {
        "name": "Macintosh HD",
        "filesystem": "apfs"
      },
      {
        "name": "Macintosh HD",
        "filesystem": "apfs"
      }
    ]
  }
}
```

### Code of Conduct

Dilyxs:
Fixes #2887 
### Current Situation
Whenever a user has a command that is larger than the max_size, the server rejects the upload(return error) and simply blocks atuin sync, problem is that this is blocking, the user cannot sync afterwards at all.

### Solution
Before attempting to upload syncs, isolate the large commands and make them into tombstone empty commands and warn the user that certain commands were tombstoned.

### Next Steps
Currently the UX warning displays this happens and gives away the command record ID, not the command itself(at this point it was already encrypted), a UX PR would potentially display what the long command was.

Also, as pointed out in the issue, most people would probably not want atuin to even remember these very large commands, a config that has an internal max_size would a great addition, this could also lead to the user eventually using what not save within atuin.

 Would love to implement this if maintainers this it's good idea...

## Checks
- [ x ] I am happy for maintainers to push small adjustments to this PR, to speed up the review cycle
- [ x ] I have checked that there are no existing pull requests for the same thing


Dilyxs:
closing as it is not possible to ever encrypt/decrypt record with server, client must do it, this approach will fail.

PR #4065, opened 2026-09-04
feat: simplify semantic captures

taylordotfish:
* Don't store the prompt and rendered command string; we don't use them anyway.
* Don't store history ID as an `Option`; it's always `Some`.
* Always emit the command capture even if the output is empty, for uniformity -- necessary for upcoming output capture features.


markovejnovic (approved):
lgtm

PR #4069, opened 2026-09-04
feat: add [output_capture] settings

markovejnovic:
Adds an `[output_capture]` section to `config.toml`.

```toml
[output_capture]
enabled = true
max_output_size = "1MB"
sync = false
max_disk_usage = "10%"   # or "10GB" / "unlimited"
```

Users can tweak the behavior of `output_capture` with this.

## Review Guide

- `atuin_common::byte_size.rs` adds a `ByteSize` type which is a type representing the unit of a byte size. It supports a `.human()` method for rendering `1.24GB`.
- `atuin_common::percent.rs` is a utility which supports parsing and handling a `"10%"` string.
- `atuin_client::settings::disk_usage_limit.rs` allows the user to pass in

```toml
max_disk_usage = "10%"
max_disk_usage = "10GB"
max_disk_usage = "unlimited"
```

- `atuin_client::settings::output_capture.rs` defines the `[output_capture] toml table
- `atuin_daemon::output_capture::backend::fjall` got a new `Gc` sidecar process tokio task which does the old entry garbage collection.


ellie:
output capture -> output

ellie (approved):

crates/atuin-pty-proxy/src/capture.rs
taylordotfish, line 1: Can we have `CommandCaptureTracker::new` take the `CaptureConfig` directly? I think we can get rid of `CaptureLimit` then and just directly create the `bounded_buffer::Limit`.

crates/atuin-common/src/units/percent.rs
```diff
@@ -0,0 +1,334 @@
...
+    fn add_assign(&mut self, rhs: Self) {
+        *self = *self + rhs;
+    }
+}
+
+impl SubAssign for Percent {
+    fn sub_assign(&mut self, rhs: Self) {
+        *self = *self - rhs;
+    }
+}
```
taylordotfish: derive_more

crates/atuin-common/src/units/percent.rs
```diff
@@ -0,0 +1,334 @@
...
+impl Mul<f32> for Percent {
+    type Output = f32;
+
+    fn mul(self, rhs: f32) -> f32 {
+        let share = self * f64::from(rhs);
+        f32::try_conv_approx(share).unwrap_or(if share < 0.0 {
+            f32::NEG_INFINITY
+        } else {
+            f32::INFINITY
+        })
```
taylordotfish:
```suggestion
        #[expect(clippy::cast_possible_truncation, reason = "saturation is intended")]
        let result = share as f32;
        result
```

crates/atuin-common/src/units/percent.rs
```diff
@@ -0,0 +1,310 @@
...
+impl Mul<Percent> for f32 {
+    type Output = Self;
+
+    fn mul(self, rhs: Percent) -> Self {
+        rhs * self
+    }
+}
+
+/// A share of an integer, truncated toward zero and saturating at the type's bounds.
+macro_rules! share_of_int {
```
taylordotfish: nit: would prefer to call this `impl_int_mul` or similar

taylordotfish (approved):

PR #4077, opened 2026-09-07
fix(client): support opencode V2 plugin loader in opencode hook

CubergDev:
## Problem
`atuin hook install opencode` writes a plugin with only a named export (`export const AtuinPlugin`). The V2 loader (opencode2 beta) rejects it:

> PluginModule.LoadError: Plugin must export a default definition with an id and an effect or setup function (Missing key at ["default"])

The plugin then never records anything. V1 (`1.18.x`) and V2 share the same global plugin path (`~/.config/opencode/plugins/atuin.ts`), so one template must satisfy both loaders.

## Fix
Dual V1+V2 entrypoint in `crates/atuin/contrib/opencode/atuin.ts` (V1 factory untouched):
- keeps the named `AtuinPlugin` export (pre-1.18.29) and adds `export default { id: "atuin", setup, server }`
- V2 `setup()` ports the same propose/start/finish lifecycle to `tool.execute.before` / `shell.create.before` / `tool.execute.after` (V2 shell hook carries no call ID, so proposals are claimed by matching `command`; entries are owner-tagged so a denied/duplicate command can never steal another call's entry or exit code)
- defensive by design: missing domains disable the V2 path, every hook is failure-swallowed, async cleanup awaits disposes — a history miss never breaks a tool call

## Validation
- bun harness with a fake `atuin` shim, 16/16 pass: V1 start/end, V2 start/end, denied commands produce zero atuin calls, same-command denied keeps owner's exit, concurrent identical commands each end once, timeout text maps to 124, malformed inputs never throw
- live: fresh opencode2 beta-19192 server loads the file with zero `failed to load plugin` (previously failed on every start)
- `hook.rs` tests unaffected (no test pins template content; install path logic untouched)

### DIFF
--- a/crates/atuin/contrib/opencode/atuin-v2.ts
+++ b/crates/atuin/contrib/opencode/atuin-v2.ts
@@ -1,0 +1,232 @@
+/**
+ * Atuin plugin for the OpenCode V2 tool/shell hook API.
+ *
+ * Install with `atuin hook install opencode-v2`, then restart OpenCode.
+ * This replaces the V1 template at opencode/plugins/atuin.ts.
+ * Do not place both templates in an autoload directory.
+ */
+
+import { spawn } from "node:child_process";
+
+const ATUIN_AUTHOR = "opencode";
+const ATUIN_TIMEOUT_MS = 10_000;
+const BASH_TOOL = "bash";
+const EXIT_ABORTED = 130;
+const EXIT_TIMED_OUT = 124;
+
+// Bound pending V2 proposals when the host omits completion events.
+const MAX_PROPOSED = 128;
+
+interface Proposal {
+	command: string;
+	intent?: string;
+}
+
+interface Entry {
+	historyId: string;
+	cwd: string;
+}
+
+interface AtuinResult {
+	code: number | null;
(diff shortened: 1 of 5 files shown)


carlosr301101 (changes requested):
Docs (agent-hooks.md) still only cover opencode, not opencode-v2.

crates/atuin/contrib/opencode/atuin-v2.ts
```diff
@@ -0,0 +1,232 @@
...
+ * Install with `atuin hook install opencode-v2`, then restart OpenCode.
+ * This replaces the V1 template at opencode/plugins/atuin.ts.
+ * Do not place both templates in an autoload directory.
+ */
+
+import { spawn } from "node:child_process";
+
+const ATUIN_AUTHOR = "opencode";
+const ATUIN_TIMEOUT_MS = 10_000;
+const BASH_TOOL = "bash";
```
carlosr301101, line 13: V2 = shell; bash is id from V1. 

https://opencode.ai/v2/docs/tools/

carlosr301101 (changes requested):

PR #4079, opened 2026-09-08
fix(pty-proxy): make PTY proxy aware of other PTYs

### LINKED ISSUE #4062: [Bug]: pty-proxy redraw isssue

When activating atuin using `ctrl+r` inside a pane in a terminal multiplexer that is not tmux[^special], there is a redraw issue where the pty-proxy is not  using the correct width of the area it is occupying inside the pane.

The pty-proxy history pane is using the correct width, so it is only when you dismiss the history popup you get an incorrect repaint.

[^special]: There is special handling baked in for tmux

### What happened?

I experience this inside [herdr](https://herdr.dev/) which has a sidebar, that is copied incorrectly when I activate atuin using `ctrl+r`

### Step 1

Here is a new `herdr` session where I have executed `fastfetch -s break` 3 times to get some content in the pane:

Image

### Step 2

I then press `ctrl+r` to activate  `atuin`:

Image

### Step 3

I then press `esc` to exit  `atuin`:

Image

Notice the sidebar is incorrectly copied into the space in the pane atuin previously occupied.

Note this is the same issue as #3566, but with better wording and better screenshots, as the issue was incorrectly closed by @markovejnovic who thought the issue was related to clap due to my bad description and screenshots.

I've also posted a PR that solves this issue here: #3679 

### Atuin doctor output

```yaml
Atuin Doctor
Checking for diagnostics


{
  "atuin": {
    "version": "18.21.0",
    "commit": "795d438923376ede63f6de73d7fa60d659449c09",
    "sync": {
      "auth_state": "Self-hosted (authenticated)",
      "auto_sync": true,
      "last_sync": "2026-09-04 11:59:53.184242 +00:00:00"
    },
    "sqlite_version": "3.46.0",
    "daemon_enabled": true
  },
  "shell": {
    "name": "zsh",
    "default": "zsh",
    "plugins": [
      "atuin"
    ],
    "preexec": "built-in"
  },
  "system": {
    "os": "Darwin",
    "arch": "arm64",
    "version": "26.6.2",
    "disks": [
      {
        "name": "Root",
        "filesystem": "apfs"
      },
      {
        "name": "Root",
        "filesystem": "apfs"
      },
      {
        "name": "Projects",
        "filesystem": "apfs"
      }
    ]
  }
}
```

### Code of Conduct

### LINKED ISSUE #4078: [Bug]: pty-proxy never starts in fish — documented `| source` init defeats its own `test -t 0` guard

Following the fish instructions in `docs/docs/reference/pty-proxy.md` — adding
`atuin pty-proxy init fish | source` to the `is-interactive` block of `~/.config/fish/config.fish` —
I expected the shell to be re-exec'd inside `atuin pty-proxy`, so `ATUIN_PTY_PROXY_ACTIVE=1` and
command output capture works.

### What happened?

Nothing happens. No error, no proxy: `ATUIN_PTY_PROXY_ACTIVE` stays unset, and `AtuinOutput` /
the MCP `atuin_output` tool returns "No captured output found" for every command.

`atuin pty-proxy init fish` emits a block guarded by:

```fish
```

Piping it into `source` means `source` reads the script from the pipe, so fd 0 is the pipe rather
than the terminal. `test -t 0` is false, and the whole block — including `exec atuin pty-proxy` —
is skipped silently.

Minimal reproduction (fish 4.8.1). `config.fish`:

```fish
if test -t 0; set -g D yes; else; set -g D no; end
echo 'if test -t 0; set -g P yes; else; set -g P no; end' | source
printf 'direct=%s piped=%s\n' "$D" "$P" > /tmp/probe.txt
```

Run `script -qec 'fish -i -c true' /dev/null`, then `cat /tmp/probe.txt`:

```
direct=yes piped=no
```

fish is the only shell whose documented init pipes the snippet into the reader: zsh and bash use
`eval "$(...)"`, and Nushell pre-generates a file and uses `source <path>`. Those keep fd 0 on the
tty, which is why only fish is affected.

Setting `[pty_proxy] enabled = true` does not help either — `atuin init fish | source` is also
documented as a pipeline, so the same guard fails there.

Workaround, avoiding the pipe — `psub` hands `source` a file path, so fd 0 stays the terminal.
Verified working: the shell re-enters with `ATUIN_PTY_PROXY_ACTIVE=1`, and output capture then
returns real output.

```fish
source (atuin pty-proxy init fish | psub)
```

Possible fixes: update the fish instructions in `docs/docs/reference/pty-proxy.md` and
`docs/docs/ai/command-output.md` to the `psub` form, and/or drop `test -t 0` from the fish guard in
`crates/atuin/src/shell/atuin.fish` — `status is-interactive` plus `test -t 1` already covers the
intent.

---

*Investigated and reproduced with Claude Code; all output above is from my machine.*

### Atuin doctor output

```yaml
{
  "atuin": {
    "version": "18.19.0",
    "commit": "fb33978d9439ba7e97123c1adc932c6a74e3c8ae",
    "sync": {
      "auto_sync": false,
      "last_sync": "1970-01-01 0:00:00.0 +00:00:00"
    },
    "sqlite_version": "3.46.0",
    "daemon_enabled": true
  },
  "shell": {
    "name": "fish",
    "default": "bash",
    "plugins": [
      "atuin"
    ],
    "preexec": "built-in"
  },
  "system": {
    "os": "TUXEDO OS",
    "arch": "x86_64",
    "version": "24.04",
    "disks": [
      {
        "name": "/dev/mapper/system-root",
        "filesystem": "ext4"
      },
      {
        "name": "/dev/nvme0n1p1",
        "filesystem": "ext3"
      },
      {
        "name": "/dev/nvme0n1p2",
        "filesystem": "vfat"
      }
    ]
  }
}
```

### Code of Conduct

taylordotfish:
This PR fixes rendering issues with the PTY proxy when used alongside multiplexers like `tmux` and `screen` or any other program that creates its own PTY.

### Background

Atuin's PTY proxy does not work well when used with other programs that create PTYs – particularly terminal multiplexers. Atuin has a workaround for tmux, but other programs cause serious rendering issues, as seen in #4062.

The underlying problem is that as soon as another PTY sits in between the Atuin client and PTY proxy – that of `tmux` or `screen`, for example – Atuin's prediction of the rendered state of the terminal, which it uses to restore the screen when the search UI is closed, is no longer accurate, and Atuin ends up drawing incorrect data in an incorrect position.

### The fix

This PR makes it so Atuin only connects to a PTY proxy if it is *directly* running under the PTY proxy – that is, if its terminal is the PTY proxy's terminal. If something like tmux is sitting in between, Atuin now treats this as though no PTY proxy is running at all, and (assuming `pty_proxy.enabled` is true) will spawn a new one in each tmux pane.

The underlying mechanism for the fix is a change in the PTY proxy's socket path. Previously, each socket path was made unique by inserting the PTY proxy's PID into the filename. Now, a unique, deterministic identifier corresponding to the child PTY created by the proxy is included instead. This means we no longer rely on passing the socket path in an environment variable (though this is still done for compatibility). Instead, Atuin derives the expected socket path from the ID of its own terminal and checks if there is a live socket there; if there is, we know we must be running under the PTY proxy on that socket. This check is immune to collisions from reused PTY numbers – because the path is uniquely scoped to our own terminal, it cannot be taken by someone else. The check is accomplished using a new internal command, `atuin __internal pty-proxy-active`, which has been made to be as fast as possible by skipping tokio, settings, and logging initialization.

### Testing

In addition to the unit tests, I've performed smoke testing with various combinations of nested shells and tmux sessions; everything works as expected with correct rendering.

Fixes #4062; also fixes #4078 (fish init script issue).

### DIFF
--- a/crates/atuin-common/src/os/unix/tty.rs
+++ b/crates/atuin-common/src/os/unix/tty.rs
@@ -1,0 +1,175 @@
+//! TTY-related utilities.
+
+use std::os::fd::{AsFd, BorrowedFd};
+
+use rustix::fs::Dev;
+
+/// A unique identifier for a terminal device.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct TtyId {
+    /// Which filesystem the terminal lives on.
+    ///
+    /// Corresponds to `st_dev` in `struct stat`.
+    pub dev: Dev,
+
+    /// The device ID of the terminal.
+    ///
+    /// Corresponds to `st_rdev` in `struct stat`.
+    pub rdev: Dev,
+}
+
+impl TtyId {
+    /// Get the terminal ID of a file descriptor.
+    ///
+    /// Returns [`None`] if the file is not a terminal.
+    #[must_use]
+    pub fn from_fd(fd: impl AsFd) -> Option<Self> {
+        let fd = fd.as_fd();
+        if !rustix::termios::isatty(fd) {
+            return None;
+        }
+        let stat = rustix::fs::fstat(fd).ok()?;
+        Some(Self {
+            dev: stat.st_dev,
+            rdev: stat.st_rdev,
+        })
+    }
+
+    /// Get the ID of the terminal that this process is attached to.
+    ///
+    /// Returns [`None`] if none of this process's standard file descriptors are attached to a
+    /// terminal.
+    #[must_use]
+    pub fn current() -> Option<Self> {
+        use rustix::stdio::{stderr, stdin, stdout};
+        first_tty_id([stdin(), stdout(), stderr()])
+    }
+}
+
+/// Get the ID of the first FD in `fds` that is a terminal.
+fn first_tty_id<'a>(fds: impl IntoIterator<Item = BorrowedFd<'a>>) -> Option<TtyId> {
+    fds.into_iter().find_map(TtyId::from_fd)
+}
+
+#[cfg(test)]
+mod tests {
+    use std::ffi::OsStr;
+    use std::fs::File;
+    use std::os::unix::ffi::OsStrExt;
+    use std::path::PathBuf;
+
+    use rstest::{fixture, rstest};
+    use rustix::pty::{OpenptFlags, grantpt, openpt, ptsname, unlockpt};
+
+    use super::*;
+
+    /// Open a fresh pty pair, returning the master, the slave, and the slave's path.
+    #[fixture]
+    fn pty() -> (File, File, PathBuf) {
+        let master = openpt(OpenptFlags::RDWR | OpenptFlags::NOCTTY).unwrap();
+        grantpt(&master).unwrap();
+        unlockpt(&master).unwrap();
+        let name = ptsname(&master, Vec::new()).unwrap();
+        let path = PathBuf::from(OsStr::from_bytes(name.as_bytes()));
+        let slave = File::options().read(true).write(true).open(&path).unwrap();
+        (File::from(master), slave, path)
+    }
+
+    #[rstest]
+    fn from_fd_of_a_pty_matches_a_stat_of_its_path(pty: (File, File, PathBuf)) {
+        use std::os::unix::fs::MetadataExt;
+
+        let (_master, slave, path) = pty;
+        let meta = std::fs::metadata(&path).unwrap();
+
+        let id = TtyId::from_fd(&slave).expect("a pty slave is a terminal");
+
+        // `Dev` is already u64 on Linux but is narrower on macOS, so the conversion is not
+        // redundant everywhere it compiles.
+        #[allow(clippy::useless_conversion, reason = "Dev is not u64 on every platform")]
+        {
+            assert_eq!(u64::try_from(id.dev).unwrap(), meta.dev());
+            assert_eq!(u64::try_from(id.rdev).unwrap(), meta.rdev());
+        }
+    }
+
+    /// Serialises the tests that swap this process's stdin.
+    static STDIN_LOCK: parking_lot::Mutex<()> = parking_lot::Mutex::new(());
+
+    /// Replace stdin with `fd` for as long as the returned guard lives.
+    fn with_stdin(fd: BorrowedFd<'_>) -> impl Drop {
+        struct Restore {
+            saved: std::os::fd::OwnedFd,
+            _lock: parking_lot::MutexGuard<'static, ()>,
+        }
+        impl Drop for Restore {
+            fn drop(&mut self) {
+                rustix::stdio::dup2_stdin(&self.saved).unwrap();
+            }
+        }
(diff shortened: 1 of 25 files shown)


crates/atuin-common/src/os/unix/tty.rs
```diff
@@ -0,0 +1,175 @@
...
+    pub fn current() -> Option<Self> {
+        use rustix::stdio::{stderr, stdin, stdout};
+        first_tty_id([stdin(), stdout(), stderr()])
+    }
+}
+
+/// Get the ID of the first FD in `fds` that is a terminal.
+fn first_tty_id<'a>(fds: impl IntoIterator<Item = BorrowedFd<'a>>) -> Option<TtyId> {
+    fds.into_iter().find_map(TtyId::from_fd)
+}
```
markovejnovic: nit-3: I'd probably inline this.

markovejnovic (approved):
lgtm

crates/atuin-common/src/os/unix/tty.rs
taylordotfish: yeah it was so it could be tested but honestly something this simple doesn't need a unit test

taylordotfish: done

ches:
This change appears to have broken my fish initialization entirely upon upgrade to 18.22.0. I'll try to minimize and create a new issue, dropping breadcrumbs here to start.

I have a `~/.config/fish/conf.d/atuin.fish` (might be significant that it's a `conf.d` snippet) that looks like this:

```fish
if status is-interactive && command --query atuin && test -z "$fish_private_mode"
    # Using the pty proxy (formerly "Hex") eliminates jumpy rendering
    atuin pty-proxy init fish | source

    atuin init fish | source
end
```

As per #4078 the pty-proxy initialization was not actually effective—I'd added it recently without yet exercising the feature.

After the upgrade, opening a terminal in every emulator I tried (iTerm 2, Terminal.app, Zed's terminal—no multiplexer in the picture) shows a `^D` printed immediately and if I let the terminal stay open after shell exit I see:

```
warning: fish could not read response to Primary Device Attribute query after waiting for 10 seconds. This is often due to a missing feature in your terminal. See 'help terminal-compatibility' or 'man fish-terminal-compatibility'. This fish process will no longer wait for outstanding queries, which disables some optional features.
fish_config theme choose: failed to find '[unknown]' section (implied by $fish_terminal_color_theme) in 'tomorrow' theme
fish_config theme choose: hint: if your terminal does not report colors, pass --color-theme=light or --color-theme=dark when using color-theme-aware themes
Welcome to fish, the friendly interactive shell
Type help for instructions on how to use fish

ches at Kiki in ~
↪ 

ches at Kiki in ~
↪ 

[Process completed]
```

(`fish_config theme choose tomorrow` is run in my main `config.fish`).

Workaround is of course to just disable the proxy init for the time being.

PR #4082, opened 2026-09-08
refactor(daemon): split output capture into fjall and nop backends

markovejnovic:
Splits the daemon's output capture into a facade plus pluggable backends, so a follow-up PR can disable output capture by selecting the no-op backend.


crates/atuin-daemon/src/output_capture/backend/mod.rs
```diff
@@ -0,0 +1,63 @@
...
+        }
+    }
+
+    async fn get(&self, id: HistoryId) -> Result<Option<CommandCapture>, GetOutputError> {
+        match self {
+            Self::Fjall(backend) => backend.get(id).await,
+            Self::Nop(backend) => backend.get(id).await,
+        }
+    }
+}
```
taylordotfish: `enum_dispatch`? we're doing something very similar for `SearchEngine` iirc

PR #4090, opened 2026-09-09
feat: rework inspector and add output view

ellie:
This change incorporates much of the feedback we have had about the inspector so far; users can easily explore past occurrences of a command, or switch to view other commands from the same session. 

We also show the actual output of the command here, too

## Checks


ellie:
@greptile-apps plz

PR #4092, opened 2026-09-09
feat(pty-proxy): keep the start and end of command captures that exceed the max size

taylordotfish:
Instead of keeping just the first 1 MiB of a command capture that exceeds the limit, this PR keeps the first 512 KiB at the start and the last 512 KiB at the end. The code is set up in a way that will make it very easy to make this limit customizable.

The start and end chunks are snapped to whole lines – if the start chunk ends in the middle of a line, or the end chunk starts in the middle of a line, the partial lines are removed. This is not just to make the output nicer but is also important for secret redaction.

`GetCommandOutputResponse` can now contain negative line numbers; these correspond to lines in the end chunk of an output whose middle was discarded. We cannot convert these to unsigned indices as we don't know how many lines were clipped, so they have to stay negative. This mirrors how the line range was expressed in the `GetCommandOutputRequest`.

This PR bumps the fjall schema version, since it adds and removes fields. Note that we have not actually released a version that uses any schema yet.

This PR fixes an issue in `PyStyleIdxRange` where `(0, -1000)` would resolve to `0..1` even for slices of length less than 1000.


crates/atuin-daemon/proto/history.proto
```diff
@@ -179,9 +169,10 @@ message CommandCapture {
   //
   // Note this output is mostly *VISIBLE* output. Many characters are stripped, including, but not
   // limited to `\r` characters, some control bytes and newlines trimmed at the end of the output.
-  string output = 1;
+  string output_start = 1;
+  optional string output_end = 2;
   // Metadata on this particular command capture. Logically required.
-  CommandCaptureMeta meta = 2;
+  CommandCaptureMeta meta = 3;
```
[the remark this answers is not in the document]
taylordotfish: Not an issue. We haven't published a release since the protocol version was incremented to 2.

crates/atuin/src/command/mod.rs
```diff
@@ -126,18 +126,26 @@ fn semantic_command_capture_sink() -> Option<atuin_pty_proxy::CommandCaptureSink
...
+                        *output = redacted;
+                    }
+                };
+
+                let mut output_start = capture.output_start;
+                redact(&mut output_start);
+                let mut output_end = capture.output_end;
+                if let Some(output_end) = output_end.as_mut() {
+                    redact(output_end);
                 }
```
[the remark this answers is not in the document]
taylordotfish: Disregard, commented on the wrong issue.

crates/atuin-common/src/string/bounded_buffer.rs
```diff
@@ -0,0 +1,497 @@
...
+    /// How many bytes from the end of the buffer to keep.
+    pub end: usize,
+}
+
+/// A string buffer limited to a certain length.
+#[derive(Clone)]
+pub struct BoundedBuffer {
+    start: String,
+    /// Invariant: `end` is valid UTF-8.
+    end: VecDeque<u8>,
```
markovejnovic, line 18: Could be nice to have a `DequeString`

crates/atuin-pty-proxy/src/capture.rs
```diff
@@ -284,6 +300,12 @@ mod tests {
     const ROWS: u16 = 24;
     const COLS: u16 = 80;
 
+    /// Roomy enough that nothing below is truncated unless the case asks for it.
+    const LIMIT: CaptureLimit = CaptureLimit {
+        start_bytes: 64 * 1024,
+        end_bytes: 64 * 1024,
```
markovejnovic, line 306:
```suggestion
        start_bytes: ByteSize::kib(64),
        end_bytes: ByteSize::kib(64),
```

Coming near you! https://github.com/atuinsh/atuin/pull/4086

crates/atuin-pty-proxy/src/screen.rs
```diff
@@ -93,10 +93,20 @@ impl Parser {
     /// in the terminal -- otherwise our emulator would badly drift from the parent terminal.
     const SCROLLBACK_CAPACITY: usize = 50;
 
+    /// The maximum size of a command capture.
+    ///
+    /// 1 MiB total, split across the start and end.
+    const CAPTURE_LIMIT: capture::CaptureLimit = capture::CaptureLimit {
+        start_bytes: 512 * 1024, // 512 KiB
```
markovejnovic, line 100:
```suggestion
        start_bytes: ByteSize::kib(512),
```

markovejnovic (approved):

PR #4097, opened 2026-09-10
fix(pty-proxy): resolve zsh --shell to an absolute path with :c

davidolrik:
Super small fix for #3606 - It just adds `:c` to whatever $ZSH_ARGZERO has been resolved to.

This turns bare word `zsh` in to an absolute path, and leaves already absolute paths alone (even paths that don't exist which should fail anyway).

There is already work in progress to fix this issue in #3757, but this 2-byte fix silences a super annoying warning and keeps $SHELL as is.

## Checks


taylordotfish:
Thank you!

PR #4101, opened 2026-09-10
feat: improve `atuin_common::ansi::to_plain_text`

taylordotfish:
With this PR, `ansi::to_plain_text` (used by Atuin AI):

* Is now up to 5.6x faster
* Now uses up to 432× less memory
* No longer has an arbitrary limit on the maximum number of rows (this was just because the emulator would use so much memory the process would OOM)

**Benchmark:**

| Input | Old | New | Speedup | Old RSS Δ | New RSS Δ | RSS factor |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| 600 B, 10 lines | 23 µs | 23 µs | 1.0× | 136 KB | 120 KB | 1.1× less |
| 61 KB, 1000 lines | 1.95 ms | 1.22 ms | 1.6× | 6.1 MB | 180 KB | 34× less |
| 35 KB, SGR-colored | 1.94 ms | 1.08 ms | 1.8× | 5.8 MB | 160 KB | 36× less |
| 100 KB single line (soft wrap) | 1.86 ms | 1.63 ms | 1.1× | 3.5 MB | 240 KB | 15× less |
| 1 MB, 20k `\r` progress rewrites | 22.2 ms | 6.3 ms | 3.5× | 38 MB | 88 KB | 432× less |
| 3.3 MB, 50k lines | 285 ms | 63 ms | 4.5× | 66 MB | 3.3 MB | 20× less |
| 110 MB, 1.6M lines | 11.65 s | 2.07 s | 5.6× | 175 MB | 108 MB | 1.6× less |

(This benchmark is available in the “Add benchmark” commit in this branch.)

Note that the last two rows undersell the performance increase: the old behavior capped the number of lines at 16,384, while the new behavior returns all 50k/1.6M lines and yet is *still* faster and uses less memory.


crates/atuin-common/src/ansi.rs
```diff
@@ -1,77 +1,115 @@
...
-    let contents = parser.screen().contents();
-    let trimmed = contents.trim_end();
-    let mut out = String::with_capacity(trimmed.len());
-    let mut first = true;
-    for line in trimmed.lines() {
-        if !first {
-            out.push('\n');
+            contents
+                .write_formatted_basic(&mut self.buffer, &mut self.state)
+                .expect("writing to a String cannot fail");
```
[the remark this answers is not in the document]
taylordotfish: We already buffer the full raw contents of stdout and stderr before feeding them to the vt100 emulator, so memory use can only increase by the size of those buffers, as it is very unlikely that emulation would produce output that greatly exceeds the original raw size of the buffers (in fact, in many cases, the rendered plain text output will be considerably smaller). This is the same increase that you would get from, e.g., simply cloning the strings. If we get OOMs from this it is an indication that we need to add output limiting earlier in the pipeline, so that the raw stdout and stderr themselves have already been limited, because we shouldn't be using 50% system memory on those buffers.

crates/atuin-ai/src/tools/mod.rs
```diff
@@ -92,6 +92,21 @@ pub enum ToolOutcome {
...
+    /// [`ansi::to_plain_text`] is more efficient, it imposes no limit of its own; to ensure we
+    /// don't return too much data, we apply the cap ourselves.
+    ///
+    /// 16384 times 120 (the width of the emulated terminal, from `PREVIEW_WIDTH`) is approximately
+    /// 2,000,000 (2MB), so we use that as our limit here.
+    ///
+    /// Note that we keep the *end* of output that exceeds the limit. This matches what
+    /// [`ansi::to_plain_text`] did previously -- it would process the full output but only keep the
+    /// last 16384 rendered lines.
+    const MAX_STRUCTURED_OUTPUT_SIZE: usize = 2_000_000;
```
markovejnovic, line 108:
```suggestion
    const MAX_STRUCTURED_OUTPUT_SIZE: ByteSize = ByteSize::mb(2);
```

markovejnovic: https://github.com/bernedom/SI

markovejnovic (approved):

PR #4105, opened 2026-09-11
chore(deps): dependency version bumps

markovejnovic:
## Drop-in major bumps (no code change)

`base64` 0.22→0.23 · `frizbee` 0.12→0.13 · `rstest` 0.26→0.27 · `signal-hook` 0.3→0.4 · `tower-http` 0.6→0.7 · `tui-textarea-2` 0.10→0.13 · `yaml-rust2` 0.11→0.12 · `zstd` 0.13→0.14 ·

---

### `argon2` 0.5 → 0.6
- `PasswordHasher::hash_password` **dropped its salt argument**. The PHC output string (`$argon2id$v=19$m=…$salt$hash`) is compatible.

### `rmcp` 2.1 → 3.3
- `ServerHandler::call_tool` now returns **`CallToolResponse`** (an enum: `Complete(CallToolResult)` / `InputRequired` / `Task`) instead of `CallToolResult` directly.

### `eye_declare` 0.6 → 0.7
- `InputEvent` is now `#[non_exhaustive]`, so the exhaustive `match event { Key … , Paste … }` no longer compiles.

### `tree-sitter` 0.26 → 0.27
- `QueryMatch::captures` became a **private field**; a public `captures()` accessor replaced direct field access.

---

## Held (a dependency of ours pins the old major)

Both carry an inline comment in `Cargo.toml` and were verified against the blocker's latest release:

- **`libsqlite3-sys`** (0.30 → 0.38 available): `sqlx-sqlite` 0.9 (our sqlx) requires `>=0.30.1, <0.38`, and the single `links = "sqlite3"` forbids a second native version. Bumping is impossible until sqlx widens its range.
- **`rusty_paseto`** (0.8 → 0.10 available): `rusty_paserk` 0.5.0 (its own latest) still requires `rusty_paseto ^0.8`. Bumping only duplicates majors.

---

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,11 +28,13 @@ atuin-vt100 = "0.19"
 
 ansi-to-tui = { version = "8.0.1", default-features = false }
 arboard = { version = "3.4", default-features = false }
+# held at 0.5: rusty_paserk 0.5.0 pins argon2 0.5; bumping would add a 2nd copy
 argon2 = "0.5"
 async-stream = "0.3"
 async-trait = "0.1.58"
 axoupdater = "0.10"
 axum = "0.8"
+# held at 0.22: sqlx-postgres 0.9 pins base64 0.22; bumping would add a 2nd copy
 base64 = "0.22"
 bytesize = { version = "2.7", features = ["serde"] }
 chrono = "0.4"
@@ -70,11 +72,11 @@ divan = { version = "5.0.1", package = "codspeed-divan-compat" }
 easy-cast = { version = "0.5.4", features = ["assert_nonzero"] }
 enum_dispatch = "0.3"
 eventsource-stream = "0.2"
-eye_declare = "0.6.4"
-eye_declare_engine = { version = "0.6.4", features = ["test-util"] }
+eye_declare = "0.7.1"
+eye_declare_engine = { version = "0.7.1", features = ["test-util"] }
 eyre = "0.6"
 fjall = { version = "3.1" }
-frizbee = "0.12.0"
+frizbee = "0.13.0"
 fs-err = "3.1"
 futures = "0.3"
 futures-util = "0.3"
@@ -98,7 +100,6 @@ metrics = "0.24"
 metrics-exporter-prometheus = { version = "0.18", default-features = false }
 minijinja = "2.9.0"
 minspan = "0.1.5"
-nix = { version = "0.30.1", features = ["signal"] }
 norm = { version = "0.1.1", features = ["fzf-v2"] }
 notify = "8"
 opentelemetry = "0.32"
@@ -116,7 +117,6 @@ pulldown-cmark = "0.13.0"
 rand = { version = "0.8.5", features = ["std"] }
 ratatui = "0.30.0"
 ratatui-core = "0.1"
-ratatui-widgets = "0.3"
 regex = "1.10.5"
 reqwest = {
     version = "0.13",
@@ -124,10 +124,10 @@ reqwest = {
     features = ["json", "native-tls", "stream", "gzip", "zstd"],
 }
 reqwest-middleware = "0.5"
-rmcp = { version = "2.1.0", default-features = false, features = ["server", "transport-io"] }
+rmcp = { version = "3.3.0", default-features = false, features = ["server", "transport-io"] }
 rmp = "0.8.14"
 rpassword = "7.0"
-rstest = "0.26.1"
+rstest = "0.27.0"
 runtime-format = "0.1.3"
 rustix = { version = "1.1.4", features = ["process", "fs", "pty", "termios", "stdio"] }
 rusty_paserk = { version = "0.5.0", default-features = false, features = ["v4", "serde"] }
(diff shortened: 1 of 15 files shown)


crates/atuin-ai/src/tui/app.rs
```diff
@@ -1017,6 +1017,8 @@ impl App for AiApp {
                     InputEvent::Paste(s) => {
                         self.input.get_mut().insert_str(s);
                     }
+                    // eye_declare 0.7 made InputEvent #[non_exhaustive].
```
markovejnovic:
```suggestion
```

eye_declare 0.7 made InputEvent #[non_exhaustive].

crates/atuin-ai/src/mcp.rs
```diff
@@ -103,15 +104,17 @@ impl ServerHandler for AtuinMcp {
...
             ToolOutcome::Error(text) => CallToolResult::error(vec![ContentBlock::text(text)]),
             // The atuin tools only produce Success/Error; fall back to the
             // generic formatting should that ever change.
             outcome @ ToolOutcome::Structured { .. } => {
                 CallToolResult::success(vec![ContentBlock::text(outcome.format_for_llm(None))])
             }
-        })
+        };
+        // rmcp 3.x wraps the tool result in a CallToolResponse (Complete/InputRequired/Task).
+        Ok(result.into())
```
markovejnovic:
```suggestion
        Ok(result.into())
```

rmcp 3.x wraps the tool result in a CallToolResponse (Complete/InputRequired/Task)

crates/atuin-client/src/import/nu_histdb.rs
```diff
@@ -36,8 +35,10 @@ impl From<HistDbEntry> for History {
         let ts_ns = (histdb_item.start_timestamp % 1000) * 1_000_000;
         // a corrupt row must not take down the whole import. the epoch sorts to the
         // bottom of history and is obviously not a real time
-        let timestamp = OffsetDateTime::from_timespec(i128::from(ts_secs), i128::from(ts_ns))
-            .unwrap_or(OffsetDateTime::UNIX_EPOCH);
+        let timestamp = OffsetDateTime::from_unix_timestamp_nanos(
+            i128::from(ts_secs) * 1_000_000_000 + i128::from(ts_ns),
+        )
```
markovejnovic: Can't this like use `std::Duration` or something?

crates/atuin-client/src/import/resh.rs
```diff
@@ -111,7 +110,10 @@ impl Importer for Resh {
                 let secs: i64 = realtime.try_cast_to(Floor).ok()?;
                 let nanosecs: i64 =
                     (realtime.fract() * 1_000_000_000_f64).try_cast_to(Nearest).ok()?;
-                OffsetDateTime::from_timespec(i128::from(secs), i128::from(nanosecs)).ok()
+                OffsetDateTime::from_unix_timestamp_nanos(
+                    i128::from(secs) * 1_000_000_000 + i128::from(nanosecs),
+                )
```
markovejnovic: Same as previous comment.

crates/atuin-server/src/handlers/user.rs
```diff
@@ -248,7 +246,7 @@ pub async fn login(
 
 fn hash_secret(password: &str) -> String {
     let arg2 = Argon2::new(Algorithm::Argon2id, Version::V0x13, Params::default());
-    let salt = SaltString::generate(&mut OsRng);
-    let hash = arg2.hash_password(password.as_bytes(), &salt).unwrap();
+    // argon2 0.6's hash_password generates a random salt internally (via OsRng).
+    let hash = arg2.hash_password(password.as_bytes()).unwrap();
```
markovejnovic:
```suggestion
let hash = arg2.hash_password(password.as_bytes()).unwrap();
```

argon2 0.6's hash_password generates a random salt internally (via OsRng).

Cargo.toml
```diff
@@ -149,7 +149,9 @@ strum_macros = "0.28"
 sysinfo = "0.39.6"
 tempfile = { version = "3.19" }
 thiserror = "2"
-time = { version = "0.3.47", features = ["serde-human-readable", "macros", "local-offset"] }
+# 0.3.55 fixes the from_unix_timestamp_nanos overflow (time-rs/time#802) that
+# OffsetDateTimeExt's helpers rely on upstream to reject.
```
markovejnovic:
```suggestion
```

crates/atuin-ai/src/tui/app.rs
markovejnovic: Dropped.

crates/atuin-ai/src/mcp.rs
markovejnovic: Dropped.

crates/atuin-client/src/import/nu_histdb.rs
markovejnovic: Reverted to main's `from_timespec` here — the inlined `secs*1e9 + nsecs` this was on is gone. It takes a signed `(i128 secs, i128 nsecs)` pair rather than `std::Duration` because unix timestamps can be negative (pre-1970) and `Duration` is unsigned, so it can't represent them. The ms→(secs,nsecs) split here is main's existing code, left as-is.

crates/atuin-client/src/import/resh.rs
markovejnovic: Same — reverted to main's `from_timespec`; see the nu_histdb thread for why not `std::Duration`.

crates/atuin-server/src/handlers/user.rs
markovejnovic: argon2 is now held at 0.5 (rusty_paserk pins it, so bumping would add a second copy) — this migration and its comment are gone from the PR entirely.

taylordotfish (approved):
approved conditional on smoke testing

markovejnovic:
> approved conditional on smoke testing

smoke tests look good

Issue #2593, opened 2025-02-20
[Bug]: Import for ZSH loads first history file found


rodrigomideac:
### What did you expect to happen?

When running `atuin import zsh` it should import the history that `history` shows and all these commands should be available for atuin.

### What happened?

 `atuin import zsh` imported a legacy history file that I had on `~/.zhistory`, instead of using my `/home/myusername/.zsh_history` that was set on `HISTFILE`.

### Atuin doctor output

```yaml
{
  "atuin": {
    "version": "18.4.0",
    "sync": null,
    "sqlite_version": "3.46.0"
  },
  "shell": {
    "name": "zsh",
    "default": "zsh",
    "plugins": [
      "atuin"
    ],
    "preexec": "built-in"
  },
  "system": {
    "os": "Manjaro Linux",
    "arch": "x86_64",
    "version": "25.0.0",
    "disks": [
      {
        "name": "/dev/nvme0n1p2",
        "filesystem": "ext4"
      },
      {
        "name": "/dev/nvme0n1p1",
        "filesystem": "vfat"
      }
    ]
  }
}
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

rodrigomideac:
It probably is related to processing the first element (.zhistory) here:

https://github.com/atuinsh/atuin/blob/c6d50bb863d7904582e0f4fbc00dbae8aa2b7107/crates/atuin-client/src/import/zsh.rs#L29

I am open to implement any fix that you suggest!

monzool:
Surprised me too. Installed atuin and after import I had only some old legacy entries in the history.

I would say if `HISTFILE` is set, it would be a sane default to use the file it points to, as import file

```sh
❱ man zshparam | col -b | grep -A 2 HISTFILE
       HISTFILE
              The file to save the history in when an interactive shell exits.  If unset, the history is not saved.
```

Alternatively just import **all** history files for those how have a long pedigree with zsh

jamiechicago312:
Taking a look at fixing this issue 👍

markovejnovic:
Closed #4113 as it wasn't a fix and was rather just a docs + test PR.

From my limited rapid research (ie. asking Claude), it seems that Zsh (and, according to Claude, other shells) do not export `HISTFILE`, but rather keep it as a local environment variable. Reasonable from the shell's perspective, I suppose.

[...]

PR #4113, opened 2026-09-11
docs(import): clarify Zsh HISTFILE selection and add regression tests

jamiechicago312:
Related to #2593.

When Zsh's `HISTFILE` is set but not exported, `atuin import zsh` can select an old `.zhistory` file. Exported `HISTFILE` already takes precedence; the child process can't see a shell parameter that hasn't been exported.

This documents how to pass the current `HISTFILE` for one import, clarifies the fallback order, and adds isolated path-selection regression tests. It also includes `test-guide.md` with disposable Docker reproduction steps. Import behavior and fallback ordering are unchanged.

## Testing

I tested the original fallback selection, explicit forwarding, a path containing spaces, and a missing explicit file in Docker. All behaved as expected: the intended file was imported when passed explicitly, and the missing file failed without importing a fallback.

Automated checks passed: 48 importer tests, 426 client tests, cargo check, nightly formatting, client rustdoc, and Vale. Client Clippy runs (including tests) exited successfully with an existing configuration warning about `time::Duration::new`.

## Checks


jamiechicago312:
Logs from testing 
[20260911-logs-atuin-issue-2593.rtf](https://github.com/user-attachments/files/32126541/20260911-logs-atuin-issue-2593.rtf)

jamiechicago312:
I used this [test-guide ](https://github.com/jamiechicago312/atuin/blob/4ca9951cb71fa8be583f2156a051eeea53e133c9/test-guide.md) to check this PR works. Attached are the logs of my testing. 
[20260913-logs-pr4113.rtf](https://github.com/user-attachments/files/32165133/20260913-logs-pr4113.rtf)

markovejnovic:
Not sure if this PR was completely written by AI as the PR body does not reflect what the commit source is. I'll provide feedback regardless. `testing-guide.md` doesn't exist.

This PR is only docs + tests, and doesn't actually solve the underlying issue. However, I do think there's a way to get the desired behavior of sourcing `HISTFILE` by spawning a `zsh` subshell and probing its environment via an inline `echo`.

I have done some large amount of work refactoring shell behavior, but we're not there yet. I will provide some further guidance in the original issue.

PR #4116, opened 2026-09-12
fix(opencode): support opencode V2 plugin API

candrewlee14:
The opencode template only exports a V1 named plugin function (`export const AtuinPlugin`), so opencode 2.x refuses to load it and no agent commands are recorded:

```
failed to load plugin target=.../plugins/atuin.ts cause="... Plugin must export a default definition with an id and an effect or setup function. (cause: SchemaError(Missing key at [\"default\"]))"
```

This default-exports one definition serving both APIs, per https://opencode.ai/v2/docs/build/plugins/migrate-v1: V1 (>= 1.18.29) calls `server()`, V2 reads `id` + `setup()` and ignores `server()`.

Notes on the V2 port:
- Same propose/claim/finish flow. V2's `shell.create.before` carries no tool call ID, only the command and its resolved cwd, so the claim matches the oldest pending proposal with equal command text (V1's `shell.env` allowed an exact call-ID lookup).
- Accepts the shell tool under either the legacy `bash` id or the newer `shell` id.
- Users also need the V2 SDK resolvable from the plugin file, e.g. `npm --prefix ~/.config/opencode install --save-exact @opencode/plugin` (the installer manages no node deps, same as the V1 `@opencode-ai/plugin` import before it).

Verified: strict `tsc` clean on the template; `setup()` exercised against stubbed tool/shell hooks with a stub `atuin` binary (start with `--intent`, end with correct exit incl. 130 on abort, denied/unmatched commands unrecorded). The V1 `server()` half is the previous logic verbatim and is shape-checked only — I have no V1 binary to run it against.

### DIFF
--- a/crates/atuin/contrib/opencode/atuin.ts
+++ b/crates/atuin/contrib/opencode/atuin.ts
@@ -7,29 +7,40 @@
  * Install with:
  *   atuin hook install opencode
  *
+ * Then make sure the plugin SDK is installed:
+ *
+ *   npm --prefix ~/.config/opencode install --save-exact @opencode/plugin
+ *
  * Then restart opencode.
+ *
+ * This file supports both opencode V1 (>= 1.18.29) and V2 from a single
+ * default export: V1 calls `server()` and uses the returned hooks, while V2
+ * reads the `id` and `setup()` definition and ignores `server()`.
+ * See https://opencode.ai/v2/docs/build/plugins/migrate-v1.
  */
 
-import type { Plugin } from "@opencode-ai/plugin";
+import { Plugin as V2Plugin } from "@opencode/plugin";
+import type { Plugin as V1Plugin } from "@opencode-ai/plugin";
 import { spawn } from "node:child_process";
 
 const ATUIN_AUTHOR = "opencode";
 const ATUIN_TIMEOUT_MS = 10_000;
 
-// opencode's shell tool keeps the id `bash` for compatibility, even though its
-// module is named `shell`.
-const BASH_TOOL = "bash";
+// opencode's shell tool kept the id `bash` for compatibility for a long time;
+// newer releases call it `shell`. The V2 hooks accept either; the V1 hooks
+// keep the original `bash` check.
+const SHELL_TOOLS = new Set(["bash", "shell"]);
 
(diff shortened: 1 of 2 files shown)


candrewlee14:
Thanks for the review — both points addressed or explained:

**1. Unmanaged SDK dependency — fixed in 914fc87.** opencode only manages dependencies for *package* plugins (`opencode plugin add`); this is a *file* plugin, so its imports resolve from `~/.config/opencode/node_modules` and nothing installs them (the old template had the same situation with `@opencode-ai/plugin`). Shelling out to npm from the Rust installer would be fragile (npm may not exist, offline machines, bun-vs-npm), so I documented the one-time `npm --prefix ~/.config/opencode install --save-exact @opencode/plugin` step in the opencode section of `agent-hooks.md` and in the template header. Note the V1 import is `import type` (erased at runtime) and `Plugin.define` is identity, so `@opencode/plugin` is the single runtime dependency on both opencode versions.

**2. Identical concurrent commands — API limitation, not fixable client-side.** V2's `shell.create.before` event is `{ command, cwd, timeout, shell, env }` with no tool call ID (verified against `@opencode/plugin@2.0.2` types and the 2.0.2 server dispatch, which builds exactly that object). Any correlation must therefore be by command text; oldest-pending-first is the best available mapping, and the ambiguity window is limited to genuinely in-flight duplicates. The alternatives are worse: skipping the shell hook loses the resolved cwd (tool args only carry an optional unresolved `workdir`), and starting the entry in `execute.before` would record commands the user goes on to deny. This tradeoff is noted in the code comment at the claim step.

candrewlee14:
Live verification update: the V2 port has been running in my own opencode session since 17:19 UTC today, and `atuin search --author opencode` shows every bash tool call from that session recorded with correct command text (including multi-line commands verbatim), cwd, duration, and exit status — e.g. 20 consecutive tool calls from 11:45 to 12:01 local, no gaps, no misattributions. This confirms the shell tool does route through the hooked `Shell.create` in 2.0.2, so the propose (tool before) → claim (shell create.before, oldest-matching command) → finish (tool after) flow works end to end in production, not just against stubs. The concurrent-identical-commands caveat from my previous comment stands as a known API-limited edge, but it has not manifested across a heavy real session.

PR #4117, opened 2026-09-12
feat(hooks): capture Antigravity (agy) shell commands

candrewlee14:
Antigravity (`agy`) loads `PreToolUse`/`PostToolUse` hooks from `hooks.json`, but `atuin hook install agy` answers `unknown agent` — and pointing its hook command at the Claude handler would silently record nothing: every field that matters differs (camelCase `hookEventName`, the tool call nested under `toolCall` with the command at `toolCall.args.CommandLine`, no tool-use id, no numeric exit code), and Antigravity requires a decision answer on stdout that atuin never prints.

This adds an `agy`/`antigravity` agent:

- Dedicated wire parser: starts entries for `run_command` with a non-empty `CommandLine`, correlates start/end through `conversationId` plus `stepIdx`, derives exit 0/1 from the `error` string.
- Stdout contract: `{"decision": "allow"}` on `PreToolUse` (Atuin observes but never gates, including for tools it ignores), `{}` afterwards.
- Installer writes entries under a named `atuin` container in `~/.gemini/config/hooks.json` (Antigravity maps hook names to event configurations, not a top-level `hooks` object), only for the two events Antigravity sends.
- Registers `antigravity` as a known agent; docs updated alongside the rest.

Verified: 17 new unit tests (parse matrix, container shape, idempotent install) pass with the existing 53 hook tests and the full `atuin-client` suite; `cargo clippy --tests -- -D warnings` clean; `cargo +nightly fmt --check` clean. Also smoke-tested the built binary end to end: a `PreToolUse`/`PostToolUse` pair for `echo smoke-test-123` printed the right stdout contract and produced a correctly closed history entry (since removed).

### DIFF
--- a/crates/atuin-client/src/history.rs
+++ b/crates/atuin-client/src/history.rs
@@ -26,7 +26,8 @@ pub use capture::CommandCapture;
 /// Known AI agent author values. Used by [`History::is_agent`] to guess who ran a command when the
 /// entry does not state it, and so when matching against [`AuthorPattern::AllAgent`] and
 /// [`AuthorPattern::AllUser`].
-pub const KNOWN_AGENTS: &[&str] = &["claude-code", "codex", "copilot", "opencode", "pi"];
+pub const KNOWN_AGENTS: &[&str] =
+    &["antigravity", "claude-code", "codex", "copilot", "opencode", "pi"];
 
 /// The spelling of [`AuthorPattern::AllUser`] on the command line and in the MCP tool schema.
 pub const AUTHOR_FILTER_ALL_USER: &str = "$all-user";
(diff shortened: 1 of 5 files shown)


candrewlee14:
Thanks for the review — addressed in 6fbc997:

**Stacked opencode commits — separated.** This branch was cut from the opencode PR branch, so the diff included it. Rebased onto `main`; this PR is now just the Antigravity commit. The opencode points (uninstalled runtime SDK) belong to #4116, where the SDK install step is already documented.

**Test style — fixed where the rule applies.** The four bare `#[test]`s in `hook.rs` are now `#[rstest]`. The remaining `#[test]`s in `event.rs` live inside a `proptest!` block — those are property tests that take generated arguments, so they can't move to `rstest` without losing coverage; that matches the pre-existing convention in this file.

**Orphaned entries/temp files — shared design, not agy-specific.** A denied command opens an entry (and a `/tmp/atuin-hook-*` id file) at `PreToolUse` that no `PostToolUse` ever closes. That is exactly how the Claude Code and Codex handlers already behave — hooks fire before the permission prompt on every agent, and the handler has no reaper. Changing that contract (e.g. a TTL sweeper for orphan id files) would affect all agents and deserves its own issue, not this PR. The agy docs section does disclose the observable part (a denied command can leave its entry open).

PR #4128, opened 2026-09-12
feat(config): accept unit-aware durations in config

markovejnovic:
Added some new types which define `Duration` entries and use them in settings, enabling type-safe accesses.

Unfortunately, in order to allow for backwards compatibility, wherein certain settings could be expressed as:

```toml
i_am_a_thing_in_minutes = 1
i_am_a_thing_in_seconds = 60
```

I had to write some different deserializers. There's a convenient crate for this [`serde_with`](https://docs.rs/serde_with/latest/serde_with/index.html) so I pulled that in


crates/atuin-common/src/time/duration.rs
```diff
@@ -199,13 +204,289 @@ impl DurationExt<Self> for time::Duration {
     }
 }
 
+/// A [`Duration`](std::time::Duration) guaranteed to be non-zero.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct NonZeroDuration(std::time::Duration);
+
+impl NonZeroDuration {
```
ellie, line 214: 🙌

ellie: we probably want to flag that previously many settings read negative as disabled, but will now error. worth checking we haven't told users to do this in the past (docs etc)

crates/atuin-client/src/settings.rs
```diff
@@ -1017,7 +1027,8 @@ pub struct Settings {
     #[serde(default)]
     pub sync_protocol: SyncProtocol,
 
-    pub sync_frequency: String,
+    #[serde_as(as = "AsDuration<Seconds>")]
+    pub sync_frequency: Duration,
```
ellie, line 1038: let's make sure that setting this to zero does not disable it. I for one have had 0 in my config for a long time, because it meant Atuin would sync after every single command.

ellie (commented):
generally looks good, though I am a bit concerned about the NonZeroDuration change in a few places

primarily

 - logs.retention
 - logs.search.retention
 - logs.daemon.retention
 - logs.ai.retention
 - ai.session_continue_minutes
 - sync_frequency (zero once meant "sync after every command")

I think zero retention especially makes semantic sense. Huge value add here, but I want to make sure we're not accidentally sneaking in a regression

markovejnovic:
Yep, I'm addresing those, will reply after I push

crates/atuin-common/src/time/duration.rs
markovejnovic: I want to avoid this regression -- I will keep things backwards compatible.

crates/atuin-client/src/settings.rs
```diff
@@ -663,8 +670,10 @@ pub struct Ai {
     /// Path to the AI sessions database.
     pub db_path: String,
 
-    /// The maximum time in minutes that an AI session can be automatically resumed.
-    pub session_continue_minutes: i64,
+    /// The maximum time an AI session can be automatically resumed (a bare number is minutes).
+    /// `0` disables auto-resume.
+    #[serde_as(as = "AsDisableableDuration<Minutes>")]
+    pub session_continue_minutes: Option<NonZeroDuration>,
```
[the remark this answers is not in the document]
markovejnovic: fixed. confirm

crates/atuin-client/src/settings.rs
```diff
@@ -607,9 +612,10 @@ pub struct Logs {
     #[serde(default)]
     pub level: LogLevel,
 
-    /// Default retention days for log files. Defaults to 4.
+    /// Default retention for log files (a bare number is days). Defaults to 4 days.
     #[serde(default = "Logs::default_retention")]
-    pub retention: u64,
+    #[serde_as(as = "AsDuration<Days>")]
+    pub retention: NonZeroDuration,
```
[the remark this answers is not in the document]
markovejnovic: fixed. confirm

crates/atuin-common/src/time/duration.rs
markovejnovic: @greptileai confirm this is true

crates/atuin-client/src/settings.rs
markovejnovic: good catch -- `0ms` was compatible, but i fixed so that `0` also works now

@greptileai confirm my claim

markovejnovic:
Alrighty, everything we talked about should be addressed -- no regressions

PR #4130, opened 2026-09-12
refactor(client): rename SyncEngine to SyncSession

markovejnovic:
In the subsequent PR of the stack, I will use the word `SyncEngine` to mean a persistent synchronization engine which runs checks in the background. Consequently, I had to rename the current SyncEngine (which isn't really an engine) to something else. SyncSession is the best I could come up with


crates/atuin-daemon/src/components/sync.rs
```diff
@@ -207,7 +207,7 @@ async fn do_sync_tick(
 
     // Perform the sync
     let res = async {
-        let engine = SyncEngine::builder()
+        let engine = SyncSession::builder()
```
ellie:
```suggestion
        let session = SyncSession::builder()
```

ellie (approved):
lgtm, probs want to rename the variables as well as the type but up to you 🤷‍♀️

PR #4131, opened 2026-09-12
refactor(daemon): replace sync event-bus with a standalone engine

markovejnovic:
Part 4/4. Base: `refactor/sync-session-rename`. This is the split-out core of #4104 (kept open separately).

Moves sync out of the event-bus Component model into a self-contained `crate::sync` engine: a background tokio task owned by a handle that aborts on drop, modelled on the output-capture GC task. Freshly-synced history is written straight into the shared search index instead of broadcasting `HistorySynced`, dropping the bus hop and the search side's DB round-trip. The unused `HistorySynced` / `SyncCompleted` / `SyncFailed` events and the sync `Component` impl are removed.

When indexing downloaded records the engine re-reads them through `load_active`, so deleted rows are filtered out — keeping tombstones and create+delete-in-one-batch churn out of the search index. A regression test drives the index directly to cover it.

## Stack (bottom → top)
1. #4128 — unit-aware durations (base: `main`)
2. #4129 — backoff sleeps before each attempt
3. #4130 — rename client `SyncEngine` → `SyncSession`
4. #4131 — event-bus → standalone sync engine

Splits the original event-bus refactor (#4104, left open separately) into reviewable pieces. Review/merge bottom-up.

https://claude.ai/code/session_01Ss7FrZLSKJEnKNLUXfXixH


crates/atuin-daemon/src/sync/engine/worker.rs
```diff
@@ -0,0 +1,258 @@
...
+
+impl Worker {
+    pub async fn new(
+        handle: DaemonHandle,
+        index: Arc<RwLock<SearchIndex>>,
+    ) -> Result<Self, StartError> {
+        let host_id = Settings::host_id().await.map_err(StartError::HostId)?;
+
+        let encryption_key = handle.encryption_key();
+        // TODO(markovejnovic): We should definitely not be creating new stores, but this is pending
```
ellie, line 91: wdym?

crates/atuin-daemon/src/sync/engine/worker.rs
```diff
@@ -0,0 +1,258 @@
...
+            };
+            let interval = interval.get();
+
+            if settings.auto_sync {
+                let backoff = Backoff::Exponential {
+                    initial: interval,
+                    max: MAX_BACKOFF,
+                    factor: BACKOFF_FACTOR,
+                };
+                let _ = backoff.retry(|| self.sync_tick(&settings), MAX_BACKOFF).await;
```
ellie, line 128: is there a world where the backoff timeout could lead to a partially successful sync tick? eg, data is downloaded but not build, timeout is hit, build is cancelled, we end up with data in the store but not built

ellie, line 128: this also returns Err once we hit the max backoff. Existing behaviour (and the ideal) is to hit the backoff max, and then keep re-attempting at that interval until success.

with this change, we fail, start ramping up, then eventually hit the max backoff and reset back to the start again.

crates/atuin-daemon/src/sync/engine/worker.rs
markovejnovic: We're invoking the `::new` constructors, even though that just opens the stores. The reality is we want to have some sort of `app_ctx.stores()` or some sort of `AppStores` reference object that gets injected in here, rather than going through the whole settings + creation path

crates/atuin-daemon/src/sync/engine/worker.rs
markovejnovic: Hmmmm... Since I want to unblock the PR, let's discuss this later -- I'll implement it as you suggest, but I'd push back on the definition of _ideal_ here.

I was coming at it from the perspective of transient issues -- if we find a transient issue, we want to backoff, but then (assuming the issue is transient), we'll bounce back to the minimum timeout duration. The transient issue is -- well -- transient, so retrying more often seems like a good idea. And if it's not transient, then the behavior doesn't matter since we'll never actually succeed to retry and we'll be pinned at the timeout ceiling.

Could go either way, the way I see it

crates/atuin-daemon/src/sync/engine/worker.rs
```diff
@@ -0,0 +1,258 @@
...
+        // TODO(markovejnovic): The fact that we rebuild the SyncSession on every single iteration
+        //                      of the loop feels wrong to me, but the legacy behavior was this, and
+        //                      in favor of not breaking anything, I prefer to keep it this way.
+        let session = SyncSession::builder()
+            .store(self.handle.store().clone())
+            .client_source(ClientSource::FromSettings {
+                settings,
+                caps: Some(self.handle.caps().clone()),
+            })
+            .build()
```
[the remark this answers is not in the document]
markovejnovic: This is not exactly correct. Capabilities have a system for automatically refreshing on a difference between the server.

crates/atuin-daemon/src/sync/engine/worker.rs
markovejnovic: Yes, but no worse than before. Any sort of failure short-circuits the function which causes partial builds... I'll try to fix that in this PR.

PR #4132, opened 2026-09-13
feat(search): allow repeated exit-code filters

### LINKED ISSUE #2701: Allow passing `--exit` and `--exclude-exit` multiple times
It would be very helpful to support specifying arguments multiple times. For example, I would like to do:

`atuin search --exclude-exit 0 --exclude-exit 130 --delete-it-all`

I see related issues #112 and #121, they are both closed as completed, but I don't see any PRs related to them, and running `atuin search --exit 0 --exit 1` returns `error: the argument '--exit <EXIT>' cannot be used multiple times`. Is this still an open issue or am I missing something?

Atuin version: `18.5.0`

pederbe:
### Problem
The CLI's `--exit` / `--exclude-exit` flags only accepted a single value each, unlike author/shell, which already supported repeats. `Sqlite::search` could apply one equality and one inequality but nothing more.

### Change
- Changed `exit` / `exclude_exit` from `Option<i64>` to `Vec<i64>` in the CLI, matching the author/shell pattern. `--exit 1 --exit 2` now matches either code; repeated `--exclude-exit` removes every listed code.
- Changed `OptFilters` to take borrowed `&[i64]` slices for these fields, with an empty slice meaning no restriction. Borrowing keeps `OptFilters` `Copy` and works with the deletion loop that reuses the same filters.
- The database builds one `IN` predicate for non-empty exit lists and one `NOT IN` for exclusions. Both combine with `AND` alongside the existing filters. Exclusions win over overlapping inclusions; duplicates are harmless.

Closes #2701

## Checks


markovejnovic:
LGTM! Thank you!

PR #4140, opened 2026-09-14
fix(nu): preserve multiline Up navigation

okuuva:
On Nushell 0.115.2 and newer, try `movelineup` after `menuup` and before opening Atuin. Keep the existing chain on older versions, where blocked cursor movement prevents fallback.

Requires the [no-op cursor movement fallback fix in Reedline][1].
Nushell picked it up in [nushell/nushell#19013][2].

Related: https://github.com/atuinsh/atuin/issues/798

## Checks

[1]: https://github.com/nushell/reedline/pull/1208
[2]: https://github.com/nushell/nushell/pull/19013


okuuva:
Current `main` and nightlies for `nu` identify as `0.115.2`. The version gate would break for older local builds and nightlies <18 but not sure if that's really a blocker for merging this. Using the latest shiny means accepting the risk that stuff might break.
Will obviously work even if the next published `nu` version is `0.116.0` instead of 0.115.2`.

PR #4143, opened 2026-09-14
chore: less strict versions

markovejnovic:
This PR removes the explicit version pinning we have in `Cargo.toml`, loosening it in favor of "pinning" the `Cargo.toml` on the:

- `MAJOR` if the version is `>= 1.x.y`.
- `MINOR` if the version is `< 1.x.y`.

This aligns with how Cargo defines "breaking changes"


markovejnovic:
Also adds `locked-tripwire` to prevent accidental `cargo install`s which risk supply-chain attacks on our users.

PR #4161, opened 2026-09-16
perf(atuin-client): read tail idx via index-only scalar query in push_record

markovejnovic:
## Finding

`HistoryStore::push_record` and `push_batch` (history/store.rs) allocate the next append index as `store.last(&series)?.idx + 1`. `last()` runs `select <all columns> ... order by idx desc limit 1` and fully materializes the tail `Record`: it loads and copies the `data` + `cek` blobs and parses the `id` and `host` UUIDs from text — all of which are thrown away, since only `idx` is used. The packfile packer's `ceiling` read (packfile/packer.rs) and `SqliteStore::len` did the same.

`push_record` runs on every `atuin history end`, and `push_batch` on every sync page, so this sits on the append hot path.

## Fix

Add `SqliteStore::tail_idx(&series) -> Result<Option<u64>>`:

```sql
select idx from store where host = ?1 and tag = ?2 order by idx desc limit 1
```

The `record_uniq (host, tag, idx)` unique index is covering for this query, so it is satisfied without touching a row — no blob load, no UUID parses, no `Record` construction. Routed `push_record`, `push_batch`, the packer ceiling read, and `len()` through it via the `atuin_common::db` scalar helper (raw `sqlx::query` is clippy-disallowed). `last()` is kept for callers that need the whole record.

Both approaches read the *same* populated in-memory series (1000 records) in one bench file, parametrized by the tail record's data-blob size (256 B ≈ a typical shell entry, 64 KiB ≈ a dotfiles/scripts record). Each timed sample issues 64 probes to amortize the fixed `block_on`/pool-checkout cost that the two paths share.

Divan medians (per 64-probe sample; per-probe in parens):

| blob size | `last()` (old) | `tail_idx` (new) | delta |
|-----------|----------------|------------------|-------|
| 256 B     | 1.281 ms (20.0 µs) | 1.207 ms (18.9 µs) | new 1.06× faster (−6%) |
| 8192 B    | 1.358 ms (21.2 µs) | 1.211 ms (18.9 µs) | new 1.12× faster (−11%) |
| 65536 B   | 1.546 ms (24.2 µs) | 1.349 ms (21.1 µs) | new 1.15× faster (−13%) |

The scalar probe stays flat (~19 µs/probe) across blob sizes while `last()` grows with the discarded blob — the divergence confirms the wasted work is the row/blob materialization. The win is smaller at typical sizes (the shared SQLite round-trip / pool checkout dominates the ~19 µs baseline) but is consistent: the scalar path was faster in every cell across repeated runs.

Reproduce:

```
cargo bench -p atuin-client --bench benchmarks -- tail_idx
```

## Tests

`cargo test -p atuin-client record::sqlite_store history::store packfile` — all pass. `cargo build -p atuin-client` clean; clippy adds no new warnings.


markovejnovic:
this is not really worth it

PR #4162, opened 2026-09-16
perf(atuin-client): parse config.toml and scan env once in Settings::new

markovejnovic:
## Finding

`Settings::new()` runs on every shell hook, and each call went through
`build_config`, which re-parsed `config.toml` and re-scanned the environment
**three times**:

1. a `data_dir`-only pre-parse (`DataDirOnly`) to resolve the data dir,
2. a `build_cloned()` to read the resolved path strings, and
3. a final `build()` after setting expanded-path overrides.

Each of those is a full `config` build: the `File` source re-reads and re-parses
the TOML from disk and the `Environment` source iterates all of `std::env::vars()`.

## Fix

`build_config` now materializes the layered config **once**:

- Build the merged config (defaults + file + env) with the default data dir, and
  only redo the merge when a custom `data_dir` / `ATUIN_DATA_DIR` is actually in
  effect (it seeds the other path defaults).
- Expand `~`/`$VAR` in the resolved path values **in place** on the built config,
  instead of reading them back and rebuilding a second time to apply overrides.
- Guard both `create_dir_all` calls behind a cheap existence check, keeping the
  already-initialized common case off the syscall.

The two-phase `data_dir` resolution is preserved, including the historical
precedence where `ATUIN_DATA_DIR` in the environment wins over the config file
value, and honoured only when a config file exists. Every default value is
unchanged. In the common case (no custom `data_dir`) this is a single config
build; a custom data dir costs two (vs three before).

`get_config_value` still receives a fully path-expanded `Config`, so
`atuin config get --resolved <path-key>` is unaffected.

Both the old three-parse path and the new single-parse path are replicated over
an identical realistic ~30-line `config.toml` in one bench file, and asserted to
resolve to the same paths before timing.

| bench | median |
| ----- | ------ |
| old (3 parses) | 150.9 µs |
| new (1 parse)  | 56.7 µs |

**~2.66x faster (-62%)**, stable across runs (second run: old 154 µs, new 57.6 µs).

Reproduce:

```
cargo bench -p atuin-client --bench settings
```

## Tests

- `cargo test -p atuin-client` — 460 passed (incl. a new `set_config_string`
  test covering nested-key navigation).
- `cargo build -p atuin-client` and `cargo clippy -p atuin-client --benches --tests`
  clean.


markovejnovic:
Yes, this idea is wise, but not sufficiently generic to merge as-is. I would like to have an `AppCtx` type which I've repeatedly tried (and failed, mostly due to scope and priorities) to merge. That would be the total, generic solution here.

PR #4163, opened 2026-09-16
perf(atuin): build fuzzy highlighter once per frame, not per row

markovejnovic:
## Finding

The search-results highlighter rebuilt its whole fuzzy engine for **every visible row on every keystroke/frame**.

- `search/history_list.rs` calls `get_highlight_indices(command)` once per row while rendering the results list.
- The db fuzzy path (`search/engines/db.rs`) did `FzfV2::new()` + `FzfParser::new()` + `parser.parse(query)` **per row**. `norm`'s own docs note that `FzfV2::new()` "will immediately allocate around 5kb of heap memory, so it's recommended to call this once and reuse the same instance."
- The daemon path (`search/engines/daemon.rs`) compiled a `frizbee::Matcher::from_query(...)` **and** normalized the query with `normalize_diacritics()` **per row**.

All of that setup is query-invariant: it only depends on the search input, which is constant across every row of a frame.

## Fix

Split the query-invariant setup out of the per-row path. Each engine now builds a `PreparedHighlighter` **once per frame** (`AnySearchEngine::prepare_highlighter`), and the render loop calls `prepared.highlight_indices(command)` per row, reusing the scorer/matcher across all rows. Empty search input short-circuits to "no highlight", skipping matcher construction and per-row diacritic normalization entirely.

Output is **byte-identical**:
- Reusing one `FzfV2` across candidates is the crate's recommended usage (its scratch slabs are overwritten per call).
- An empty query already produced no ranges (`Fzf::distance` returns early without touching `ranges`; `frizbee`'s empty pattern yields empty `indices`), so the short-circuit changes nothing.
- The existing daemon highlight regression tests (accented / multibyte / CJK / long-query) pass unchanged.

Both approaches run over the **same** 30 realistic command strings and a typical multi-token query in one bench file, so the relative comparison is robust to machine load from parallel builds. `atuin` is a binary crate with no lib target, so the two approaches are replicated self-contained against the same `norm` fzf engine `engines/db.rs` imports.

| approach | median |
| --- | --- |
| `old_per_row` (build scorer+parser per row) | ~11.35 µs |
| `new_hoisted` (build once, reuse across rows) | ~4.21 µs |

**~2.7x faster (~-63%)** for a 30-row frame. Ratio held at 2.6–2.7x across repeated runs under varying load.

Reproduce:

```
cargo bench -p atuin --bench search_highlight
```


markovejnovic:
Perhaps not worth it.

PR #4164, opened 2026-09-16
perf(atuin): avoid per-char String allocation when drawing command cells

markovejnovic:
Rendering a history row draws the command one terminal cell at a time:
`command()` in `crates/atuin/src/command/client/search/history_list.rs`
iterates `display.char_indices()` and calls `self.draw(&str, ..)` per
character. Each cell allocated a fresh heap `String` via `ch.to_string()`
— one allocation per visible character, per row, on every redraw.

## Fix

`draw()` already takes `&str`, so encode the char into a 4-byte stack
buffer instead of allocating:

```rust
let mut ch_buf = [0u8; 4];
self.draw(ch.encode_utf8(&mut ch_buf), Style::from_crossterm(char_style));
```

| approach                        | median   |
| ------------------------------- | -------- |
| `old_to_string` (`ch.to_string()`) | 1.54 µs |
| `new_encode_utf8` (`encode_utf8`)  | 193.7 ns |

**~7.95x faster** (−87% median) over a full row.


crates/atuin/src/command/client/search/history_list.rs
```diff
@@ -350,7 +350,9 @@ impl DrawState<'_> {
                 }
                 char_style.attributes.set(style::Attribute::Bold);
             }
-            self.draw(&ch.to_string(), Style::from_crossterm(char_style));
+            // Encode into a stack buffer instead of allocating a String per cell.
```
markovejnovic:
```suggestion
```

PR #4166, opened 2026-09-16
perf(atuin-daemon): append finished history by reference, dropping a full clone

markovejnovic:
## Finding

When a command finishes, `HistoryJournal::finish` (crates/atuin-daemon/src/history_journal.rs) appends the entry to the record store and then still uses it twice more — to update the in-memory search index and to broadcast a `CmdEvent::Finished`. Because `HistoryStore::push` takes the `History` **by value**, finish was forced to `push(history.clone())`: a full deep clone of the `History` (owned `command`, `cwd`, `session`, `author`, `intent`, `shell` strings) on every single command completion, purely to hand serialization an owning `HistoryRecord::Create`.

The serialization itself never needed ownership: `History::serialize` and `HistoryRecord::serialize` both already take `&self`.

## The fix

- Add `HistoryRecord::serialize_create(&History) -> Result<DecryptedData>`, producing the exact same bytes as `HistoryRecord::Create(history.clone()).serialize()` without constructing an owning record. `HistoryRecord::serialize`'s `Create` arm now delegates to it, so the wire format is unchanged (guarded by the existing byte-snapshot test `test_serialize_deserialize::case::create`).
- Add `HistoryStore::push_ref(&self, history: &History)`, which serializes from the borrow and shares the append loop with `push` via a new private `push_serialized`. Owned `push()` now delegates to `push_ref`.
- In `finish`, call `self.history_store.push_ref(&history)` and drop the `.clone()`.

Serialized bytes are byte-identical — the record written to the store is unchanged.

Both approaches over identical, realistic finished commands in one file: `old` = clone + serialize (what value-taking `push` required), `new` = `serialize_create` from a borrow. Medians:

| batch | old (clone+serialize) | new (borrow) | delta |
|-------|-----------------------|--------------|-------|
| 1     | 418.9 ns              | 296.6 ns     | ~1.41x faster (-29%) |
| 100   | 38.87 µs              | 28.41 µs     | ~1.37x faster (-27%) |

(A second run reproduced the same ratios: 379.9 ns → 270.5 ns, 36.74 µs → 25.37 µs.)

Reproduce:

```
cargo bench -p atuin-client --bench journal_finish
```

## Tests

- `cargo build -p atuin-client -p atuin-daemon` clean.
- `cargo test -p atuin-client --lib history::` — 48 passed (includes the byte-identity snapshot).
- `cargo test -p atuin-daemon` — all suites passed.


markovejnovic:
This is correct, however, the scope and complexity does not justify the pay-off. As we migrate more and more stuff into the daemon, hopefully this will become more and more redundant.

PR #3796, opened 2026-07-30
TUI v2

markovejnovic:
## Goals

- [ ] #3611
- [ ] #2731
- [ ] #3213
- [ ] #2943
- [ ] #2663
- [ ] #1324
- [ ] #2398
- [ ] #798
- [ ] #3045
- [ ] #2959
- [ ] #1468
- [ ] #1828
- [ ] #2620
- [x] #3477
- [ ] #798
- [ ] #2771
- [x] #2568
- [ ] #2845
- [ ] #967
- [ ] #1761
- [ ] #3539
- [ ] #2683
- [ ] #2592
- [x] #1144
- [ ] #1746
- [ ] #1836
- [ ] #3599
- [x] #3830
- [x] #2773
- [ ] #3866 (UI should tell you if you have the daemon enabled but it's not running—right now you only get an error if you have `search_mode = "daemon-fuzzy"` *and* you type at least one character in the search box)
- [ ] #2333
- [ ] #2214 (this is important; it has an accessibility purpose)
- [ ] #1839
- [ ] #3978
- [ ] #4167

PR #4167, opened 2026-09-16
perf(atuin): cache resolved syntax styles per frame instead of per-char theme lookups

markovejnovic:
## Finding

The interactive search redraws every visible row on each keystroke and each scroll. In `DrawState::command` (`crates/atuin/src/command/client/search/history_list.rs`), the command column is styled **one character at a time**, and every character called `Theme::as_style(meaning)`:

```rust
let mut char_style = source_byte
    .and_then(|b| syntax.get(b))
    .map_or(style, |&meaning| self.theme.as_style(meaning));
```

`Theme::as_style` is not free (`crates/atuin-client/src/theme.rs`): it calls `closest_meaning`, which does a `HashMap::contains_key` and can recurse through meaning fallbacks, then indexes the `styles` HashMap. So each highlighted character paid at least two HashMap probes, repeated for the whole visible screen on every redraw.

## Fix

Resolve the handful of syntax meanings (`Base`, the `Syntax*` set) once per row into a small fixed table, plus the highlight color once per row, then match per character:

```rust
let syntax_styles = (!syntax.is_empty()).then(|| SyntaxStyles::resolve(self.theme));
let highlight_style = row_highlighted.then(|| self.theme.as_style(Meaning::AlertWarn));
```

The per-character loop becomes a cheap match over a fixed enum set instead of a HashMap probe. The table is only built for rows that are actually syntax-highlighted, so the highlighting-off path keeps doing zero per-character resolution. `theme.rs` is untouched.

Both approaches run over the **identical** per-row classification vectors (a realistic screenful: 48 rows of classified shell commands, ~2400 character resolutions), so the relative comparison is robust to load from other parallel builds. `Theme::as_style` lives in `atuin-client`, so the proving bench lives there too and replicates the per-row-table logic.

| approach | median |
| --- | --- |
| `per_char_as_style` (old: HashMap probe per char) | **22.37 µs** |
| `per_row_style_table` (new: resolve once per row, match per char) | **6.08 µs** |

**~3.7x faster** (stable at 3.7-3.8x across repeated runs).

### Reproduce

```
cargo bench -p atuin-client --bench benchmarks -- theme_style
```


markovejnovic:
This is a really good PR, but it's unfortunately a little bit of a hack. This should be revisited in #3796.

PR #4171, opened 2026-09-17
fix(hook): accept a string tool_response from codex

### LINKED ISSUE #4169: [Bug]: Codex completion events with string tool_response are silently ignored

Shell commands ran in Codex should appear in Atuin history when the daemon is enabled

The hooks are configured in `~/.codex/hooks.json`: both `PreToolUse` and `PostToolUse` run `atuin hook codex` with matcher `^Bash$`. Both hooks are enabled and trusted in Codex, and Atuin has `[daemon] enabled = true`.

### What happened?

With Atuin 18.22.0 and Codex CLI 0.154.0, shell commands do not appear in Atuin history when the daemon is enabled. Codex sends `tool_response` as a string, but Atuin expects an object and silently ignores the completion event. The daemon therefore never saves the command.

For example, after `printf probe`, Codex sends a `PostToolUse` payload like this (IDs shortened):

```json
{
  "hook_event_name": "PostToolUse",
  "tool_name": "Bash",
  "tool_use_id": "example-call",
  "session_id": "example-session",
  "cwd": "/tmp",
  "tool_input": {"command": "printf probe"},
  "tool_response": "probe"
}
```

I had Codex try to fix this problem. It noticed that after sending the matching start event, changing only this line in the completion payload made Atuin accept it and save the command:

```diff
-  "tool_response": "probe"
+  "tool_response": {"exitCode": 0}
```

### Atuin doctor output

```yaml
{
  "atuin": {
    "version": "18.22.0",
    "commit": "f98361cf8d33326fb70e7a242ff37a58845f5173",
    "sync": null,
    "sqlite_version": "3.46.0",
    "daemon_enabled": true
  },
  "shell": {
    "name": "bash",
    "default": "bash",
    "plugins": [
      "atuin",
      "bash-preexec"
    ],
    "preexec": "bash-preexec"
  },
  "system": {
    "os": "Ubuntu",
    "arch": "x86_64",
    "version": "24.04",
    "disks": [
      {
        "name": "/dev/mapper/ubuntu--vg-ubuntu--lv",
        "filesystem": "ext4"
      },
      {
        "name": "/dev/sda2",
        "filesystem": "ext4"
      },
      {
        "name": "/dev/sda1",
        "filesystem": "vfat"
      }
    ]
  }
}
```

### Code of Conduct

maskjelly:
Closes #4169.

Codex serializes `tool_response` as the tool's output string; the wire type required an object, so the `PostToolUse` payload failed to deserialize and `HookEvent::from_json_str` skipped it as a data error. The matching `PreToolUse` had already opened the entry, so with the daemon owning the store the command never persisted — `atuin hook codex` looked like it worked while history silently lost every Bash command.

**Reproduced** against this tree with `target/debug/atuin`, daemon disabled, piping the issue's exact PreToolUse + PostToolUse payloads: before the fix the completion left the hook's `example-call` id file behind and the row stayed `printf probe | exit=-1 | duration=-1` (daemon-owned stores only persist on the End event, which is exactly the reported loss). With `"tool_response":{"exitCode":0}` as a control, the row finished `exit=0`. After the fix the string form behaves identically to the object form: id file removed, `exit=0`.

**Fix:** `WireToolResponse` is now an untagged enum accepting either the string (Codex) or the object with `exitCode` (Claude Code), with an `exit_code()` accessor. A bare string carries no exit code — Codex routes failures through `PostToolUseFailure`, which the parser already maps to exit 1, so a `PostToolUse` string means the command ran, the same treatment an object without `exitCode` gets. Scoped to the hook parser; no shared types changed.

**Tests:** `completion_accepts_string_tool_response` feeds the issue's exact raw JSON through the same `from_json_str` the hook uses and asserts `End { exit: 0 }`; reverting only the two production hunks fails it with `None`. The object form was already covered and still passes.

**Gates:** `cargo +nightly fmt --all --check` clean; `cargo clippy -p atuin -- -D warnings` and `--tests` clean; `cargo test -p atuin --bin atuin` 538 passed / 0 failed. I could not exercise a live Codex 0.154.0 against a running daemon in this environment; the parser is the shared choke point for both daemon and local paths, and the manual CLI run above covers the completion end to end.


crates/atuin/src/command/client/hook/wire.rs
```diff
@@ -58,7 +58,29 @@ pub struct WireToolInput {
...
+    },
+}
+
+impl WireToolResponse {
+    /// The exit code the agent reported, if any.
+    ///
+    /// A bare string carries no exit code. Codex reports failures through a
+    /// separate `PostToolUseFailure` event, so a `PostToolUse` with no exit code
+    /// means the command ran; callers default the missing value to zero, exactly
+    /// as they do for an object without `exitCode`.
```
markovejnovic:
```suggestion
    /// Get the exit code the agent reported, if any.
```

crates/atuin/src/command/client/hook/wire.rs
```diff
@@ -58,7 +58,29 @@ pub struct WireToolInput {
 
 /// See [`WireHookEvent::tool_response`].
 #[derive(Debug, Deserialize)]
-pub struct WireToolResponse {
-    #[serde(rename = "exitCode", default)]
-    pub exit_code: Option<i64>,
+#[serde(untagged)]
+pub enum WireToolResponse {
+    /// Codex serializes the response as the tool's output string. It carries no
+    /// exit code; see [`WireToolResponse::exit_code`].
```
markovejnovic:
```suggestion
    /// Some harnesses (ie. Codex) only return the tool output string.
```

crates/atuin/src/command/client/hook/wire.rs
```diff
@@ -58,7 +58,29 @@ pub struct WireToolInput {
...
+    Output(String),
+    /// Claude Code serializes the response as an object carrying the exit code.
```
markovejnovic:
```suggestion
    /// Other harnesses return more details.
    /// TODO(markovejnovic): Should we split the enum per-harness?
```

crates/atuin/src/command/client/hook/event.rs
```diff
@@ -80,7 +80,8 @@ impl From<WireHookEvent> for Option<HookEvent> {
                 })
             }
             HookEventName::PostToolUse => {
-                let exit = wire.tool_response.and_then(|response| response.exit_code).unwrap_or(0);
+                let exit =
```
markovejnovic, line 85:
```suggestion
                // TODO(markovejnovic): Is it safe to assume that no exit code
                //                      means "success"?
                let exit =
```

crates/atuin/src/command/client/hook/wire.rs
```diff
@@ -58,7 +58,25 @@ pub struct WireToolInput {
...
+    /// TODO(markovejnovic): Should we split the enum per-harness?
+    Object {
+        #[serde(rename = "exitCode", default)]
+        exit_code: Option<i64>,
+    },
+}
+
+impl WireToolResponse {
+    /// The exit code the agent reported, if any.
+    /// Get the exit code the agent reported, if any.
```
markovejnovic:
```suggestion
```

markovejnovic:
Thank you for your PR!

Issue #3319, opened 2026-03-21
[Bug]: Migration error 20260224000100


Grimblee:
### What did you expect to happen?

Hello !

For some reason my atuin setup is in error, I tried deleting the records.db as mentioned in a previous issue but that didn't fix it.

I tried reinstalling, I also reset my data and tried only copying the history.db* files, but it persists.

I'm not using the sync function.
How can I recover my history data ?

Thank you,

### What happened?

Error: migration 20260224000100 was previously applied but is missing in the resolved migrations

Caused by:
    migration 20260224000100 was previously applied but is missing in the resolved migrations

Location:
    crates/atuin/src/command/client.rs:144:18

### Atuin doctor output

```json
{
  "atuin": {
    "version": "18.12.1",
    "commit": "42dd242541d1db21c676e213a98d50ee74bd553d",
    "sync": null,
    "sqlite_version": "3.46.0"
  },
  "shell": {
    "name": "zsh",
    "default": "zsh",
    "plugins": [],
    "preexec": null
  },
  "system": {
    "os": "CachyOS Linux",
    "arch": "x86_64",
    "version": "rolling",
    "disks": [
      {
        "name": "/dev/mapper/luks-eabc0411-aef5-450a-9a80-60f7c463abf2",
        "filesystem": "btrfs"
      },
      {
        "name": "/dev/mapper/luks-eabc0411-aef5-450a-9a80-60f7c463abf2",
        "filesystem": "btrfs"
      },
      {
        "name": "/dev/mapper/luks-eabc0411-aef5-450a-9a80-60f7c463abf2",
        "filesystem": "btrfs"
      },
      {
        "name": "/dev/mapper/luks-eabc0411-aef5-450a-9a80-60f7c463abf2",
        "filesystem": "btrfs"
      },
      {
        "name": "/dev/mapper/luks-eabc0411-aef5-450a-9a80-60f7c463abf2",
        "filesystem": "btrfs"
      },
      {
        "name": "/dev/mapper/luks-eabc0411-aef5-450a-9a80-60f7c463abf2",
        "filesystem": "btrfs"
      },
      {
        "name": "/dev/mapper/luks-eabc0411-aef5-450a-9a80-60f7c463abf2",
        "filesystem": "btrfs"
      },
      {
        "name": "/dev/nvme0n1p1",
        "filesystem": "vfat"
      }
    ]
  }
}
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

[...]

PR #4173, opened 2026-09-17
fix: clarify migration errors caused by conflicting Atuin versions

ellie:
## Summary
- Explain that a missing migration likely means multiple Atuin versions or a downgrade.
- Show the running executable path and recovery steps without suggesting database deletion.
- Preserve underlying errors and add regression tests.

Related to #3319.

## Validation
- `cargo test -p atuin-common --features sqlite migration --lib`
- `cargo check -p atuin-client -p atuin-server`
- `cargo +nightly fmt`


ellie:
Local validation completed:

- `cargo build -p atuin` and `cargo build --workspace` pass. The full CLI build caught two remaining `sqlx::Result` callers in KV/scripts; fixed and pushed.
- All 1,645 library tests pass across common, client, KV, and scripts.
- Tested the built CLI against isolated temporary databases, injecting fake applied migration `20990101000000` into history, records, meta, KV, and scripts individually. All five fail with the new guidance, actual executable path, and original migration ID.
- Injected a checksum mismatch: original error remains, without misleading version guidance.
- Removed test-only faults and verified normal startup and saved history remain intact. No real user databases were touched.
- Formatting passes. Clippy for the affected crates (including tests) exits successfully; it reports an existing invalid disallowed-method entry for `time::Duration::new` in `clippy.toml`.

Manual test script and output are retained locally at `/tmp/test-atuin-migrations.py` and `/tmp/atuin-migration-e2e.log`.

crates/atuin-client/src/database.rs
```diff
@@ -354,7 +354,7 @@ impl Sqlite {
     }
 
     #[instrument(level = "trace", skip_all, err)]
-    async fn setup_db(pool: &SqlitePool) -> Result<()> {
```
markovejnovic, line 357: Can we not do this? @taylordotfish and I have been working really hard to avoid adding more `eyre` to the codebase. We've even managed to remove it from `atuin-common` and even some other places.

I'd suggest making a new `thiserror::Error` type called `DbSetupError` and returning that. Here's a convenient template:

```rust
#[derive(Debug, Clone, Error)]
pub struct DbSetupError {
  #[error("unexpected sqlx error")]
  Sqlx(#[from] #[source] sqlx::Error),
  #[error("This copy of Atuin cannot open a database likely updated by a newer version.\nYou may \
         have multiple versions of Atuin installed, and your shell is running an older \
         copy.\nRunning executable: {executable}\n\nCheck `atuin --version` and `which -a atuin` \
         (`where.exe atuin` on Windows).\nUpdate Atuin and remove older copies from your PATH, \
         then restart your shells and any Atuin daemon.\nDo not delete your database files or \
         migration records: this may lose your history.")]
  Migration(#[from] #[source] MigrationError),
}

File: crates/atuin-common/src/db/mod.rs [hunk 2]
```diff
@@ -93,6 +93,40 @@ where
...
+    #[error(
+        "This copy of Atuin cannot open a database likely updated by a newer version.\nYou may \
+         have multiple versions of Atuin installed, and your shell is running an older \
+         copy.\nRunning executable: {executable}\n\nCheck `atuin --version` and `which -a atuin` \
+         (`where.exe atuin` on Windows).\nUpdate Atuin and remove older copies from your PATH, \
+         then restart your shells and any Atuin daemon.\nDo not delete your database files or \
+         migration records: this may lose your history."
+    )]
+    IncompatibleVersion {
+        executable: String,
```
markovejnovic:
```suggestion
        executable: PathBuf,
```

I'm guessing?

crates/atuin-kv/src/database.rs
```diff
@@ -40,7 +40,7 @@ impl Database {
         Ok(self.sqlite.info().await.version?)
     }
 
-    async fn setup_db(pool: &SqlitePool) -> Result<()> {
+    async fn setup_db(pool: &SqlitePool) -> eyre::Result<()> {
```
markovejnovic: See my other comment please

crates/atuin-scripts/src/database.rs
```diff
@@ -40,7 +40,7 @@ impl Database {
         Ok(self.sqlite.info().await.version?)
     }
 
-    async fn setup_db(pool: &SqlitePool) -> Result<()> {
+    async fn setup_db(pool: &SqlitePool) -> eyre::Result<()> {
```
markovejnovic: See my other comment please

crates/atuin-common/src/db/mod.rs
```diff
@@ -93,6 +93,40 @@ where
...
+    Other(sqlx::migrate::MigrateError),
+}
+
+impl From<sqlx::migrate::MigrateError> for MigrationError {
+    fn from(error: sqlx::migrate::MigrateError) -> Self {
+        match error {
+            sqlx::migrate::MigrateError::VersionMissing(_) => Self::IncompatibleVersion {
+                executable: std::env::current_exe()
+                    .map(|path| path.display().to_string())
+                    .unwrap_or_else(|_| "<unable to determine executable path>".to_owned()),
```
markovejnovic: I would consider this an anti-pattern. We lose resolution. Why not store `PathBuf` in the error?

crates/atuin-common/src/db/mod.rs
```diff
@@ -93,6 +93,40 @@ where
...
+#[derive(Debug, Error)]
+pub enum MigrationError {
+    #[error(
+        "This copy of Atuin cannot open a database likely updated by a newer version.\nYou may \
+         have multiple versions of Atuin installed, and your shell is running an older \
+         copy.\nRunning executable: {executable}\n\nCheck `atuin --version` and `which -a atuin` \
+         (`where.exe atuin` on Windows).\nUpdate Atuin and remove older copies from your PATH, \
+         then restart your shells and any Atuin daemon.\nDo not delete your database files or \
+         migration records: this may lose your history."
+    )]
```
markovejnovic: I would bubble this up one layer, out of the low-level `atuin-common` utilities. The general rule of thumb I follow:

1. If the logic could live in some sort of general purpose library, it belongs in `atuin-common`.
2. If the logic is really `atuin`-app specific, then the code should live in `atuin-client`.

`atuin-common` has become a crate of useful utilities `atuin` finds useful, but is ultimately decoupled from atuin (for the most part). I think it should stay that way.

I also find it **really** strange that this error lost resolution on whatever the body of `sqlx::migrate::MigratError::VersionMissing(<what-is-this>)` is.

I would **strongly** suggest you remove this utility here, since it seems to me that the error returned by 

```rust
            ::sqlx::Connection::clear_cached_statements(&mut *conn)
                .await
                .map_err(::sqlx::migrate::MigrateError::Execute)?;
            ::core::result::Result::<(), $crate::db::MigrationError>::Ok(())
```

already has information on that the error was _due_ to the migration, so can we perhaps just match on that one layer above, in `atuin-client/src/database.rs`?

crates/atuin-common/src/db/mod.rs
markovejnovic: Also this feels really strange -- why does the error hold information on the active application?

markovejnovic (changes requested):
Left some feedback. Let's talk in person if my suggestions lack clarity

markovejnovic:
If you want, I can also just make the changes I want in this PR if it's easier for you

markovejnovic (approved):

PR #4182, opened 2026-09-18
ci: cut build-job time

markovejnovic:
image


ellie (approved):

markovejnovic:
Merge note: branch protection currently requires `build (depot-ubuntu-24.04)` and `build (depot-macos-26)`. After this PR the release build job is just `build` (Linux, 8 vCPU) and the release profile is type-checked inside `check (<os>)` on all three OSes, so the required checks need updating to `build` (and the macOS build entry dropped) when this lands.

PR #4193, opened 2026-09-19
Remove dotfiles

ellie:
Dotfiles was an interesting idea at the time, but I never felt that it was one worth continuing development of. There are much better solutions for dotfile management, many of which are much more comprehensive (see chezmoi, etc)

## Checks


markovejnovic (approved):

alifeee:
bit confused after installing Atuin, but thank you for writing clearly in the release notes the deprecation. I was able to use `atuin dotfiles alias list` to export my aliases and store them in my local dotfile

PR #4207, opened 2026-09-21
docs: add apt/Debian installation

nc7s:
Kinda weird it only occurred to me after maintaining it in Debian for 2 years, but anyway.

## Checks


markovejnovic:
Hi @nc7s. Thank you for the PR! We just had a conversation and as per #4219, we've decided that our docs will only recommend installation methods that enable users to receive the latest Atuin.

Unfortunately, since Debian isn't a rolling release distro, running Debian Stable, for example, results in a what-we-consider-very old version of Atuin (18.6). Between 18.6 and 18.23, there have been some massive improvements, so we do not feel comfortable recommending Debian on our docs page.

I recognize this kind of sucks, but since Atuin really moves super quickly, I cannot recommend using an older Atuin to our users, when we ship features and bugfixes so quickly that all older releases are significantly lower quality than newer ones.

I'm going to close this PR for now, and I apologize for that. I'll also reply to your Forum messages too, soon-ish.

nc7s:
Got it. I do understand the considerations here. No need to apologize, as we both know what it means to support more. Though like in #4219, I suggest that a clear support policy be set up. Like it or not, packagers have a reference to follow.

Otherwise, for example, I understand unwilling to support a "stable" variant of a distro, which might lack very far behind; but Debian unstable is quite fast on keeping up with your releases these days: 18.20.0 took 4 days, 18.21.0 - 3 days, 18.22.0 is a bit longer with 11 days but that came with architectural change, 18.23.0 - 0 day. I'm not saying that Debian unstable must be supported, but an unclear policy would invite questions like "why is the rolling part of my distro ineligible if I keep it on the roll?"

markovejnovic:
Makes sense, we're working on shipping a large feature so I'll pause the effort this week, but I'll try to get to it Friday or so.

PR #4219, opened 2026-09-22
docs: limit installation methods to those that install the latest version

taylordotfish:
After a team discussion, we've decided to limit the installation methods mentioned in the docs only to those that will always install the latest version (potentially lagging by a few days).

This PR removes mentions of installation methods that do not meet this requirement:

* XBPS (Void Linux): currently v18.16.1.
* Nixpkgs: although unstable is relatively up-to-date (v18.21.0), stable is currently v18.15.2. Only the Nixpkgs mention is removed; installation of the repository as a flake is still listed.


markovejnovic:
Should we even recommend _anything_ outside of our `curl` script? I don't want to have to update all the docs all the time and since we're the bleeding release, anything that isn't using our `curl` directly will be stale, by definition, even for a brief period.

ellie:
Yes - many users will not curl | bash, and will prefer using a package manager, no matter how much we encourage it. 

It is better that we document those methods as much as possible, including which require bash-preexec setup separately or other such accommodations.

markovejnovic (approved):

nc7s:
Throwing in my 2¢: consider setting up a formal support policy for versions, so both users and packagers know if a version has fallen out of upstream support, and base their choices on it.

